PR #10, opened 2015-05-11
Human-readable decode for id-ctrl and id-ns

jderrick:
This patchset adds a human readable decode for the various bitfields in id-ctrl and id-ns. It also formats the LBA Format decode in id-ns. It is enabled with the -H switch.


keithbusch:
Thanks, I've no objection to the idea, but I'm not sure about this commit log: one is an upstream merge, one only has a "conflict" in the commit log with no diff, and one is a duplicate. Can we clean this pull req up to only contain what you want to committed?

PR #15, opened 2015-06-22
fixed: segmentation faults in argconfig argument parsing and help message

kaoudis:
When attempting to use this tool (on kernel ver 4.1.0-rc8+ with Debian), I noticed that the argconfig help message seg faulted whenever I passed in a nonstandard option instead of, well, printing a help message. I also noticed that from time to time argument parsing itself would seg fault, ran valgrind on the nvme executable, and realized it was reading into memory it didn't own in quite a few places. Thus, I've fixed a few layers deep of seg faults. 

Scenarios where seg faults were previously occurring:

*nvme attach-ns -e*, where -e isn't an option to this particular sub command

*nvme attach-ns -c /dev/nvme0 --namespace-id=1*, where both options are legal

*nvme attach-ns -c /dev/nvme0 -p*, where one option is legal and one is not

(replace attach-ns and options with requisite valid and unvalid options for all nvme subcommands except help). 

I'm not entirely sure if I've gotten all the issues with argconfig, but it's better than it was. Feel free to try this changeset out as you will.


kaoudis:
accidentally removed one commit too many. fixing and reopening in a sec.

kaoudis:
fixed and reopened.

PR #16, opened 2015-06-22
add Documentation contents that should not be propagated in repo cloning to .gitignore

kaoudis:
Add a gitignore to Documentation/ so that contents of the folder stay constant regardless of make [clobber/doc] status.


kaoudis:
gotcha. changed it

PR #17, opened 2015-06-23
fix bad memory management in argconfig, add nvme help strings to subopt help menu, add tab completions for zsh and bash

kaoudis:
Commit 1 
--------------

In the case of a bad option parse or the user providing an unsupported option to any NVMe sub-command (e.g., read), argument parsing and argconfig's help-menu printing segfaulted due to multiple cases of improper memory management. 

- Add an argument *const unsigned int opt_arr_len* to option and suboption parsing as well as argconfig_parse's subcommand help menu function. Since we already know how many (max) possible arguments could be used in the NVMe sub-command functions where parsing is called, there's no reason to look through an indeterminate amount of memory to figure out what the options are and how many there are, esp since this can cause segmentation faults when argconfig wanders into memory it does not own.

- Replace *strcat ()* (which by definition appends strings to the end of a buffer, doesn't copy into a buffer as the original author may have intended) with printf straight to stdout. Also, remove extraneous string buffers. Previous behaviour: append (strcat) help strings to end of buffer, extending into unallocated memory. Note that *nothing* is ever stored in the buffer in the original version of this code; the NVMe sub-command help string is printed by a different function called by the help menu function, then an attempt is made to populate the buffer with options (which results in text being appended to instead of put into the buffer, because strcat concatenates).

The above items eliminate many (hopefully all?) of the memory management issues with argconfig.

- Finally, get rid of some weird spacing issues in argconfig.c that made the code harder to look over in vim (lines > 80 characters are now split up). 

Things that would fail before adding this patch:

*nvme read --e-*

*nvme attach-ns --namespace-id=1 -?*

*nvme help ?*

*nvme attach-ns*

Commit 2 
---------------

 - Add NVMe sub-option function descriptions to sub-command help messages. This makes sub-command help messages actually helpful. 

- Prettify option output in help messages; group options by what they do, instead of listing long options and short options randomly.

- In argconfig.c, clean up help menu printing function since we really don't need it to re-parse the full list of options for the given nvme sub-command (now, we just print the char * option string constructed in nvme.c to console).

Try out the improved subcommand help menus by doing something like *nvme read --e-*.

Commit 3
--------------
add completions/*, where _nvme is our zsh tab completions and bash-nvme-completions.sh is our bash completions. Also add completions/readme, since this is a little non-intuitive to set up. Adding tab completions ought not to be done automatically, since it requires editing the ~/.+rc file for the shell in question. 

Try out both sets if you like! The zsh set provides a description for each option since it's supported by compinit, but the bash set does not. 

In zsh, hit TAB once after typing nvme to see a list of subcommands and brief descriptions, and once a subcommand has been chosen (e.g., your current line looks something like nvme read) hit TAB again to see all the options to read and brief descriptions of each.

In bash, TAB twice after typing nvme to see a list of subcommands, and once a subcommand has been chosen (e.g., your current line looks something like nvme read) you can keep hitting TAB TAB after entering each option and value to see the full options list for that command again.

### DIFF
--- a/completions/README
+++ b/completions/README
@@ -1,0 +1,60 @@
+Kelly Kaoudis, kelly.n.kaoudis at intel.com, June 2015
+
+Setting Up NVMe Tab Autocompletion for bash or zsh
+==================================================
+
+If your working shell is bash...
+--------------------------------
+the following gets bash autocompletion to behave properly
+#echo "bind 'set show-all-if-ambiguous on'" >> ~/.bashrc
+#echo "bind 'set show-all-if-unmodified on'" >> ~/.bashrc
+#echo "bind 'set completion-ignore-case on'" >> ~/.bashrc
+#echo "bind 'set completion-map-case on'" >> ~/.bashrc
+
+add NVMe autocompletion script to your autocompletes directory
+#cp `pwd`/bash-nvme-completion.sh /etc/bash_completion.d/nvme
+
+make sure this bash knows where everything is
+#source /etc/bash_completion.d/nvme && source ~/.bashrc
+
+you should be able to autocomplete with the nvme utility now
+(double TABs still apply)! If autocompleting has disappeared,
+just re-source nvme and .bashrc.
+
+You may also need to uncomment the "enable bash completion in interactive
+shells" part of /etc/bash.bashrc, that is:
+
+if [ -f /usr/share/bash-completion/bash_completion ]; then
+	. /usr/share/bash-completion/bash_completion
+elif [ -f /etc/bash_completion ]; then
+	. /etc/bash_completion
+fi
+
+(don't bother with the shopt part, your Bash version might not support shopt).
+
+if your working shell is zsh...
+-------------------------------
+create the zsh completions directory if you don't have it
+#if [ ! -e "~/.zsh" ]; then
+#	mkdir ~/.zsh
+#	mkdir ~/.zsh/completion
+#fi
+
+#cp `pwd`/_nvme ~/.zsh/completion/_nvme
+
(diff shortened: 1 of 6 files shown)


sbates130272:
Kelly

Thanks for all the work on this. Looks good to me but I am going to let Keith do the pull since he has been working on this with you.

Stephen

kaoudis:
No worries, it was fun! Sounds good to me.

Kelly

kaoudis:
Hey Keith! What might need to be changed in order for you to do a merge? :)

keithbusch:
I don't think we should be modifying the argconfig tool this much. If there are problems with that, can we make that a separate task? I think we just want to get the help strings in for each parameter so it shows useful information with this initial task.

kaoudis:
Sure, I can make a separate request for that if it makes more sense. Trouble is, just adding help strings requires the argconfig tool to not segfault. Pretty much all I've done to argconfig is replace the parts that weren't working with working code.

kaoudis:
The global options stuff that doesn't depend on argconfig can definitely be separated out, though!

keithbusch:
The segfault shouldn't be happening in the latest version of this tool AFAIK. The strcat overrun fault won't occur because the buffer should be large enough now (not a real "fix", just a stop-gap).

kaoudis:
But the buffer issue is nothing is ever put in the buffer, not that it's too small... everything is just appended to the end of the buffer into memory we don't own (since strcat doesn't copy, it just concatenates the second string to the first, and since the buffer is null terminated in the original version iirc it counts as a string)

kaoudis:
I'll open a new PR shortly.

PR #19, opened 2015-07-04
prepare Makefile for easier packaging

mihu:
The changes enable the packaging mechanism to tune the compiler and linker flags, install files to requested folders and install documentation separately.

I can provide the rest of the deb package files if there is interest in them.


sbates130272:
Hi

Thanks. Looks like the .PHONY line needs to be updated with the new install-* targets. Aside from that, this looks reasonable to me. Can you update the PR?

Stephen

mihu:
ah, silly me. It's fixed.

sbates130272:
Thanks! Pulled.

PR #21, opened 2015-07-18
add additional smart log

mihu:
derived from:
Intel Solid-State Drive Data Center P3700 Series Product Specification

Mostly copy&paste, but it seems to work:)

### DIFF
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -196,7 +196,7 @@ enum {
 };
 
 #pragma pack(push,1)
-typedef struct nvme_additional_smart_log_item {
+struct nvme_additional_smart_log_item {
 	__u8			key;
 	__u8			_kp[2];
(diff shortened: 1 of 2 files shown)


keithbusch:
Thanks for the patch. I guess it's okay to have vendor specific commands here and assume the users knows what they're doing if they run it.

As far as coding style, this project inherited the linux kernel coding guidlines, so we don't typedef structs here. Otherwise looks good to me.

mihu:
Ok, removed the typedef. Is there any spec on additional smart log items? The only device I have with NVMe is the P3700.

keithbusch:
As far as I know, we have decoding for all standards defined log pages up to the 1.2 specification. 

Vendor specific log pages have to be documented by the vendor, assuming they want that information public.

PR #23, opened 2015-07-23
Allow longer timeout for format

paul-grabinar:
Some NVMe drives take longer than 60 seconds to perform a format on a
namespace. Increase the timeout for the format to 120 seconds.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -51,6 +51,8 @@
 
 #define min(x, y) (x) > (y) ? (y) : (x)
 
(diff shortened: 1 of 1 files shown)


paul-grabinar:
On 28/07/15 18:47, keithbusch wrote:
>
> Merged #23 <https://github.com/linux-nvme/nvme-cli/pull/23>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/linux-nvme/nvme-cli/pull/23#event-367502987>.
>

Thanks Keith!

PR #25, opened 2015-08-10
Allow for firmware activation that needs a reset

paul-grabinar:
On a firmware activate, some drives require a reset before the new
firmware can be used. Such drives return a status 0x10b to the activate
request.
Rather than treat this as an error, tell this user the activation was
successful, but a reset is required.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2172,8 +2172,12 @@ static int fw_activate(int argc, char **argv)
 	if (err < 0)
 		perror("ioctl");
 	else if (err != 0)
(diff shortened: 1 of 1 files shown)


paul-grabinar:
On 10/08/15 15:34, keithbusch wrote:
>
> Merged #25 <https://github.com/linux-nvme/nvme-cli/pull/25>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/linux-nvme/nvme-cli/pull/25#event-378057825>.
>
That was fast, I didn't even get time to drop you an email!
Thanks.

PR #26, opened 2015-08-12
use simpler, less error-prone option parsing; update man pages & add man page for smart-log-add

kaoudis:
Argconfig was simultaneously more powerful and more bug-prone, even after some intense bug chasing, than needed for nvme sub-commands currently part of the application. Therefore, simply use getopt_long_only (since this was the only part of the argconfig_parse function code we were really using, anyway). 

Add a small 'help' function to replace the used parts of the argconfig_print_help function. Put the rest of the help information into the man pages, update all man pages and add one for smart-log-add (which didn't have a manpage yet).

Also, update zsh tab completion strings (completions/_nvme) since they were slightly out of sync with the man pages.

Let me know if anything should be changed :)


keithbusch:
This fell off my radar ...

I think we should stick with the parser we're using, but it'd certainly be nice if it was simpler to setup. Lots of structures to define with rigid relations among structures, and duplicating entries for short and long opts is a bit annoying. Macro magic will help.

PR #29, opened 2015-09-09
Added a decode option for all mandatory and optional Feature Identifiers fields

samiWaheed:
- Using --human-readable (or -H) option with the get-feature command will decode
the fields of the specified FID.


keithbusch:
Thanks for the patch

PR #32, opened 2015-09-21
Fixing source headers

hyandell:
Per https://github.com/linux-nvme/nvme-cli/commit/3fdbf125a2448a52a64d40dd46923ce223871aa9 - project was relicensed to GPL 2.0 after discussion with contributors. Updating source headers to reflect the change made to the LICENSE file.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5,17 +5,19 @@
  *
  * Written by Keith Busch <keith.busch@intel.com>
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


keithbusch:
It was a mistake to mark the nvme.c with the non-GPL license, but I don't think we can change the argconfig without the sign-off from the copyright holder. I think we're perfectly within license compliance just by linking two different programs the way we've done it, so we just need to fix the comment in nvme.c and we should be okay.

keithbusch:
Merged PR from original author to change the license.

PR #36, opened 2015-10-01
annotate-list-ns.py to annotate list-ns output.

ymmt2005:
Hi, team.  Thanks for the great tool.

I am willing to contribute a small script to annotate list-ns command
output for humans.  The script annotates IDENTIFY NAMESPACE
result according to the version 1.2 of the NVMe specifications.

The script works well for both Python 2.7 & Python 3.4.

The result looks like:
```
# of logical blocks: 781422768
# of logical blocks after format: 781422768
# of logical blocks in use: 0
Support for deallocated or unwritten logical block error: False
Support for NAWUN/NAWUPF/NACWU: False
Support for thin provisioning: False
# of LBA formats: 4
Metadata at the end of LBA: True
LBA format ID (LBAF): 0
Metadata transfer using a separate buffer: True
Metadata transfer using extended data LBA: True
Support for PI transfer at the tail of metadata: True
Support for PI transfer at the head of metadata: True
Support for protection information Type 3: True
Support for protection information Type 2: True
Support for protection information Type 1: True
Protection information is not enabled.
Shared namespace: No
Reservation is not supported.
Support for format progress indicator: False
% remains to be formatted: 0
Namespace Atomic Write Unit Normal (NAWUN): 0
Namespace Atomic Write Unit Power Fail (NAWUPF): 0
Namespace Atomic Compare & Write Unit (NACWU): 0
Namespace Atomic Boundary Size Normal (NABSN): 0
Namespace Atomic Boundary Offset (NABO): 0
Namespace Atomic Boundary Size Power Fail (NABSPF): 0
NVM capacity in bytes: 0
Namespace GUID: 00000000000000000000000000000000
Namespace EUI64: 0000000000000000
LBAF0
    Relative performance: Good
    LBA data size: 512
    Metadata size: 0
LBAF1
    Relative performance: Degraded
    LBA data size: 512
    Metadata size: 8
LBAF2
    Relative performance: Best
    LBA data size: 4096
    Metadata size: 0
LBAF3
    Relative performance: Better
    LBA data size: 4096
    Metadata size: 8
```

### DIFF
--- a/scripts/annotate-list-ns.py
+++ b/scripts/annotate-list-ns.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
+import re
+import sys
+
+# Constants
+HELP_DESCRIPTION = '''\
+Annotate list-ns output for humans.
+The output should be given via stdin like this:
+
+    sudo nvme list-ns -n 1 | annotate-list-ns.py'''
+RE_LINE = re.compile(r'^\[\s*(\d+)\]:0x([0-9a-fA-F]+)$')
+RELATIVE_PERFORMANCE = {
+    0: 'Best', 1: 'Better', 2: 'Good', 3: 'Degraded',
+}
+
+
+# Functions
+def parse_input(file):
+    data = [0] * 1024
+    for line in file:
+        m = RE_LINE.match(line)
+        if m is None:
+            sys.exit('Invalid input: ' + line)
+        n = int(m.group(1))
+        v = int(m.group(2), base=16)
+        data[1023-n] = v
+    return ''.join(format(i, '032b') for i in data)
+
+
+def decode(data, byte_offset, bit_offset, nbits):
+    pos = 1024 * 4 * 8 - byte_offset*8 - bit_offset - nbits
+    return int(data[pos:(pos+nbits)], base=2)
+
+
+def annotate(data):
+    '''Annotate list-ns output according to NVMe spec 1.2 Figure 92 & 93.'''
+    print('# of logical blocks:', decode(data, 0, 0, 8*8))
+    print('# of logical blocks after format:', decode(data, 8, 0, 8*8))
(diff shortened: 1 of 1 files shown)


keithbusch:
It's been a while since I've read python code, but I think the usage is wrong, or your controller doesn't actually support the command correctly.

 ` nvme list-ns -n 1 | annotate-list-ns.py`

The "`list-ns`" sub-command shouldn't do what the program thinks it does. It should return an array of namespace identifiers instead.

I actually can't even figure out what input the program is expecting in order for this to work. :)

The output example you've provided also looks similar to the "`id-ns`" subcommand when using the verbose option "`--human-readable`". Maybe we can improve that output instead of bringing in a new script?

ymmt2005:
@keithbusch 
What on earth?

This is what I'm getting from `nvme list-ns -n 1`:
```
ubuntu:~$ sudo ./nvme list-ns /dev/nvme0 -n 1
[   0]:0x2e9390b0
[   2]:0x2e9390b0
[   6]:0x3100300
[   7]:0x1f
[  32]:0x2090000
[  33]:0x3090008
[  34]:0xc0000
[  35]:0x10c0008
[  97]:0xba4e42
[  98]:0x1
```

What went wrong?

ymmt2005:
Ah, you suggested, my controller is to blame...

Thank you.  I am happy to use `id-ns --human-readable`.

keithbusch:
Aha, yeah, your controller doesn't support the list namespace command. It is interpreting the command as an identify namespace. I have a controller like that as well. The list namespace command should return an array of namespace identifiers from a subsystem, and they should be ascending order. I have one controller with attached namespaces with id's 1, 3, 4, and 6. The output looks like this:

```
# nvme list-ns /dev/nvme0 -n 0
[   0]:0x1
[   1]:0x3
[   2]:0x4
[   3]:0x6
```

See NVMe specification section 5.11, figure 86 for details. The list-ns uses CNS value 02h, but a lot of 1.0 controllers don't interpret this correctly.

ymmt2005:
Thank you.  Closing.

PR #38, opened 2015-12-03
Factor out ID printing code into common.c

pmmccorm:
The printing output should be identical. The id structure
printing functions can be more or less used by lsnvme.

Signed-off-by: Patrick McCormick <patrick.m.mccormick@intel.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,14 +26,16 @@ NVME-VERSION-FILE: FORCE
 -include NVME-VERSION-FILE
 override CFLAGS += -DNVME_VERSION='"$(NVME_VERSION)"'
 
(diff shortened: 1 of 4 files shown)


keithbusch:
Looks fairly uncontroversial to me, though I think we should prefer flags defined as an enum. Do you have a planned usage for the terse flag?

pmmccorm:
I actually had flags defined as an enum and then changed it. I can change it back.

For terse I was thinking:
```
lsnvme -v : terse output
lsnvme -a : entire struct
lsnvme -ah: entire struct, human readable
```

Something like that.

pmmccorm:
Updated with enum change.

PR #44, opened 2015-12-16
Don't cast pointers, cast values.

wdoekes:
The compiler rightly warned about an issue in ed3848130, but the fix was
wrong.

There is no guarantee that an int is exactly 32 bits and no guarantee
that a long long is exactly 64 bits. The former one is guaranteed to be
at least 16 bits (but we can safely assume 32 for non-ancient machines).

Imagine an u32 array:

    0... 1... 2... 3... 4... 5...

Now cast the pointer to the zeroth element to a 64-bits-int-pointer;
64bit sized int are quite common.

    0.......  1.......  2.......

Writing to 0, 1 and 2 would put values in 0, 2 and 4 (on little endian).
And if there is enough data, it would overwrite the stack.

This change moves the casting to the point where the individual values
are copied. That's safe.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -52,6 +52,7 @@
 #include "src/argconfig.h"
 #include "src/suffix.h"
 
+#define array_len(x) ((size_t)(sizeof(x) / sizeof(0[x])))
 #define min(x, y) (x) > (y) ? (y) : (x)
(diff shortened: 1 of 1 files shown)


keithbusch:
Thanks for the fix!

PR #50, opened 2016-01-21
Fixing Memory Leak

lakshminarayanand:
In submit_io  buffers "buffer and mbuffer" allocated are not getting freed at the fail cases and hence might lead to memory leak 
This patch free the buffer before the return of the function .


keithbusch:
Being a user space application that typically executes on the order of milliseconds, we got a bit lazy with resource management and let the operating system reclaim memory and file descriptors when the program closes.

Cleaning up that sloppiness is a good idea. Maybe this program will add an interactive mode in the future where resource management will be more important.

PR #51, opened 2016-01-21
Memory Leak Fix at nvme_passthru

lakshminarayanand:
In nvme_passthru  buffers "data and metadata" allocated are not getting freed at the fail cases and hence might lead to memory leak 
This patch free the buffer before the return of the function .


keithbusch:
This one has formatting errors and missing braces on an if condition. I fixed and merged this by hand, so closing the pull request.

PR #56, opened 2016-02-07
Add starter doc for AlpineLinux

FlorianHeigl:
I've included two command examples, i hope that's not too much bloat. normally people will try to avoid using udev, but it's not really necessary either assuming everyone can ls /dev/nvme* ;)


sbates130272:
Thanks! The udev thing is well known. It is a side effect of how we do nvme list. At somepoint we might change how we do nvme list and this issue will resolve.

PR #60, opened 2016-02-26
Splitting responsibilities

keithbusch:
This looks like a large change, but it's really just moving where things are defined. This splits the ioctl submission and structure printing into their own files and should make it easier to maintain as separate entities.

There shouldn't be any functional change here, but sending a pull request for potential second opinions.

Last patch just cleans up whitespace damage and irking formatting issues. Maybe we should have a reference to a preferred coding style. I initially adopted the linux kernel's, but we don't necessarily have to do that.


sbates130272:
Looks good. I am keen to stick to kernel coding style though I am not sure the libs we use for the command line parsing conform to that. For the nvme-* files though we can push for that style.

keithbusch:
Speaking of the parsing component, I have a patch that makes it much simpler to define long and short options together. Right now it's only for the nvme version of this. Would you prefer these go to the original arg-parse project instead of continuing the nvme specifics we currently have?

sbates130272:
Keith - Feel free to just change it in nvme-cli. The master repo for this is diverged quite a bit at this point.

keithbusch:
In that case, any qualms if I change the indent style to match?

sbates130272:
No. Please do ;-).

keithbusch:
Closing request, applied directly.

PR #63, opened 2016-03-02
Convert build system to autotools.

### LINKED ISSUE #59: add autoconf/autotools ?
We all loathe autotools for good reasons, but it is does provide some nice features that nvme-cli will want or already implements itself:

1) easy cross compilation
2) configurable locations for everything (prefix, destdir, sysconfdir)
3) detection of libraries (libudev), headers (nvme.h or nvme-ioctl.h), and struct members
4) handles version substitution in source files

Out of all these I think number three can provide us some real benefits: we can generate defines for certain fields in struct nvme_id_ctrl, and what ioctls are available (eg use for #57).

Is this something people are interested in? I can issue a PR if so.

pmmccorm:
A pretty straight-forward conversion, fixes #59. The Documentation
makefile is left alone (we just ship the manpages in there and don't
do any processing at build/install time).

Change details:
- add local copy of nvme_ioctl.h (from kernel 4.5 rc6), remove local
  ioctl defs from nvme.h. Use autoconf to prefer system header, fall
  back to internal header.
- Rename and update rpm spec file to work with autotools and other
  misc cleanups. Tested on RHEL7.0.
- Debian packaging needs some tweak to work still.

Signed-off-by: pmmccorm patrick.m.mccormick@intel.com


pmmccorm:
@leitao This commit almost certainly breaks the debian packaging targets; I don't know a single thing about debian packaging so I just left them as is (hoping you can follow this up with a (hopefully) simple fix).

pmmccorm:
Rebased on latest HEAD, added a "--enable-debug" option to configure script.
@keithbusch , @sbates130272 does this look OK?

keithbusch:
HI Patrick, looks okay to me. I'll pull this in locally and give it a better look next Monday.

Just a question, though: are there no pattern matching rules for the man pages? It'd be nice not to have to individually specify them. One less thing to remember when adding a new command.

sbates130272:
Looks good to me.

pmmccorm:
I actually tried this (using $(wildcard ..) and autogen.sh fails. The problem is Makefile.am isn't a gnu Makefile, it is a template that gets translated by automake. Some googling turned up:

https://www.gnu.org/software/automake/manual/html_node/Wildcards.html

Which explains why I have never seen wildcards or pattern substitution targets in Makefile.am.

pmmccorm:
Rebased on HEAD and fixed merge conflict.

pmmccorm:
Rebased and updated gitignore to catch the autools files a little better.

pmmccorm:
Rebased on latest. @keithbusch Did you get a chance to try this out?

keithbusch:
I did, and seems good to me. I wanted to tag a new release before pulling anything drastic since I messed up the previous.

Is a "make clean" supposed to leave uncommitted artifacts? I find there are several untracked files in the repository after trying to clean a built project.

pmmccorm:
OK great, I'll keep rebasing it periodically.

I'm not sure on the "make clean": it is auto-generated. "make distclean" might be what you want: it removes more (configure script, Makefile, etc).

pmmccorm:
Rebased on latest master.

pmmccorm:
Rebased to latest master and fixed merge conflict. @keithbusch, are there any more issues/changes/questions with this PR?

keithbusch:
Okay, sorry this is taking me so long, I'm a bit paranoid about changing anything drastic anymore.

I think this is pretty much good to go, as far as I can tell. I'm not sure how this affects the packaging targets for some of the distro support, though, so I still need to look into that.

Before we can apply this, we also do need to update the README.md so it has the correct build information for automake method.

pmmccorm:
I can include the updated README.md in this PR that is probably where it belongs.

pmmccorm:
Updated the REDME.md and included. I also cleaned up some of the misc usage examples that were scattered throughout the distro support section. Made a note about libudev support.

pmmccorm:
Rebased on latest. I'm going to stop tracking this. When/if you are ready to merge ping this PR and I'll rebase again, or close this PR if you don't intend to merge.

PR #67, opened 2016-03-04
Fixed option checking for format to match NVMe spec limits.

braschul:
I noticed the format command help text text was a little off from the spec and the range checking was also not respecting the spec defined values.  Updated the help text and option range checking.  Please let me know if these changes are out of line.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1388,7 +1388,8 @@ static int format(int argc, char **argv)
 
 	get_dev(1, argc, argv);
 
-	if (cfg.ses > 2) {
+	/* ses & pi checks set to 7 for forward-compatibility */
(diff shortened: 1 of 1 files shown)


braschul:
Ok - that's fair.  Mind if I put a comment in those checks to say as much?  Any comments on my help-text changes?

keithbusch:
The help text update looks good to me. Also okay on proposed code comments.

braschul:
Ok - should be good now?

keithbusch:
Looks fine to me. I'll pull the series. Normally we'd ask this to be a single commit to keep the logs clean, but I don't have the moral backing to do so based on my own commit history. :)

PR #77, opened 2016-03-22
Fix the bug of Model Number and display the info of Serial Number

seehearfeel:
According to the NVMe Specification, there is an error when display Model Number.
In addition, currently it does not display the info of Serial Number, so add it.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -579,7 +579,7 @@ void show_nvme_id_ctrl(struct nvme_id_ctrl *ctrl, unsigned int mode)
 	printf("vid     : %#x\n", le16toh(ctrl->vid));
 	printf("ssvid   : %#x\n", le16toh(ctrl->ssvid));
 	printf("sn      : %-20.20s\n", ctrl->sn);
-	printf("mn      : %-20.20s\n", ctrl->mn);
+	printf("mn      : %-40.40s\n", ctrl->mn);
 	printf("fr      : %-.8s\n", ctrl->fr);
 	printf("rab     : %d\n", ctrl->rab);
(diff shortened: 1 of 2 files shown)


keithbusch:
Oh no, why did we write it with hard-coded lengths instead of a `printf("%.*s\n", (int)sizeof(field), field)` Thanks for the catch.

I don't know if we want to include the serial number in the list output, though. That output is already too long for most terminal displays. We should be reducing the information it prints since it's unnecessarily duplicated across multi-namespace devices.

seehearfeel:
Thanks for your reply.

In my opinion, it is a fixed value of (int)sizeof(field) due to (int)sizeof(list_item.ctrl.mn) is always 40, because there is a member declaration:
char            mn[40];
in the struct nvme_id_ctrl at the line 66 of linux/nvme.h file.

Generally, the Serial Number is a basic and important info for a disk like Model Number and Firmware Revision. So I think there is no need to use (int)sizeof(field) and display the Serial Number is necessary.

keithbusch:
Right, understood. I just didn't want to have to write "40" for each time we want to print it (or "20" for mn, and "8" for fr, for that matter). We already wrote the size into the structure, so why not let the print format specifier figure out the length on it's own? The linux kernel prints these fields in sysfs that way.

seehearfeel:
OK, I see, thanks for your suggestion.

I have modified the related code of Serial Number, Model Number and Firmware Revision,
other fields are not be changed because the length can not be properly figured out by
the print format specifier, (int)sizeof(field) is too big and (int)strlen(field) is random.
In addition, fix a minor spelling mistake "Namepace" to "Namespace".

Please review, thanks!

PR #78, opened 2016-04-06
remove wrong tail (%)

ywang19:
Signed-off-by: ywang19 <yaguang.wang@intel.com>

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -807,7 +807,7 @@ void show_intel_smart_log(struct nvme_additional_smart_log *smart, unsigned int 
 	printf("timed_workload_media_wear       : %3d%%       %.3f%%\n",
 		smart->timed_workload_media_wear.norm,
 		((float)int48_to_long(smart->timed_workload_media_wear.raw)) / 1024);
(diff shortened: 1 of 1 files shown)


nvme-print.c
```diff
@@ -807,7 +807,7 @@ void show_intel_smart_log(struct nvme_additional_smart_log *smart, unsigned int
 	printf("timed_workload_media_wear       : %3d%%       %.3f%%\n",
 		smart->timed_workload_media_wear.norm,
 		((float)int48_to_long(smart->timed_workload_media_wear.raw)) / 1024);
-	printf("timed_workload_host_reads       : %3d%%       %lu%%\n",
+	printf("timed_workload_host_reads       : %3d%%       %lu%\n",
```
keithbusch: This is a percentage value, right? Why do you want to remove the percent sign? And if you want to remove it, you need to remove both "%%" instead of just one, yeah?

ywang19: yeah, should remove "%%". removed them below.

keithbusch:
Thanks, could you squash this into a single commit, and send new pull request?

PR #79, opened 2016-04-12
Rename nvme_log() to nvme_get_log()

minggr:
This kills off the old nvme_get_log() and rename
nvme_log() to nvme_get_log().

Signed-off-by: Ming Lin <ming.l@ssi.samsung.com>


keithbusch:
HI Ming, I'm still considering whether or not to apply this. Having the second API accepting the full CDW10 field was for convenience to extensions testing features in current TPs under review that use reserved bits in CDW10.

keithbusch:
Alright, there's a pure pass thru command if anyone needs to use reserved bits. I'll merge this new log API, and just change it again if reserved bits get defined for specific sub-commands. Thanks for the patch.

PR #83, opened 2016-05-06
nvme.spec.in: Add a build/install requirement on systemd

aakefbs:
systemd and systemd-dev include libudev on RHEL7 based
distributions. I'm not sure about SLES12, if SLES12 should
have different dependencies, we would need to add a distribution
detection and different requirements for RHEL and SLES.


morbidrsa:
On SLES we build without the systemd requirement but have a BuildRequires: pkgconfig(libudev) which should have the same effect.

keithbusch:
I usually generate rpm's from a non-RH based distribution, too. It works today, but it'd fail with such build requirement.

PR #87, opened 2016-05-24
Add LightNVM adminstration and diagnose support

MatiasBjorling:
Add support for identify geometry, get bad block interface and
administration interface for LightNVM. This patch adds the following
commands:

 - (lnvm-id-ns) Identify geometry of a LightNVM device.
 - (lnvm-diag-bbtbl) Retrieve bad block table from LightNVM device.
 - (lnvm-info) Get information about modules loaded, and available
               devices.
 - (lnvm-init) Initialize LightNVM device with a media manager.
 - (lnvm-create) Create targets on top of registered devices.
 - (lnvm-remove) Remove a registered target.
 - (lnvm-factory) Factory initialize a LightNVM device.

Signed-off-by: Matias Bjørling <m@bjorling.me>

### DIFF
--- a/Documentation/nvme-lnvm-create.txt
+++ b/Documentation/nvme-lnvm-create.txt
@@ -1,0 +1,60 @@
+nvme-lnvm-create(1)
+===================
+
+NAME
+----
+nvme-lnvm-create - Instantiate a target on top of a LightNVM enabled device
+
+SYNOPSIS
+--------
+[verse]
+'nvme lnvm-create' [--device-name=<DEVICE> | -d <DEVICE>]
+			[--target-name=<TARGET> | -n <TARGET>]
+			[--target-type=<TARGETTYPE> | -t <TARGETTYPE>]
+			[--lun-begin <NUM> | -b <NUM>]
+			[--lun-end <NUM> | -e <NUM>]
+
+DESCRIPTION
+-----------
+Instantiate a target on top of a LightNVM enabled device. This exposes the
+physical for the user to use.
+
+The target name is the name of which the media is exposed as in
+/dev/<targetname>
+
+The target type is the target to be instantiated. Typically pblk or rrpc.
+
+LUN begin and end defines the range of LUNs to use for a target instantiation.
+
+OPTIONS
+-------
+--device-name=<DEVICE>::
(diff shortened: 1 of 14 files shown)


keithbusch:
Do you want to add some man pages to the documentation? If you just want to write them in asciidoc, I can generate the rest. Or if you want to skip adding docs, that's fine too.

In the very near future, I actually do want to have vendor, command set, and protocol specifics go into their own class of sub-commands. For example, anything Intel specific would be under "`nvme intel <command>`", lightnvm can be "`nvme lnvm <command>`", etc.... so that the top level help from "`nvme help`" won't pollute the screen with things that don't apply to the nvme core specification.

So, we can merge this sooner and deprecate it later, or we can try to create an infrastructure  for command groups (for lack of a better term).

MatiasBjorling:
Let me fix up the asciidoc. I would love to have it merged first, and then we can work out the refactoring of the pieces.

I started doing the nvme lnvm <command> layout, but stopped as it grew in complexity. :)

MatiasBjorling:
Ok, added all the .txt files for the lnvm-* commands. :)

and also got the set bb tbl command integrated.

keithbusch:
Cool, looks good.

MatiasBjorling:
Thanks Keith

keithbusch:
On Wed, Jun 01, 2016 at 12:14:52PM -0700, Matias Bjørling wrote:
> Let me fix up the asciidoc. I would love to have it merged first, and then we can work out the refactoring of the pieces.
> 
> I started doing the nvme lnvm <command> layout, but stopped as it grew in complexity. :)

Just FYI, if you haven't seen recently, I did move the lnvm part to
an extension, though I preserved backward compatibility on command
execution as you originally wrote it.

The main difference is that all lnvm commands will not appear in top level
"nvme help". You will see an entry for "lnvm" as a plugin, and "nvme
lnvm help" will show all the lnvm sub-commands. I hope that works for you.

It was kind of getting complicated, but I think I have a fairly decent
design that makes this easy and maintainable, but would be happy to hear
you've any other thoughts or ideas on it.

MatiasBjorling:
Thanks Keith. Just checked it and it looks good.

PR #91, opened 2016-06-03
 Fix leak of mbuffer on error exit path

ColinIanKing:
building with clang scan-build picked up a memory leak:
nvme.c:2176:3: warning: Potential leak of memory pointed to by 'mbuffer'
   fprintf(stderr, "failed to read data buffer from input file\n");
   ^~~~~~~

instead of returning, jump to free_and_return which performs the
correct cleanup.  Also fix the indentation of the return at the end of
the function.

Signed-off-by: Colin Ian King <colin.king@canonical.com>


keithbusch:
We've been lazily letting exit clean up all resources since it is just a one-shot program. But agree we should clean these up.

PR #92, opened 2016-06-21
Enhanced both reset and subsystem-reset to return an error message

samiWaheed:
- Enhanced both the reset and subsystem-reset functions to return an error message on older kernels with no ioctl support.  Current behavior is to return general_help() message with no indication that the ioctl call failed.
- Added "zeroes based" to the block count documentation for write-uncorr and write-zeroes

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1277,14 +1277,26 @@ static void clear_args(int argc, char **argv)
 
 static int subsystem_reset(int argc, char **argv, struct command *cmd, struct plugin *plugin)
 {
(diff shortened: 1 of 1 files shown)


keithbusch:
Thanks for the updates. Will apply.

In the future, we should strive to be more disciplined about splitting different fixes into different patches (the help text for zeroes based values is separate from the ioctl errors).

samiWaheed:
Will do for future patches.  Thx.

PR #94, opened 2016-06-24
Added "-s" option to the set-feature command to save the specified feature

samiWaheed:


keithbusch:
Two things: please add this to the man page, and change the type to "int". no_argument requires int type, and bool is the wrong size for that. I found out the hard way that those parameters were corrupting memory before changing them all to ints...

keithbusch:
Thanks for fixing the set feature man page. I'm not sure how to squash these through the github interface, so we'll just pull this as two commits.

PR #99, opened 2016-07-05
Enhanced error reporting for invalid commands and/or extensions. 

samiWaheed:
List reason why command line failed before displaying the usage message


keithbusch:
Looks good.

PR #102, opened 2016-07-13
Fix read/write memory corruption issues

hrasiq:
When issuing a read to the device, if we pass a data size smaller than the
current sector size, the ioctl() will overflow the data buffer. This will
also happen when reading more than one block from the device, if the buffer
does not have enough bytes to store the result. This patch adds a simple
check using the requested block count to correctly allocate a buffer of
minimum size.

Signed-off-by: Heitor Ricardo Alves de Siqueira <halves@linux.vnet.ibm.com>


keithbusch:
Yeah, it's a bit too easy to break things using the NVME_IOCTL_SUBMIT_IO. The interface assumes too much with no good way to verify the user's allocated buffer size is appropriate from the kernel. So helping out from user space is a good idea. Thanks for the fix.

PR #108, opened 2016-08-22
nvme-cli: Implement list command that doesn't rely on libudev

ScottyBauer:
The previous list command required libudev. We can accomplish the same
thing with standard functions, and in the future remove the libudev
dependency

### DIFF
--- a/Documentation/nvme-list.1
+++ b/Documentation/nvme-list.1
@@ -38,7 +38,6 @@ nvme-list \- List all recognized NVMe devices
 .sp
 Scan the sysfs tree for NVM Express devices and return the /dev node for those devices as well as some pertinent information about them\&.
 .sp
-This command requires udev\&.
 .SH "OPTIONS"
 .sp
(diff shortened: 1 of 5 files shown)


keithbusch:
Awesome, this will fix several users' concerns.

I'm thinking we can drop the libudev optional dependency entirely and have only your method so that binaries can work across different systems.

The man page in Documentation/nvme-list.txt declares the udev dependency, which we can drop with this fix.

Finally, a possibly cleaner method might use 'scandir' to collect dirents instead of implementing looping over entries on your own, but no big deal.

ScottyBauer:
Alrighty, 

Removed the ifdefs from nvme.c and removed the udev stuff from the documentation. There is still use of udev code in fabrics. c so we cannot remove the libudev checks in the make file yet.

Tested it on cent OS 7 vm with one device.

keithbusch:
This looks good to me. I'll give it a test today as well and merge if no problems.

PR #111, opened 2016-08-29
purge `readdir_r` -- which is deprecated in recent glibc.

iori-yja:
This commit allows nvme-cli to be built with recent gcc/glibc (6.1.1/2.24).


keithbusch:
Thanks, pulled.

PR #117, opened 2016-09-19
Implement json output for list that matches 'isdtc show -output json --intelssd

ScottyBauer:
…--intelssd'

    New command:
    nvme list -json -intelssd or
    nvme list -json  or
    nvme list

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ NVME_DPKG_VERSION=1~`lsb_release -sc`
 
 OBJS := argconfig.o suffix.o parser.o nvme-print.o nvme-ioctl.o \
 	nvme-lightnvm.o fabrics.o json.o plugin.o intel-nvme.o \
-	lnvm-nvme.o memblaze-nvme.o
+	lnvm-nvme.o memblaze-nvme.o nvme-models.c
 
 nvme: nvme.c nvme.h $(OBJS) NVME-VERSION-FILE
 	$(CC) $(CPPFLAGS) $(CFLAGS) nvme.c -o $(NVME) $(OBJS) $(LDFLAGS)
--- a/nvme-models.c
+++ b/nvme-models.c
@@ -1,0 +1,96 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <string.h>
+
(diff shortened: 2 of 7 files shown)


ScottyBauer:
The only difference between this and isdct is we won't report  "Bootloader is up-to-date ..." That seems strange to be looking for that in this tool. If we really need it I can go ahead and put it in.

ScottyBauer:
I'll have to re-word the commit it should be -intel not -intelssd.

Anyway, another thing to ponder before merging is do we want to keep -json or use the -output-format JSON that the rest of the code base uses?

keithbusch:
Don't worry about the bootloader up-to-date part. That only works if the firmware and bootloader are built into the program, which is not the case for the open tooling. We are good to just display what the device returns.

Let's also remove the "intel" specific list option. If we need an intel specific one, we'll put it in the intel vendor plugin.

As for reporting models would it be feasible to pull these out of the pci.ids file? This is how 'lspci' does it, and would be convenient to use so we don't need to update multiple repositories with model name decoding. It may be a bit more work, though, since it isn't straight forward to get the SVID:SSID from a block handle.

jderrick:
Those should be in /sys/class/nvme/nvme0/device/subsystem_{device,vendor}

ScottyBauer:
Thanks Jon, that helped a bunch.

Keith, with the removal of the intel flag, can we rely on the bootloader and the health status being in the same location for devices manufactured by other companies? I'm essentially asking if I need to remove that whole chunk of code and place it in the vendor plugin?

keithbusch:
The health and bootloader are vendor specific and don't need to be reported in the generic version of this command.

ScottyBauer:
Changes from V1->V2:

Added Function to pull out controller names via /usr/share/hwdata/pci.ids

output looks like:
{
  "PCIe Nvme  DC P3700 SSD CVFT516500..etc":
        etc
}

It does not include the Manufacturer ie, Intel, Samsung, etc. We *can* get that info from pci.ids, but it adds a little more complexity to the parsing code due to re-use of manufacture numbers within the file.

ScottyBauer:
I'd like to get this branch in a solid state, I have a bunch of other work that is branched on top of this one. Once this is stable I can rebase all that other work and get it here for review.

ScottyBauer:
v2->v3 
Rebased ontop of Christoph's changes. Changed  from -json to the more standard -output-format json|normal that is used throughout the file.

keithbusch:
This feature was pulled in a different request; closing.

PR #121, opened 2016-09-26
Jsonify SMART/SMART-ADD and fw-info slots

ScottyBauer:
This branch and patch is based on top of pull request:
https://github.com/linux-nvme/nvme-cli/pull/117

Once that is merged this diff should be cleaned up and can then be reviewed.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -719,15 +719,6 @@ static void get_registers(struct nvme_bar **bar)
 	*bar = membase;
 }
 
-#define MAX_LIST_ITEMS 256
-struct list_item {
-	char                node[1024];
-	struct nvme_id_ctrl ctrl;
-	int                 nsid;
-	struct nvme_id_ns   ns;
-	unsigned            block;
-};
-
 static void print_list_item(struct list_item list_item)
 {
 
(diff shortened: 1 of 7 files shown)


keithbusch:
Can we have the list command use the argconfig parsing? We need all commands to have consistent help messages, and argconfig provides that.

ScottyBauer:
It's currently using argconfig, I just forgot to provide a help message. I'll fix that now.

ScottyBauer:
It's fixed.

./nvme list -help works

I've removed the 2 other pull requests since we don't have to support json for those commands.

Here are the commands that have modifications:
nvme list -output-format json
nvme error-log /dev/nvme0 -output json
nvme smart-log /dev/nvme0 -output json
nvme smart-log-add /dev/nvme0 -output json

ScottyBauer:
forgot this command:
nvme fw-log /dev/nvme0 -output-format json

nvme.c
```diff
@@ -819,6 +810,27 @@ static int list(int argc, char **argv, struct command *cmd, struct plugin *plugi
...
+	const struct argconfig_commandline_options opts[] = {
+		{"output-format", 'o', "FMT",
+		 CFG_STRING, &cfg.output_format,
+		 required_argument, "Output Format: normal|json"},
+		{NULL}
+	};
+	argconfig_parse(argc, argv, desc, opts, &cfg, sizeof(cfg));
+	fmt = validate_output_format(cfg.output_format);
+
+	if (fmt != JSON || fmt != NORMAL)
```
keithbusch: Should be && instead of || if checking for an error condition; the format can't possibly be both. :)

ScottyBauer: pesky boolean logic, thanks.

ScottyBauer:
Okay give this a quick run. I tested on your server and my local box.  There is a built version in /tmp on the server if you don't want to pull and build.

PR #137, opened 2016-10-26
Implementation for SED/Opal 

ScottyBauer:
Don't worry about pulling/reviewing this yet, Keith. Just getting everything staged.

This will change a few times as upstream potentially has us modify API. Once sed has landed then we can review this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ NVME_DPKG_VERSION=1~`lsb_release -sc`
 
 OBJS := argconfig.o suffix.o parser.o nvme-print.o nvme-ioctl.o \
 	nvme-lightnvm.o fabrics.o json.o plugin.o intel-nvme.o \
-	lnvm-nvme.o memblaze-nvme.o nvme-models.o
+	lnvm-nvme.o memblaze-nvme.o nvme-models.o sed.o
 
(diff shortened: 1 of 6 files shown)


cristim:
I think this can be reopen now, the sed patches landed a couple of months ago.

whirm:
Any plans on getting this functionality merged?

ScottyBauer:
No. Sed-opal is storage agnostic and doesn't belong in nvme-cli. https://github.com/Drive-Trust-Alliance/sedutil/issues/90  is in the midst of incorporating OPAL_SAVE into their tooling, https://github.com/ScottyBauer/sed-opal-temp/ is another set of tooling you can use as well.

whirm:
@ScottyBauer great thanks! Is sed-opal-temp functional? Does it support suspension?

ScottyBauer:
@whirm Yes, it will allow you to set a key in the kernel so you can unlock from S3. There are some examples in scripts/ on how to setup your drive, if it's not already opal-enabled. If you setup your drive via DTA sed-util you'll have to send the _hashed_ password via the sed-util tooling.

whirm:
@ScottyBauer I'll have a look ASAP. Thanks again!

PR #144, opened 2016-12-09
JSON format fixes

apage43:
Some fixes for issues prevented the list output's JSON format being parseable in some cases:

* Fix the 'Firmware' field in the JSON output mode possibly reading past the end of the `.fr` field in the `nvme_id_ctrl` struct.
* Print the list as a JSON array of objects (as the "Devices" key of a root JSON object) rather than an object with the product name as the key, to avoid the issue of duplicate keys when two cards have the same product name
* Fix some issues right-trimming field whitespace

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -63,14 +63,15 @@ void d_raw(unsigned char *buf, unsigned len)
 static void format(char *formatter, size_t fmt_sz, char *tofmt, size_t tofmtsz)
 {
 
-	snprintf(formatter,fmt_sz, "%-*.*s",
+	fmt_sz = snprintf(formatter,fmt_sz, "%-*.*s",
 		 (int)tofmtsz, (int)tofmtsz, tofmt);
 	/* trim() the obnoxious trailing white lines */
-	while (--fmt_sz) {
+	while (fmt_sz) {
 		if (formatter[fmt_sz - 1] != ' ' && formatter[fmt_sz - 1] != '\0') {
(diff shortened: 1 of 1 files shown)


keithbusch:
Hmm ... your NVMe vendor is blatantly violating the spec if you've got strings with such characters. From NVMe 1.2.1, section 1.5:

"Some parameters are defined as a string of ASCII characters. ASCII data fields shall contain only code
values 20h through 7Eh."

I'm not sure we should encourage such behavior. You should file a bug with your vendor.

apage43:
! Thanks, @keithbusch. I checked over things again, and saw that I didn't see that in the non-JSON output mode. Turns out my device wasn't broken, the JSON output mode could just read past the end of the field. Removed the escaping commit since it shouldn't be needed and added a fix for that.

keithbusch:
Ah, okay, looks like the code failed to truncate. That part looks fine then.

On the printing, I am hoping not to change the json code since we "borrowed" that from fio. Can we leave the root as an object, and instead append an array as the top level value? The device name can just be an attribute instead of the object name if that's causing trouble.

apage43:
@keithbusch -- Can do that. I did originally did that when testing this so that I didn't have to modify `json.c`. This patch does already move devicename into the object as an attr so that we still have it in the output.

apage43:
Removed commit that edited `json.c`, now printing the array as the `Devices` key of the root level object.

keithbusch:
Looks good to me, merge applied.

PR #145, opened 2016-12-14
Necessary to decode the status code of reservation conflict to the string

yongseokoh:
Translating the status code of reservation conflict (e.g., 0x83) to "RESERVATION_CONFLICT" is very useful when we make use of reservation support NVMe SSDs.


keithbusch:
You've got an error in the string decode with a misspelled function such that it doesn't compile. I'll fix it up and push the commit.

keithbusch:
Pushed commit with your attribution; closing.

PR #152, opened 2016-12-22
There are two block-count parameter

andyandroid:
There are two block-count parameter, change the last one to block-amount. Change write to read


keithbusch:
I see the man page is a little messed up, but the patch here doesn't really fix the issue since the program doesn't accept a parameter named "block-amount". Just delete the first and incorrect instance of the block-count parameter and change write to read, yeah?

keithbusch:
Fixed here: https://github.com/linux-nvme/nvme-cli/commit/31d34552f3fdfcbca2c18ea2753f1ad805ef1439

PR #155, opened 2017-02-01
get-features: Don't try to read the data buffer for sel == 3

amluto:
When sel == 3, the controller returns capabilities in the completion
entry and does not produce a feature value blob.  Stop allocating
space for the blob and displaying it -- the result is garbage.

Signed-off-by: Andy Lutomirski <luto@kernel.org>


keithbusch:
Thanks!

I think we may want decode this select method differently as well to make it a little easier to read the result as a bit mask and not the value of the feature.

PR #157, opened 2017-02-06
Moving the bash completion files to the standard

leitao:
Currently it is not recommend to put the bash completion files inside
/etc, but on the new /usr/share/bash-completion, as stated in :

https://lintian.debian.org/tags/package-installs-into-obsolete-dir.html
https://bugs.debian.org/776954

Signed-off-by: Breno Leitao <breno.leitao@gmail.com>


leitao:
Thanks @keithbusch . Version 1.1 uploaded in Debian already.

PR #158, opened 2017-02-15
lightnvm: default to -1 on lun begin and end

MatiasBjorling:
The default behavior when initializing a target without lun end and
begin is to use only the first lun of the OCSSD. Going forward, the
user shall either explicit specific the begin and end, or if not defined
for kernel 4.11 and newer, the full SSD is initialized.

Signed-off-by: Matias Bjørling <matias@cnexlabs.com>


MatiasBjorling:
Hi Keith :) Will you pick up this small fix. I got tired of writing -b -e on each creation command.

PR #162, opened 2017-03-02
fabrics: Handle space-padded TRSVCID and TRADDR fields

rolandd:
The TRSVCID and TRADDR fields in the discovery log page are defined
as ASCII strings, which according to the NVMe standard means they
should be space-padded rather than NUL-terminated.

The current nvme-cli code will print all the spaces and possibly some
garbage from the next field.  For example this causes "connect-all"
to write strings that get rejected with "malformed IP address passed."

Fix this by only writing the contents of these fields until the first
space character.

Signed-off-by: Roland Dreier <roland@purestorage.com>

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -354,6 +354,17 @@ out:
 	return error;
 }
 
+static int space_strip_len(int max, const char *str)
+{
+	int i;
+
+	for (i = max - 1; i >= 0; i--)
+		if (str[i] != '\0' && str[i] != ' ')
+			break;
+
+	return i + 1;
+}
+
 static void print_discovery_log(struct nvmf_disc_rsp_page_hdr *log, int numrec)
 {
 	int i;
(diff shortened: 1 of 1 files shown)


keithbusch:
There's no such TRSVCID or TRADDR that could have space in the middle,, right? I'm confident that's a safe assumption for TRADDR, but I'm not sure what a TRSVCID looks like (the spec points to a non-existent document).

jpfreyen:
On Mon, 2017-03-06 at 09:20 -0800, Keith Busch wrote:
> TRSVCID

TRSVCID is the port number, which by default for an RDMA connection in fabrics
is 4420, so if the port number is not passed in nvme-cli it will use 4420.  I
would think it would be ok?

jpfreyen:
On Mon, 2017-03-06 at 11:16 -0800, Roland Dreier wrote:

TRSVCID is defined for RDMA in the Fabrics ECNs, http://www.nvmexpress.org/wp-content/uploads/NVMe_over_Fabrics_1.0_Errata.zip

It's either a port number ("decimal number represented as an ASCII string with no leading zeroes") or "16 case insensitive ASCII hexadecimal digits (‘0’-‘9’, ‘a’-‘f’, ‘A-F) in groups of 4 separated by hyphens (‘-‘), e.g., dead-beef-0123-3210" for the AF_IB case.

Current draft of FC-NVMe says "Transport Service ID: shall be set to “none”"

With all that said, it's easy enough to print up to the last space instead of stopping at the first space... should I update my patch?

I think that would be good.

rolandd:
OK, updated the commit in this pull request to print up until the last non-space character.

PR #165, opened 2017-03-16
Fixup Security send/recv to read data from disk

ScottyBauer:
We forgot to actually read the data from the disk into the buffer.

Reported-by: Thomas Bowen <thomas.t.bowen@intel.com>


keithbusch:
Looks like we found someone with a drive capable capable of exercising this feature. :)

PR #167, opened 2017-03-28
fabrics: Fix disconnect_by_device from disconnecting instance 0 on error

lsgunth:
I was a bit surprised that running:

nvme disconnect -d /dev/nvme1

actually disconnected nvme0 and I could never disconnect nvme1. Turns
out the code was not expecting a full path and silently used instance
zero if it failed to parse any arguments.

This patch fixes it so that it ignores any path component and fails if
sscanf doesn't match any items.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -31,6 +31,7 @@
 #include <asm/byteorder.h>
 #include <inttypes.h>
 #include <linux/types.h>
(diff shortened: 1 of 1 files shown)


keithbusch:
Nice catch. Thanks, Logan!

PR #169, opened 2017-04-06
Fixed problem with security-recv binary output not working correctly

ScottyBauer:
Security Receive sent the wrong pointer, specifically a pointer to a
stack location. When using the Dump raw command we would dump stack
contents, not the buffer received from the controller.

Signed-off-by: Thomas Bowen <thomas.r.bowen@intel.com>
Signed-off-by: Scott Bauer <Scott.Bauer@intel.com>


ScottyBauer:
Woops I think its over 60 lines the message let me fix that

keithbusch:
Too late! :)

The character limit is really just a guideline anyway. It makes  git logs look good on some terminals, but it's not a big deal.

PR #170, opened 2017-04-11
lightnvm: allow to init target on factory mode

javigon:
Allow to drop the target recovery mechanism when creating a target in
order to start in factory mode. This facilitates development and makes
it possible to evaluate an OCSSD in different states.

Signed-off-by: Javier González <javier@cnexlabs.com>


javigon:
Matias just picked the kernel side of it for 4.12

PR #186, opened 2017-06-06
fix use-after-free in pci.ids parsing

apage43:
If a device that is _not_ in pci.ids has its name looked up after a device
that _was_ present in pci.ids, these pointers are still set even though
they've been free'd, and then get passed to snprintf resulting in a
use-after-free.

### DIFF
--- a/nvme-models.c
+++ b/nvme-models.c
@@ -32,11 +32,17 @@ static char *class_final;
 static void free_all(void)
 {
 	free(device_top);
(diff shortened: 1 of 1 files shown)


apage43:
can repro this on a system with two nvme devices of different vendor/device/sub ids. if the one listed _second_ isn't in `pci.ids` then `nvme list -o json` will crash

PR #189, opened 2017-06-16
Implement Intel Vendor specific logs 

ScottyBauer:


ScottyBauer:
Tested on FD/CD ill test on mansion beach here after lunch

PR #196, opened 2017-06-27
nvme-cli : add help text for get-ns-id command.

CKulkarni1111:
This patch adds the help text for get-ns-id command.

Signed-off-by: Chaitanya Kulkarni <chaitanya.kulkarni@hgst.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -836,8 +836,10 @@ static int list(int argc, char **argv, struct command *cmd, struct plugin *plugi
 		return -EINVAL;
 
 	n = scandir(dev, &devices, scan_dev_filter, alphasort);
-	if (n < 0)
+	if (n < 0) {
+		fprintf(stderr, "no NVMe device(s) detected.\n");
 		return n;
+	}
 
 	list_items = calloc(n, sizeof(*list_items));
 	if (!list_items) {


keithbusch:
I'm a bit confused by the first patch here. Right now the syntax is `nvme get-ns-id /dev/nvme0n1` without any arguments in order to get the NSID of unnamed parameter /dev/nvme0n1. What is the new argument gaining us?

CKulkarni1111:
@keithbusch You are right, we don't need the block device option, the idea to have help text printed when "nvme get-ns-id" executed.

In current implementation this command doesn't print the help text : -
$ ./nvme get-ns-id -h
-h: No such file or directory
$ ./nvme get-ns-id
get-ns-id: Invalid argument
$

With this patch it is adding a new option which we don't need, all we need is a help text : -
$ ./nvme get-ns-id -h
Usage: nvme get-ns-id <device> [OPTIONS]

Get namespace ID of a particular NVMe Block Device.

Options:
  [  --block-dev=<NAME>, -d <NAME> ]    --- Name of block device to get
                                            Namespace ID.  _<----- need to get rid of this._

 
The desired output we need to print the help text is : -

$ ./nvme get-ns-id -h
Usage: nvme get-ns-id <device> [OPTIONS]

Get namespace ID of a particular NVMe Block Device.

Options:

Please let me know if this is workable solution to print only help text.  I'll make these changes and send a new pull request.

PR #198, opened 2017-06-27
Handle Unsupported Smart Log per namespace

narasimhan-v:
Some controllers do not support smart log per namespace.
This can be checked in id-ctrl.
Added a check to handle that.

Signed-off-by: Narasimhan V <sim@linux.vnet.ibm.com>

### DIFF
--- a/tests/nvme_test.py
+++ b/tests/nvme_test.py
@@ -321,6 +321,20 @@ class TestNVMe(object):
             - Returns:
                 - 0 on success, error code on failure.
         """
+        smart_log_supported_per_ns_cmd = "nvme id-ctrl " + self.ctrl
+        proc = subprocess.Popen(smart_log_supported_per_ns_cmd,
+                                shell=True,
+                                stdout=subprocess.PIPE)
+        err = proc.wait()
+        if nsid != '0xFFFFFFFF':
+            smart_log_per_ns = 0
+            for line in proc.stdout:
+                if line.startswith('lpa'):
+                    if bin(int(line.split(' ')[-1], 16))[2:][-1] == '0':
+                        smart_log_per_ns = -1
+                        break
+            if smart_log_per_ns != 0:
+                return 0
         smart_log_cmd = "nvme smart-log " + self.ctrl + " -n " + str(nsid)
         print smart_log_cmd
         proc = subprocess.Popen(smart_log_cmd,


narasimhan-v:
```
Running nvme_smart_log_test.py

/sys/devices/pci0002:01/0002:01:00.0/nvme/nvme0
nvme_smart_log_test.TestNVMeSmartLogCmd.test_smart_log ... Executing command: nvme id-ctrl -H /dev/nvme0
Executing command: nvme smart-log /dev/nvme0 -n 0xFFFFFFFF
data_units_read 101807
data_units_written 502230
host_read_commands 901171
host_write_commands 1125118
Executing command: nvme list-ns /dev/nvme0
Executing command: nvme id-ctrl -H /dev/nvme0
ok

----------------------------------------------------------------------
Ran 1 test in 0.039s

OK
```

narasimhan-v:
@keithbusch Can you review this one please

keithbusch:
Is string pattern matching the preferred way to do this? I would have parsed the binary stream to look for the capability field so that the test won't break if we change the output string. It's been about years since I wrote a line of python, though ... Off the top of my head, I can't recommend how to unmarshal.

narasimhan-v:
I have pushed a newer commit with the suggested changes.

keithbusch:
Not quite what I was thinking. I mean I'd run `nvme id-ctrl -r`  and parse that binary stream.

narasimhan-v:
Does `nvme id-ctrl -r /dev/nvme0` and `nvme id-ctrl /dev/nvme0` not give the same values in different formats ?

What I have done is:
```
# nvme id-ctrl /dev/nvme0 | grep ^lpa
lpa     : 0x2
```

keithbusch:
Sorry, I meant '-b' for the short option. The long option is --raw-binary. This is very different than string parsing. String parsing means we're stuck with the current output text. While I have no intention of changing it, I like having the option.

narasimhan-v:
```
linux-kvnj:~ # nvme id-ctrl /dev/nvme0 | grep ^lpa
lpa     : 0x1
linux-kvnj:~ # nvme id-ctrl /dev/nvme0 -r > raw.txt; hexdump -s 0x105 -C raw.txt
00000105  01
```

Is this what you expect ?

keithbusch:
As mentioned, -r was the wrong option. I meant -b. My controller looks something like this:

```
# nvme id-ctrl -b /dev/nvme0 | hexdump
0000000 8086 8086 5546 424d 3235 3635 3030 4a30
0000010 3832 4130 2020 2020 4e49 4554 204c 5353
0000020 5044 4445 4431 3431 4730 2041 2020 2020
0000030 2020 2020 2020 2020 2020 2020 2020 2020
...
```

keithbusch:
huh, interestingly -r flags the same binary flag. The long options may be parsed based on partial matching if it isn't ambiguous. Just learned something new!

narasimhan-v:
Yes, -b and -r prints the same.
And, I vote for `nvme id-ctrl /dev/nvme0 | grep ^lpa` instead of `nvme id-ctrl -b /dev/nvme0 | hexdump` for 2 reasons:
1. It is easier to parse and understand, so maintainability of code.
2. The normal output is expected be consistent with output of -r (My assumption).

keithbusch:
My suggestion really isn't to parse hexdumps. It's about deserializing the stream into a language native struct.

narasimhan-v:
@keithbusch Can you please explain why you suggest that method and not the method of parsing the normal output ?

I see couple of issues with the approach. Please correct me if I am wrong.

1. Will Deserializing the stream take care of endianness ?
2. How do we take care if the struct changes in future ?

keithbusch:
The bit's meaning will never change, but reserve the right to change the string associated with it. Parsing the bit stream is future proof; parsing strings is not.

narasimhan-v:
@keithbusch Sorry for asking it pretty late, any pointers / examples on how `deserializing` is done ?
I have not done it, sorry if it is a dumb question.

hreinecke:
As said, recommendation is to use the binary stream. The stream itself is defined to be in little endian, so there won't be any issues with different architectures.

PR #205, opened 2017-07-14
Intel's Plugin - Latency Bug Fix

tbjones95:
The lat-stats function within Intel's plugin reported only "Bucket 1" values instead of all 3 buckets values. 

I also added JSON output functions for both lat-stats and temp-stats reports.


keithbusch:
Hi Tanner,
It looks like you created a pull request with some unintended merges,. Could you rebase this so you only have the new commits you actually want?

PR #206, opened 2017-07-14
Updating for new Features

tbjones95:
Updating for new features to some Intel drives including latency statistics, SMART attributes, and JSON output.


keithbusch:
There seems to be unrelated formatting changes that shouldn't be included here, and the formatting isn't correct anyway.

In any case, the formatting changes in the second patch appear to be a helper or some sort so that the user doesn't need to know which lbaf goes with which format size. I think that's just confusing having these mutually exclusive options, and it's probably going to do the wrong thing depending on metadata format settings.

hreinecke:
Closing.

PR #217, opened 2017-08-12
nvme-cli/write: add checking for data file while doing write/compare operation

liangxiao1:
If don't provide --data= or -d when do write/compare operation, read function
will wait for user inputting without any prompt. Now add checking before reading.

Signed-off-by: Xiao Liang <xiliang@redhat.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2566,6 +2566,11 @@ static int submit_io(int opcode, char *command, const char *desc,
 		return EINVAL;
 	}
 
+	if (!dfd)       {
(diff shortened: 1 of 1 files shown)


keithbusch:
This will break usage like:

`echo "hello world" | nvme write /dev/nvme0n1`

If you want to simply prompt for input, that's fine, but we shouldn't be returning an error when the data input file is set to STDIN.

liangxiao1:
Yes, I will this pull request. Thanks

PR #218, opened 2017-08-15
Fixing macro definitions to avoid issues when these are used in expression

mandarcthorat:
C operator precedence puts ternary operator below many arithmetic operators
http://en.cppreference.com/w/c/language/operator_precedence

Without these changes following expression will return 2 instead of expected answer 5
printf("Sum = %d", min(2,3)+max(2,3));


keithbusch:
Indeed, thanks for the fix. You'd think these macros would have had a generic definition in a libarary!

PR #224, opened 2017-09-13
Added handling for temperature threshold values below 0C(273K)

mandarcthorat:
Added handling for temperature threshold values below 0C(273K) in human readable option and removed buffer check for human readable output as Temperature threshold feature does not use Memory Buffer for Attributes

 Changes to be committed:
	modified:   nvme-print.c
	modified:   nvme.c

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1281,9 +1281,9 @@ static int get_feature(int argc, char **argv, struct command *cmd, struct plugin
 			printf("get-feature:%#02x (%s), %s value:%#08x\n", cfg.feature_id,
 				nvme_feature_to_string(cfg.feature_id),
 				nvme_select_to_string(cfg.sel), result);
-			if (cfg.human_readable && buf)
-				nvme_feature_show_fields(cfg.feature_id, result, buf);
-			else if (buf)
+			if (cfg.human_readable)
+				nvme_feature_show_fields(cfg.feature_id, (cfg.cdw11 & 0x3f0000) | result, buf);
+			if (buf)
 				d(buf, cfg.data_len, 16, 1);
 		} else if (buf)
 			d_raw(buf, cfg.data_len);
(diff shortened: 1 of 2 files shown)


nvme.c
```diff
@@ -1281,9 +1281,9 @@ static int get_feature(int argc, char **argv, struct command *cmd, struct plugin
 			printf("get-feature:%#02x (%s), %s value:%#08x\n", cfg.feature_id,
 				nvme_feature_to_string(cfg.feature_id),
 				nvme_select_to_string(cfg.sel), result);
-			if (cfg.human_readable && buf)
-				nvme_feature_show_fields(cfg.feature_id, result, buf);
-			else if (buf)
+			if (cfg.human_readable)
+				nvme_feature_show_fields(cfg.feature_id, (cfg.cdw11 & 0x3f0000) | result, buf);
```
keithbusch: What's this cdw11 part for?

mandarcthorat: This bits In cdw11 for get-feature 4 command need to provided by user to get feature value for specific Threshold type select(THSEL) and Threshold temperature select(TMPSEL)

keithbusch: Not sure why you are trying to get that from CDW11 of the submission command, then. The spec says it's already in CQE.DW0, which is what "result" points to:

"If a Get Features command is submitted for this feature, the temperature threshold selected by Command Dword 11 is returned in Dword 0 of the completion queue entry for that command"

mandarcthorat: Yes, the spec is saying that the temperature threshold is returned in Dword 0. 

Please note that there is temperature threshold field in CDW11(TMPTH). This is causing confusion.

Please let me know if I should revert this

keithbusch: Yes, please undo that part. The temperature threshold field defined in CDW11 is only used for setting the feature, not for the getting it.

PR #229, opened 2017-10-23
Include sizing data in json output that's included in the non-json results

steve-gh:


keithbusch:
This is mixing declarations with code. That won't work on some c compiler settings. Let's just get lba, nsze and nuse defined at the top of the function, then we're okay.

keithbusch:
Looks fine, thanks for the update.

steve-gh:
Moved variable definitions to top of function, please take another look. Thanks!

PR #230, opened 2017-11-01
nvme/id-ns: do not try to get namespace id from non-block device.

liangxiao1:
namespace id's default value is 0 and changed to 15 by get_nsid()
when device is not a block device.

With this change:
`# ./nvme id-ns /dev/nvme0 -n 0`
Error: requesting namespace-id from non-block device
NVMe Status:INVALID_NS(b) **NSID:0**

Without this change:
Missleading output when user passed 0 or 15 as namespace id.
`# nvme id-ns /dev/nvme0`
Error: requesting namespace-id from non-block device
NVMe Status:INVALID_NS(b) **NSID:15**
`# nvme id-ns /dev/nvme0 -n 0`
Error: requesting namespace-id from non-block device
NVMe Status:INVALID_NS(b) **NSID:15**
`# nvme id-ns /dev/nvme0 -n 15`
NVMe Status:INVALID_NS(b) **NSID:15**

Signed-off-by: Xiao Liang <xiliang@redhat.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1178,8 +1178,11 @@ static int id_ns(int argc, char **argv, struct command *cmd, struct plugin *plug
 		flags |= VS;
 	if (cfg.human_readable)
 		flags |= HUMAN;
-	if (!cfg.namespace_id)
+	if (!cfg.namespace_id && S_ISBLK(nvme_stat.st_mode))
 		cfg.namespace_id = get_nsid(fd);
+	else if(!cfg.namespace_id)
+		fprintf(stderr,
+			"Error: requesting namespace-id from non-block device\n");`
 
 	err = nvme_identify_ns(fd, cfg.namespace_id, cfg.force, &ns);
 	if (!err) {


keithbusch:
But we want to be able to identify namespaces that are not attached to the controller, so we need to be able to issue this command to the admin character device.

liangxiao1:
Thanks, but in my change, it does not skip identify namespaces command  to the admin character device. Look at nvme_get_nsid(), I see that it directly return ENOTBLK(15) if it is character device.

And in next step of nvme.c, it use namespace id 15(which is not user passed "0")to identify namespace.
Is it expected? Thanks

nvme.c:  
`err = nvme_identify_ns(fd, cfg.namespace_id, cfg.force, &ns);`

In nvme-ioctl.c:
`        if (!S_ISBLK(nvme_stat.st_mode)) {
                fprintf(stderr,
                        "Error: requesting namespace-id from non-block device\n");
                return ENOTBLK;
        }
`

keithbusch:
Oh, right. Looks fine, then, thanks for the patch.

PR #240, opened 2017-12-20
nvme-cli: Implement self-test and self-test-log from 1.3 spec.

ScottyBauer:
Implement logic to start self-tests, and to get self-test log
pages.


ScottyBauer:
I'm going to sometime tomorrow submit a patch to move max()/min()/array_size() to a standard .h file. We've redefined it in I think 4 different files -- mostly my fault it seems.

minwooim:
@ScottyBauer ,  
Any news on this?

ScottyBauer:
We don't have any drives here that implement this so I'm hesitant to submit it without testing.

minwooim:
Oh I see. Thanks for the notification. :)

ScottyBauer:
We have an intern here, I'll have him rebase it, if you have drives that have this feature we'd love at test.

tim-oleksii:
You can send me a patch, I'll get it tested.

ScottyBauer:
@tim-oleksii please pull from here and test out for us:
https://github.com/RevanthRajashekar/nvme-cli/tree/smart_log1

Let us know if you run into any problems or questions.
Thanks

tim-oleksii:
Thank you for a patch. I've got a couple of things:

1. Status messages contradict operation.

> ./nvme device-self-test /dev/nvme0 -s 15 -n 1
prints
> Device self-test started
However, abort is executed.

2.  JSON is malformed right now. It's structure is not suited for sequential data structures, it's a mix of printf and json calls. Key value map contains duplicate keys.

3. Some values aren't reported correctly. I've got a drive that has POH (Power on hours) which is 0x100000000 according to nvme-cli, however, it was recently manufactured. Our internal tool reports something about 0x77, which, I believe, correct value.

4. Namespace is reported incorrectly.
> Namespace Identifier (NSID) : 119
But i've got '-n 1' as a parameter.

I've attached a couple of artifacts that might help debugging.
https://gist.github.com/tim-oleksii/7c3cd29af45bf9b6eaeb331c74d2eff6

tim-oleksii:
@ScottyBauer OK, it's not urgent for me to have it in nvme-cli, a patch is welcome when it's done. As my time shift is UTC +3, the feedback would be sent next day.

ScottyBauer:
Okay, he has another one to try. GCC was padding the structure which caused the offsets to move by 4 bytes.

https://github.com/RevanthRajashekar/nvme-cli/tree/smart_log2

Can you also send us that binary blob you have so we can just replay it with "real" data to do further debugging/fine tuning?

tim-oleksii:
I omitted adding binary payload as github refused to attach it, hoping that 'xxd' could be sufficient to recreate binary payload. I'll drop you a line into corporate mailbox with a file.

keithbusch:
Implemented in separate request.

PR #247, opened 2018-01-07
nvme-cli: make sanitize description const

minwooim:
Make description in sanitize() const.


minwooim:
Bad request so close it.

PR #248, opened 2018-01-07
nvme-cli: remove nsid parameter in error-log command

minwooim:
Error Information Log Page is global to controller.
So, remove unnecessary nsid parameter in nvme_error_log().

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -1409,7 +1409,7 @@ void nvme_feature_show_fields(__u32 fid, unsigned int result, unsigned char *buf
 		printf("\tNumber of IO Submission Queues Allocated (NSQA): %u\n",  (result & 0x0000ffff) + 1);
 		break;
 	case NVME_FEAT_IRQ_COALESCE:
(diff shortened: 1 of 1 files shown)


keithbusch:
I am not sure if that's true. Log pages do take namespace parameters.

minwooim:
Keith, 
Thanks for your review.

I have reviewed NVMe 1.2.1 and 1.3 again.
SMART/Health Information log page (LID 02h) is supported on a per namespace basis.
It's shown that Error Information Log page (LID 01h) is global to the controller.

Identify Controller Data Structure contains LPA field at [261] which only cares about
extended data, command effect log page, and SMART/Health information log page.

If you don't mind, would you please let me know which case namespace is taken by device
in case of error log page?

keithbusch:
Hm, it appears error log doesn't consult nsid.

minwooim:
So, Are you going to keep it that way, or not?

keithbusch:
Just need to get the Documentation/nvme-error-log.txt updated to reflect the possible parameters.

minwooim:
Totally agree with you.
I will send a patch including that one soon.

PR #251, opened 2018-01-07
nvme-cli: update print for sanitize-log command with output-format

minwooim:
Update print for sanitize-log command with normal and binary
output-format. Also move origin print code to nvme-print.c. with
human-readable option.

Make sanitize-log command call nvme_sanitize_log() so that
it looks more consistent with other XXX-log commands.


keithbusch:
This is fine as-is, though a json format would be a nice future addition.

minwooim:
I will also send a code for JSON format later time.
Thanks for your review and pull.

PR #261, opened 2018-01-12
lightnvm: enable to set OP on target creation

javigon:
On target creation, allow to define the over-provision (OP). Since OP
area has a direct impact on write amplification, users can optimize the
OCSSD behavior for their workloads.

Support in the kernel has been upstreamed in 4.16

Signed-off-by: Javier González <javier@cnexlabs.com>


MatiasBjorling:
Looks good to me.

PR #271, opened 2018-01-16
nvme/vendor: Add get log LSP/LSO fields from 1.3 spec

ScottyBauer:

### DIFF
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -776,6 +776,12 @@ enum {
 	NVME_FWACT_ACTV		= (2 << 3),
 };
 
+enum {
+	NVME_NO_LOG_LSP       = 0x0,
+	NVME_NO_LOG_LPO       = 0x0,
+	NVME_TELEM_LSP_CREATE = 0x1,
+};
+
 /* Sanitize and Sanitize Monitor/Log */
 enum {
 	/* Sanitize */
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -779,7 +779,6 @@ enum {
 enum {
 	NVME_NO_LOG_LSP       = 0x0,
 	NVME_NO_LOG_LPO       = 0x0,
-	NVME_TELEM_LSP_CREATE = 0x1,
 };
 
 /* Sanitize and Sanitize Monitor/Log */
--- a/nvme-ioctl.c
+++ b/nvme-ioctl.c
@@ -381,7 +381,8 @@ int nvme_identify_ns_descs(int fd, __u32 nsid, void *data)
 	return nvme_identify(fd, nsid, NVME_ID_CNS_NS_DESC_LIST, data);
 }
 
-int nvme_get_log(int fd, __u32 nsid, __u8 log_id, __u32 data_len, void *data)
+int nvme_get_log(int fd, __u32 nsid, __u8 log_id, __u8 lsp, __u32 lpo,
+                 __u32 data_len, void *data)
 {
 	struct nvme_admin_cmd cmd = {
 		.opcode		= nvme_admin_get_log_page,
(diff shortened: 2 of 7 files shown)


ScottyBauer:
Hi Keith,
Hope you're doing well.

For the internal Telemetry stuff they had me do I need access to the LPO/LSO fields from the 1.3 spec. 
http://nvmexpress.org/wp-content/uploads/NVM-Express-1_3a-20171024_ratified.pdf:
Figures 87 and 88 (page 99)

I tested on CDR and it looks to be working for old log commands (SMART,fw slot). 

Let me know if you want this modified anyway or you want an entire new get_log function for 1.3? I do need this to go in before we can merge the telemetry log gathering as I depend on these fields to get telemetry logs.

linux/nvme.h
```diff
@@ -776,6 +776,12 @@ enum {
 	NVME_FWACT_ACTV		= (2 << 3),
 };
 
+enum {
+	NVME_NO_LOG_LSP       = 0x0,
+	NVME_NO_LOG_LPO       = 0x0,
+	NVME_TELEM_LSP_CREATE = 0x1,
```
minwooim: The last enumeration value seems not related to this patch.

nvme-ioctl.c
```diff
@@ -393,39 +394,55 @@ int nvme_get_log(int fd, __u32 nsid, __u8 log_id, __u32 data_len, void *data)
 	__u16 numdu = numd >> 16, numdl = numd & 0xffff;
 
 	cmd.cdw10 = log_id | (numdl << 16);
+	if (lsp)
+                cmd.cdw10 |= ((int)lsp) << 8;
```
minwooim: Q. why _lsp_ has to be casted to int first?

nvme-ioctl.c
```diff
@@ -393,39 +394,55 @@ int nvme_get_log(int fd, __u32 nsid, __u8 log_id, __u32 data_len, void *data)
...
+
 	cmd.cdw11 = numdu;
+        cmd.cdw12 = lpo;
```
minwooim: LPO is 64-bit field separated to LPOL and LPOU.
Where is assign statement for cdw13?

nvme.c
```diff
@@ -426,13 +428,17 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 		__u32 log_id;
 		__u32 log_len;
 		__u32 aen;
+		__u32 lpo;
```
minwooim: LPO seems 64-bit field with LOPL + LPOU

nvme.c
```diff
@@ -441,6 +447,8 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 		{"log-len",      'l', "NUM", CFG_POSITIVE, &cfg.log_len,      required_argument, log_len},
 		{"aen",          'a', "NUM", CFG_POSITIVE, &cfg.aen,          required_argument, aen},
 		{"raw-binary",   'b', "",    CFG_NONE,     &cfg.raw_binary,   no_argument,       raw_binary},
+		{"lsp",          's', "NUM", CFG_POSITIVE, &cfg.lsp,          required_argument, lsp},
+		{"lsp",          'o', "NUM", CFG_POSITIVE, &cfg.lpo,          required_argument, lpo},
```
keithbusch: This is a u8, but CFG_POSITIVE will think it's pointing to a u32. It happens to be safe in this case because of struct padding, but it should be CFG_BYTE.

keithbusch: Oh, and I think you intended the long option should be "lpo".

keithbusch:
Thanks, Scott. Mostly fine, just a couple minor things here.

I mixed up LSP and LPO on my CFG_BYTE comment, but I hope you saw through that.

There are also a few minor formatting issues with tabs/spaces.

I've no problem with the design, so no need to have a new log API to accommodate the new params.

nvme-ioctl.c
ScottyBauer: I meant to take that out. I wasn't sure of the integer promotion rules here. For example, does the shift promote up to 32bit then cast back down or is shifting by the width of a u8 on a u8 undefined behavior? 
I didn't know the answer to this, and I didn't want to open the c-spec or go read assembly so I cast it to int. If you know the answer, I'll change it.

nvme-ioctl.c
ScottyBauer: I didn't add it. I was under the impression controllers aren't going to need to index past 4gb into a log for a long time... But I guess we'll slap it in so we do not have to do it in the future.

ScottyBauer:
Hi, I believe I addressed all the reviews here, thank you! Keith, it seems there is a way to squash that on merge into the original commit during the merge into the real nvme-cli. If you can't find it I will squash it and re-push. 

Thanks

keithbusch:
No problem, squashed on my end, will push that as-is.

There is still some mismatch on parameters for LSP and LPO, but I'll fix that on the rebase, and just note it in the commit log.

keithbusch:
Squashed rebase pushed; thanks for the adding the feature!

keithbusch:
https://github.com/linux-nvme/nvme-cli/commit/722c71bb3246a46a784861a3874aa5017b919f49

PR #277, opened 2018-01-17
nvme-cli: lnvm: Make it return error status when argconfig_parse fails

minwooim:
The return value of argconfig_parse() should be returned if it fails.
Make lnvm commands return when it fails.


MatiasBjorling:
Looks good to me. Thanks!

PR #278, opened 2018-01-18
nvme-cli: Implement get_telemetry_log from 1.3 spec

ScottyBauer:

### DIFF
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -377,6 +377,28 @@ enum {
 	NVME_NIDT_UUID		= 0x03,
 };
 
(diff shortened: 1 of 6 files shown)


keithbusch:
Thanks for doing this. I'll pull this now, but I will probably make a few minor tweaks. I think hostgen param should be a flag instead of integer, and I think we can either come up with a default output name (see intel's nlog, for example), or just write it to stdout.

ScottyBauer:
Sure thing, that makes sense. Let me know if there are any functional changes you need tested.

PR #282, opened 2018-01-19
nvme-cli: Implement nvme_get_log13

ScottyBauer:
This fixes a portion of 722c71bb:
nvme/vendor: Add get log LSP/LSO fields from 1.3 spec

This fixes up the code so it can use the same interface
log we've had for awhile. People who wish to pass log
specific parameters can use the new API.

### DIFF
--- a/intel-nvme.c
+++ b/intel-nvme.c
@@ -255,9 +255,7 @@ static int get_additional_smart_log(int argc, char **argv, struct command *cmd, 
 	if (fd < 0)
 		return fd;
 
-	err = nvme_get_log(fd, cfg.namespace_id, 0xca,
-			   NVME_NO_LOG_LSP, NVME_NO_LOG_LPO,
-			   sizeof(smart_log), &smart_log);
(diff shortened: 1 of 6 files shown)


ScottyBauer:
Since I created the get_log mess I thought I would try and clean it up for you. If you don't like this approach we can just kill this and I'll let you do it the proper way -- no offense taken if you want to do it that way.

keithbusch:
No worries, not a big deal at all. There were a few clashes recently, but it happens.

Let me double check what's going on in the mailing list before pulling this to make sure everything is sane. We may need to stop github pull requests and use only the linux-nvme mailing list for now, since a lot more people subscribe to there.

keithbusch:
I like the addition. I had to apply a vendor update that creates a conflict with this patch in wdc. though.

ScottyBauer:
I can rebase, if you want. Then from now on i'll submit through the mailing list.

ScottyBauer:
Closed since I sent to the mailing list.

PR #283, opened 2018-01-20
nvme-cli: Add support of unsigned long long to json routines

tim-oleksii:

### DIFF
--- a/json.c
+++ b/json.c
@@ -55,6 +55,19 @@ static struct json_value *json_create_value_int(long long number)
 	return value;
 }
 
+static struct json_value *json_create_value_uint(unsigned long long number)
+{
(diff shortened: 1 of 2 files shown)


keithbusch:
This seems fine, but there aren't any users of this new type. I assume these will be coming?

tim-oleksii:
Some plugins need it, for instance, 
https://github.com/linux-nvme/nvme-cli/blob/69b377995c0c05d297486f405fa54889355c8f22/wdc-nvme.c#L1133
It's certainly a bug of mistreating real value, but I haven't touched it so far. There might be a ticket to have it done, as it's complex job of checking every plugin.

keithbusch:
Okay, I think I'll apply this. I did lift the entire json component from 'fio', and was hoping we'd never have to touch it.

PR #290, opened 2018-01-30
nvme-cli: add command specific field to json output of error-log

minwooim:
Add command specific field to json output of error-log command.
This field was added in commit 662c1618 without considering json output
format.

Fixes: 662c1618("nvme-cli: add print for command specific info field of
error log page")

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -1841,6 +1841,7 @@ void json_error_log(struct nvme_error_log_page *err_log, int entries, const char
 		json_object_add_value_int(error, "lba", err_log[i].lba);
 		json_object_add_value_int(error, "nsid", err_log[i].nsid);
 		json_object_add_value_int(error, "vs", err_log[i].vs);
+		json_object_add_value_uint(error, "cs", err_log[i].cs);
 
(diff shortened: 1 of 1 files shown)


nvme-print.c
```diff
@@ -1841,6 +1841,7 @@ void json_error_log(struct nvme_error_log_page *err_log, int entries, const char
 		json_object_add_value_int(error, "lba", err_log[i].lba);
 		json_object_add_value_int(error, "nsid", err_log[i].nsid);
 		json_object_add_value_int(error, "vs", err_log[i].vs);
+		json_object_add_value_uint(error, "cs", err_log[i].cs);
```
tim-oleksii: It seems that you can switch error_count and lba as well into output it as 'uint'.

minwooim: Tim,

Thanks for your kind comment.
This patch is focused on just an addition of Command Specific field to the JSON print function. 
I think that other unsigned data fields should be converted by another patch which will cover all the print functions in nvme-print.c
Please give your opinion :)

Thanks,

PR #291, opened 2018-01-30
nvme-cli: make it have bpid in cdw10 for fw-commit

minwooim:
_bpid_ was added as a parameter in commit d52c9988 without any being
used.
Make it have inside of cdw10 before submitting a command.

Fixes: d52c9988 ("nvme-cli: update Firmware Commit with boot partition
feature")


keithbusch:
Will apply as-is. It looks like bpid should be a flag, though.

PR #300, opened 2018-01-30
nvme-cli: add initial value to namespace of list-ctrl

minwooim:
cfg.namespace_id has been used without any initialization if not given.
nvme_identify_ctrl_list() function checks given namespace_id value.

Add initial value for this namespace id.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -674,6 +674,7 @@ static int list_ctrl(int argc, char **argv, struct command *cmd, struct plugin *
 
 	struct config cfg = {
 		.cntid = 0,
+		.namespace_id = 0,
 	};
 
 	const struct argconfig_commandline_options command_line_options[] = {


nvme.c
```diff
@@ -674,6 +674,7 @@ static int list_ctrl(int argc, char **argv, struct command *cmd, struct plugin *
 
 	struct config cfg = {
 		.cntid = 0,
+		.namespace_id = 0,
```
tim-oleksii: Isn't it initialized with 0 by default wit hdesignated initializers?

keithbusch:
This is a no-op, really. Just initializing a struct like `struct _type_ _name_ = {}` sets everything to 0 already.

minwooim:
Thank you for the reviews.
Sorry for making a noise for this one. :(

PR #308, opened 2018-02-21
nvme-cli: fix endianess in WDC plugin.

tim-oleksii:


tim-oleksii:
@jeffreyalien Could you please review?

jeffreyalien:
I have review the changes and they look fine.  I plan to verify the changes too, but will probably need to get some special firmware code to set those stats to something other then 0's.  That might take a few days or so.   I'll comment as soon as I get it verified.

keithbusch:
Thanks, Jeff. I think this looks correct too just based on how it looked before, so I'll pull in now and we can make additional fixes if you find them.

tim-oleksii:
Thank you!

jeffreyalien:
I was able to verify the changes with non-zero data and all looks good.   No additional changes will be needed.

PR #314, opened 2018-03-07
Few fix related to 'nvme list-subsys'

vijaykumar-s:
Sending these changes for review:

1) Fix double free in list_subsys

    get_nvme_subsystem_info() deallocates subsys_list_item in case of error.
    This hits double free when list_subsys() as well attempts to free it.
    Removed deallocation of subsys_list_item from get_nvme_subsystem_info().

2) Log error incase of failure in get_nvme_subsystem_info()

    Currently get_nvme_subsystem_info() silently fails in case
    of error. This patch logs relevant error message incase of
    failure in get_nvme_subsystem_info().
    
Signed-off-by: Vijay Kumar <vijay.ac.kumar@oracle.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1217,18 +1217,24 @@ int get_nvme_subsystem_info(char *name, char *path,
 	int n, i, ret = 1;
 
 	item->subsysnqn = get_nvme_subsnqn(path);
-	if (!item->subsysnqn)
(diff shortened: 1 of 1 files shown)


keithbusch:
Thanks, looks good.

PR #320, opened 2018-03-16
Change default value for log_id

rrosatti:
Currently the log_id is being set as 0 for default, if no value is
passed with the '-i' parameter. In this situation, we get the error:
NVMe Status:INVALID_LOG_PAGE(2109). It happens because 00h is Reserved
according to the NVM Express Spec 1.3.

Signed-off-by: Rodrigo R. Galvao <rosattig@linux.vnet.ibm.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -621,7 +621,7 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 
 	struct config cfg = {
 		.namespace_id = NVME_NSID_ALL,
-		.log_id       = 0,
+		.log_id       = 1,
 		.log_len      = 0,
 		.lpo          = NVME_NO_LOG_LPO,
 		.lsp          = NVME_NO_LOG_LSP,


rrosatti:
The part of the Spec. I'm talking about is on page 97, the Figure 90: Get Log Page – Log Page Identifiers.
Please, correct me if I'm wrong. And any thoughts about it is very welcome! =)

Thanks,
Rodrigo

keithbusch:
IMO, this ought to be a required parameter if people are using the generic get-log.

rrosatti:
So, maybe we could use something like that:
if (!cfg.log_id) {
        fprintf(stderr, "log-id is required param\n");
        err = EINVAL;
} 
The same thing we are doing with the log_len parameter.

keithbusch:
I'm not sure if log id 0 will ever become valid, but just in case, how about we default to 0xffffffff. The log id is only 8 bits, so anything > 0xff will never be a valid log identifier. We probably should have been range checking this value before anyway.

rrosatti:
Hm right ... but, doing this we would get the same error message, correct?
NVMe Status:INVALID_LOG_PAGE(2109)

rrosatti:
The other suggestion would be just to check if the user passed a value for log_id.
Something like this:

--- a/nvme.c
+++ b/nvme.c
@@ -650,7 +650,10 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
```
       if (!cfg.log_len) {
                fprintf(stderr, "non-zero log-len is required param\n");
                err = EINVAL;
-      } else {
+      } else if (!cfg.log_id) {
+        fprintf(stderr, "log-id is required param\n");
+        err = EINVAL;
+      } else {
                unsigned char *log;

                log = malloc(cfg.log_len);
```

So, that way the user would be advised about it instead of throwing the `NVMe Status:INVALID_LOG_PAGE(2109)` error message. What you think about it?

keithbusch:
Right, but that means you can never request log-id 0. It's in the valid log range. It's just marked reserved for future use.

rrosatti:
Right, good catch! Well, so I believe for this one we do not have much to do, right? I mean, people should always specify the log id when using the get-log command.

keithbusch:
Fixed with https://github.com/linux-nvme/nvme-cli/commit/44044e6c3c6fc1d4813163d5054c0750ef88ee06.

PR #321, opened 2018-03-19
Add a "friendly" error message on create_ns()

rrosatti:
If you try to create a namespace when you already have reached the
maximum number of namespaces, we're throwing the following error: NVMe Status:NS_ID_UNAVAILABLE(116).
The idea of this commit is just to add a more friendly error message
together with the current message:

NVMe Status:NS_ID_UNAVAILABLE(116)
The number of namespaces supported has been exceeded!

Signed-off-by: Rodrigo R. Galvao <rosattig@linux.vnet.ibm.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1016,10 +1016,13 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 	err = nvme_ns_create(fd, cfg.nsze, cfg.ncap, cfg.flbas, cfg.dps, cfg.nmic, &nsid);
 	if (!err)
 		printf("%s: Success, created nsid:%d\n", cmd->name, nsid);
-	else if (err > 0)
+	else if (err > 0) {
 		fprintf(stderr, "NVMe Status:%s(%x)\n",
 					nvme_status_to_string(err), err);
-	else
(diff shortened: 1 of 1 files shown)


rrosatti:
The idea of this message is just to make it clear what's this error about. I don't know whether it's relevant or not, but taking myself as an example: I'm just starting to dive into the "nvme world" and I still do not know much about the errors and stuff. So, I had to look that up in the Spec.

keithbusch:
If people find the current decoding too cryptic (I'd agree that it isn't always immediately obvious), I would prefer seeing this addressed generically for all error codes in nvme_status_to_string().

rrosatti:
Cool! I think I can work on it, if you think it's valid.

rrosatti:
@keithbusch Keith, I'm looking at the nvme_status_to_string() code and I see it wouldn't be that easy to "add" these message errors there, right? Do you think we could work that out in another way?

I'd really like to help with something, so if you could give me some directions it would be good =)

keithbusch:
Yes, I think instead of using the stringified name of the #define's, if we use the string in NVMe standard "Description" colum of Figures 31-33 (rev1.3.a), that would be a nice improvement.

rrosatti:
Ok. I'll work on that, then!
Thanks!!

PR #334, opened 2018-04-09
nvme-cli: replace input val to returned one when printing result

minwooim:
Make it print out the returned value from the Set Feature command   
instead of the value from the user so that user can see the result  
in CDW0 directly just like get_feature().  

Signed-off-by: Minwoo Im \<minwoo.im.dev@gmail.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2586,7 +2586,7 @@ static int set_feature(int argc, char **argv, struct command *cmd, struct plugin
 		perror("set-feature");
 	} else if (!err) {
 		printf("set-feature:%02x (%s), value:%#08x\n", cfg.feature_id,
-			nvme_feature_to_string(cfg.feature_id), cfg.value);
+			nvme_feature_to_string(cfg.feature_id), result);
 		if (buf) {
 			if (cfg.feature_id == NVME_FEAT_LBA_RANGE)
 				show_lba_range((struct nvme_lba_range_type *)buf,


keithbusch:
While I'd agree this is the behavior we'd want to see out of controllers, the only feature the specification requires the value of the result be set in CQE DW0 is feature 7, "Number of Queues". The remaining features all say only "Get Feature" returns the value in CQE DW0, but nothing about "Set Feature".

minwooim:
As you said, FID 0x7(NOQ) is the only feature taking DW0 of CQE.  
IMO, however, to avoid confusion It would be better the printed result _should_ be a value from the controller, not from the user since the user will definitely know what the input value is.  

What about to print the DW0 value only for FID 0x7 (NOQ) with no printing-out for the remaining features?

keithbusch:
If you're issuing a Number of Queues set-feature from nvme-cli, the specification says just going to get an error, so that command shouldn't be going through the success path the prints the result.

For the rest of the features, the specification for those features is either the user requested value was applied, or you get an error, so printing the user requested value on success doesn't seem like a bad thing.

minwooim:
@keithbusch , Thanks for your kindly response.
As you gave me directives, Set Feature with FID 0x7 seems not proper with nvme-cli in a normal use case like below statement in spec.
> This feature shall only be issued during initialization prior to the creation of any I/O queues.

Thanks again. :)

PR #336, opened 2018-04-09
nvme-cli: fix file descriptor leaks in subcommands

minwooim:
Fix file descriptor(fd) leaks after getting an invalid nsid from the file
descriptor in subcommands.

Signed-off-by: Minwoo Im \<minwoo.im.dev@gmail.com>


minwooim:
We can find the proper one here.
1b7a0ed435af4a14dbc9d6fd2a6d80ad803c0d5f

PR #340, opened 2018-04-13
Fixes for log page access.

AndrewMeir:
Some small corrections to vendor log page contents.


AndrewMeir:
Thanks Keith. Once again, I'm majorly impressed at your fast response time!
There is support for raw binary output when the filename arg is given. Perhaps the arg syntax could be changed to make this behaviour more intuitive and/or consistent with other commands. Feel free to suggest changes. I can also review this over the next few days.
Thanks again,
Andrew Meir

PR #342, opened 2018-04-15
nvme-cli: add support for id-nvmset subcmmand

minwooim:
From 4e48e94751bf85469faff5f45e26f91e438c11a2 Mon Sep 17 00:00:00 2001
From: Minwoo Im \<minwoo.im.dev@gmail.com>
Date: Mon, 16 Apr 2018 03:01:50 +0900
Subject: [PATCH 0/3] nvme-cli: add support for id-nvmset subcmmand

This patchset adds Identify NVM Set List command.
It also updates new fields to Identify Namespace data structure.

Minwoo Im (3):
  nvme-cli: add support for id-nvmset subcommand
  nvme-cli: update new fields in id-ns subcommand
  nvme-cli: update the document for id-nvmset subcommand

 Documentation/nvme-id-nvmset.1    | 147 +++++++
 Documentation/nvme-id-nvmset.html | 869 ++++++++++++++++++++++++++++++++++++++
 Documentation/nvme-id-nvmset.txt  |  78 ++++
 linux/nvme.h                      |  24 +-
 nvme-builtin.h                    |   1 +
 nvme-ioctl.c                      |  13 +-
 nvme-ioctl.h                      |   2 +
 nvme-print.c                      |  59 +++
 nvme-print.h                      |   2 +
 nvme.c                            |  64 +++
 10 files changed, 1257 insertions(+), 2 deletions(-)
 create mode 100644 Documentation/nvme-id-nvmset.1
 create mode 100644 Documentation/nvme-id-nvmset.html
 create mode 100644 Documentation/nvme-id-nvmset.txt

-- 
2.7.4

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -949,6 +949,63 @@ void show_nvme_id_ctrl(struct nvme_id_ctrl *ctrl, unsigned int mode)
 	__show_nvme_id_ctrl(ctrl, mode, NULL);
 }
 
+void show_nvme_id_nvmset(struct nvme_id_nvmset *nvmset)
+{
+	int i;
+
+	printf("nid     : %d\n", nvmset->nid);
+	printf(".................\n");
+	for (i = 0; i < nvmset->nid; i++) {
+		printf(" NVM Set Attribute Entry[%2d]\n", i);
+		printf(".................\n");
+		printf("nvmset_id               : %d\n",
+				le16_to_cpu(nvmset->ent[i].id));
+		printf("enduracne_group_id      : %d\n",
+				le16_to_cpu(nvmset->ent[i].endurance_group_id));
+		printf("random_4k_read_typical  : %u\n",
+				le32_to_cpu(nvmset->ent[i].random_4k_read_typical));
+		printf("optimal_write_size      : %u\n",
+				le32_to_cpu(nvmset->ent[i].opt_write_size));
+		printf("total_nvmset_cap        : %.0Lf\n",
+				int128_to_double(nvmset->ent[i].total_nvmset_cap));
+		printf("unalloc_nvmset_cap      : %.0Lf\n",
+				int128_to_double(nvmset->ent[i].unalloc_nvmset_cap));
+		printf(".................\n");
+	}
+}
+
+void json_nvme_id_nvmset(struct nvme_id_nvmset *nvmset, const char *devname)
+{
+	struct json_object *root;
+	struct json_array *entries;
+	__u32 nent = le32_to_cpu(nvmset->nid);
+	int i;
+
+	root = json_create_object();
+
+	json_object_add_value_int(root, "nid", nent);
+
+	entries = json_create_array();
+	for (i = 0; i < nent; i++) {
+		struct json_object *entry = json_create_object();
+
+		json_object_add_value_int(entry, "nvmset_id", le16_to_cpu(nvmset->ent[i].id));
+		json_object_add_value_int(entry, "endurance_group_id", le16_to_cpu(nvmset->ent[i].endurance_group_id));
+		json_object_add_value_int(entry, "random_4k_read_typical", le32_to_cpu(nvmset->ent[i].random_4k_read_typical));
+		json_object_add_value_int(entry, "optimal_write_size", le32_to_cpu(nvmset->ent[i].opt_write_size));
+		json_object_add_value_float(entry, "total_nvmset_cap", int128_to_double(nvmset->ent[i].total_nvmset_cap));
+		json_object_add_value_float(entry, "unalloc_nvmset_cap", int128_to_double(nvmset->ent[i].unalloc_nvmset_cap));
+
+		json_array_add_value_object(entries, entry);
+	}
+
+	json_object_add_value_array(root, "NVMSet", entries);
+
+	json_print_object(root, NULL);
+	printf("\n");
+	json_free_object(root);
+}
+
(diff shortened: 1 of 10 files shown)


nvme-print.c
```diff
@@ -949,6 +949,61 @@ void show_nvme_id_ctrl(struct nvme_id_ctrl *ctrl, unsigned int mode)
...
+				int128_to_double(nvmset->ent[i].unalloc_nvmset_cap));
+		printf(".................\n");
+	}
+}
+
+void json_nvme_id_nvmset(struct nvme_id_nvmset *nvmset, const char *devname)
+{
+	struct json_object *root;
+	struct json_array *entries;
+	__le32 nent = le32_to_cpu(nvmset->nid);
```
tim-oleksii: __u32 might be better, as ARCH is not guaranted to be LE

nvme.c
```diff
@@ -1759,6 +1759,70 @@ static int id_ns(int argc, char **argv, struct command *cmd, struct plugin *plug
...
+		fmt = BINARY;
+
+	err = nvme_identify_nvmset(fd, cfg.nvmset_id, &nvmset);
+	if (!err) {
+		if (fmt == BINARY)
+			d_raw((unsigned char *)&nvmset, sizeof(nvmset));
+		else if (fmt == JSON)
+			json_nvme_id_nvmset(&nvmset, devicename);
+		else {
+			printf("NVME Identify NVM Set List %d:\n", cfg.nvmset_id);
```
tim-oleksii: Should it be like this ? I mean ':' at the end? You probably meant something like 'NVM Set List: %d\n'

minwooim: The next line show-nvme_id_nvmeset() will show the remaining NVM Set list.

nvme.c
```diff
@@ -1759,6 +1759,70 @@ static int id_ns(int argc, char **argv, struct command *cmd, struct plugin *plug
...
+	};
+
+	struct config cfg = {
+		.nvmset_id = 0,
+		.output_format = "normal",
+	};
+
+	const struct argconfig_commandline_options command_line_options[] = {
+		{"nvmset_id",       'i', "NUM", CFG_POSITIVE, &cfg.nvmset_id,       required_argument, nvmset_id},
+		{"raw-binary",      'b', "",    CFG_NONE,     &cfg.raw_binary,      no_argument,       raw_binary},
```
tim-oleksii: 'raw-binary' seems superfluous and breaks single point of truth. why did you introduced it?

nvme.c
```diff
@@ -1759,6 +1759,70 @@ static int id_ns(int argc, char **argv, struct command *cmd, struct plugin *plug
...
+		{"output-format",   'o', "FMT", CFG_STRING,   &cfg.output_format,   required_argument, output_format },
```
tim-oleksii: output_format has to bring some help message but i do not see it defined/declared? could you please point out?

for instance, https://github.com/linux-nvme/nvme-cli/blob/d94fda25db81ba5b43b18c05dadff489d50ea0d2/wdc-nvme.c#L1469

minwooim: You can find it at line 155.
`static const char *output_format = "Output format: normal|json|binary";`

It has been declared for a long time as a static value in nvme.c.

nvme.c
minwooim: Why do you think it's 'superfluous' ?
I believe there might be an use case that user wants to pipe or redirect the output of this command to the other one.

nvme.c
tim-oleksii: OK

nvme.c
tim-oleksii: I think that '-o binary' can be sufficient to handle the same. Otherwise, it can rise questions why '-b' has higher priority. The result of '-o binary > /tmp/out.bin' and '-b > /tmp/out.bin' still the same.

nvme-print.c
```diff
@@ -949,6 +949,61 @@ void show_nvme_id_ctrl(struct nvme_id_ctrl *ctrl, unsigned int mode)
...
+		json_object_add_value_int(entry, "endurance_group_id", le16_to_cpu(nvmset->ent[i].endurance_group_id));
+		json_object_add_value_int(entry, "random_4k_read_typical", le32_to_cpu(nvmset->ent[i].random_4k_read_typical));
+		json_object_add_value_int(entry, "optimal_write_size", le32_to_cpu(nvmset->ent[i].opt_write_size));
+		json_object_add_value_float(entry, "total_nvmset_cap", int128_to_double(nvmset->ent[i].total_nvmset_cap));
+		json_object_add_value_float(entry, "unalloc_nvmset_cap", int128_to_double(nvmset->ent[i].unalloc_nvmset_cap));
+
+		json_array_add_value_object(entries, entry);
+	}
+
+	json_print_object(root, NULL);
```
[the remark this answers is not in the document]
minwooim: Thanks for finding this.

nvme.c
minwooim: I'm not sure about what history was there about this, but many subcommands support both of them.
I was trying to follow the previous one, but I think some needs to be revised at one time if decided not to be used.
What do you say?

tim-oleksii: I'd got with '-o', but @keithbusch  is the one to decide.

keithbusch: I'd really like to deprecate the "--raw-binary" and use the output-format instead.

minwooim: Thanks for your reviews.
I have updated pull request with deprecation of --raw-binary.

minwooim:
I have removed the --raw-binary option to deprecate. Also branch rebased.

PR #361, opened 2018-06-07
Specific Telemetry Data Area Retrieval

maximveligan:
Wrote a patch to allow users to specify which data area region they want to retrieve from the telemetry log.


nvme.c
```diff
@@ -288,47 +293,67 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
...
+	} else {
+		fprintf(stderr, "Invalid data area requested");
+		goto free_hdr;
+	}
+
+	/* In case full_size is somehow not a multiple of 512, return from this function
+	 * to prevent an infinite loop.
+	 */
+	if (full_size % bs) {
+		fprintf(stderr, "The full size of the data area requested is not rounded to the nearest block.");
```
keithbusch: Need a newline here.

nvme.c
```diff
@@ -288,47 +293,67 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
...
-	page_log = malloc(4096);
+	/* Setting offset at blocksize since the first block is the header
+	 * and is already pulled
+	 */
+	offset = bs;
+	page_log = malloc(bs);
 	if (!page_log) {
 		fprintf(stderr, "Failed to allocate %zu bytes for log\n",
-			full_size);
+			bs);
```
keithbusch: This can all go on one line now.

nvme.c
```diff
@@ -288,47 +293,67 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
...
+		if (err != bs) {
 			fprintf(stderr, "Failed to flush all data to file!");
 			break;
 		}
-		full_size -= 4096;
-		offset += 4096;
+		offset += bs;
 	}
-
+
```
keithbusch: Some white space damage here?

keithbusch:
okay, i'll squash this for you, bur we're still missing you're signed-off-by in the commit log.

nvme.c
```diff
@@ -276,10 +281,10 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
 
 	cfg.host_gen = !!cfg.host_gen;
 
-	hdr = malloc(4096);
-	memset(hdr, 0, 4096);
+	hdr = malloc(bs);
+	memset(hdr, 0, bs);
```
tim-oleksii: Might be great to check againt NULL as it's done at the line 305.

nvme.c
```diff
@@ -288,47 +293,67 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
...
+	 */
+	if (!cfg.data_area || cfg.data_area == 3) {
+		full_size = (hdr->dalb3 * bs) + offset;
+	} else if (cfg.data_area == 1) {
+		full_size = (hdr->dalb1 * bs) + offset;
+	} else if (cfg.data_area == 2) {
+		full_size = (hdr->dalb2 * bs) + offset;
+	} else {
+		fprintf(stderr, "Invalid data area requested");
+		goto free_hdr;
```
tim-oleksii: page_log leaks here if error is occurred..

nvme.c
```diff
@@ -328,6 +333,7 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
 	 */
 	if (full_size % bs) {
 		fprintf(stderr, "The full size of the data area requested is not rounded to the nearest block\n");
+		free(page_log);
```
tim-oleksii: This one is OK,  but moving free(page_log), into free_hdr might seem better as well as renaming free_hdr into free_mem. :) never mind, it's just personal preference.

maximveligan:
I've applied all changes suggested by @tim-oleksii and @keithbusch and signed off and squashed all the commits into one. Please let me know if there is more I need to change. Thanks!

nvme.c
```diff
@@ -276,10 +281,15 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
 
 	cfg.host_gen = !!cfg.host_gen;
 
-	hdr = malloc(4096);
-	memset(hdr, 0, 4096);
+	hdr = malloc(bs);
+	memset(hdr, 0, bs);
+	if (!hdr) {
```
tim-oleksii: It seems that you memset null pointer before checking it.

nvme.c
maximveligan: I agree with you. I was going to do that, but noticed that on line 289 we use goto free_hdr, and there is no page_log allocated yet. I was worried that attempting to free memory that has not been allocated would cause problems.

nvme.c
```diff
@@ -288,47 +298,64 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
...
+	default:
+		fprintf(stderr, "Invalid data area requested");
+		goto free_hdr;
+	}
+	/* In case full_size is somehow not a multiple of 512, return from this function
+	 * to prevent an infinite loop.
+	 */
+	if (full_size % bs) {
+		fprintf(stderr, "The full size of the data area requested is not rounded to the nearest block\n");
+		free(page_log);
```
tim-oleksii: This one is OK, but it might be better to stick with policy of cleaning up memory at the end of function. 

renaming free_hdr into free_mem will do the trick, as well as moving free(page_log) one line below its current location.

this is just personal preference.

maximveligan: Wouldn't putting free(page_log) inside of free_hdr and renaming it to free_mem give us issues if we attempt to goto it before page_log is malloced?

tim-oleksii: as long as page_log is null, it's ok. 


https://linux.die.net/man/3/free
The free() function frees the memory space pointed to by ptr, which must have been returned by a previous call to malloc(), calloc() or realloc(). Otherwise, or if free(ptr) has already been called before, undefined behavior occurs. If ptr is NULL, no operation is performed.

maximveligan: so you think initializing it to null would be better?

tim-oleksii: yep, it'd nice, but it's up to you.

maximveligan: I decided to just malloc page_log right after hdr, and then check for NULL on both right after. That way there's no need to initialize page_log to NULL, and there will never be a case where hdr is malloc'ed but page_log isn't.

nvme.c
```diff
@@ -358,11 +349,10 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
 		}
 		offset += bs;
 	}
-	free(page_log);
- free_hdr:
+	goto free_mem;
```
keithbusch: What's with the goto the very next line?! That just looks weird.

maximveligan: If we don't free page_log and hdr, there will be a memory leak. Beforehand only page_log was being freed at the end. Because there is already a goto that does this (free_mem), I thought it would be better to use it rather than typing out `free(page_log); free(hdr);`.

keithbusch: The code execution just calls through the label, though. You don't need to 'goto' the very next line of code as that's going to happen anyway.

nvme.c
keithbusch: err = EINVAL in this default case.

nvme.c
maximveligan: Ah got it. I thought the label was only called when you wrote goto. Sorry about that. Removed.

nvme.c
```diff
@@ -275,65 +280,78 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
...
-	err = nvme_get_telemetry_log(fd, hdr, cfg.host_gen, cfg.ctrl_init, 4096, 0);
+	memset(hdr, 0, bs);
+	err = nvme_get_telemetry_log(fd, hdr, cfg.host_gen, cfg.ctrl_init, bs, 0);
 	if (err) {
 		fprintf(stderr, "NVMe Status:%s(%x)\n",
 			nvme_status_to_string(err), err);
 		fprintf(stderr, "Failed to aquire telemetry header %d!\n", err);
-		free(hdr);
-		goto close_output;
+		close(output);
```
tim-oleksii: Could you please return close(output) into close_output or close_fd? Fallthrough label flow will close it anyway.

keithbusch:
I manually merged this one.

PR #363, opened 2018-06-08
nvme-cli: Implemented Device self test and log.

RevanthRajashekar:
Implmented device self test and log gathering described in the NVMe
1.3 specificiation.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -1224,6 +1224,74 @@ void show_smart_log(struct nvme_smart_log *smart, unsigned int nsid, const char 
 	printf("Thermal Management T2 Total Time    : %u\n", le32_to_cpu(smart->thm_temp2_total_time));
 }
 
+void show_self_test_log(struct nvme_self_test_log *self_test, const char *devname)
+{
+	int i, temp;
+	const char *test_code_res;
+	const char *test_res[10] = {
+		"Operation completed without error",
+		"Operation was aborted by a Device Self-test command",
+		"Operation was aborted by a Controller Level Reset",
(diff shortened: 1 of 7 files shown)


nvme-print.c
```diff
@@ -1224,6 +1224,74 @@ void show_smart_log(struct nvme_smart_log *smart, unsigned int nsid, const char
...
+		printf("\nPrinting the contents of REPORT-%d\n",(i+1));
+
+		temp = self_test->result[i].device_self_test_status & 0xf;
+		if (temp == 0xf){
+			printf("Entry not used (does not contain a test result)\n");
+			continue;
+		}
+		if(temp > 8)
+			temp = 9;
+		printf("Self test result - %#x: %s\n", temp, test_res[temp]);
```
keithbusch: I understand why you're over-writing 'temp' to 9 to get the "Reserved" test result, but I'd really like to see the value that the drive actually returned here.

nvme.c
```diff
@@ -1813,6 +1813,116 @@ static int get_ns_id(int argc, char **argv, struct command *cmd, struct plugin *
...
+	fd = parse_and_open(argc, argv, desc, command_line_options, &cfg, sizeof(cfg));
+	if (fd < 0)
+		return fd;
+
+	err = nvme_self_test_start(fd, cfg.namespace_id, cfg.cdw10);
+	if (!err) {
+			if ((cfg.cdw10 & 0xf) == 0xf)
+				printf("Aborting device self-test operation\n");
+			else
+				printf("Device self-test started\n");
```
keithbusch: These code blocks have an extra indentation that makes it a little odd to read.

PR #364, opened 2018-06-08
nvme-cli: Implemented Device self test and log.

RevanthRajashekar:
Implmented device self test and log gathering described in the NVMe
1.3 specificiation.


keithbusch:
Manually merged with fixups.

PR #366, opened 2018-06-10
Request to add Seagate Plugin 

xahmad:
This plugin will support the current and any future NVMe devices from Seagate.


keithbusch:
Thanks, this all looks in order to me.

PR #368, opened 2018-06-11
Making gcc compiler (5.4.0 among others) happy

xahmad:
I was getting this on gcc 5.4.0
nvme-print.c:1340:10: error: format ‘%lx’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘long long unsigned int’ [-Werror=format=]
   printf("  Power on hours (POH)         : %#"PRIx64"\n",
          ^


tim-oleksii:
It's been just fixed with https://github.com/linux-nvme/nvme-cli/commit/2e65d5ef3b994796c33dfd316bf87f6932d125c5

xahmad:
ok thank you

PR #370, opened 2018-06-11
nvme-cli: Implemented get and set timestamp features

RevanthRajashekar:
Implmented get and set timestamp feature as described in the NVMe
1.3 specificiation.


keithbusch:
Hi Revanth, this feature doesn't really deserve a top-level command like this. This is just a sub-component of the get/set feature command, which already exists, so it just needs to be augmented with the feature identifier and any other decoding specifics.

PR #383, opened 2018-06-28
nvme-cli: add support of RAE

tim-oleksii:
Currently telemetry-log cannot retain content of corresponding log page as RAE is not set.
This patch adds RAE bit support into routines to retrieve log pages.

Signed-off-by: Alexey Timofeyev <alexey.timofeyev@sk.com>


tim-oleksii:
Thanks Keith!

PR #390, opened 2018-07-15
nvme-print.c : fix ns-descs JSON output

bdeyal:
Before fix string is truncated, not enough information.
CMD: nvme ns-descs /dev/nvme0n1 -o json
> {
>     "nguid" : "d800000000000000"
> }

After Fix, string displayed correctly, JSON is in array format and with more information:
CMD ./nvme ns-descs /dev/nvme0n1 -o json
> {
>    "ns-descs" : [
>       {
>          "nidt" : 2,
>          "nidl" : 16,
>          "type" : "nguid",
>          "nguid" : "d80000000000000004009902ad005076"
>    }
>  ]
> }

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -677,56 +677,63 @@ void show_nvme_id_ns(struct nvme_id_ns *ns, unsigned int mode)
 	}
 }
 
+
 void json_nvme_id_ns_descs(void *data)
 {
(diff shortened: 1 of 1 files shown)


minwooim:
The branch that you want it to be merged from seems to have unnecessary commits which
are not related to JSON output format of ns-descs.

bdeyal:
Removed unrelated commits

PR #393, opened 2018-07-25
Remove trailing "\n" so that the output is valid json

puneetk:
## Summary of Changes
* Remove trailing "\n" so that the output is valid json


puneetk:
Thank you

PR #402, opened 2018-08-12
json.c - fix crash with empty strings, fix escape for special chars.

bdeyal:
Fix program crashed when strdup escape got "" as input.
Fix json generation  - covers more characters that needed escape

### DIFF
--- a/json.c
+++ b/json.c
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <errno.h>
 #include <stdarg.h>
+#include <ctype.h>
 #include "json.h"
 
 static inline void fail_and_notify(void)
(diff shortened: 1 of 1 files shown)


keithbusch:
what path is leading to print strings with non-printable characters?

bdeyal:
Added for completeness. This may be useful for cases where controllers (e.g. in development) return junk when ascii characters are expected. I still  want to print JSON correctly.

keithbusch:
we should just bail if any character fails 'isprint'. We don't aim to support non-compliant drives.

bdeyal:
So other than '\"' and '\\' nothing should be escaped, that means old code is OK except for segfault in case of empty string "" field

bdeyal:
Note:
In the examples below, /dev/nvme0 is a discovery controller (on an io controller everything's fine)
```
# nvme-upstream version
nvme version 1.6.19.ge3da

# nvme-upstream id-ctrl /dev/nvme0 -o json
Allocation of memory for json object failed, aborting.
Aborted (core dumped)

# nvme-fix_json version
nvme version 1.6.14.g6c31

# nvme-fix_json id-ctrl /dev/nvme0 -o json
(Prints empty strings in json OK)
```

bdeyal:
Please comment if no merge, 
I will close iif so.
Thanks.

bdeyal:
If id-ctrl JSON output for a discovery controller is not important then just leave it as it is now.

PR #409, opened 2018-09-03
Fix f28 (gcc 8.1) compilation

bdeyal:
Original:
```
nvme.c:1253:48: error: ‘snprintf’ output may be truncated before the last format character [-Werror=format-truncation=]
  snprintf(sspath, sizeof(sspath), "%s/subsysnqn", path);
                                                ^
```


bdeyal:
Verified my change compiles OK with older gcc (RHEL7 gcc 4.8), and clang as well.

bdeyal:
Fixed elsewhere. Closing pull request.

PR #413, opened 2018-09-20
attach-ns : fix segfault for ctrl list wrong input

bdeyal:
Original
```
# nvme attach-ns /dev/nvme0 -n 1 -c /dev/nvme1
```

After fix:
```
# ./nvme attach-ns /dev/nvme0 -n 1 -c /dev/nvme1
attach-ns: controller id list is required
```


bdeyal:
Also : changed the help message to controller id list.

PR #415, opened 2018-09-29
nvme-cli: make gen-hostnqn output stable, host specific IDs

zdzichu:
This changes generation of hostnqn to depend on `/etc/machine-id`. Each invocation of gen-hostnqn on the same host will generate the same ID. Different hosts still get different ids.
  This makes gen-hostnqn idempotent.

  `NVME_APPLICATION_ID` was randomly generated.


amluto:
I think there’s a better solution: teach all the code that reads /etc/nvme/host am to use the systemd-generated id if the file doesn’t exist.

zdzichu:
At the moment there's only a single place:
https://github.com/linux-nvme/nvme-cli/blob/master/fabrics.c#L465

hreinecke:
We have reworked the hostnqn generation, so this patch is no longer needed.

PR #420, opened 2018-10-14
Portability fixes

bsdimp:
There's a series of commits that fix many Linux-specific idioms where more portable constructs work just as well. It's part of a larger series that ports things to FreeBSD, but the rest of the series aren't ready yet as to complete that port some structural changes will be needed and I've not settled on the proper way to do them. These are needed regardless of any future work, however, and should be generally uncontroversial. It only makes three real code changes: uses the proper unsigned long for the command argument to ioctl where it's passed from function to function, removes a needless const on a char * argument, and gets timezone offsets from standard interfaces rather than glibc specific exported globals. Other than that, I'm just removing needless includes and preferring POSIX include files to Linux specific ones.
This is my first pull request to nvme-cli, so if there's some guidelines I've missed that I don't yet conform to, a simple pointer is all that would be needed.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -24,7 +24,6 @@
  * This program uses NVMe IOCTLs to run native nvme commands to a device.
  */
 
-#include <endian.h>
 #include <errno.h>
 #include <getopt.h>
 #include <fcntl.h>
--- a/nvme.c
+++ b/nvme.c
@@ -36,6 +36,7 @@
 #include <unistd.h>
 #include <math.h>
 #include <dirent.h>
+#include <libgen.h>
 
 #include <linux/fs.h>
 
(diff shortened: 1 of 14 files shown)


nvme.c
```diff
@@ -36,6 +35,7 @@
 #include <unistd.h>
 #include <math.h>
 #include <dirent.h>
+#include <libgen.h>
```
myfreeweb: Keep in mind that on Linux/glibc, `<libgen.h>` activates the POSIX version of basename (that mutates the string), while it seems to have the exact opposite effect on FreeBSD 12. (Just today was dealing with this in libinput :D)

So it might be a good idea to `strdup` the argument of `basename` below

bsdimp: In both it activates the non-const version of basename (I couldn't actually find a const version in Linux, but I may have looked in the wrong places). Both will modify the argument, but if you trace back where the argument actually came from, and when such modifications take place, you'll see there's nothing to worry about. It took me a bit of tracing through the code to convince myself that this is correct in both places, both in terms of implementation, and in terms of what the spec says.

keithbusch:
Looks good to me. I'll merge this in.

PR #425, opened 2018-10-24
Refactor plugins in a file hierarchy

bdeyal:
Put vendor plugins in a separate hierarchy.
Add a target in makefile that verifies builtin code does not depend on anything in the plugins.

hierarchy is simple: intel plugin files are in plugins/intel/  etc

This patch is only file structure. No functional change.

Verify build on Fedora 28 (gcc 8.1.1) , RHEL 7.x (gcc 4.8) and Ubuntu 14.04 (gcc 4.8)
On rpm systems (fedora and RHEL) rpm built cleanly.


keithbusch:
Yeah, we started out with a single 'c' file, and didn't adapt a more pleasant hierarchy as things were added. Your proposal looks fine, and maybe we can do even better with the remaining components in the future.

MatiasBjorling:
Would it make sense to place them all in a single vendor directory, all seems to only have two files.

bdeyal:
It's only two files (except for wdc) because of the flat structure. Now each vendor can restructure the code to more files as they wish. I think it is more convenient for the plugin writer to have a separate vendor location for the plugin code.

PR #433, opened 2018-11-22
id-ns: Fix printing of large ANAGRPID values

freezer5:


keithbusch:
Ha, almost everything here should probably be unsigned prints. Thanks for the fix.

PR #435, opened 2018-11-29
micron-nvme: Replace direct use of ioctl

hjmallon:
Where possible it is best to use the redirections, so that any future usb bridges etc (where the nvme command will have to do translation) will work.


keithbusch:
An additional improvement could export nvme_submit_admin_passthru().

PR #439, opened 2018-12-10
nvme-cli: Added support for virtualization-management command

Sushmakk:
Previous feedback, **(http://lists.infradead.org/pipermail/linux-nvme/2018-June/018724.html)**, on the implementation of NVMe 1.3's virtualization feature for nvme-cli requested that the change be tested against a device actually supporting the feature.  This has now been completed internally at Intel, and this pull request merges the previous patch into the latest nvme-cli sources.

### DIFF
--- a/nvme-builtin.h
+++ b/nvme-builtin.h
@@ -67,6 +67,7 @@ COMMAND_LIST(
 	ENTRY("gen-hostnqn", "Generate NVMeoF host NQN", gen_hostnqn_cmd)
 	ENTRY("dir-receive", "Submit a Directive Receive command, return results", dir_receive)
 	ENTRY("dir-send", "Submit a Directive Send command, return results", dir_send)
+	ENTRY("virt-mgmt", "Manage Flexible Resources between Primary and Secondary Controller ", virtual_mgmt)
 );
 
 #endif
--- a/nvme-ioctl.c
+++ b/nvme-ioctl.c
@@ -867,3 +867,19 @@ int nvme_self_test_start(int fd, __u32 nsid, __u32 cdw10)
 
 	return nvme_submit_admin_passthru(fd, &cmd);
 }
+
+int nvme_virtual_mgmt(int fd, __u32 cdw10, __u32 cdw11, __u32 *result)
+{
+	struct nvme_admin_cmd cmd = {
+		.opcode = nvme_admin_virtual_mgmt,
+		.cdw10  = cdw10,
+		.cdw11  = cdw11,
+	};
+	int err;
+
+	err = nvme_submit_admin_passthru(fd, &cmd);
+	if (!err && result)
+		*result = cmd.result;
+
+	return err;
+}
--- a/nvme.c
+++ b/nvme.c
@@ -2148,6 +2148,74 @@ static int get_ns_id(int argc, char **argv, struct command *cmd, struct plugin *
 	return 0;
 }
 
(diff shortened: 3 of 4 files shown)


nvme.c
```diff
@@ -2148,6 +2148,69 @@ static int get_ns_id(int argc, char **argv, struct command *cmd, struct plugin *
...
+                "3. 8 - Secondary Controller Assign\n"\
+                "4. 9 - Secondary Controller Online";
+        const char *cdw11 = "Indicate the Number of Controller Resources (NR)";
+        int fd, err;
+ 
+        struct config {
+                int     cntlid;
+                int     rt;
+                int     act;
+                __u32   cdw10;
```
keithbusch: cdw10 should just be normal stack variables rather than within the config struct. Variables in 'struct config' should only be there if they are command line parameters. For cdw11, you might want to give it a more descriptive name, like "nr".

If you want to provide cdw10 and 11 directly as an override, like for forward compatibility to access the currently reserved fields, that might be okay too.

nvme.c
```diff
@@ -2148,6 +2148,69 @@ static int get_ns_id(int argc, char **argv, struct command *cmd, struct plugin *
...
+        fd = parse_and_open(argc, argv, desc, command_line_options, &cfg, sizeof(cfg));
+        if (fd < 0)
+                return fd;
+ 
+        cfg.cdw10 = cfg.cntlid << 16;
+        cfg.cdw10 = cfg.cdw10 | (cfg.rt << 8);
+        cfg.cdw10 = cfg.cdw10 | cfg.act;
+ 
+        err = nvme_virtual_mgmt(fd, cfg.cdw10, cfg.cdw11);
+        if (err > 0) {
```
keithbusch: We should have a "success" message if !err, and print the result of the command when successful.

nvme-ioctl.c
```diff
@@ -867,3 +867,15 @@ int nvme_self_test_start(int fd, __u32 nsid, __u32 cdw10)
 
 	return nvme_submit_admin_passthru(fd, &cmd);
 }
+
+int nvme_virtual_mgmt(int fd, __u32 cdw10, __u32 cdw11)
```
keithbusch: This one should take a __u32 *result field since the virtualization mgmt command does return something informational in CQE DW0, and if successful, we want to print that out. See nvme_get_feature, as an example of what I mean.

nvme-builtin.h
```diff
@@ -67,6 +67,7 @@ COMMAND_LIST(
 	ENTRY("gen-hostnqn", "Generate NVMeoF host NQN", gen_hostnqn_cmd)
 	ENTRY("dir-receive", "Submit a Directive Receive command, return results", dir_receive)
 	ENTRY("dir-send", "Submit a Directive Send command, return results", dir_send)
+	ENTRY("virtualization-management", "Manage Flexible Resources between Primary and Secondary Controller ", virtual_mgmt)
```
keithbusch: That is a bit verbose of a name for a shell utility. Let's call it "virt-mgmt".

nvme.c
Sushmakk: I agree and I will change n to nr, and what would you suggest for rest of them too like ctrlid for c, rtype for r(resource type) and act for a. I feel instead of letters if we have descriptive names easy to use and understand

nvme-builtin.h
Sushmakk: I agree and will update it

PR #440, opened 2018-12-14
timeout option added to ns_create/delete

jaredeh:
We've been working with some devices that require longer than standard namespace operation timeouts.  Using this to work around that limitation.


keithbusch:
Sure, I understand that some implementations may find this useful.

Could you squash your patch down so it's just a single commit? Also, please update the doc/ .txt files for these commands.

kqnfd:
could anyone please tell me how to run the tests cases in python???
currently I have tried "python nvme_fw_log_test.py" on the ubantu command line but nothing happended, no prints or visible results at all.
waiting for reply from some kind guys....

PR #445, opened 2019-01-08
Format command helper to specify block size

benreese0:
Adding a helper to format command to specify block size and have nvme-cli determine the associated LBAF.

PR #449, opened 2019-01-15
Adding a Format Helper to create-ns

benreese0:
Similar to #445 however adds the helper to create-ns command instead of format. Allows user to select target block size in a create-ns command instead of requiring FLBAS to be specified directly.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1163,8 +1163,10 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 	const char *flbas = "FLBA size";
 	const char *dps = "data protection capabilities";
 	const char *nmic = "multipath and sharing capabilities";
+	const char *bs = "target block size";
 
-	int err = 0, fd;
+	int err = 0, fd, i;
+	struct nvme_id_ns ns;
 	__u32 nsid;
 
 	struct config {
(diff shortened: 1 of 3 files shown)


nvme.c
```diff
@@ -1184,13 +1189,52 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
 		{"nmic",  'm', "NUM", CFG_BYTE,        &cfg.nmic,  required_argument, nmic},
+		{"block-size",   'b', "NUM",  CFG_LONG_SUFFIX, &cfg.bs,        required_argument, bs},
 		{NULL}
 	};
 
 	fd = parse_and_open(argc, argv, desc, command_line_options, &cfg, sizeof(cfg));
 	if (fd < 0)
 		return fd;
 
+	if (cfg.flbas != 0xFF && cfg.bs != 0x00) {
```
keithbusch: What happens if flbas == 0xff && cfg.bs == 0?

nvme.c
```diff
@@ -1184,13 +1189,52 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
+		if (err) {
+			if (err < 0)
+				perror("identify-namespace");
+			else
+				fprintf(stderr,
+					"NVME Admin command error:%s(%x)\n",
+					nvme_status_to_string(err), err);
+			return err;
+		}
+		for (i=0; i<16; ++i) {
```
keithbusch: Formatting: spaces between operators:

`for (i = 0; i < 16; ++i)`

nvme.c
```diff
@@ -1184,13 +1189,52 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
+			if ( (1 << ns.lbaf[i].ds) == cfg.bs && ns.lbaf[i].ms == 0) {
+				cfg.flbas = i;
+				break;
+			}
+		}
+		if (cfg.flbas == 0xff) {
```
keithbusch: Just move this check outside this section, and that should fix the issue in the "flbas == 0xff && cfg.bs == 0" issue.

nvme.c
```diff
@@ -1184,13 +1189,52 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 		{"flbas", 'f', "NUM", CFG_BYTE,        &cfg.flbas, required_argument, flbas},
 		{"dps",   'd', "NUM", CFG_BYTE,        &cfg.dps,   required_argument, dps},
 		{"nmic",  'm', "NUM", CFG_BYTE,        &cfg.nmic,  required_argument, nmic},
+		{"block-size",   'b', "NUM",  CFG_LONG_SUFFIX, &cfg.bs,        required_argument, bs},
```
keithbusch: We've been trying to keep all the columns aligned in the arguments since it normally helps with reading these.

PR #453, opened 2019-01-30
Added abort command w/ documentation

bdeyal:


keithbusch:
What is the justification for this? The user space doesn't know which command id's are outstanding on which submission queues, there's no way this could be used correctly or to any good outcome.

bdeyal:
Testing.

I used admin-passthru to send the abort command (QA + testing stuff), wrapped it in a script to save some calculations, and thought maybe a built-in syntactic-sugar would be useful to others.

I agree that there's no use for this command for sysadmins and  in production environment.

keithbusch:
Just use the admin-passthru. There is little chance such a convenience function could be used correctly in any operational environment, so I'd rather not make it "easy" to use.

bdeyal:
ok.

PR #468, opened 2019-04-03
Fix possible memory leak on get-internal-log

kheitke:


keithbusch:
Heh, why is this variable even malloc'ed instead of on the stack? Merging anyway.

PR #471, opened 2019-04-06
ioctl: Free memory if get_property for fabrics failed

minwooim:
```
The given *pbar is allocated inside this function without freeing it
will be freed from the caller in case of show-regs subcommand for NVMe-oF
controller.

If Get Property is failed because it has PCIe transport, then it needs
to be freed here instead of caller.

```


minwooim:
@kheitke , I think you're definitely right. I'll raise a pull request with your diff.
Thanks,

PR #474, opened 2019-04-07
[RFC] Simplify virt-mgmt subcommand options to lowercase

minwooim:
Hi,
I'd like to propse changes about virt-mgmt subcommand options here.
All the other options for subcommands are lowercase, but virt-mgmt command
has options started with uppercase with long name of it.

I'd like to have a discussion to change it to lowercase because there might
be somebody's already using these options.

Thanks,
```
Minwoo Im (2):
  nvme: Simplify description for virt-mgmt
  nvme: Simplify virt-mgmt options to lowercase

 nvme.c | 37 ++++++++++++++++++-------------------
 1 file changed, 18 insertions(+), 19 deletions(-)
```


keithbusch:
Oo, i didn't review the previous one closely enough. Yeah, I don't like the original naming either: it doesn't fit existing convention. Since we didn't tag a release, I think we can get away with changing the param names.

PR #488, opened 2019-04-25
nvme-print: Few code clean-ups here

minwooim:
Hi @keithbusch ,

Minor patches are here without any functional changes.  The first one
introduces a common API to print out nvme status in case of error.  The second
makes function which returns string type to have const with it. The third and
fourth make printing nvme status to be in case of return value is positive.

```
Minwoo Im (4):
  print: Introduce show_nvme_status to print nvme status
  print: Add const to return type of string
  get-telemtry-log: Print nvme status in case err > 0
  sanitize: Print nvme status in case ret > 0

 nvme-print.c |  30 ++++++----
 nvme-print.h |   9 +--
 nvme.c       | 152 +++++++++++++++++++++------------------------------
 3 files changed, 85 insertions(+), 106 deletions(-)
```

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -78,6 +78,12 @@ void d_raw(unsigned char *buf, unsigned len)
 		putchar(*(buf+i));
 }
 
+void show_nvme_status(__u16 status)
+{
+	fprintf(stderr, "NVMe status: %s(%#x)\n",
+			nvme_status_to_string(status), status);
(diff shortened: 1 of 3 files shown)


minwooim:
Hi @keithbusch ,

I think this PR can be merged which is not related to return value
that what we have discussed about.

Thanks,

PR #489, opened 2019-04-25
nvme: Return negative error value

minwooim:
Hi @keithbusch ,

IMHO, It would be nicer if we have a ruled return types for
subcommands like the following:
  - subcommand plugin handler returns
    - positive: nvme status
    - negative: internal errno value
    - 0: success

```
Some subcommands does not currently returning negative value, but some
do.  It needs to be consistant that when it returns a positive value, it
can mean the nvme status, if it's negative, then it can mean an internal
error has occurred.

```

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -337,7 +337,7 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct 
 
 	if (!cfg.file_name) {
 		fprintf(stderr, "Please provide an output file!\n");
-		err = EINVAL;
+		err = -EINVAL;
 		goto close_fd;
 	}
 
@@ -347,7 +347,7 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct 
 	if (!hdr || !page_log) {
 		fprintf(stderr, "Failed to allocate %zu bytes for log: %s\n",
 				bs, strerror(errno));
-		err = ENOMEM;
+		err = -ENOMEM;
 		goto free_mem;
 	}
 	memset(hdr, 0, bs);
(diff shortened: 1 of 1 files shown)


keithbusch:
For the exit status? That's an unsigned byte, and should be the errno value by convention. Making it negative will have `echo $?` report the wrong status.

minwooim:
@keithbusch ,

You're right. if we go through `echo $?` then it will show -err value like 254
or some reduced value from 255.

Then, how do you think about changing all the return values to a positive
values because we always print the nvme status error if it occurs which
means we don't need to pass the nvme status code to the actual main().

What do you say?

minwooim:
I mean there are few cases that return negative value with errr value.
I think it needs to be uniformized.

keithbusch:
Oh, good point. I think we should return 0 if we got a valid response, not the nvme status.

minwooim:
Alright, then I'll prepare V2 patchsets with zero-returning in case
nvme completion has been successfully done.

minwooim:
@keithbusch ,

I have one doubt here.  How do you think a program will recognize that the
given subcommand succeeded or not in case nvme cqe has been received well.

If we are going to return 0 for both success case and nvme error cases,
the only thing we know is the print-out messages of nvme status.  But there
are only few "success" prints, most of them do not print out "success".

I think a program would like to invoke a subcommand and want to know the
status of it.

Do you want it to be done by parsing the print-out messages only ? not by
the return value of this program.

Thanks ;)

keithbusch:
Since we only have 8 bits for exit status, and nvme status are 16 bits, status collisions are possible, so you won't be able to tell what the status means.

The way I look at it, a 0 status code would mean the program was able to send the requested command and receive a reply. A non-zero status means some other error happened preventing a transaction with the device. from occuring

minwooim:
Hah, my bad ;)
Even I told you 255 which is 8 bits, I was thinking it could handle nvme error
status which is 16 bits. 
Let me prepare V2 patcheset for this one. 

Thanks!

minwooim:
PR #492 has been raised for V2. so closed.

PR #492, opened 2019-04-27
[RFC V2] errno clean-ups

minwooim:
Hi @keithbusch ,                                                              
                                                                              
Here's a V2 patchset with something new added more than before.  First of all,
this patchset will make every functions' return type below:                   
                                                                              
```
  0     on success                                                            
  > 0   nvme-specific status code                                             
  < 0   internal error defined in errno.h                                     
```
                                                                              
All the functions will return like above, and it's going to be finally        
returned by main().  As we discussed, the program should return 0 on success, 
otherwise positive error value on error.                                      

```                                                                              
  0     on success || nvme error                                              
  > 0   internal error defined in errno.h                                     
```
                                                                              
Also this patchset prevents errno from being used without system calls which  
might update the errno itself when it fails because we don't want to make     
sure when errno has been updated inside of a function.  All the errno value   
is now returned as a return value, not errno itself.                          
                                                                              
```                                                                           
Minwoo Im (11):                                                               
  nvme: subsys: Return proper error value if fails                            
  nvme: get_ns: Use 'errno' only right after system call                      
  ioctl: verify_chr: Return negative err if fails                             
  nssr,reset,ns-rescan: Return errno from ioctl to caller                     
  check-arg: Do not use 'errno' without system call                           
  ioctl: Do not return ioctl() directly                                       
  open_dev: Return specific negative errno value                              
  open_dev: Close fd leaked if failure occurs                                 
  nvme: Introduce pr_err() to print out error code                            
  nvme: Return negative error value in falure from subcmd                     
  main: Return 0 for >= 0 case, otherwise err                                 
                                                                              
 nvme-ioctl.c |  65 ++++++++--                                                
 nvme.c       | 339 +++++++++++++++++++++++++++------------------------       
 2 files changed, 231 insertions(+), 173 deletions(-)                         
```

### DIFF
--- a/nvme-ioctl.c
+++ b/nvme-ioctl.c
@@ -24,12 +24,12 @@ static int nvme_verify_chr(int fd)
 
 	if (err < 0) {
 		perror("fstat");
-		return errno;
+		return -errno;
 	}
 	if (!S_ISCHR(nvme_stat.st_mode)) {
 		fprintf(stderr,
 			"Error: requesting reset on non-controller handle\n");
-		return ENOTBLK;
+		return -EINVAL;
 	}
 	return 0;
 }
--- a/nvme-ioctl.c
+++ b/nvme-ioctl.c
@@ -41,7 +41,13 @@ int nvme_subsystem_reset(int fd)
 	ret = nvme_verify_chr(fd);
 	if (ret)
 		return ret;
-	return ioctl(fd, NVME_IOCTL_SUBSYS_RESET);
+
+	ret = ioctl(fd, NVME_IOCTL_SUBSYS_RESET);
+	if (ret < 0) {
+		perror("subsystem-reset");
+		return -errno;
+	}
+	return ret;
 }
 
 int nvme_reset_controller(int fd)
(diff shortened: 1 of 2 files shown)


minwooim:
Hi, @keithbusch ,
As you can see [1] discussion for the blktests patch, @ChaitanayaKulkarni 
has proposed that the error value returned by main() could be nvme
status code itself with some mapping of it.

I have agreed what you have mentioned in previous discussion, that
returning 0 in case the program has successfully received the completion.

I was thinking that the nvme status code can be checked by stderr output
which will be commonized in a single format.  But @ChaitanayaKulkarni 
looks think differently here.

Could you please give your comment on this ?
Thanks,

[1] http://lists.infradead.org/pipermail/linux-nvme/2019-May/023909.html

keithbusch:
Sure, okay, that's fine if we want an nvme-status-to-errno conversion.

minwooim:
I'll prepare a patch into linux-nvme mailing list for this.

minwooim:
This PR has been posted to linux-nvme mailing list to discuss
about the return value.

http://lists.infradead.org/pipermail/linux-nvme/2019-May/023979.html

minwooim:
PR: https://github.com/linux-nvme/nvme-cli/pull/512

PR #504, opened 2019-05-22
fix json print for list-subsys command

YoavCohen:
The bug was the instead of adding paths to to the subsystem object it
was added as a new subsystem


keithbusch:
Nice catch

PR #514, opened 2019-06-21
fix false error in nvme_get_property

BS-Kim:
When the target is NVMe Over fabrics device, the return value is possibly FAIL even though the device's status is normal. The reason is that get property CMD reponds with FAIL if the user tries to get reserved property such as INTMS, INTMC, and etc.

(NVMeOF's Property table is not the same as NVMe controller register table.)

Therefore, even if the return value is not successful, the loop must go on.


BS-Kim:
At today, the pull request(related with this fix) is merged.
close.

PR #519, opened 2019-07-10
nvme-list: Add PCI B:D.f to list output

freezer5:
The PCI address helps to distinguish between local and fabrics devices, and facilitates driver bind/unbind using sysfs.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -2166,21 +2166,22 @@ static void show_list_item(struct list_item list_item)
 		nsze, s_suffix);
 	sprintf(format,"%3.0f %2sB + %2d B", (double)lba, l_suffix,
 		le16_to_cpu(list_item.ns.lbaf[(list_item.ns.flbas & 0x0f)].ms));
-	printf("%-16s %-*.*s %-*.*s %-9d %-26s %-16s %-.*s\n", list_item.node,
+	printf("%-16s %-*.*s %-*.*s %-9d %-26s %-16s %12s %-.*s\n", list_item.node,
             (int)sizeof(list_item.ctrl.sn), (int)sizeof(list_item.ctrl.sn), list_item.ctrl.sn,
             (int)sizeof(list_item.ctrl.mn), (int)sizeof(list_item.ctrl.mn), list_item.ctrl.mn,
-            list_item.nsid, usage, format, (int)sizeof(list_item.ctrl.fr), list_item.ctrl.fr);
+            list_item.nsid, usage, format, list_item.pci_bdf,
+	    (int)sizeof(list_item.ctrl.fr), list_item.ctrl.fr);
 }
 
 void show_list_items(struct list_item *list_items, unsigned len)
 {
 	unsigned i;
 
-	printf("%-16s %-20s %-40s %-9s %-26s %-16s %-8s\n",
-	    "Node", "SN", "Model", "Namespace", "Usage", "Format", "FW Rev");
-	printf("%-16s %-20s %-40s %-9s %-26s %-16s %-8s\n",
+	printf("%-16s %-20s %-40s %-9s %-26s %-16s %-12s %-8s\n",
+	    "Node", "SN", "Model", "Namespace", "Usage", "Format", "PCI Address", "FW Rev");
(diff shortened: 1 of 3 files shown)


minwooim:
Did I miss something here?  I have tested this patch, but I can't see any expected
results from it.

- This is previous format for nvme-list
```bash
Node             SN                   Model                                    Namespace Usage                      Format           FW Rev  
---------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
/dev/nvme0n1     foo                  QEMU NVMe Ctrl                           1         536.87  MB / 536.87  MB    512   B +  0 B   1.0     
```

- This is based on your patch
```bash
Node             SN                   Model                                    Namespace Usage                      Format           PCI Address  FW Rev  
---------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- ------------ --------
/dev/nvme0n1     foo                  QEMU NVMe Ctrl                           1         536.87  MB / 536.87  MB    512   B +  0 B           nvme 1.0     
```

freezer5:
@minwooim , thanks for checking! Can you please share the major/minor numbers of /dev/nvme0 and the value of the symbolic link in /sys/dev/char/MAJ:MIN ?
This is what it looks like on my system with a Hitachi drive:
`$ ls -l /dev/nvme0`
`crw------- 1 root root 238, 0 Jul 14 09:23 /dev/nvme0`
`$ ls -l /sys/dev/char/238\:0`
`lrwxrwxrwx 1 root root 0 Jul 14 09:24 /sys/dev/char/238:0 -> ../../devices/pci0000:3a/0000:3a:00.0/0000:3b:00.0/0000:3c:06.0/0000:43:00.0/nvme/nvme0`

minwooim:
@freezer5 , You can try with the nvme virtual device target in qemu. 
Anyway, here's what you have requested..

```bash
root@host:~# ll /dev/nvme0
crw------- 1 root root 248, 0 Jul 14  2019 /dev/nvme0
root@host:~# ll /sys/dev/char/248\:0
lrwxrwxrwx 1 root root 0 Jul 14  2019 /sys/dev/char/248:0 -> ../../devices/pci0000:00/0000:00:04.0/nvme/nvme0/
root@host:~#
```

freezer5:
@minwooim , thanks! I pushed some changes to this pull request, hopefully this should fix it.

minwooim:
Please squash all those three commits into a single one.

minwooim:
Before reviewing the code, I'd like to ask first: why do we really need this one?
Now "nvme list" command shows namespace nodes regardless to the type
(e.g. PCIe, fabrics).  PCIe address is only just for PCIe device.

freezer5:
Sure, I'll squash.
The main reason for doing it is that "nvme list" shows only the logical elements without their physical counterparts, so it's hard to associate a namespace with the actual device providing it. lspci shows the physical device information, nvme shows the logical device(s), and nothing connects the two.
It's useful when you want to unbind/rescan a particular device at the PCIe level, remove and replace a physical device, or associate log messages about a PCIe device with corresponding NVMe namespaces when diagnosing an I/O fault.
Fabrics devices are ephemeral by nature, I would love to see connection information (traddrs) per controller but I'm not sure if there's a stable method of obtaining this information when the connection is already up.

minwooim:
On 19-07-14 06:39:52, freezer5 wrote:
> Sure, I'll squash.
> The main reason for doing it is that "nvme list" shows only the logical elements without their physical counterparts, so it's hard to associate a namespace with the actual device providing it. lspci shows the physical device information, nvme shows the logical device(s), and nothing connects the two.
> It's useful when you want to unbind/rescan a particular device at the PCIe level, remove and replace a physical device, or associate log messages about a PCIe device with corresponding NVMe namespaces when diagnosing an I/O fault.
> Fabrics devices are ephemeral by nature, I would love to see connection information (traddrs) per controller but I'm not sure if there's a stable method of obtaining this information when the connection is already up.

You're right that it would be great if the PCI device address is
provided to remove/rescan or something like them.

I'm saying that we need to consider other types of NVMe devices such as
fabrics.  We just can't provide "PCIe Address" in the header of "nvme
list" command.

freezer5:
OK. Thanks for the feedback. I'll try to rework it into some other format.

PR #525, opened 2019-07-22
nvme: add Verify command

kheitke:
Reviewed-by: Edmund Nadolski <edmund.nadolski@intel.com>
Reviewed-by: Revanth Rajashekar <revanth.rajashekar@intel.com>
Signed-off-by: Kenneth Heitke <kenneth.heitke@intel.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -220,10 +220,10 @@ static int get_smart_log(int argc, char **argv, struct command *cmd, struct plug
 	else
 		perror("smart log");
 
- close_fd:
+close_fd:
 	close(fd);
 
- ret:
+ret:
 	return nvme_status_to_errno(err, false);
 }
 
@@ -421,14 +421,14 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct 
 		offset += bs;
 	}
 
- close_output:
+close_output:
 	close(output);
- free_mem:
+free_mem:
 	free(hdr);
 	free(page_log);
- close_fd:
+close_fd:
 	close(fd);
- ret:
+ret:
 	return nvme_status_to_errno(err, false);
 }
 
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -4656,6 +4656,83 @@ static int write_cmd(int argc, char **argv, struct command *cmd, struct plugin *
...
+	err = nvme_verify(fd, cfg.start_block, cfg.block_count, control,
+			cfg.ref_tag, cfg.app_tag, cfg.app_tag_mask);
+	if (err < 0)
+		perror("verify");
+	else if (err != 0)
+		show_nvme_status(err);
+	else
+		printf("NVME Verify Success\n");
+
+ close_fd:
```
minwooim: Can we just have it without any whitespace for the consistency?
We have different styles with functions so that we just can uniform it
to the kernel style which is without whitespace.

Could you please do that?

nvme-ioctl.c
```diff
@@ -178,6 +178,13 @@ int nvme_compare(int fd, __u64 slba, __u16 nblocks, __u16 control, __u32 dsmgmt,
 		       reftag, apptag, appmask, data, metadata);
 }
 
+int nvme_verify(int fd, __u64 slba, __u16 nblocks, __u16 control,
+	      __u32 reftag, __u16 apptag, __u16 appmask)
+{
+	return nvme_io(fd, nvme_cmd_verify, slba, nblocks, control, 0,
```
keithbusch: The kernel only allows read, write, and compare with this function's ioctl, right?

nvme.c
kheitke: Will do

nvme-ioctl.c
kheitke: I'll have to double check. I have two versions of this patch; one that uses the "deprecated" ioctl which requires a kernel change (which I posted but was rejected) and one that uses the passthrough ioctl. I may have posted the wrong patch. Thanks.

PR #532, opened 2019-07-25
nvme-vendor: fix c99 declarations in vendor plugins

simson-suse:
We can't use c99 declarations with older compilers.

Signed-off-by: Simon Schricker <sschricker@suse.de>


keithbusch:
Indeed we can't, thanks for the fix!

PR #534, opened 2019-07-28
nvme-format: reject format to char device without an explicit NSID

freezer5:
When formatting a char device and NSID is not specified, check ID_CTRL.FNA bit 0.
If FNA is enabled, then default to NSID_ALL; otherwise, reject the command.
See issue #510


keithbusch:
I applied this one manually. Just needed to fixup some code style formatting.

I think this change should be safe, and hopefully this won't break any scripts out there.

PR #536, opened 2019-08-01
Mask out invalid options during discovery.

simson-suse:
These are old patches by Hannes we forgot to send upstream. I had to fiddle a bit to make them apply again so I will ask Hannes to review them too.

A few things:
* I am not sure how to properly indent in the `build_options` function, because of the inconsistent use of whitespace.
* There are some new options, which weren't there when Hannes made the patches initially, which may need to be masked out as well.
* I think I also found the original discussion regarding these patches, as a bit of background:
http://lists.infradead.org/pipermail/linux-nvme/2019-February/022249.html

### DIFF
--- a/Documentation/nvme-connect-all.txt
+++ b/Documentation/nvme-connect-all.txt
@@ -98,7 +98,9 @@ OPTIONS
 
 -k <#>::
 --keep-alive-tmo=<#>::
-	Overrides the default keep alive timeout (in seconds).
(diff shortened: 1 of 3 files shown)


hreinecke:
For all patches:

Reviewed-by: Hannes Reinecke <hare@suse.com>

keithbusch:
Thanks, looks good to me too.

PR #538, opened 2019-08-01
Intel: Add feature enable lat-stats-tracking

Kchaumein:
Adds the lat-stats-tracking feature to the Intel plugin for nvme log pages.

### DIFF
--- a/plugins/intel/intel-nvme.c
+++ b/plugins/intel/intel-nvme.c
@@ -374,6 +374,13 @@ struct intel_lat_stats {
 	__u32	data[1216];
 };
 
+struct latency_bucket {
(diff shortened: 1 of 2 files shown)


keithbusch:
Some of this looks okay, but this many commits appear like they should have been squashed into fewer commits. They also do not have any change logs that justify the commit and we do not have any sign-offs.

The code style is not consistent with this project. We follow Linux kernel-style. So things like CamelCasedNames are not consistent, for example, and we use tabs instead of spaces.

Finally, if we're going to have vendor specific get/set features, I would prefer these all appear under a get/set-feature sub-command rather than give each feature top-level support.

PR #547, opened 2019-08-05
[Log] modify mistake

djukyo:
1. a mistake for command "nvme memblaze smart-log-add /path";
2. Modify the output prompt: GiB --> GB;
3. The data is right.

Signed-off-by: 丁剑 <djukyo@gmail.com>


keithbusch:
So you're saying that the data is not in more common base 2 GiB units, but really in the base 10 GB? Could we get a review from the vendor, please?

keithbusch:
Or if there's a public spec from the vendor that explains that, please provide a link.

PR #548, opened 2019-08-10
Intel: Added json format output to temp-stats.

gcoakes:
It seemed odd that temp-stats was the only command from the Intel plugin which does not have a json output format. I added it. Also, it seemed as if several of the documentation pages were simply copies of the SMART log documentation, and there were some references to it that were never updated to refer to temp stats.

@keithbusch I've been told you are the primary maintainer from Intel for nvme-cli. Is there anything else that might be desired in the output?

### DIFF
--- a/Documentation/nvme-intel-temp-stats.1
+++ b/Documentation/nvme-intel-temp-stats.1
@@ -28,7 +28,7 @@
 .\" * MAIN CONTENT STARTS HERE *
 .\" -----------------------------------------------------------------
 .SH "NAME"
-nvme-intel-temp-stats \- Send NVMe SMART log page request, returns result and log
+nvme-intel-temp-stats \- Send NVMe temperature stats log page request, returns result and log
 .SH "SYNOPSIS"
 .sp
 .nf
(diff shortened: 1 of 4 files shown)


keithbusch:
Yeah, thanks for the patch. IMO, every output should have a json option. If I were to do this over again, json would have been the default.

For consistency with the other command json options, though, the option should be to deprecate the '`-r`' binary option and instead use the '`-o [json|binary]`' option.

gcoakes:
I rebased onto the changes made in the 1.9 release. I've implemented the argument parsing for intel temp-stats following the convention of the base nvme commands (i.e.: --output-format [normal|json|binary]). Additionally, I've done the same for every other command that had --raw-binary implemented.

I don't currently have the resources to validate this with any rigor. I'll comment back when I have a chance to sit down with some Intel drives tomorrow to test these changes.

Please let me know if there is anything else that should be different. I'm also interested in if there should be json output implemented for any of the other commands.

keithbusch:
Thanks, I'll give this a look today.

PR #578, opened 2019-09-24
Add an example for attach-ns

rgirasegit:
Add an example to man page of attach-ns command showing syntax and information to use hexadecimal value for Namespace identifier and controller identifier.

### DIFF
--- a/Documentation/nvme-attach-ns.txt
+++ b/Documentation/nvme-attach-ns.txt
@@ -30,7 +30,12 @@ OPTIONS
 
 EXAMPLES
 --------
-No examples yet.
+	Attach namspace to the controller:
+
+		# nvme attach-ns /dev/nvme1  -n 0x2 -c 0x21
+        
+		Note: Namespace identifier and controller identifier(s) should be in hexadecimal.
+
 
 NVME
 ----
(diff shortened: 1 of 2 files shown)


keithbusch:
The example fine, but on the explanation, the nsid and ctrl id don't need to be in hexidecimal. It will be interpreted that way if you give a 0x prefix, octal with a 0 prefix, and decimal otherwise. The ctrl id can also be a list, like '1,2,10' for a multi-ctrl subsystem.

rgirasegit:
> The example fine, but on the explanation, the nsid and ctrl id don't need to be in hexidecimal. It will be interpreted that way if you give a 0x prefix, octal with a 0 prefix, and decimal otherwise. The ctrl id can also be a list, like '1,2,10' for a multi-ctrl subsystem.

Thank you Keith for reviewing. I have resubmitted patch. Please review it.

keithbusch:
Eh, I should have checked first. This doesn't look like anything in the .txt was changed, and instead you've added an undesirable file to the code base. I'll fix it up.

rgirasegit:
> Eh, I should have checked first. This doesn't look like anything in the .txt was changed, and instead you've added an undesirable file to the code base. I'll fix it up.

Keith, my bad. Should I resubmit it?

PR #581, opened 2019-09-26
intel: Add function lat-stats to cli

Kchaumein:
Adds the lat-stats function to intel plugin.
If lat-stats-tracking is enabled, lat-stats will print out latency
statistics according to the user's desired format.

Auto detects intel drives to print the correct format.

Can output as raw binary, a formatted histogram, or as formatted json.


benreese0:
Hi @keithbusch , how does the latest version of this code look to you?

Issue #349, opened 2018-05-01
add cls and nlw to Intel id-ctrl


hemiant:
Please add Intel Vendor Specific to intel id-ctrl:
- Current PCIe Link Speed field (CLS)
- Negotiated Link Width (NLW)

You can see I added four new lines below in to intel_id_ctrl:
- json_object_add_value_int(root, "cls", vs[24]);
- json_object_add_value_int(root, "nlw", vs[25]);
- printf("cls     : %d\n", vs[24]);
- printf("nlw     : %d\n", vs[25]);

`static void intel_id_ctrl(__u8 *vs, struct json_object *root)
{
        char bl[9];
        char health[21];

        memcpy(bl, &vs[28], sizeof(bl));
        memcpy(health, &vs[4], sizeof(health));

        bl[sizeof(bl) - 1] = '\0';
        health[sizeof(health) - 1] = '\0';

        if (root) {
                json_object_add_value_int(root, "ss", vs[3]);
                json_object_add_value_string(root, "health", health[0] ? health : "healthy");
                json_object_add_value_string(root, "bl", bl);
        } else {
                printf("ss      : %d\n", vs[3]);
                printf("cls     : %d\n", vs[24]);
                printf("nlw     : %d\n", vs[25]);
                printf("health  : %s\n", health[0] ? health : "healthy");
                printf("bl      : %s\n", bl);
        }
}`

10x

benreese0:
This can be closed now, right?

PR #584, opened 2019-09-27
Intel plugin: Adding fields to id-ctrl VU region

benreese0:
Updating fields in 'nvme intel id-ctrl' to parse/print data already provided by the drive.

### DIFF
--- a/plugins/intel/intel-nvme.c
+++ b/plugins/intel/intel-nvme.c
@@ -58,7 +58,6 @@ struct nvme_additional_smart_log {
 	struct nvme_additional_smart_log_item	host_bytes_written;
 };
 
-#pragma pack(push,1)
 struct nvme_vu_id_ctrl_field { //CDR MR5
 	__u8			rsvd1[3];
 	__u8			ss;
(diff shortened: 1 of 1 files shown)


benreese0:
This should also cover #349

plugins/intel/intel-nvme.c
```diff
@@ -58,25 +58,70 @@ struct nvme_additional_smart_log {
 	struct nvme_additional_smart_log_item	host_bytes_written;
 };
 
+#pragma pack(push,1)
```
keithbusch: That's a microsoft windows compiler thing. Some new gcc's support it, but let's stick with the native gcc dialect if we can. This structure here is entirely made of u8's, though, so I don't think there's anyway it can be unaligned, right?

benreese0: I was copying the structure used for nvme_additional_smart_log_item that appears to originate from way back at 61fcd8b637cc8a742beafc9e3bd756fd7d8d2dc, but you're right, it doesn't really make sense for this set of u8's.

benreese0:
Thanks @keithbusch!

PR #594, opened 2019-10-18
Systemd ids

amluto:
More improvements to hostid and hostnqn.

Please review this carefully.  Also, if you merge this, please consider changing the 'luto.us' bit to something a bit more sensible.  Maybe com.github.linux-nvme?

I'm still dubious about the way that newlines in the ids are handled, but my patches don't *change* that per se.

### DIFF
--- a/Documentation/nvme-show-hostid.txt
+++ b/Documentation/nvme-show-hostid.txt
@@ -1,0 +1,29 @@
+nvme-show-hostid(1)
+===================
+
+NAME
+----
+nvme-show-hostid - Generate a host NVMe ID
+
+SYNOPSIS
+--------
+[verse]
+'nvme show-hostid'
+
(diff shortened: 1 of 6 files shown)


keithbusch:
Could you post this to the linux-nvme mailing list instead? I think some of the people who would have the most interest in this area don't subscribe to github.

keithbusch:
Closing not to reject, but to direct this to linux-nvme@infradead.org because the subscribers there really should weigh in, and I can just as easily apply patches posted there when ack'ed.

amluto:
Emailed here:

https://lore.kernel.org/linux-nvme/cover.1576726427.git.luto@kernel.org/T/#t

amluto:
@keithbusch, this has been languishing on the list for quite a while now.  Can you take a look, please?

PR #596, opened 2019-10-28
nvme-cli: Add support for nvme abort admin command

ikegami-t:
Currently the abort command can be submitted by using nvme admin-passthru.
But it is better to be supported by the nvme abort command directly so add it.

### DIFF
--- a/nvme-builtin.h
+++ b/nvme-builtin.h
@@ -74,6 +74,7 @@ COMMAND_LIST(
 	ENTRY("dir-receive", "Submit a Directive Receive command, return results", dir_receive)
 	ENTRY("dir-send", "Submit a Directive Send command, return results", dir_send)
 	ENTRY("virt-mgmt", "Manage Flexible Resources between Primary and Secondary Controller ", virtual_mgmt)
+	ENTRY("abort", "Submit a Abort admin command ", abort_cmd)
 );
 
 #endif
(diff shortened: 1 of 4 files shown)


keithbusch:
User space doesn't assign command IDs, though, so there's no good reason to provide a convenient function that can't be reliably used.

ikegami-t:
Okay I see. I am thinking to also assign command IDs by user space so let me consider it at first.

keithbusch:
The kernel driver, at least, will not allow you to assign command ids from user space. The tags are owned by the kernel because it is used to index other data structures that we have to coordinate with the block layer.

ikegami-t:
Noted. I will do reconsider other way to abort. Thank you.

keithbusch:
The nvme admin-passthru is the recommended way issue an abort from user space if you *really* want to do that. The only reason I've ever found to use it, though, is to trigger hardware, like a protocol analyzer or a firmware internal log snapshot, which doesn't really require a valid sq:cid tuple if you have the ability to instrument the external devices to do something in response to it.

ikegami-t:
Understood well so I will try to use the name admin-passthru to abort as advised. At first I thought to use it to test commands if aborted correctly. Also I thought that it can be used for an error case to abort command hanged up without timeout waiting. Thanks again.

PR #603, opened 2019-11-11
Get DUI for SN-340 failing with Data Area 4 - Buffer Misalignment

jeffreyalien:


keithbusch:
i manually merged this one to avoid picking up the unnecessary merge commit.

PR #604, opened 2019-11-13
nvme-cli: New Command Implemented: Format Progress Indicator

RevanthRajashekar:
If a format operation is in progress, this feature indicates
the percentage of the namespace that remains to be formatted.

Signed-off-by: Revanth Rajashekar <revanth.rajashekar@intel.com>


keithbusch:
This just 'id-ns', we don't need a too level command for individual fields.

PR #606, opened 2019-11-15
nvme.c: create-ns arguments are added

sc108-lee:
ANA Group Identifier, NVM Set Identifier
Its based on NVMe 1.4 Spec


keithbusch:
Could you update the man page in Documentation/nvme-create-ns.txt with the new options, please? Also, please keep the line width in code to <= 80 if possible.

sc108-lee:
I update documents and code <= 80 as well.

keithbusch:
i manually applied this patch. The manpage in the pull-request had a "_" when it should have been a "-", and fixed some whitespace issues. Thanks for the patch!

PR #607, opened 2019-11-15
nvme.c: format help modification(clear explaination)

sc108-lee:
former commit was unclear (bit rule was worng for ses[0-3])
so reordering with bit order, show bit numbers for each options

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2663,7 +2663,7 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 		"Can also be used to change LBAF to change the namespaces reported physical block format.";
 	const char *namespace_id = "identifier of desired namespace";
 	const char *lbaf = "LBA format to apply (required)";
-	const char *ses = "[0-2]: secure erase";
+	const char *ses = "[0-3]: secure erase";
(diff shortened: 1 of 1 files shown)


keithbusch:
The SES field does actually occupy bits 0-2 of the SES portion of CDW10 carved out for SES, though. Yes, those are bits 09:11 offset from CDW10, but why is reporting that useful for a user?

sc108-lee:
humm... 
In my opinion. It is unclear using that expression.
[0-2] -> it looks like using 3bit bit0, bit1, bit2.
so I make it clear with real bits. (And I also reordered in bits order.)

If you don't agree with that, I can modify as [0-2]
How do you think? Just let me know.

keithbusch:
Why would a user need to know or even care what arbitrary base was used to define the bit offsets? The only thing this was intended to convey was the valid value range because there is no native type that is 3 bits to naturally constrain. 

Personally I don't really care for the existing help text, but defining it by dword offsets doesn't seem helpful. It would look confusing if two fields in different dwords had overlapping offsets.

sc108-lee:
ok I don't mind with this changes, I just suggest this option.
then I will modify as original one, only for ses

sc108-lee:
I got it. dont need to change this, I will close it

PR #616, opened 2019-11-26
Micron plugin updates

hanumanthuh:
Additional commands to pull drive log information from Micron drives

### DIFF
--- a/plugins/micron/micron-nvme.c
+++ b/plugins/micron/micron-nvme.c
@@ -5,6 +5,9 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <time.h>
+#include <string.h>
+#include <libgen.h>
 #include <sys/stat.h>
 #include "nvme.h"
 #include "nvme-print.h"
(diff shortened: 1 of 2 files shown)


keithbusch:
Can we stick with the native coding style? Tabs vs spaces, approrpriate placing of {} braces, etc...

We're following linux kernel codings style, and I'm going to try to enforce that going forward. There's a reference here:

https://www.kernel.org/doc/html/v5.3/process/coding-style.html

hanumanthuh:
Thanks for your inputs. I have modified the sources accordingly and with some more fixes for smart attributes in our drives. Please use https://github.com/linux-nvme/nvme-cli/pull/616/commits/0d832c6c1ec4cc318007fad263b33e60adde0c70 this commit for merging.

keithbusch:
Getting new compiling warnings (which we treat as errors) in micron_clear_pcie_correctable_errors() where it is not providing sufficient space for the strTempFile. I was able to fix up the errors, but there really is no error handling or reporting at all here: the 'err' return value is initialized to 0, and it is never set to anything but 0 (success) no matter what may have failed while running this. You can fix that up if you want later. I've just fixed up the warnings and pushed out what we have. It'd also be great if these commands had man pages too.

PR #617, opened 2019-11-26
Memblaze add new command

djukyo:
Memblaze add new command for download firmware.
Please help revirew.


plugins/memblaze/memblaze-nvme.c
```diff
@@ -481,3 +481,129 @@ static int set_additional_feature(int argc, char **argv, struct command *cmd, st
...
+	};
+
+	struct config cfg = {
+		.fw     = "",
+		.select = "\0",
+	};
+
+	OPT_ARGS(opts) = {
+		OPT_FILE("fw",     'f', &cfg.fw,     fw),
+		OPT_LIST("select", 's', &cfg.select, select),
```
keithbusch: Why is this a list option? That's what you'd use if you want to have multiple comma separated values, like "--select=OOB,EEP".

djukyo: I updated prompt.
--select=OOB,EEP,ALL

keithbusch: > I updated prompt.
> --select=OOB,EEP,ALL

I still don't understand. It doesn't look like you support a list at all. You're enforcing it be exactly 3 characters long rather than a comma separated value. Why not just make this an OPT_STRING instead?

djukyo:
change to OPT_STRING, and update it.
Please review again.

PR #640, opened 2020-01-07
Add  vs-error-reason-identifier WDC plugin command

jeffreyalien:
This commit contains the following changes:
Add support for WDC plugin command - vs-error-reason-identifier
Reverse enable/disable bit on the vs-telemetry-controller-option WDC plugin command
Add documentation file for vs-error-reason-identifier WDC plugin command
Minor corrections to the documentation files for vs-fw-activate-history and clear-fw-activate-history WDC plugin commands


jeffreyalien:
Thanks Keith.  Could you also regen the Documentation files when you  get a chance?

PR #644, opened 2020-01-16
fix nvme disconnect by device failure

sulakshm:
Signed-off-by: Lakshmi Narasimhan Sundararajan <lns@portworx.com>

When nvme disconnect with device path is issued, it fails as below.

```
root:github.com/linux-nvme/nvme-cli# ./nvme disconnect -d /dev/nvme0n1
Failed to disconnect by device name: /dev/nvme0n1
```

Looking  at the  code and debugging, I find there is an added suffix to device name and the match only happens for nvme0 and not nvme0n1.
In fabrics.c:ctrl_instance() method:
nvme ctrl_instance passed device /dev/nvme0n1, basename nvme0n1, instance 0, tmp 'd' nvme0

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -269,7 +269,7 @@ static int ctrl_instance(char *device)
 	if (ret <= 0)
 		return -EINVAL;
 	if (snprintf(d, sizeof(d), "nvme%d", instance) <= 0 ||
-	    strcmp(device, d))
+	    strncmp(device, d, strlen(d)))
 		return -EINVAL;
 	return instance;
 }


keithbusch:
The disconnect command is supposed to target a controller, so sending this command to a namespace handle creates ambiguities if it's multipathed. That function is trying to retrieve the controller instance, but what you're doing would sometimes return the subsystem instance, and that's not always the same thing as the controller.

sulakshm:
Hi Keith,
A controller may hold multiple subsystems, but shouldn't disconnect be able to target a particular  subsystem within? Infact disconnect with nqn targets a subsystem within the controller, isn't it? In any case, if it is ambiguous while multipath'd, then failing is correct stating removal is ambiguous when multipath'ed. But in this case it is a simple single instance with single subsystem holding single namespace.  This should not fail either with nqn on the subsystem or the mapped device path. I believe for this case, the added checks are correct.

keithbusch:
A subsystem contains one or more controllers, not the other way around.

sulakshm:
So disconnect by device as reported above is not supposed to work and failure is expected?

keithbusch:
Yes, I believe this is working as designed.

Issue #646, opened 2020-01-20
tests failing to handle string in python3 standards


narasimhan-v:
```
======================================================================
ERROR: nvme_error_log_test.TestNVMeErrorLogCmd.test_get_error_log
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/nvme-cli-master/tests/nvme_error_log_test.py", line 63, in test_get_error_log
    assert_equal(self.get_error_log_ctrl(), 0)
  File "/root/nvme-cli-master/tests/nvme_error_log_test.py", line 59, in get_error_log_ctrl
    return self.get_error_log()
  File "/root/nvme-cli-master/tests/nvme_test.py", line 434, in get_error_log
    err_log_entry_count = int(line.split(" ")[5].strip().split(":")[1])
TypeError: a bytes-like object is required, not 'str'

----------------------------------------------------------------------
```
```
======================================================================
ERROR: TestNVMeIdentifyNamespace (nose2.loader.LoadTestsFailure)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/six.py", line 693, in reraise
    raise value
  File "/usr/lib/python3.6/site-packages/nose2/plugins/loader/testclasses.py", line 149, in _loadTestsFromTestClass
    for name in names])
  File "/usr/lib/python3.6/site-packages/nose2/plugins/loader/testclasses.py", line 149, in <listcomp>
    for name in names])
  File "/usr/lib/python3.6/site-packages/nose2/plugins/loader/testclasses.py", line 193, in __init__
    self.obj = cls()
  File "/root/nvme-cli-master/tests/nvme_id_ns_test.py", line 43, in __init__
    self.ns_list = self.get_ns_list()
  File "/root/nvme-cli-master/tests/nvme_test.py", line 181, in get_ns_list
    ns_list.append(string.replace(line.split(":")[1], '\n', ''))
AttributeError: module 'string' has no attribute 'replace'

----------------------------------------------------------------------
```

PR #647, opened 2020-01-20
Tests: Hande string encoding and handling failures

narasimhan-v:
Handling string encoding to utf-8 and also string handling in
tests.

Signed-off-by: Narasimhan V <sim@linux.vnet.ibm.com>


narasimhan-v:
fixes https://github.com/linux-nvme/nvme-cli/issues/646

PR #654, opened 2020-01-27
nvme-cli: Add script to determine host NQN

hreinecke:
Add a script to create a deterministic hostnqn which remains stable across reinstalls.
* Based on the system-UUID via dmidecode
* verifies UUID format
* tries to catch fake/lazy UUIDs

### DIFF
--- a/scripts/gen-hostnqn.sh
+++ b/scripts/gen-hostnqn.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+LC_ALL=C
+
+UUID=$(dmidecode -s system-uuid | tr -d '[:space:]')
+
+if [ -z "$UUID" ] ; then
+	>&2 echo "No UUID found, can't determine hostnqn."
+	exit 1
+fi
+
+# convert UUID to lower-case only:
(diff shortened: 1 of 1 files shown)


simson-suse:
As a side note: I originally named the script `det-hostnqn`, to distinguish it from the already existing `nvme-cli gen-hostnqn`, which creates a random hostnqn. But I don't have a strong opinion on this.

hreinecke:
I know. But 'nvme-det-hostnqn' is not very clear about what it's doing. 'nvme-gen-hostnqn' is more in-line with existing naming schemes. But yes, there is the confusion between 'nvme-gen-hostnqn' and 'nvme gen-hostnqn'.
Let Keith decide :-)

keithbusch:
I like it! Maybe the existing gen-hostnqn should attempt to get system uuid before falling back to random uuid.

keithbusch:
Yeah, I think I'll try system's uuid first instead of random with the tool's gen-hostnqn command. I initially didn't want to decode /sys/firmware/ tables, but it looks like someone already puts the result in a convenient place: /sys/devices/virtual/dmi/id/product_uuid.

PR #672, opened 2020-03-04
Fixing Alignment Error In intel smart-log-add

benreese0:
Fixing alignment in nvme_additional_smart_log_item introduced by switch
from '#pragma pack' semantics to '__attribute__((pack))'.

Per https://gcc.gnu.org/onlinedocs/gcc/Common-Type-Attributes.html each
union or struct requires it's own '__attribute__((pack))'.


benreese0:
Thanks Keith!

PR #673, opened 2020-03-06
Change 0xC2 Log Page retrieval to use UUID index value

rbates0119:
Add get_wdc_uuid_index function 

Add new function to find index to wdc uuid
Fix comments

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -858,7 +858,8 @@ static __u64 wdc_get_drive_capabilities(int fd) {
 			break;
 		case WDC_NVME_SN200_DEV_ID:
 			capabilities = (WDC_DRIVE_CAP_CAP_DIAG | WDC_DRIVE_CAP_INTERNAL_LOG | WDC_DRIVE_CAP_CLEAR_PCIE |
-					WDC_DRIVE_CAP_DRIVE_LOG | WDC_DRIVE_CAP_CRASH_DUMP | WDC_DRIVE_CAP_PFAIL_DUMP);
+				        	WDC_DRIVE_CAP_LOG_PAGE_DIR | WDC_DRIVE_CAP_DRIVE_LOG |
+			        		WDC_DRIVE_CAP_CRASH_DUMP | WDC_DRIVE_CAP_PFAIL_DUMP);
 
 			/* verify the 0xCA log page is supported */
 			if (wdc_nvme_check_supported_log_page(fd, WDC_NVME_GET_DEVICE_INFO_LOG_OPCODE) == true)
@@ -1038,6 +1039,54 @@ static int wdc_create_log_file(char *file, __u8 *drive_log_data,
 	return 0;
 }
 
+static const uint8_t zero_uuid[16] = { 0 };
+
+static int get_wdc_uuid_index(int fd, __u8* uuid_ix, uuid_t uuid)
+{
+	*uuid_ix = 1;
+	int ret = -1;
+
+	struct nvme_id_ctrl ctrl;
+	struct nvme_id_uuid_list uuid_list;
+
+	/* Determine if the controller supports a UUID List by checking CTRATT */
+	ret = nvme_identify_ctrl(fd, &ctrl);
+
+	if (ret == 0) {
+
+		if ((ctrl.ctrattr & NVME_CTRL_CTRATT_UUID_LIST) == NVME_CTRL_CTRATT_UUID_LIST) {
+			/* UUID list is supported.  Get the WDC UUID entry */
+			ret = nvme_identify_uuid(fd, &uuid_list);
+
+			if (ret == 0) {
+            	/* The 0th entry is reserved so start at 1 */
+				for (*uuid_ix = 1; *uuid_ix < NVME_MAX_UUID_ENTRIES; (*uuid_ix)++) {
+
+					/* The list is terminated by a zero UUID value */
+					if (memcmp(uuid_list.entry[*uuid_ix].uuid, zero_uuid, sizeof(zero_uuid)) == 0) {
+						/* if we reached the end then we did not find the uuid */
+						*uuid_ix = 0;
+						break;
+					} else {
+
+						if (memcmp(uuid_list.entry[*uuid_ix].uuid, uuid.b, sizeof(uuid.b)) == 0) {
+
+							/* found the uuid.  */
+							break;
+						}
+					}
+				}
+			} else{
+				/* bad return code from identify_uuid. */
+			}
+		} else {
+			/* UUID list not supported. */
+		}
+	}
+
+	return ret;
+}
+
 static bool get_dev_mgment_cbs_data(int fd, __u8 log_id, void **cbs_data)
 {
 	int ret = -1;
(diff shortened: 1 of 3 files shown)


jeffreyalien:
I've reviewed these changes and didn't find any issues with it.

plugins/wdc/wdc-nvme.c
```diff
@@ -1038,6 +1039,84 @@ static int wdc_create_log_file(char *file, __u8 *drive_log_data,
 	return 0;
 }
 
+static int wdc_get_log(int fd, __u32 nsid, __u8 log_id, bool rae, __u8 uuid_ix,
```
keithbusch: This looks like a rather generic routine, and very similar to existing common code. I think the common case only misses the uuid index. Any chance we can move this routine to the common case? I don't want every vendor duplicating this.

rbates0119: Yes we could move it to nvme-ioctl.c and give it a generic name like nvme_get_log_from_uuid.  The other option would be to change the existing nvme_get_log routine to take a uuid parameter and call nvme_get_log14 as the routine I added does.  The problem with that option is that there are 28 calls to nvme_get_log so we would have to change a lot of code to implement that way.  A third option would be to remove the new routine and call nvme_get_log14 directly from get_dev_mgment_cbs_data. Please let me know your thoughts.

plugins/wdc/wdc-nvme.c
```diff
@@ -1038,6 +1039,54 @@ static int wdc_create_log_file(char *file, __u8 *drive_log_data,
...
+				for (*uuid_ix = 1; *uuid_ix < NVME_MAX_UUID_ENTRIES; (*uuid_ix)++) {
+
+					/* The list is terminated by a zero UUID value */
+					if (memcmp(uuid_list.entry[*uuid_ix].uuid, zero_uuid, sizeof(zero_uuid)) == 0) {
+						/* if we reached the end then we did not find the uuid */
+						*uuid_ix = 0;
+						break;
+					} else {
+
+						if (memcmp(uuid_list.entry[*uuid_ix].uuid, uuid.b, sizeof(uuid.b)) == 0) {
```
keithbusch: what is uuid.b? In my implementation, the uuid_t is just a u8[16] type, it doesn't contain a member called 'b'.

rbates0119: I used the uuid_t as it is defined in nvme.h:
typedef struct {
	uint8_t b[16];
} uuid_t;
 But that is because LIBUUID is not defined.  I assume you would be expecting that library to be included so it will use a definition in uuid.h.  Where do I get libuuid?

keithbusch: I see. This type should be treated as opaque: we can't peek into its implementation for portability reasons since we prefer to let the definition come from elsewhere. Use &uuid instead of referencing the member by name.

FWIW, you can get the appropriate library using your package manager's (dnf, yum, apt, zypper, etc...) install with libuuid-devel or uuid-dev.

rbates0119: Ok got it.  I got uuid-dev installed also.

plugins/wdc/wdc-nvme.c
```diff
@@ -1038,6 +1039,84 @@ static int wdc_create_log_file(char *file, __u8 *drive_log_data,
...
+		offset += xfer_len;
+		ptr += xfer_len;
+	} while (offset < data_len);
+
+	return 0;
+}
+
+static const uint8_t zero_uuid[16] = { 0 };
+
+static int get_wdc_uuid_index(int fd, __u8* uuid_ix, uuid_t uuid)
```
keithbusch: This function actually has a bit too much indentation and exceeds the standard 80-column width. I think we can shorten it up a bit if we reorganize the parts. What do you think of the following?

```
static int get_wdc_uuid_index(int fd, __u8* uuid_ix, uuid_t uuid)
{
        struct nvme_id_uuid_list uuid_list = { 0 };
        struct nvme_id_ctrl ctrl;
        int i, ret = -1;

        *uuid_ix = 0;

        /* Determine if the controller supports a UUID List by checking CTRATT */
        ret = nvme_identify_ctrl(fd, &ctrl);
        if (ret)
                return ret;

        if (!(ctrl.ctrattr & NVME_CTRL_CTRATT_UUID_LIST))
                return 0;

        /* UUID list is supported.  Get the WDC UUID entry */
        ret = nvme_identify_uuid(fd, &uuid_list);
        if (ret)
                return ret;

        /* The 0th entry is reserved so start at 1 */
        for (i = 1; i< NVME_MAX_UUID_ENTRIES; i++) {
                struct nvme_id_uuid_list_entry *entry = &uuid_list.entry[i];

                /* The list is terminated by a zero UUID value */
                if (memcmp(entry->uuid, zero_uuid, sizeof(zero_uuid))) {
                        if (memcmp(entry->uuid, &uuid, sizeof(uuid_t)) == 0) {
                                *uuid_ix = i;
                                return 0;
                        }
                } else
                        break;
        }

        return 0;
}
```

plugins/wdc/wdc-nvme.c
```diff
@@ -1055,9 +1136,23 @@ static bool get_dev_mgment_cbs_data(int fd, __u8 log_id, void **cbs_data)
 	}
 	memset(data, 0, sizeof (__u8) * WDC_C2_LOG_BUF_LEN);
 
+	/* TODO:  get wdc uuid value */
```
keithbusch: This presents a bit of a problem for including this feature at this time. Without a valid uuid to check, running this command is not going to get the correct result.

plugins/wdc/wdc-nvme.c
rbates0119: I see you set the default value for uuid_ix to zero.  Jeff said it should be set to one unless we find a specific uuid entry otherwise I am ok with it..

keithbusch: Sure, but the calling function already overrides it to 1 already.

I'm not sure of the rationale for that, though. If you can't find the desired uuid, then the function should abandon the action with an appropriate error.

rbates0119: Actually I do see a problem with your suggested code. 
The following line checks to see if we reached the end without finding an entry.  The line after needs to be in the else as follows.
if (memcmp(entry->uuid, zero_uuid, sizeof(zero_uuid))) {
  uuid_ix = 0;
} else {
if (memcmp(entry->uuid, &uuid, sizeof(uuid_t)) == 0) {
/* entry found*/
  uuid_ix = i;
  return 0;
}

keithbusch: > Actually I do see a problem with your suggested code.
> The following line checks to see if we reached the end without finding an entry. The line after needs to be in the else as follows.
> if (memcmp(entry->uuid, zero_uuid, sizeof(zero_uuid))) {
> /* entry not found _/ uuid_ix = 0; } else { if (memcmp(entry->uuid, &uuid, sizeof(uuid_t)) == 0) { /_ entry found*/
> uuid_ix = i;
> return 0;
> }

I don't understand. If the entry's uuid doesn't match the zero uuid, why does that mean "entry not found"? Isn't the opposite supposed to be the case?

plugins/wdc/wdc-nvme.c
rbates0119: Yeah, we set the default value to 1 since we thought the uuid index for WD would always be 1 and it would return the wdc values unless a uuid was found but now we understand that it could also be 0.  We are putting this case on hold until we get more firmware implementation details.  Closing this pull request.

keithbusch: Isn't this something an OEM can change at their discretion?  The only thing an initial vendor can do is provide their own uuid value. Generally speaking, you don't want to run a uuid specific extension against the wrong device. It will be interpreted incorrectly, so I believe we need a valid uuid check for this to function reliably.

PR #674, opened 2020-03-09
Update Makefile to remove old destination folder

rbates0119:
The install path has been changed recently.  This change removes the old install folder if it exists.


rbates0119:
Tried to add reviewers but could not.  Re-opening

PR #683, opened 2020-03-19
Pt 4562

rbates0119:
Fix invalid option reporting for vs-telemetry-controller-option WDC VUC.


jeffreyalien:
I've review these changes and approve them as coded.

PR #692, opened 2020-04-14
nvme-print: fix namespace descriptor human printing

birkelund:
9b2ce6ca3ba7 fixed the len accumulation, but only for the JSON printing.
Fix it for human output as well.

Reported-by: Narae Park <narae7.park@samsung.com>


keithbusch:
Great catch, thanks for the fix.

PR #695, opened 2020-04-17
Change command name

Dequan-Dera:
Change command 'stat' to 'smart-log-add' and its description

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -3776,6 +3776,9 @@ static void show_registers_cmbsts(__u32 cmbsts)
 
 static void show_registers_pmrcap(__u32 pmrcap)
 {
+	printf("\tController Memory Space Supported (CMSS)		   : "\
+	       "Referencing PMR with host supplied addresses is %s\n",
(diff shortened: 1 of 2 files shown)


keithbusch:
This is highly unusual. For the sake of backward compatibility, you should make the "old" name of this command an alias to the new name. Sound okay?

Dequan-Dera:
Add new command named 'smart-log-add', make old 'stat' an alias to the new name.

keithbusch:
Sorry, I should have been more clear. On the "ENTRY" line, there is an optional 4th argument for alias names. This way it doesn't show up in the "help" so it discourages new use on the now deprecated name, but it allows existing scripts to run uninterrupted.  The end result should look something like this:

`ENTRY("smart-log-add", "Retrieve Dera SMART Log, show it", get_status, "stat")`

Dequan-Dera:
Add new command named 'smart-log-add', make old 'stat' an alias to the new name. Sorry for ignorance of great character of ENTRY. Thank you.

keithbusch:
squashed and committed via command line

PR #701, opened 2020-04-27
nvme-cli:lightnvm:fix fd leak when execute ioctl failed

wubo009:
If the execution of ioctl fails, the fd is not closed before returning,
and there is a resource leak.

Signed-off-by: Wu Bo <wubo40@huawei.com>


MatiasBjorling:
Thanks Wu Bo.

Reviewed-by: Matias Bjørling <matias.bjorling@wdc.com>

PR #702, opened 2020-04-29
nvme-connect-all: add -m /  -matching option

mwilck:
Optionally have `connect-all` connect only to such controllers that have a matching `traddr` field. 

This avoids confusing error messages and failed connect attempts for NVMeoFC autoconnect. See https://github.com/linux-nvme/nvme-cli/commit/f989634e0595b1f184a0acb125c4460b1b31960a commit message for a detailed rationale.

### DIFF
--- a/Documentation/nvme-connect-all.1
+++ b/Documentation/nvme-connect-all.1
@@ -2,12 +2,12 @@
 .\"     Title: nvme-connect-all
 .\"    Author: [FIXME: author] [see http://www.docbook.org/tdg5/en/html/author]
 .\" Generator: DocBook XSL Stylesheets vsnapshot <http://docbook.sf.net/>
-.\"      Date: 04/24/2020
+.\"      Date: 04/29/2020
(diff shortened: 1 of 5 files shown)


keithbusch:
I think this is fine, but is traddr the only thing you may ever want to connect matching records? Why not say connect only a specific transport or address family, for example?

mwilck:
I wanted to keep it simple, and restrict it to the cases that I can actually test. Can a discovery controller return discovery log entries for a different `trtype` or `adrfam`? AFAICT the spec doesn't forbid it, but I wonder it makes any sense, and if any real-world system would do it. (On the contrary, the different `traddr` records are actually observed on every NetApp target).

For `connect-all` on a given `traddr` / `host-traddr` pair, trying to connect to a controller with a different transport type is doomed to fail anyway, or am I missing something? So, we might as well skip such entries in `connect-all` always (if any discovery controller returns them), not only with `--matching`.

PR #704, opened 2020-05-01
nvme-cli: controller list argument is required in attach-ns/detach-ns commands

KLZ-0:
Empty controller-id list in the ioctl call for attach/detach of namespace results effectively in no operation, yet the nvme cli reports falsely success. The controller id list must be specified with at least 1 controller in the list.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1058,7 +1058,7 @@ static int nvme_attach_ns(int argc, char **argv, int attach, const char *desc, s
 	__u16 ctrlist[2048];
 
 	const char *namespace_id = "namespace to attach";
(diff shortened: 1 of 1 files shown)


keithbusch:
I agree it's a rather pointless no-op to have zero controllers listed in an attach/detach command, but the spec allows such a list, and spec compliance testing requires a controller process such a command. If the controller returns success for the requested command, then that's what nvme-cli will return, so it's not really "falsely" success.

I would be okay with emitting a warning for this, though, but I don't think we should prevent a spec compliant command from being sent (we use this tool for compliance testing).

PR #705, opened 2020-05-04
remove block device checking for namespace

tkulasek:
In SPDK project (see spdk.io) we have implemented the ability
to set up devices via CUSE device exposing controller and
namespaces as character devices.

Namespace devices are also exposed as character devices, so
nvme-cli tools will not recognize them as valid devices.

To allow to use nvme-cli tool with SPDK NVMe CUSE devices
we can't assume that namespaces are block devices.

Change-Id: I52aa79d24002b8dec10e6fdd0cb9a71bb6750358
Signed-off-by: Tomasz Kulasek <tomaszx.kulasek@intel.com>

### DIFF
--- a/nvme-ioctl.c
+++ b/nvme-ioctl.c
@@ -70,12 +70,6 @@ int nvme_get_nsid(int fd)
 	if (err < 0)
 		return -errno;
 
-	if (!S_ISBLK(nvme_stat.st_mode)) {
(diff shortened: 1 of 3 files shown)


keithbusch:
Sure, I suppose we'll get an error eventually if the commands are run on an inappropriate device anyway, so no need to error out early.

Issue #709, opened 2020-05-15
After format re-read partition failed 


sc108-lee:
since checking block device is removed by below commit.
https://github.com/linux-nvme/nvme-cli/commit/e770466615096a6d41f038a28819b00bc3078e1d#diff-a71a06d2f8c91ea6e57b2f28e8813cfdR2862

There is always fail for format with character device. 
like nvme /dev/nvme0
case 1. nvme /dev/nvme0 -n 0xffffffff (solved)
case 2. nvme /dev/nvme0 -n 1 ( prev lbaf == new lbaf ) (solved)
    case 1, 2 solved by #708 
    https://github.com/linux-nvme/nvme-cli/pull/708/commits/e9e0de444b76cea8133b42b62192ea99d81e7515

**case 3: nvme /dev/nvme0 -n 1 ( prev lbaf != new lbaf )**
always fail to issue ioctl(fd, BLKRRPART) on character device.
Do we need this failure checking???
If the device AER enabled by set-feature, it would be handled by kernel.

I would love to hear opinion @tkulasek @minwooim

keithbusch:
We'd also need the device to actually send the AER for the kernel to react, which some do, some don't. Instead of the RRPART ioctl, nvme-cli should send NVME_IOCTL_RESCAN.

sc108-lee:
hummm.
@keithbusch I agree to send NVME_IOCTL_RESCAN for character device with ALL NS.
But for case 3, It would be different with intention. ( target ns 1, but rescan all )
Not sure this is right way to do this.

And also for spdk device will be a character device.
would be fine to do NVME_IOCTL_RESCAN for that?

keithbusch:
There's no harm in having the kernel revalidate all the disks for this exceptional condition.

No clue about spdk's handling.

sc108-lee:
@keithbusch thanks for confirming this.
since there is no harm for this, I raise the PR for this.
https://github.com/linux-nvme/nvme-cli/pull/712

tkulasek:
SPDK operates in the user space and extending to handle additional ioctls should be not the problem in the future.

PR #712, opened 2020-05-19
format: rescan after format for character device #709

sc108-lee:
if the aer support & enabled, kernel will do so skipping rescan or re-read

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2923,12 +2923,22 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 		nvme_show_status(err);
 	else {
 		printf("Success formatting namespace:%x\n", cfg.namespace_id);
-		if (cfg.lbaf != prev_lbaf && ioctl(fd, BLKRRPART) < 0) {
-			fprintf(stderr, "failed to re-read partition table\n");
-			err = -errno;
-			goto close_fd;
+		if (cfg.lbaf != prev_lbaf){
+			if (S_ISCHR(nvme_stat.st_mode)){
+				if(ioctl(fd, NVME_IOCTL_RESCAN) < 0){
+					fprintf(stderr, "failed to rescan namespaces\n");
+					err = -errno;
+					goto close_fd;
+				}
+			}
+			else{
+				if(ioctl(fd, BLKRRPART) < 0) {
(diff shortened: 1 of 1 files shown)


keithbusch:
I think you're making this a little more difficult than it needs to be. The rescan is harmless, there's no need for all these checks to verify if it's 100% necessary. Just kick the rescan and then all scenarios are covered.

sc108-lee:
@keithbusch 
As I remember, there is problem to do rescan while rescaning is happening from sometimes.
so thats why I blocked for AER support. 

if we just do rescan after format success, it will be run like this.
format -> rescan
AER -> rescan

thats why I add checking AER enabled or not.

keithbusch:
The rescan occurs within a work queue, so the rescan is single threaded for the controller. If AER and userspace both request a rescan, only one can run. Should be fine.

keithbusch:
And FWIW, the command effects log also kicks rescan work in some cases in recent kernels, which can also coincide with AER handling and/or userspace requests. As far as I know, this works well, but if it's a problem, then we need to take it to the nvme kernel mailing list.

sc108-lee:
humm.
I agree if there is problem, need to take on kernel side.
But I still think this is not difficult and also only get-feature command is added for this logic.
since id-ctrl already used on format command.

just do rescan
on kernel side, do id-ctrl, id-ns-list, id-ns-descs.. ect more logic.

AER skipping logic
just do get-feature

isnt it much simple for the device ?

keithbusch:
The kernel has to do all those things anyway and its infrastructure already exists to do it. Whether it's triggered by AER or userspace is irrelevant. It's a harmless action that is guaranteed to get the format's updated LBA results (as long as the kernel is new enough to support that ioctl).

The get-feature isn't enough to know the controller hasn't internally masked the event. We *should* get the event, but there's enough controllers that won't send one.

sc108-lee:
As far as I know AER supported and enabled by set-feature, then It must send it.
If it is not, isn't it spec violation??
should we consider for that case??

sc108-lee:
@keithbusch Modified and rebased.
modified to do rescan for character device

keithbusch:
thanks, manually applied just to address some minor formatting issues.

PR #719, opened 2020-05-22
Add option to never use fabrics for the show-regs command.

gcoakes:
Using fabrics commands on devices that conform to NVMe spec 1.2 is causing the error count in log page 1 to increment. This patch adds the option to not use fabrics when showing registers, so we can avoid those errors.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2518,7 +2518,7 @@ static int show_registers(int argc, char **argv, struct command *cmd, struct plu
 					"in binary or human-readable format";
 	const char *human_readable = "show info in readable format in case of "\
 					"output_format == normal";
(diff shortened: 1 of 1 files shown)


keithbusch:
Could you squash intermittent commits and fixups into a single commit?

To address the issue you're observing, I would prefer checking if the device is PCI first before using the fabrics properties command instead of requiring an option.

gcoakes:
@keithbusch, that definitely seems like the better approach. I wasn't entirely certain how to go about determining whether it was a PCI device first. Any suggestions? I would rather put the optimal solution into place now than to have to deprecate an option later.

keithbusch:
If the device sysfs property "transport" exists, you can read this file and see if it says "pcie" to confirm. If the property does not exist, then it must be "pcie".

PR #721, opened 2020-05-26
Adding ScaleFlux subcommands

jyang-sfx:
This is from ScaleFlux Inc to add its sub-commands.

### DIFF
--- a/plugins/scaleflux/sfx-nvme.c
+++ b/plugins/scaleflux/sfx-nvme.c
@@ -1,0 +1,877 @@
+#include <fcntl.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <linux/fs.h>
+#include <inttypes.h>
+#include <asm/byteorder.h>
+#include <sys/ioctl.h>
+#include <sys/sysinfo.h>
+
+#include "linux/nvme_ioctl.h"
+
+#include "nvme.h"
+#include "nvme-print.h"
+#include "nvme-ioctl.h"
+#include "nvme-status.h"
+#include "json.h"
+#include "plugin.h"
+
+#include "argconfig.h"
+#include "suffix.h"
+
+#define CREATE_CMD
+#include "sfx-nvme.h"
+
+#define SFX_PAGE_SHIFT                      12
+#define SECTOR_SHIFT						9
+
+#define SFX_GET_FREESPACE           _IOWR('N', 0x240, struct sfx_freespace_ctx)
+#define IDEMA_CAP(exp_GB) 			(((__u64)exp_GB - 50ULL) * 1953504ULL + 97696368ULL)
+
+
+enum {
+    SFX_LOG_LATENCY_READ_STATS  = 0xc1,
+    SFX_LOG_SMART               = 0xc2,
+    SFX_LOG_LATENCY_WRITE_STATS = 0xc3,
+    SFX_LOG_QUAL                = 0xc4,
+    SFX_LOG_MISMATCHLBA         = 0xc5,
+    SFX_LOG_MEDIA               = 0xc6,
+    SFX_LOG_BBT                 = 0xc7,
+    SFX_LOG_IDENTIFY            = 0xcc,
+    SFX_FEAT_ATOMIC             = 0x01,
+};
+
+enum sfx_nvme_admin_opcode {
+    nvme_admin_query_cap_info   = 0xd3,
+    nvme_admin_change_cap       = 0xd4,
+    nvme_admin_sfx_set_features = 0xd5,
+    nvme_admin_sfx_get_features = 0xd6,
+};
+
+struct sfx_freespace_ctx
+{
+    __u64 free_space;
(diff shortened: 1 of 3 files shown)


plugins/scaleflux/sfx-nvme.c
```diff
@@ -0,0 +1,877 @@
...
+    SFX_LOG_MEDIA               = 0xc6,
+    SFX_LOG_BBT                 = 0xc7,
+    SFX_LOG_IDENTIFY            = 0xcc,
+    SFX_FEAT_ATOMIC             = 0x01,
+};
+
+enum sfx_nvme_admin_opcode {
+    nvme_admin_query_cap_info   = 0xd3,
+    nvme_admin_change_cap       = 0xd4,
+    nvme_admin_sfx_set_features = 0xd5,
```
keithbusch, line 49: Just FYI, nvme spec reserved feature id's 0xc0 - 0xff for vendor specific features just so we wouldn't need vendor specific feature opcodes. Not that I'd stop you from doing what you're doing, but just pointing that out. :)

plugins/scaleflux/sfx-nvme.c
```diff
@@ -0,0 +1,877 @@
...
+    __u64 app_written; /* app data written in 4K */
+};
+
+struct nvme_capacity_info {
+    __u64 lba_sec_sz;
+    __u64 phy_sec_sz;
+    __u64 used_space;
+    __u64 free_space;
+};
+#pragma pack(push,1)
```
keithbusch, line 69: I know I've taken similar before, but i'd rather not use Microsoft dialects anymore. The gcc style is "__attribute__((packed))".

plugins/scaleflux/sfx-nvme.c
```diff
@@ -0,0 +1,877 @@
...
+struct nvme_additional_smart_log_item {
+    uint8_t            key;
```
keithbusch, line 71: I'd really like to keep the entire code base consistent with the linux c style, which uses tabs over spaces. Please try to use that convention throughout. The patch appears to be using a mixture of both conventions.

plugins/scaleflux/sfx-nvme.c
```diff
@@ -0,0 +1,877 @@
...
+};
+
+int nvme_change_cap(int fd, __u32 nsid, __u64 capacity)
+{
+    struct nvme_admin_cmd cmd = {
+    .opcode      = nvme_admin_change_cap,
+    .nsid	     = nsid,
+    };
+
+    cmd.cdw10 = (capacity & 0xffffffff);
```
keithbusch, line 119: you can just set these in the struct declaration above instead of outside it.

plugins/scaleflux/sfx-nvme.c
```diff
@@ -0,0 +1,877 @@
...
+    .opcode      = nvme_admin_sfx_get_features,
+    .nsid        = nsid,
+    .cdw10       = fid,
+    };
+
+    err = nvme_submit_passthru(fd, NVME_IOCTL_ADMIN_CMD,&cmd);
+    if (!err && result) {
+        *result = cmd.result;
+    }
+    printf("%d %d %p\n", *result,
```
keithbusch, line 150: Since 'result' is assigned only on success, you are potentially printing an uninitialized value. But the only caller prints it on success anyway, so I'm not sure why you're printing it here.

plugins/scaleflux/sfx-nvme.c
jyang-sfx: OK, thanks.

plugins/scaleflux/sfx-nvme.c
jyang-sfx: will do.

plugins/scaleflux/sfx-nvme.c
jyang-sfx: will do.

plugins/scaleflux/sfx-nvme.c
jyang-sfx: will change.

plugins/scaleflux/sfx-nvme.c
jyang-sfx: will remove.

keithbusch:
Looks fine. I'll squash the two commits into one for the final inclusion.

Could I get your "Signed-off-by" tag to affirm your certificate of origin? No need to send an new patch; just reply to this pull request with it so I can pick it up.

jyang-sfx:
This is singed of by James Yang from ScaleFlux Inc. Do I need to use git commit -s to add a new file? Thanks.

keithbusch:
`git commit -s`  is the preferred way and that automatically appends it to the commit message.

But just to manually write it, the format is:

`Signed-off-by: My Name <email@address.org>`

Of course, you must substitute your real information into the above template.

jyang-sfx:
Signed-off-by: James Yang <james.yang@scaleflux.com>

keithbusch:
Thank you, applied.

PR #722, opened 2020-05-27
Change 0xC2 log page retrieval to use UUID index

rbates0119:
Signed-off-by: Randy Bates <randall.bates@wdc.com>

### DIFF
--- a/nvme-ioctl.c
+++ b/nvme-ioctl.c
@@ -438,30 +438,7 @@ int nvme_get_log13(int fd, __u32 nsid, __u8 log_id, __u8 lsp,
 int nvme_get_log(int fd, __u32 nsid, __u8 log_id, bool rae,
 		 __u32 data_len, void *data)
 {
-	__u32 offset = 0, xfer_len = data_len;
-	void *ptr = data;
-	int ret;
-
-	/*
-	 * 4k is the smallest possible transfer unit, so by
-	 * restricting ourselves for 4k transfers we avoid having
-	 * to check the MDTS value of the controller.
-	 */
-	do {
-		xfer_len = data_len - offset;
-		if (xfer_len > 4096)
-			xfer_len = 4096;
-
-		ret = nvme_get_log13(fd, nsid, log_id, NVME_NO_LOG_LSP,
-				     offset, 0, rae, xfer_len, ptr);
-		if (ret)
-			return ret;
-
-		offset += xfer_len;
-		ptr += xfer_len;
-	} while (offset < data_len);
-
-	return 0;
+	return nvme_get_log_from_uuid(fd, nsid, log_id, rae, 0, data_len, data);
 }
 
 int nvme_get_log_from_uuid(int fd, __u32 nsid, __u8 log_id, bool rae, __u8 uuid_ix,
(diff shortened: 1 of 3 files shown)


nvme-ioctl.c
```diff
@@ -464,6 +464,36 @@ int nvme_get_log(int fd, __u32 nsid, __u8 log_id, bool rae,
 	return 0;
 }
 
+int nvme_get_log_from_uuid(int fd, __u32 nsid, __u8 log_id, bool rae, __u8 uuid_ix,
```
keithbusch, line 467: This is essentially the same as nvme_get_log(). The only difference is uuid parameter. Could you make reuse of this function instead of duplicating 99% of it?

plugins/wdc/wdc-nvme.c
```diff
@@ -1054,6 +1054,101 @@ static int wdc_create_log_file(char *file, __u8 *drive_log_data,
 	return 0;
 }
 
+bool wdc_get_dev_mng_log_entry(
+    __u32 log_length,
```
keithbusch, line 1058: This project's coding conventions is to use the linux kernel style. Please use tabs instead of spaces.

plugins/wdc/wdc-nvme.c
```diff
@@ -1054,6 +1054,101 @@ static int wdc_create_log_file(char *file, __u8 *drive_log_data,
...
+    struct wdc_c2_log_subpage_header *p_next_log_entry = NULL;
+
+    if (*p_p_found_log_entry == NULL) {
+    	fprintf(stderr, "ERROR : WDC - wdc_get_dev_mng_log_entry: No ppLogEntry pointer.\n");
+        return false;
+    }
+
+    *p_p_found_log_entry = NULL;
+
+    // Ensure log data is large enough for common header
```
keithbusch, line 1079: Also from linux style, comments use the /* */ convention

plugins/wdc/wdc-nvme.c
```diff
@@ -1091,28 +1187,45 @@ static bool get_dev_mgment_cbs_data(int fd, __u8 log_id, void **cbs_data)
...
 
-	while (length < le32_to_cpu(hdr_ptr->length)) {
-		sph = (struct wdc_c2_log_subpage_header *)(data + length);
+	if (found) {
+		*cbs_data = (void *)&sph->data;
+	} else {
 
-		if (le32_to_cpu(sph->entry_id) == log_id) {
+		/* not found with uuid = 1 try with uuid = 0 */
+		if (uuid_ix)
```
keithbusch, line 1211: I don't follow this logic. uuid_ix was initialized to 1, and it there's nothing preceeding this that modified it. When is this 'if' not taken?

rbates0119: True... will remove the if

plugins/wdc/wdc-nvme.c
rbates0119: will replace all // comments with /* */

plugins/wdc/wdc-nvme.c
rbates0119: will replace spaces with tabs

nvme-ioctl.c
rbates0119: change nvme_get_log to call nvme_get_log_from_uuid

PR #729, opened 2020-05-28
Modify JSON output routines to dump object into stream

tim-oleksii:


minwooim:
May I ask why the outputs are dumped to stream?  Maybe it can be described
in the commit message.  And I can see the merge commit which looks like you
can rebase it based on master.

PR #732, opened 2020-06-09
Fix non-qword-aligned PMRMSC read

adamdjudge:
PMRMSC is defined at a non-qword-aligned offset, so depending on the architecture used, accessing this register can cause some controllers to crash. I also noticed the CMSE bitmask in nvme_show_registers_pmrmsc was off by one and corrected it.

### DIFF
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -170,7 +170,8 @@ enum {
 	NVME_REG_PMRSTS = 0x0e08,	/* Persistent Memory Region Status */
 	NVME_REG_PMREBS = 0x0e0c,	/* Persistent Memory Region Elasticity Buffer Size */
 	NVME_REG_PMRSWTP= 0x0e10,	/* Persistent Memory Region Sustained Write Throughput */
-	NVME_REG_PMRMSC = 0x0e14,	/* Persistent Memory Region Controller Memory Space Control */
+	NVME_REG_PMRMSCL= 0x0e14,	/* Persistent Memory Region Controller Memory Space Control Low */
+	NVME_REG_PMRMSCH= 0x0e18,	/* Persistent Memory Region Controller Memory Space Control High */
 	NVME_REG_DBS	= 0x1000,	/* SQ 0 Tail Doorbell */
 };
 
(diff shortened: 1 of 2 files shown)


keithbusch:
Right, NVM 1.4 ECN 006 is addressing this. But this program's mmio read is done as 2 32-bit operations anyway, so that shouldn't incur any crashes, right?

Okay on the CMSE bitmask fix.

adamdjudge:
Ah, so it is, I didn't notice.. Hm, it was still causing some devices to crash though.

keithbusch:
And this patch fixes that crash? The memory access should be identical, afaics, so I'm curious what the root cause of the crash may be.

BTW, ECN 006 (currently yet to be ratified) defines the top half of the qword as PMRMSCU (U for "upper") rather than PMRMSCH.

adamdjudge:
Actually I got it backwards: on one drive, reading these two dwords separately worked, but somehow only when the upper dword was read first, then the lower, which might explain why mmio_read64 didn't work. No idea why this is (I wasn't working on this drive directly), but unless you think we should wait on making changes to this until after the ECN is ratified and/or if more drives have a problem with this, I'll swap the order -- and replace 'high' with 'upper' since I blanked on that.

keithbusch:
Ah, we intended to do a lo-hi read, just like the what the kernel driver does. It looks like a compiler may reorder the access from the implementation we currently have, so we do want a fix for that.

keithbusch:
The result should look something like the following using volatile reads into placeholder variables to prevent reordering:
```
        const volatile __u32 *p = addr;
        __u32 low, high;

        low = readl(p);
        high = readl(p + 1);

        return low + ((__u64)high << 32);
```

adamdjudge:
Could the compiler have optimized mmio_read64 to just do one 64-bit read rather than two 32-bit reads? I believe the person whose drive this was crashing mentioned that it was specifically an unaligned qword read causing it, which was avoided by removing that call to mmio_read64, so that's the only explanation I can think of. If that's the case then I'd imagine this would be a more reliable fix.

adamdjudge:
I'm pretty sure this is indeed the case, as I found this in the disassembly of nvme-print.o:
`21a0:   48 8b 87 14 0e 00 00    mov    0xe14(%rdi),%rax`
It seems the compiler was "helpful" enough to optimize away the call to mmio_read64 altogether! I'm going to close this pull request and make a new one for that fix to separate the reads, and probably do a separate one for the CMSE bitmask fix.

PR #739, opened 2020-06-16
Tp4053 tp4056

sc108-lee:
This commits are development for TP4053, 4056 ratified version.

### DIFF
--- a/Documentation/nvme-changed-zone-list-log.1
+++ b/Documentation/nvme-changed-zone-list-log.1
@@ -1,0 +1,102 @@
+'\" t
+.\"     Title: nvme-changed-zone-list-log
+.\"    Author: [FIXME: author] [see http://www.docbook.org/tdg5/en/html/author]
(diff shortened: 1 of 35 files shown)


keithbusch:
you may have suspected that this was also being developed in parallel elsewhere.

While the code looks mostly fine for what you were doing, there are few things that I wanted out of ZNS enabling. Mainly, being a different command set, I would have preferred to not overload the top-level commands with ZNS specifics, so ZNS really needs to go under its own hierarchy, such that invoking the command looks like "`nvme zns <cmd> /dev/nvmeX`". This is ultimately what was merged, and I would like to continue the same model with future command sets.

The most glaring issue here, though is that you've provided no way to get the result from the append operation (this has a kernel dependency on 64-bit return types in io ioctls). That's important to return, otherwise there's no good way to know where your data ended up being placed.

But anyway, basic support does exist in the tree now, and would prefer to iterate improvements and addressing any gaps from the new baseline.

PR #744, opened 2020-06-18
Drv 849 update fb log page

rbates0119:


rbates0119:
@jeffreyalien, Pull request created.

PR #745, opened 2020-06-19
Tp4053

sc108-lee:
Rebased #739 commits only picking up 4053 changes.
Fixing up bugs on upstream zns features and Add some more feature on it.

I got a question for your changes @keithbusch 
why does split report_zones, zone_mgmt_recv ?
The difference is print for report result or not.


nvme-print.c
```diff
@@ -2934,9 +2940,38 @@ void json_nvme_zns_id_ns(struct nvme_zns_id_ns *ns,
...
 
+static void show_nvme_id_ns_zoned_zoc(__le16 ns_zoc)
+{
+	__u16 zoc = le16_to_cpu(ns_zoc);
+	__u8 rsvd = (zoc & 0xfc) >> 2;
+	__u8 ze = (zoc & 0x2) >> 1;
+	__u8 vzc = zoc & 0x1;
+	if (rsvd)
+		printf("  [7:2] : %#x\tReserved\n", rsvd);
+	printf("  [1:1] : %#x\tZone Excursions are %sEnabled\n",
```
MatiasBjorling, line 2951: Zone Active Excursions

nvme-print.c
```diff
@@ -4363,7 +4422,8 @@ static void nvme_show_plm_config(struct nvme_plm_config *plmcfg)
 
 static void nvme_show_iocs_vector(struct nvme_iocs_vector *iocs_vector)
 {
-	printf("\tNVM Command Set is selected: %s\n", iocs_vector->nvm ? "True":"False");
+	printf("\tNVM   Command Set is selected: %s\n", iocs_vector->nvm ? "True":"False");
+	printf("\tZoned Command Set is selected: %s\n", iocs_vector->zoned ? "True":"False");
```
MatiasBjorling, line 4428: Zoned Namespace Command Set

plugins/zns/zns.c
```diff
@@ -223,7 +223,20 @@ static int zone_mgmt_send(int argc, char **argv, struct command *cmd, struct plu
...
 		return errno;
 
-	if (cfg.data_len) {
+	if (!cfg.zsa) {
+		fprintf(stderr, "zone send action must be specified\n");
+		err = -EINVAL;
+		goto close_fd;
+	}
+
+	if (cfg.zsa == 0x10){
```
MatiasBjorling, line 233: Define it and use it instead of 0x10.

plugins/zns/zns.c
```diff
@@ -223,7 +223,20 @@ static int zone_mgmt_send(int argc, char **argv, struct command *cmd, struct plu
 	if (fd < 0)
 		return errno;
 
-	if (cfg.data_len) {
+	if (!cfg.zsa) {
```
MatiasBjorling, line 227: May want to limit it to the only supported zone send actions. Then we avoid one trying to issue reserved actions.

plugins/zns/zns.c
```diff
@@ -223,7 +223,20 @@ static int zone_mgmt_send(int argc, char **argv, struct command *cmd, struct plu
...
+		fprintf(stderr, "zone send action must be specified\n");
+		err = -EINVAL;
+		goto close_fd;
+	}
+
+	if (cfg.zsa == 0x10){
+		if(!cfg.data_len){
+			cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
```
MatiasBjorling, line 235: If zdes = 0, then the command returns a status of invalid field in command. 

One could stop here without going further and allocate buffers (of zero size).

MatiasBjorling, line 235: If the data_len is not there, let's exit. The user should know that it supplies the data and the length of the data. What could happen, would be that a user thought it was 64bytes, but the data file provided was only 32 bytes, then the 32bytes after are essential unmapped, and would be garbled information that would be added to the zone.

Might be good to check the file size as well - but I am not sure if there is a precedence for that in nvme-cli. It is a bit on the defensive programming side.

nvme-print.c
```diff
@@ -1773,17 +1773,23 @@ static void nvme_show_id_ctrl_cmic(__u8 cmic)
 static void nvme_show_id_ctrl_oaes(__le32 ctrl_oaes)
```
MatiasBjorling, line 1773: Humen -> Human

"Add human readable for the Zoned Namespace Command Set specific Identify Namespace data structure"

nvme-print.c
```diff
@@ -1773,17 +1773,23 @@ static void nvme_show_id_ctrl_cmic(__u8 cmic)
...
 	__u32 lbasin = (oaes & 0x2000) >> 13;
 	__u32 egealpcn = (oaes & 0x4000) >> 14;
-	__u32 rsvd1 = oaes & 0xFF;
+	__u32 rsvd2 = oaes & 0xFF;
 
 	if (rsvd0)
-		printf(" [31:10] : %#x\tReserved\n", rsvd0);
+		printf(" [31:28] : %#x\tReserved\n", rsvd0);
+	printf("[27:27] : %#x\tZone Information Change Notices %sSupported\n",
+			zone_info_change_notice, zone_info_change_notice ? "" : "Not ");
```
MatiasBjorling, line 1790: Use zicn shorthand.

nvme-print.c
```diff
@@ -2957,6 +2996,10 @@ void nvme_show_zns_id_ns(struct nvme_zns_id_ns *ns,
 		printf("lbafe %02d: zsze:%"PRIx64" zdes:%u%s\n", i,
```
MatiasBjorling, line 2996: Pretty print for this as well?

keithbusch:
> I got a question for your changes @keithbusch
> why does split report_zones, zone_mgmt_recv ?
> The difference is print for report result or not.

One is specific to a known report, the other is for future extensions that we couldn't possibly know how to decode today, but would still like to be able to access as conveniently as possible. The generic one could check the zra and decode known ones like the 'report-zones' does if that's desirable.

nvme-print.c
MatiasBjorling: How about "Zone Active Excursions management is required"?

sc108-lee: @MatiasBjorling 
when its set to ‘1’, then a controller may transition the state. so I don't think required is right.
if this mentioned as required, it seems user need to do something for it.

nvme-print.c
sc108-lee: @MatiasBjorling thanks for reviewing. Didn't realize modified as zoned namespace command set :). will modify this.

plugins/zns/zns.c
sc108-lee: @MatiasBjorling forgot to modify as NVME_ZNS_ZSA_SET_DESC_EXT. ! thanks

plugins/zns/zns.c
sc108-lee: since Invalid Zone State Transition (device should return this error code) are exist, I think It can be better with not checking reserved actions. And also there is other command for each action. action limit by each command, without limit for send to check the device return right failure.

sc108-lee: And also if the command extended with other actions for the future, then will always need to be changed. if does that.

plugins/zns/zns.c
sc108-lee: @MatiasBjorling 
Great Idea

sc108-lee: @MatiasBjorling I think its better to let it be like this. because I develop based on nvme-cli as is. for example, write support stdin input for data & metadata. and It can issue smaller data then data size & meta data size.
So I did same things for extended descriptor.

nvme-print.c
sc108-lee: This is human readable format. so use Zone Information Change notice.

nvme-print.c
sc108-lee: @MatiasBjorling 
	if (human)
		show_nvme_id_ns_zoned_zoc(ns->zoc);
	printf("ozcs:   %u\n", le16_to_cpu(ns->ozcs));
	if (human)
		show_nvme_id_ns_zoned_ozcs(ns->ozcs);

already added

sc108-lee:
@keithbusch Disappointed this PR already issued, but other PR merged after this PR with same contents for zone extended descriptor print. I understand last time, because you want it to make as "nvme zns cmd". So I did modify as you developed. Do you prefer for this kind of PR to be multiple commits ?

nvme-print.c
sc108-lee: Done support human readable format

sc108-lee:
@keithbusch @MatiasBjorling modified. thanks for reviewing

nvme-print.c
MatiasBjorling: Agree - I noodled over the weekend. I think it should be:
Zone Active Excursions: Yes (Host support required) / No
Variable Zone Capacity: Yes (Host support required) / No

nvme-print.c
MatiasBjorling: For the variable ;)

MatiasBjorling:
Hi Seongno, 

It is natural to split up a PR per feature. One could split up in fixes and additions. However, for this, it is fine as is I believe.

nvme-print.c
MatiasBjorling: zone_info_change_notice  -> zicn

nvme-print.c
```diff
@@ -2934,9 +2940,38 @@ void json_nvme_zns_id_ns(struct nvme_zns_id_ns *ns,
...
+static void show_nvme_id_ns_zoned_ozcs(__le16 ns_ozcs)
+{
+	__u16 ozcs = le16_to_cpu(ns_ozcs);
+	__u8 rsvd = (ozcs & 0xfe) >> 1;
+	__u8 razb = ozcs & 0x1;
+
+	if (rsvd)
+		printf("  [7:1] : %#x\tReserved\n", rsvd);
+	printf("  [2:2] : %#x\tRead Across Zone Boundaries is %sSupported\n",
+		razb, razb ? "" : "Not ");
```
MatiasBjorling, line 2967: Read Across Zone Boundaries: Yes/No

nvme-print.c
```diff
@@ -2947,16 +2982,31 @@ void nvme_show_zns_id_ns(struct nvme_zns_id_ns *ns,
...
 	printf("frl:    %u\n", ns->frl);
 
-	for (i = 0; i <= id_ns->nlbaf; i++)
-		printf("lbafe %02d: zsze:%"PRIx64" zdes:%u%s\n", i,
-			(uint64_t)le64_to_cpu(ns->lbafe[i].zsze),
-			ns->lbafe[i].zdes, i == lbaf ? " (in use)" : "");
+	for (i = 0; i <= id_ns->nlbaf; i++){
+		if (human)
+			printf("LBA Format Extension %2d : Zone Size: %"PRIx64" blocks"
+					"Zone Descriptor Extension Size: %-5d %s\n",
```
MatiasBjorling, line 2998: Add bytes (or b, or B) after Zone Descriptor Extension Size. Such that we know it is bytes, and already been multiplied by 64 bytes.

nvme-print.c
```diff
@@ -4065,6 +4115,22 @@ const char *nvme_status_to_string(__u32 status)
 		return "RESERVATION_CONFLICT: The command was aborted due to a conflict with a reservation held on the accessed namespace";
 	case NVME_SC_FORMAT_IN_PROGRESS:
 		return "FORMAT_IN_PROGRESS: A Format NVM command is in progress on the namespace.";
+	case NVME_SC_ZONE_BOUNDARY_ERROR:
+		return "ZONE_BOUNDARY_ERROR: Invalid Zone Boundary crossing";
```
MatiasBjorling, line 4119: Note that there are different descriptions for the zone boundary errors - nothing we should do about it here. Just wanted to let you know for future patches (that may implement copy)

plugins/zns/zns.c
MatiasBjorling: I am not sure about the precedence around this one. Keith can probably decide.

plugins/zns/zns.c
```diff
@@ -223,7 +224,26 @@ static int zone_mgmt_send(int argc, char **argv, struct command *cmd, struct plu
...
+	if (cfg.zsa == NVME_ZNS_ZSA_SET_DESC_EXT){
+		if(!cfg.data_len){
+			cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
+			if (cfg.data_len == 0){
+				fprintf(stderr, 
+				"Zone Descriptor Extensions are not supported\n");
+				goto close_fd;
+			}
+			else if (cfg.data_len < 0) {
+				fprintf(stderr, "Failed to retrive valid zdes\n");
```
MatiasBjorling, line 242: Failed to retrieve the Zone Description Zone Description Extension Size? (if it is being written out in the previous error message, I think the same should be done here so it is consistent.

plugins/zns/zns.c
```diff
@@ -244,6 +264,13 @@ static int zone_mgmt_send(int argc, char **argv, struct command *cmd, struct plu
 			goto close_ffd;
 		}
 	}
+	else{
+		if (strlen(cfg.file) || cfg.data_len){
+			fprintf(stderr, "data is valid for set extended descriptor\n");
```
MatiasBjorling, line 269: is valid? looks like it should be reworded.

sc108-lee:
@MatiasBjorling resolved for modified review :)

nvme-print.c
```diff
@@ -1773,17 +1773,23 @@ static void nvme_show_id_ctrl_cmic(__u8 cmic)
...
 	__u32 plealcn = (oaes & 0x1000) >> 12;
 	__u32 lbasin = (oaes & 0x2000) >> 13;
 	__u32 egealpcn = (oaes & 0x4000) >> 14;
-	__u32 rsvd1 = oaes & 0xFF;
+	__u32 rsvd2 = oaes & 0xFF;
 
 	if (rsvd0)
-		printf(" [31:10] : %#x\tReserved\n", rsvd0);
+		printf(" [31:28] : %#x\tReserved\n", rsvd0);
+	printf("[27:27] : %#x\tZone Information Change Notices %sSupported\n",
```
MatiasBjorling, line 1789: Sorry for finding this late.

Isn't this field called "Zone Descriptor Changed"?

Should it be "Zone Descriptor Changed Event %sSupported"?

nvme-print.c
```diff
@@ -2934,9 +2940,38 @@ void json_nvme_zns_id_ns(struct nvme_zns_id_ns *ns,
...
 
+static void show_nvme_id_ns_zoned_zoc(__le16 ns_zoc)
+{
+	__u16 zoc = le16_to_cpu(ns_zoc);
+	__u8 rsvd = (zoc & 0xfc) >> 2;
+	__u8 ze = (zoc & 0x2) >> 1;
+	__u8 vzc = zoc & 0x1;
+	if (rsvd)
+		printf("  [7:2] : %#x\tReserved\n", rsvd);
+	printf("  [1:1] : %#x\tZone Excursions: %s\n",
```
MatiasBjorling, line 2951: "Zone Active Excursions" is missing.

Rename "Yes" response to "Yes (Host support required)"

nvme-print.c
```diff
@@ -2934,9 +2940,38 @@ void json_nvme_zns_id_ns(struct nvme_zns_id_ns *ns,
...
+		ze, ze ? "Yes" : "No");
+	printf("  [0:0] : %#x\tVariable Zone Capacity: \n",
+		vzc, vzc ? "Not " : "");
```
MatiasBjorling, line 2954: Make it Yes (Host support required)/No for variable zone capacity.

MatiasBjorling:
Misses the three fixes above.

nvme-print.c
sc108-lee: oh... mistake. I didn't save the files.. haha thanks for checking it again

nvme-print.c
sc108-lee: for this 
LBA Status Information Notices event(spec) -> LBA Status Information Notices %sSupported\n"(nvme-cli)
Zone Descriptor Changed Notices(spec) -> Zone Information Change Notices %sSupported\n" (nvme-cli)

MatiasBjorling: LBA Status Information Notices has "Information" in it, that is not true to Zone Descriptor Changed Notices.

I believe it should be "Zone Descriptor Changed Notices %sSupported", else an unknown person would try to lookup Zone Information Changed Notices in the spec, and come up empty-handed.

nvme-print.c
```diff
@@ -2934,9 +2940,38 @@ void json_nvme_zns_id_ns(struct nvme_zns_id_ns *ns,
...
+static void show_nvme_id_ns_zoned_zoc(__le16 ns_zoc)
+{
+	__u16 zoc = le16_to_cpu(ns_zoc);
+	__u8 rsvd = (zoc & 0xfc) >> 2;
+	__u8 ze = (zoc & 0x2) >> 1;
+	__u8 vzc = zoc & 0x1;
+	if (rsvd)
+		printf("  [7:2] : %#x\tReserved\n", rsvd);
+	printf("  [1:1] : %#x\tZone Active Excursions: %s\n",
+		ze, ze ? "Yes" : "No");
```
MatiasBjorling, line 2952: Please modify to

"Yes (Host support required)" for both ZAE and VZC.

sc108-lee: @MatiasBjorling it does not always required.

MatiasBjorling: Why is it not always required? The host will lose data if it does not implement support.

MatiasBjorling:
What this compile tested?

`nvme-print.c: In function ‘show_nvme_id_ns_zoned_zoc’:
nvme-print.c:2953:9: error: too many arguments for format [-Werror=format-extra-args]
  printf("  [0:0] : %#x\tVariable Zone Capacity: \n",
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    CC plugins/wdc/wdc-utils.o
    CC plugins/huawei/huawei-nvme.o
    CC plugins/netapp/netapp-nvme.o
    CC plugins/toshiba/toshiba-nvme.o
    CC plugins/micron/micron-nvme.o
    CC plugins/seagate/seagate-nvme.o
    CC plugins/virtium/virtium-nvme.o
    CC plugins/shannon/shannon-nvme.o
    CC plugins/dera/dera-nvme.o
    CC plugins/scaleflux/sfx-nvme.o
    CC plugins/transcend/transcend-nvme.o
    CC plugins/zns/zns.o
    CC util/argconfig.o
plugins/zns/zns.c: In function ‘zone_mgmt_send’:
plugins/zns/zns.c:235:19: error: implicit declaration of function ‘get_zdes_bytes’ [-Werror=implicit-function-declaration]
    cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
                   ^~~~~~~~~~~~~~
plugins/zns/zns.c: At top level:
plugins/zns/zns.c:489:12: error: static declaration of ‘get_zdes_bytes’ follows non-static declaration
 static int get_zdes_bytes(int fd, __u32 nsid)
            ^~~~~~~~~~~~~~
plugins/zns/zns.c:235:19: note: previous implicit declaration of ‘get_zdes_bytes’ was here
    cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
                   ^~~~~~~~~~~~~~
plugins/zns/zns.c: In function ‘zone_append’:
plugins/zns/zns.c:790:4: error: implicit declaration of function ‘elapsed_utime’ [-Werror=implicit-function-declaration]
    elapsed_utime(start_time, end_time));
    ^~~~~~~~~~~~~
plugins/zns/zns.c:789:37: error: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘int’ [-Werror=format=]
   printf(" latency: zone append: %llu us\n",
                                  ~~~^
                                  %u
    elapsed_utime(start_time, end_time));
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    CC util/suffix.o
    CC util/json.o
Makefile:95: recipe for target 'plugins/zns/zns.o' failed
make: *** [plugins/zns/zns.o] Error 1
make: *** Waiting for unfinished jobs....
Makefile:95: recipe for target 'nvme-print.o' failed
make: *** [nvme-print.o] Error 1
`

nvme-print.c
sc108-lee: About. VZC
If the Variable Zone Capacity bit is set to ‘1’ in the Zone Operation Characteristics field
in the Zoned Namespace Command Set specific Identify Namespace data structure
(refer to Figure 8), then the zone capacity may change upon successful completion of
a Zone Management Send command specifying the Zone Send Action of Reset Zone
(refer to section 4.3.1.4).

**It just means can be modified zone cap by zome mgmt send. If the host manages the capacity with kinda file system, then support required. But not all of host manages this. It means, its only required if its been handled by host.** 

About. ZAE
If set to ‘1’, then a controller may transition a zone in the ZSIO:Implicitly Opened state, the ZSEO:Explicitly Opened state, or the ZSC:Closed state to the ZSF:Full state due to a
vendor specific excursion event.

**It is also same as VZC, if all of zone status is managed by host, then it is required. But incase of implictly using zns, It is not required.**

MatiasBjorling:
It is also missing the -H option wrt to reporting of the human readable print.

sc108-lee:
@MatiasBjorling thanks for checking many things, while rebasing my older commits made many mistake.

sc108-lee:
@MatiasBjorling 
	VERBOSE	= 1 << 0,	/* verbosely decode complex values for humans */
already supported as verbose on zns id_ns.

@keithbusch Shell we make this same argument?
verbose, human-readable?
It would be better not to be confused.

nvme-print.c
MatiasBjorling: Any host, that uses a drive that has a Zone Operation Characteristic quirks, i.e., is "1", must implement support. If not, the host is broken and will get errors if it is specifically written to understand this behavior.

MatiasBjorling:
@sc108-lee 

VERBOSE is the correct as a flag - but not as a cli option.

See id_ns for an example of how to parse. The flags are set as follows in the id_ns function within nvme.c

`  err = flags = validate_output_format(cfg.output_format);
        if (flags < 0)
                goto close_fd;
        if (cfg.raw_binary)
                flags = BINARY;
        if (cfg.vendor_specific)
                flags |= VS;
        if (cfg.human_readable)
                flags |= VERBOSE;
`

sc108-lee:
@MatiasBjorling 
Thats nvme id-ns..
see nvme zns id-ns below. original commit does use verbose currently.
(For your information) my original commits were using -H option.
Inside of nvme-cli mixed up already (some command are using --verbose, others are using --human-readable). -> thats why I asked to @keithbusch to align this not to be confused.

https://github.com/linux-nvme/nvme-cli/blob/68bb51ae69e5b375545e9003a07b9159a878e793/plugins/zns/zns.c#L85

MatiasBjorling:
@sc108-lee 

That is only for a little subset of the commands. E.g., must subsys and list. For the commands that we are implementing in this PR. It's id-ctrl and id-ns. For those, -v means vendor specific, and -H means human readable. 

For the commands that you implement human-readable for - it should be changed to the above.

sc108-lee:
@MatiasBjorling yes, as I mentioned before. thats why I asked @keithbusch opinion to make this aligned as same parameter...
you can see six commands are using verbose.
list, list subsys, self-test-log
wdc command.
zns id-ns, report zone

MatiasBjorling:
Yep - If it is, don't worry. I can submit a PR if you don't want to make the change.

keithbusch:
I prefer the wording for 'verbose', but it's too late to go back on the 'old' wording and consistency is probably more important at this point.  Maybe at some point we can have both options.

nvme-print.c
sc108-lee: @MatiasBjorling 
spec says "due to a vendor specific excursion event", so it is not always...
It might be needed for some, but not always....
By device behavior, it won't be needed.

sc108-lee:
@keithbusch rebased on master.
https://github.com/linux-nvme/nvme-cli/pull/745#discussion_r443356192
Any opinion?. otherwise PR has been ready.

MatiasBjorling:
Hi Seongno,

"@MatiasBjorling
spec says "due to a vendor specific excursion event", so it is not always...
It might be needed for some, but not always....
By device behavior, it won't be needed."

If it not needed, the device should _NOT_ report as having the ZOC. In that case, it should be "No".

MatiasBjorling:
Please also fix the compiler error:

plugins/zns/zns.c: In function ‘zone_mgmt_send’:
plugins/zns/zns.c:235:19: error: implicit declaration of function ‘get_zdes_bytes’ [-Werror=implicit-function-declaration]
    cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
                   ^~~~~~~~~~~~~~
plugins/zns/zns.c: At top level:
plugins/zns/zns.c:489:12: error: static declaration of ‘get_zdes_bytes’ follows non-static declaration
 static int get_zdes_bytes(int fd, __u32 nsid)
            ^~~~~~~~~~~~~~
plugins/zns/zns.c:235:19: note: previous implicit declaration of ‘get_zdes_bytes’ was here
    cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
                   ^~~~~~~~~~~~~~
plugins/zns/zns.c: In function ‘zone_append’:
plugins/zns/zns.c:790:4: error: implicit declaration of function ‘elapsed_utime’ [-Werror=implicit-function-declaration]
    elapsed_utime(start_time, end_time));
    ^~~~~~~~~~~~~
plugins/zns/zns.c:789:37: error: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘int’ [-Werror=format=]
   printf(" latency: zone append: %llu us\n",
                                  ~~~^
                                  %u
    elapsed_utime(start_time, end_time));
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

sc108-lee:
> Hi Seongno,
> 
> "@MatiasBjorling
> spec says "due to a vendor specific excursion event", so it is not always...
> It might be needed for some, but not always....
> By device behavior, it won't be needed."
> 
> If it not needed, the device should _NOT_ report as having the ZOC. In that case, it should be "No".

Due to spec, just I give you a example of my idea..
in case of ZSIO -> write cmd (fully written)  -- controller move state --> ZSF
If the system knows device would do that, it wouldnt be need any handle.
But in this case also controller move the status, so ZAE(yes).
If you are not satisfied with my explanation, I don't mind to modify as what you recommend.
But I'm still thinking this is not always required. (since this related with vendor)
even if it does, this tool is user level tool, so it is not much needed for tool user.

Many commits are merged after I did make PR, I will not make PR as bunch of commit for TP based from know on.....
will separate commits by each modification.

sc108-lee:
> Please also fix the compiler error:
> 
> plugins/zns/zns.c: In function ‘zone_mgmt_send’:
> plugins/zns/zns.c:235:19: error: implicit declaration of function ‘get_zdes_bytes’ [-Werror=implicit-function-declaration]
> cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
> ^~~~~~~~~~~~~~
> plugins/zns/zns.c: At top level:
> plugins/zns/zns.c:489:12: error: static declaration of ‘get_zdes_bytes’ follows non-static declaration
> static int get_zdes_bytes(int fd, __u32 nsid)
> ^~~~~~~~~~~~~~
> plugins/zns/zns.c:235:19: note: previous implicit declaration of ‘get_zdes_bytes’ was here
> cfg.data_len = get_zdes_bytes(fd, cfg.namespace_id);
> ^~~~~~~~~~~~~~
> plugins/zns/zns.c: In function ‘zone_append’:
> plugins/zns/zns.c:790:4: error: implicit declaration of function ‘elapsed_utime’ [-Werror=implicit-function-declaration]
> elapsed_utime(start_time, end_time));
> ^~~~~~~~~~~~~
> plugins/zns/zns.c:789:37: error: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘int’ [-Werror=format=]
> printf(" latency: zone append: %llu us\n",
> ~~~^
> %u
> elapsed_utime(start_time, end_time));
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Thanks for checking. fixed.

MatiasBjorling:
@sc108-lee 

> > Hi Seongno,
> > "@MatiasBjorling
> > spec says "due to a vendor specific excursion event", so it is not always...
> > It might be needed for some, but not always....
> > By device behavior, it won't be needed."
> > If it not needed, the device should _NOT_ report as having the ZOC. In that case, it should be "No".
> 
> Due to spec, just I give you a example of my idea..
> in case of ZSIO -> write cmd (fully written) -- controller move state --> ZSF
> If the system knows device would do that, it wouldnt be need any handle.
> But in this case also controller move the status, so ZAE(yes).
> If you are not satisfied with my explanation, I don't mind to modify as what you recommend.
> But I'm still thinking this is not always required. (since this related with vendor)
> even if it does, this tool is user level tool, so it is not much needed for tool user.
> 

That is not a ZAE event. A write that transitions a zone from Impl. Opened state to Full state, is described as a valid transition and is not a ZAE. See section 2.4.2 in the specification. 

Please go ahead and make the modification.

MatiasBjorling:
@sc108-lee 

It doesn't compile on my system

In file included from nvme-status.c:5:0:
nvme.h:114:41: error: ‘struct timeval’ declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 unsigned long long elapsed_utime(struct timeval start_time,
                                         ^~~~~~~
Makefile:95: recipe for target 'nvme-status.o' failed
make: *** [nvme-status.o] Error 1

sc108-lee:
@MatiasBjorling 
I used older version of gcc, didnt catch this. cool.

MatiasBjorling:
Cool, it now compiles on my system.

Please go ahead and add the "Host support required" text to ZAE and VZC.

sc108-lee:
> @sc108-lee
> 
> > > Hi Seongno,
> > > "@MatiasBjorling
> > > spec says "due to a vendor specific excursion event", so it is not always...
> > > It might be needed for some, but not always....
> > > By device behavior, it won't be needed."
> > > If it not needed, the device should _NOT_ report as having the ZOC. In that case, it should be "No".
> > 
> > 
> > Due to spec, just I give you a example of my idea..
> > in case of ZSIO -> write cmd (fully written) -- controller move state --> ZSF
> > If the system knows device would do that, it wouldnt be need any handle.
> > But in this case also controller move the status, so ZAE(yes).
> > If you are not satisfied with my explanation, I don't mind to modify as what you recommend.
> > But I'm still thinking this is not always required. (since this related with vendor)
> > even if it does, this tool is user level tool, so it is not much needed for tool user.
> 
> That is not a ZAE event. A write that transitions a zone from Impl. Opened state to Full state, is described as a valid transition and is not a ZAE. See section 2.4.2 in the specification.
> 
> Please go ahead and make the modification.

**5.4 Zone Active Excursions**
A Zone Active Excursion is orthogonal to the Finish
Zone Recommended feature (refer to section 5.3), and can occur at any point when a zone has
transitioned to one of the listed zone states

5.3 issue write commands to that zone such that the zone transitions to the ZSF:Full state

MatiasBjorling:
Seongno, no, you are misreading the spec. 

You first have to look at section 2.4, specifically 2.4.2 There the valid state changes are described. There you see, that the transition that ZSIO to ZSF can be done by a write (and a couple other states), or a ZAE. So when transition from ZSIO to ZSF, this is NOT considered a ZAE event.

sc108-lee:
Based on 2.4.2 I agree, but for this feature 5.4 is main explanation.
Anyway I will follow your recommendation, since Its not much effect the command.
BTW, my name is Steven.. haha

MatiasBjorling:
Hi Steven,

My apologies!! I thought you were another person, which I've known in at a past workplace. His name is Seongno Lee, so I automatically made the connection without verifying.

sc108-lee:
@MatiasBjorling no worries, it happens haha. anyway nice conversation with you.
lets wait for merging by maintainer (@keithbusch )

MatiasBjorling:
@keithbusch Looks good to me. 

Thanks, Steven.

linux/nvme.h
```diff
@@ -418,8 +419,13 @@ struct nvme_id_ns {
 };
 
 struct nvme_iocs_vector {
-	__le64	nvm  : 1;
-	__le64	rsvd : 63;
+	union {
+		__le64 nvm   : 1;
+		__le64 rsvd1 : 1;
+		__le64 zoned : 1;
+		__le64 rsvd3 : 61;
```
keithbusch, line 426: We can't use bitfields in a program where bit order matters because there's no guarantee from C what order it will lay them out. C99 reference, Section 6.7.2 paragraph 10:

"The order of allocation of bit-fields within a unit (high-order to low-order or low-order to high-order) is implementation-defined."

PR #759, opened 2020-06-25
nvme-print: streamline show-regs output

birkelund:
Fix mixed tabs/spaces use in show-regs pretty print and align tabulated
output.


MatiasBjorling:
I prefer the short-hand to be straight after the long name, followed by an adequate number of spaces. Keith?

birkelund:
Makes no difference to me, I just streamlined the minority with the majority for that command, but I guess most other pretty prints are as you describe.

MatiasBjorling:
It is a good cleanup. The reporting is all over the place.

PR #760, opened 2020-06-26
Zns refactoring

sc108-lee:
1. To do exception handling, the function needs nsid.
so nsid exception moved before using nsid

2. The data_len should be in bytes while setting zone descriptor.
And the logic already exist as a function, so use the function for it.


keithbusch:
Yikes, this doesn't even compile...

sc108-lee:
@keithbusch sorry for mistake. I can see that you committed fix. thanks

PR #765, opened 2020-06-30
Add zone-mgmt-recv parameters to issue command

sc108-lee:


keithbusch:
On second thought, why call this "partial" for the generic zone-mgmt-recv? That parameter is defined to "partial" only for specific receive actions.

sc108-lee:
@keithbusch 
original value name was (zrass)
zrass was not clear because the 16th bit is Zone Receive Action Specific Features.
And zrasf is taken by Zone Receive Action Specific Field.
so I just use that partial.
should we change this to other name?

keithbusch:
I would like to change it because this is the generic command interface and 'partial' is the name of the field when it is specific to specific actions. The generic name, "Zone Receive Action Specific Features" unfortunately shares the same acronym as another field, "Zone Receive Action Specific Field".

sc108-lee:
zras-feat, zras-field
how does it sound?

Issue #768, opened 2020-07-01
Negative error code should be handled by zns routines


ygoldfill:
routines that use nvme_submit_io_passthru for example handle negative error code returned from the driver.
nvme_zns_mgmt_send and nvme_zns_mgmt_recv assume just nvme error is returned

sc108-lee:
point taken, updated by below PR !
https://github.com/linux-nvme/nvme-cli/pull/769

PR #769, opened 2020-07-03
Handling system error #768

sc108-lee:


minwooim:
It would be great if it would have some commit descriptions related
to the issue posted.

Anyway, The changes are looking good to me.

sc108-lee:
@keithbusch Is there any problem with this?

PR #770, opened 2020-07-03
Zns completions

sc108-lee:
zsh (subcommand level support)
bash (Add subcommand and options level support, Add plugin feature with function)
modify report zone option

### DIFF
--- a/completions/bash-nvme-completion.sh
+++ b/completions/bash-nvme-completion.sh
@@ -12,10 +12,68 @@ _cmds="list id-ctrl id-ns list-ns id-iocs nvm-id-ctrl create-ns delete-ns \
 	write-uncor copy reset subsystem-reset show-regs discover \
 	connect-all connect disconnect version help \
 	intel lnvm memblaze list-subsys endurance-event-agg-log \
-	lba-status-log resv-notif-log"
+	lba-status-log resv-notif-log zns"
+
+plugin_list_opts_zns () {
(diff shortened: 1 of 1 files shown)


sc108-lee:
@keithbusch 
Do you prefer using plugin function inside of case instead of using checking function existence logic??

shmily1012:
Hi There,

I am wondering if you have plan to include the 'ZRWA' related feature in TP4076.

Thanks,
Alex

MatiasBjorling:
Hi @shmily1012 

The ZRWA technical proposal is in the progress of being standardized within the NVMe ZNS Task Group.  Support will be added when the TP has been ratified and integrated into the ZNS Command Set specification. 

Best, Matias

sc108-lee:
Hi @shmily1012 
As Matias mentioned, TP is under progress.
once TP is ratified, I might be able to do that feature.
Thanks

shmily1012:
Hi @sc108-lee  and @MatiasBjorling,
Sounds good. Thanks for the quick response. 
Regards,
Alex

sc108-lee:
@keithbusch if you have any other reason to not using this, let me know :)
it could be issued as "plugin_list_opts_zns ${words[2]} $prev" under the case.
but I add like this to skip cases for the plugins

sc108-lee:
since there is other PR has been accepted, this PR means nothing.
so close it.

PR #773, opened 2020-07-10
add path to nvme list

tomzawadzki:
In SPDK project (spdk.io) it is possible to represent
devices using CUSE as character devices in "/dev/spdk/" directory.
No sysfs entries are generated for those devices.

nvme-cli default behavior is to scan the sysfs for entries on
newer kernels. Previously legacy_list() was used to determine
the topology, but that worked only on "/dev/" path.

To enable listing SPDK devices following changes were made:
- pass path parameter to list command
- before checking sysfs, check the path using legacy_list()
- expand arguments of scan_subsystems and related functions to accept the path

Example:
nvme list -d /dev/spdk/

Signed-off-by: Tomasz Zawadzki <tomasz.zawadzki@intel.com>

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -1699,7 +1699,7 @@ int fabrics_disconnect_all(const char *desc, int argc, char **argv)
 		for (j = 0; j < s->nr_ctrls; j++) {
 			struct nvme_ctrl *c = &s->ctrls[j];
 
-			if (!strcmp(c->transport, "pcie"))
+			if (!c->transport || !strcmp(c->transport, "pcie"))
 				continue;
 			err = disconnect_by_device(c->name);
 			if (err)
--- a/fabrics.c
+++ b/fabrics.c
@@ -1687,7 +1687,7 @@ int fabrics_disconnect_all(const char *desc, int argc, char **argv)
 	if (err)
 		goto out;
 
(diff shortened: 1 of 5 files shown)


minwooim:
If we are going to support a kind of generic usage for the scanning location,
It definitely needs to provide help commands by adding a parameter for the
dev_path.

@keithbusch ,
One thing want to ask, Does nvme-cli have some plan to support non-kernel
driver-based device like SPDK for generic in the built-in plugin?
Maybe we can make it as a plugin.

tomzawadzki:
@minwooim Thanks for the input, the help command for `-d` parameter has been updated through out the series.

My understanding is that plugins in nvme-cli for the most part handle vendor extensions for the commands. SPDK NVMe driver does not by itself add such commands and provides interface compliant with the specification. 

Fortunately nvme-cli already can point to specific devices for most commands (see examples in https://spdk.io/doc/nvme.html#nvme_cuse). Exception to this was `list` command that either used hardcoded `/dev/` path or sysfs to search for devices.
The goal of the series is to bring `list` in line with other commands to enable passing a path to search for devices.

Tagging @keithbusch

jimharris:
Thanks @keithbusch!

PR #774, opened 2020-07-11
nvme-print: Unify nsid format to decimal

minwooim:
Hi,

From the "nvme list" command, NSID format is not the same between verbose
and non-verbose mode.

I prefer nsid as a decimal format so I posted this patch converting the hex format
to the decimal for the verbose-mode, but If it's should be in hex format, I can re-post
another patch to convert non-verbose mode printing format.

Thanks,

---
When printing the ns list with "nvme list -v", the nsid is not matched
with the nsid from the non-verbose mode which prints in decimal format.


minwooim:
Hi @keithbusch ,

Can we go with this patch? I think the nsid should be printed out in decimal format
OR it should be printed out with 0x prefix in case of hexadecimal.

Please have a look.
Thanks,

minwooim:
Closed this MR not to modify the actual format of the output value.

https://github.com/linux-nvme/nvme-cli/pull/850 posted.

PR #776, opened 2020-07-13
Add WDC Plugin Capabilities Command

bjpaupor:
New command to list the capabilities of the nvme device with respect to the other commands in the WDC plugin.


bjpaupor:
@keithbusch Could you take a look at this when you get a chance?

Issue #703, opened 2020-04-30
Read Nlog in nvme cli fails due to invalid field


Sumithra92:
Commands in nvme cli to read internal NLogs (Tried with v1.8/v1.9/v1.11.1): 
**nvme intel internal-log /dev/nvme0 -l 0 -r 1 -m 1**

The above command fails with the following error message:

failed on cmd.data_len 4096 cmd.cdw13 0 cmd.cdw12 28 cmd.cdw10 1024 err 4002 remaining size 1024
--
NVMe Status:INVALID_FIELD: A reserved coded value or an unsupported value in a defined field(4002)

keithbusch:
does your drive actually support this vendor specific log page?

Sumithra92:
> does your drive actually support this vendor specific log page?

Hi Keith - Yes. The drive supports nlogs.

Sumithra92:
Hi! 
I do apologize for the long inactivity. We figured out the issue for now and are making necessary changes to the product. This issue can be closed for now.

PR #781, opened 2020-07-22
Error Message Suggesting Telemetry

Dishoungh:
For some newer drives that don't support internal_log, print out an
error message recommending telemetry. This should prevent reoccurrence of #703

Signed-off-by: Dishoungh White II <47036723+Dishoungh@users.noreply.github.com>

### DIFF
--- a/plugins/intel/intel-nvme.c
+++ b/plugins/intel/intel-nvme.c
@@ -1138,6 +1138,13 @@ static int get_internal_log(int argc, char **argv, struct command *command,
 		err = get_internal_log_old(buf, output, fd, &cmd);
 		goto out;
 	}
+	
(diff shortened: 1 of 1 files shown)


Dishoungh:
@igaw To be perfectly honest with you, I forgot I worked on this. Yes, at the time, I was working for Intel NSG as an intern (now SK Hynix I heard). At the time, I was working with @benreese0 (don't know if this is him but I think it is). Sorry, I don't know any way to point you where I got that information. All I can go from is memory and all I remember was that certain types of Intel (or SK Hynix; I'll just call it Intel still to make it easier) SSD SKUs could use that internal_log command and some couldn't.

igaw:
Let's close this PR. We can still reopen if we get an official confirmation.

PR #788, opened 2020-08-04
Add support for Replay Protection Memory Block (RPMB) commands

hanumanthuh:
RPMB commands with HMAC-SHA256 computed using Linux kernel crypto services and using MD5 (computed using Linux kernel crypto services) of a random number as nonce. Currently supports
 - read-counter  - read current value of write counter
 - read-data       - read data from specified RPMB target onto given file
 - read-config    - reads configuration block data from specified RPMB target onto given file
 - program-key  - configures given authentication key for an RPMB target
 - write-data      - writes given block of data onto specified RPMB target
-  write-config   - writes device configuration block data to specified RPMB target
- info                  - prints current RPMBs support info of given NVMe device

TODO:
Need to check on what should be abstraction at nvme cli  to support boot partitions (lock, unlock, enable protection etc).

### DIFF
--- a/Documentation/nvme-rpmb.1
+++ b/Documentation/nvme-rpmb.1
@@ -1,0 +1,326 @@
+'\" t
+.\"     Title: nvme-rpmb
+.\"    Author: [FIXME: author] [see http://docbook.sf.net/el/author]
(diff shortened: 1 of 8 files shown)


hanumanthuh:
@keithbusch can you please look into this ?  If this PR needs any further changes to be merged, I can make them real quick

PR #804, opened 2020-09-04
Report-zones, get-log : transfer with mdts

sc108-lee:
get log :
no need to retrieve in 4k if its possible.

report zone :
if there are lots of zones on device, getting report zone will be failed without checking limitation.
so using mdts to split the report data retrieve all zones.


nvme-topology.c
```diff
@@ -644,3 +644,80 @@ void *mmap_registers(const char *dev)
 	return membase;
 }
 
+unsigned int max_data_transfer_size(int fd){
```
keithbusch, line 647: I think it's perfectly safe to assume 4k for minimum page size. The kernel won't let one go through anyway for various reasons, and it's unlikely a device that has a larger one would be very successful in the market anyway just because x86 uses it.

Further, many kernels have the IOMEM_STRICT enabled, so you won't even be able to map the registers.

So, this function should be simplified and more reliable with that assumption in mind.

sc108-lee: Thanks for your opinion,
Base on that modified and rebased :)

PR #809, opened 2020-09-18
Zns print fixes

floatious:
Hello Keith,

Here are some minor ZNS print fixes when using human readable format.

Kind regards,
Niklas


keithbusch:
Yep, all looks good. Thanks for the fixes

PR #812, opened 2020-09-28
Improve output readability of the zns id-ns command

MatiasBjorling:
- Expose MAR/MOR in a human readable format.
- Update spaces between values to align appropriately.
- Expose the MAR/MOR/FRL/RRL fields as unsigned decimal instead of hex.

Signed-off-by: Matias Bjørling <matias.bjorling@wdc.com>


keithbusch:
I have a feeling we're going to revisit this again. The use of \t in printf formats will eventually not align as expected depending on what the preceding formats output. Will apply this now anyway.

MatiasBjorling:
Yep - you're probably right.

Although, I don't expect drives to support more than 10M Open/Active zones any time soon, heh.  Which we would need for it to break.

PR #815, opened 2020-10-02
WDC Plugin Enclosure changes

jeffreyalien:


plugins/wdc/wdc-nvme.c
```diff
@@ -992,15 +1018,71 @@ static int wdc_get_pci_ids(uint32_t *device_id, uint32_t *vendor_id)
...
+
+	*vendor_id = (uint32_t) ctrl.vid;
+
+	return ret;
+}
+
+static bool wdc_check_power_of_2(int num)
+{
+	int first_set = 1;
+
```
Jeemon-Joy, line 1042: Better you can have below logic
`return (num && ( !(num & (num-1))));`

jeffreyalien: Jeemon-Joy,
Just to confirm, you're suggesting I change the wdc_check_power_of_2 function to the following, correct?
static bool wdc_check_power_of_2(int num)
{
        return (num && ( !(num & (num-1))));
}

jeffreyalien: Updated the wdc_check_power_of_2 function with your suggested change.

PR #818, opened 2020-10-19
Fix to WDC Plugin vs-internal-log command for SN730 Data Area 3

jeffreyalien:


plugins/wdc/wdc-nvme.c
```diff
@@ -2114,22 +2115,23 @@ static int wdc_do_cap_dui(int fd, char *file, __u32 xfer_size, int data_area, in
...
-		cap_dui_length = le32_to_cpu(log_hdr_v4->log_size_sectors) * WDC_NVME_SN730_SECTOR_SIZE;
+		cap_dui_length_v4 = le64_to_cpu(log_hdr_v4->log_size_sectors) * WDC_NVME_SN730_SECTOR_SIZE;
 
 		if (verbose) {
 			fprintf(stderr, "INFO : WDC : Capture V4 Device Unit Info log, data area = %d\n", data_area);
 			fprintf(stderr, "INFO : WDC : DUI Header Version = 0x%x\n", log_hdr_v4->hdr_version);
 			fprintf(stderr, "INFO : WDC : DUI Product ID = 0x%x/%c\n", log_hdr_v4->product_id, log_hdr_v4->product_id);
-
+			fprintf(stderr, "INFO : WDC : DUI log size sectors = 0x%x\n", log_hdr_v4->log_size_sectors);
+			fprintf(stderr, "INFO : WDC : DUI cap_dui_length = 0x%lx\n", (unsigned long int)cap_dui_length_v4);
```
keithbusch, line 2135: If it's a 64-bit value, casting to long on a 32-bit arch is usually going to yield a 32-bit value.

jeffreyalien: Thanks for the comment Keith.  I probably should verify these changes on a 32 bit machine (if I can find one).   I don't think we have any in our stock of test systems.

jeffreyalien:
Keith, please review the latest commit.  I think this should resolve your comment.

plugins/wdc/wdc-nvme.c
```diff
@@ -2132,7 +2133,7 @@ static int wdc_do_cap_dui(int fd, char *file, __u32 xfer_size, int data_area, in
 			fprintf(stderr, "INFO : WDC : DUI Header Version = 0x%x\n", log_hdr_v4->hdr_version);
 			fprintf(stderr, "INFO : WDC : DUI Product ID = 0x%x/%c\n", log_hdr_v4->product_id, log_hdr_v4->product_id);
 			fprintf(stderr, "INFO : WDC : DUI log size sectors = 0x%x\n", log_hdr_v4->log_size_sectors);
-			fprintf(stderr, "INFO : WDC : DUI cap_dui_length = 0x%lx\n", (unsigned long int)cap_dui_length_v4);
+			fprintf(stderr, "INFO : WDC : DUI cap_dui_length = 0x%llX\n", cap_dui_length_v4);
```
keithbusch, line 2136: I believe the 64 bit print formats really do need to use the PRIx64 rather than llx due to compatibility across archs.

jeffreyalien: Keith, 
How does this look?  I copied an example from nvme-print.c.   I tried to do it without having to cast but couldn't get that to compile.  

fprintf(stderr, "INFO : WDC : DUI cap_dui_length = 0x%"PRIx64"\n", (uint64_t)cap_dui_length_v4);

PR #819, opened 2020-10-20
scaleFlux: add led control sub command

whou-sfx:


whou-sfx (commented):
@keithbusch , I hit a wired issue when pull request to merge my forked repo to upstream repo. 
There are many changed files listed in draft pull request are not the ones I make. And the diff shows are already in head of upstream repo and should not list here. 
Do you have any suggestion?

plugins/scaleflux/sfx-nvme.c
```diff
@@ -594,6 +602,143 @@ static int sfx_get_bad_block(int argc, char **argv, struct command *cmd, struct
...
+			fprintf(stderr, "readlink fail, errno =%d \n", errno);
+			return -1;
+		}
+	} else {
+		snprintf(tg_name, tg_size, "%s", link_name);
+	}
+	return 0;
+}
+
+static int nvme_block_from_char(char *char_dev, char *blk_dev)
```
keithbusch: This doesn't actually work with the linux driver. And there should already be a function that does this, if I'm not mistaken.

whou-sfx: The prototype is from nvme_block_from_char() but do convert scaleflux misc char device name "sfxv*" to block device name "sfdv*n1". I will change the function to avoid confusion.

whou-sfx: Do you have any idea about why the pull request show diffs which already in master?

PR #825, opened 2020-10-28
nvme.spec.in: using specific directory for nvme command on %post

sc108-lee:
${_sbindir} could be placed in out of default bin directories
so just 'nvme gen-hostnqn' would not work with 'nvme command not found'


sc108-lee:
if default directory is set out of runnable directory, then the command always failed with  'nvme command not found'

PR #827, opened 2020-10-29
nvme-print: Adding typedef enum nvme_feat

ErwanAliasr1:
This patch switch the feature_id into an typedef enum called nvme_feat.

Having typedef enum ease the implementation as Werror=switch compile flag will detect forgotten case statements.

A build failure looks like :
	nvme-print.c:4525:2: error: enumeration value ‘NVME_FEAT_RRL’ not handled in switch [-Werror=switch]

This patch adds the NVME_FEAT_SANITIZE & NVME_FEAT_RRL inside nvme_feature_show_fields().

A NVME_FEAT_NONE=0 value is added to give a neutral/default value when needed.

Signed-off-by: Erwan Velu <e.velu@criteo.com>


keithbusch:
> A build failure looks like :
> nvme-print.c:4525:2: error: enumeration value ‘NVME_FEAT_RRL’ not handled in switch [-Werror=switch]

Is there a compiler actually generating this warning with the current code base?

I would rather just add a "default:" in the switch.

ErwanAliasr1:
Gcc does and with the existing code base.

 With a switch on a typedef enum, the default case isn't suitable as it
will mask the missing members.

Without default, If some adds a new member in nvme_feat, the compiler will
automatically catch the missing case statements.

Very useful for the code quality and feature support.

Le jeu. 29 oct. 2020 à 19:15, Keith Busch <notifications@github.com> a
écrit :

ErwanAliasr1:
I've been checking, this is at least supported since GCC 4.0, and surely earlier. So sounds safe to use this feature.

keithbusch:
Gcc doesn't produce such a warning on the current code base. I think you are saying it will produce such a warning if you use a named enum type within the switch, and you want to warning to happen to compel it to be handled.

So that's all fine. The only thing you should change is that the patch should align to the local coding style, which comes from the Linux kernel:

https://www.kernel.org/doc/html/latest/process/coding-style.html#typedefs

The examples use 'struct', but the same reasoning applies to 'enum'. So, I'm fine to name the enums and group them for their different categories, but let's not hide the fact that these are 'enum' types through a typedef.

ErwanAliasr1:
About the GCC feature :
I've been taking this patch and removed the NVME_FEAT_SANITIZE case from nvme_feature_show_fields(), then rebuild on two different gcc release (4.8 on centos & 9.3 on Ubuntu)

So to me that's perfectly supported and functional. If a case is missing in a statement, this generates not a warning but an error.

[root@host:/tmp/nvme-cli]# gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,java,fortran,ada,go,lto --enable-plugin --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.8.5 20150623 (Red Hat 4.8.5-39) (GCC) 
[root@94-40-c9-84-08-e4.inventory.criteo.prod /tmp/nvme-cli]# make
    CC nvme-print.o
nvme-print.c: In function 'nvme_feature_show_fields':
nvme-print.c:4527:2: error: enumeration value 'NVME_FEAT_SANITIZE' not handled in switch [-Werror=switch]
  switch (fid) {
  ^
make: *** [nvme-print.o] Error 1
[root@host: /tmp/nvme-cli]# 


e.velu@R90YT7WC:~/Devel/nvme-cli (evelu-feature)$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/9/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:hsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 9.3.0-17ubuntu1~20.04' --with-bugurl=file:///usr/share/doc/gcc-9/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++,gm2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-9 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-9-HskZEa/gcc-9-9.3.0/debian/tmp-nvptx/usr,hsa --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04) 
e.velu@R90YT7WC:~/Devel/nvme-cli (evelu-feature)$ make
    CC nvme-print.o
nvme-print.c: In function ‘nvme_feature_show_fields’:
nvme-print.c:4527:2: error: enumeration value ‘NVME_FEAT_SANITIZE’ not handled in switch [-Werror=switch]
 4527 |  switch (fid) {
      |  ^~~~~~
make: *** [Makefile:96 : nvme-print.o] Erreur 1
e.velu@R90YT7WC:~/Devel/nvme-cli (evelu-feature)$

ErwanAliasr1:
About the typedef enum, that's pretty common in the kernel like 
https://elixir.bootlin.com/linux/v5.10-rc1/source/include/uapi/linux/net.h#L48

I don't really see what the issue is here.
This is naming the enum and makes it a type that can be checked across functions like switch().

The current enum wasn't typed and was mixing various concepts in it.
I'd love to offer other enum types so each one is managing a single data type.

Does this makes sense for you ?

keithbusch:
> About the GCC feature :
> I've been taking this patch and removed the NVME_FEAT_SANITIZE case from nvme_feature_show_fields(), then rebuild on two different gcc release (4.8 on centos & 9.3 on Ubuntu)

There's no question about the gcc capability. My question was if the warning was somehow producing on the current code base, which it's not. You have to modify the code to raise such an error.

> So to me that's perfectly supported and functional. If a case is missing in a statement, this generates not a warning but an error.

All warnings are errors here.

keithbusch:
> About the typedef enum, that's pretty common in the kernel like
> https://elixir.bootlin.com/linux/v5.10-rc1/source/include/uapi/linux/net.h#L48
> 
> I don't really see what the issue is here.

I don't care about subsystem specific exceptions in the kernel. We have a decided on a coding style that has a long standing document explaining what is acceptable to help drive consistency.

> This is naming the enum and makes it a type that can be checked across functions like switch().

The same check works with declaring the type as 'enum nvme_feature_id' too.

But the nvme spec doesn't enumerate all possible values for a feature id either. There are 256 possible values, but only a subset are enumerable in the spec. The rest are still valid values that would fit within the enum definition.

ErwanAliasr1:
> My question was if the warning was somehow producing on the current code base, which it's not.

Oh sorry, I didn't understood your remark this way.
That's more or less true.
Yes the actual code doesn't fail because of the data type, the swtich is done on a int so gcc as no idea of what's missing.
If I switch to the enum, it does fail, this is how I discovered the two missing one in nvme_feature_show_fields().
So having this feature sounds safer to me to get an explicit case per value of the enum.

ErwanAliasr1:
@keithbusch is it good for you ?

ErwanAliasr1:
lgtm ? :pray:

keithbusch:
I'm just catching up this week. But I'm still mulling over the implementation. The fact you're passing through an enum type to functions and as a command line parameter type suggests that possible values are constrained to the enum, which isn't really true. I think I would prefer to leave most of the code as-is, and you can cast the generic int/u32 value to the enum type within the functions before the switch. Does that sound okay?

ErwanAliasr1:
The existing code is already using an enum to store these values. When I went to the code, I was surprised to see an enum mixing several types of variables and reproducing several time the same values.
So my proposal is to keep the enum but name it so we can leverage the code verification. I initially didn't get that all values are not in this enum but I see of value of doing it this way.
If one is adding a new value into the enum, the compilation process will ensure that all values are properly managed by the switches. I also see a value having types to ease the code reading.
The idea would be to move the other values (like NVME_LOG* or NMVE_FWACT*) into their own structures.

keithbusch:
We just used an unnamed enum to hold constant values. It's better than a bunch of \#defines.

ErwanAliasr1:
So why not having named enum per constant types.

keithbusch:
I'm fine with naming the enums and grouping them for their field. libnvme does that, too.

I just don't think it's appropriate to use these enum types as command line or function parameters because they can't enumerate legal values, and they're the wrong size. For example, an nvme feature id is 8 bits, but an enum is an int.

ErwanAliasr1:
> For example, an nvme feature id is 8 bits, but an enum is an int.

But the feat was a __u32 isn't it ? 

WDYT about using an  __attribute__ ((__packed__)) ?

ErwanAliasr1:
with the packed attribute, the sizeof equals 1

ErwanAliasr1:
I forgot to mention that as we don't have negative values in the enum, gcc makes it as an unsigned int not an int. So it's sounds to me we would be fine.

ErwanAliasr1:
@keithbusch I pushed the latest commit with the packing.
So we have an uint here.
Is everything fine for you ?

keithbusch:
Hand merged to modify commenting style (always uses /* */, per the style guide).

ErwanAliasr1:
@keithbusch Thanks for the review, comments and merge.
I'll offer other PRs in this area.

PR #833, opened 2020-11-16
fix dsm and copy range setup

birkelund:
The argconfig array helpers parse into int and unsigned long long
arrays. In the copy function we are casting an int* to __u16* and we end
up with wrong values. Instead, just use the types in the setup functions
directly and convert the individual values as part of calling the
cpu_to_le* helpers.

While copy is the one that is immediately broken on most systems, do
this for DSM as well since it breaks on systems where int is not 32
bits.

Fixes: 5dc44e6c21a1 ("add support for issuing simple copy commands")


keithbusch:
shouldn't we fix the arg parsing instead? Seems like we should prefer the protocol's natural type rather than coerce to 'int'.

birkelund:
Ugh. I guess so... I'll see what I can come up with.

PR #845, opened 2020-12-21
Micron plugin changes - version 1.0.4

hanumanthuh:
Parsing support for Extended Smart Health information log on Micron drives


keithbusch:
I merged this one by hand since it had conflicts on some of the log page updates supporting the LSP field. I think it's okay, but let me know if I messed it up.

PR #850, opened 2021-01-07
nvme-print: notify hexadecimal format to nsid header

minwooim:
The following two different commands print different format of Namespace
ID.

	nvme list:		Prints decimal nsid
	nvme list --verbose:	Prints hexadecimal nsid without 0x prefix.

It might cause confusions to users for figuring out namespace IDs.
Also, it might not possible to update the print-format because there
might many systems that parse the output of this command as a
hexadecimal.  If we put the prefix '0x' to the value, it might cause
some problems for systems to parse that value based on the previous
format.

This patch describes the format of the output value to the header:

	NSID(hex)
	------------
	48

It makes clear that the output value is in hexadecimal format.


keithbusch:
I'm not sure if the hex/dec inconsistency was intentional, but I feel that printing in decimal makes more sense since that's how the kernel's sysfs will show the nsid.

minwooim:
Agreed.  I was wondering if it's okay to modify the origin format or not.

PR #852, opened 2021-01-07
nvme: set block size to blkdev before re-read partition

minwooim:
Before re-reading partition information from a block device, to make
block size of the block device, application should do either:

	- Close current file descriptor and re-open it again.
	- Request block device to set block size by itself.

Kernel community has discussions about it, and can be found at [1].

The first option might not be good for readability in this program which
is one-shot command-based application.  All the subcommands are
constructed "fd open -> issue command -> fd close".  I don't want to
break the common style for this requirement.

This patch takes the second one and it just ioctl() to the block device
to update it's block size to a given value.

[1] https://lore.kernel.org/linux-nvme/20210105150214.GA16251@lst.de/T/#u

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2753,6 +2753,7 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 	struct nvme_id_ns ns;
 	struct nvme_id_ctrl ctrl;
 	int err, fd, i;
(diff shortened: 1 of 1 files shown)


keithbusch:
Yeah, this is fine.

PR #854, opened 2021-01-08
nvme: add support for predictable latency per NVM set log page

ganaidu707:
This is to add support for LID = 0x0A, Predictable Latency per
NVM Set, used to determine the current window for the specified
NVM Set when Predictable Latency Mode is enabled and any events
that have occurred for the specified NVM Set. For More details
see NVM Express 1.4 Spec. Section 5.14.1.10 ("Predictable Latency
Per NVM Set (Log Identifier 0Ah)")

Co-Authored-By: Karthik Balan  <karthik.b82@samsung.com>

### DIFF
--- a/Documentation/nvme-pred-lat-per-nvmset-log.1
+++ b/Documentation/nvme-pred-lat-per-nvmset-log.1
@@ -1,0 +1,112 @@
+'\" t
+.\"     Title: nvme-pred-lat-per-nvmset-log
+.\"    Author: [FIXME: author] [see http://docbook.sf.net/el/author]
+.\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
+.\"      Date: 01/07/2021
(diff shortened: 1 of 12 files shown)


keithbusch:
Ah, this is conflicting with some of your other pull's that I've merged. Perhaps I did them in the wrong order?

In any case, the name of the command line invocation, pred-lat-per-nvmset-log seems a bit long, yeah? Can we call it "predictable-lat-log" instead? There shouldn't be any confusion as to what log this refers to, I think.

ganaidu707:
Yeah, it is the wrong order, no issue, I will resolve conflicts or maybe resubmit it!

Yeah, it is a bit long, sure, I will make it as "predictable-lat-log".

Thanks, Keith.

ganaidu707:
@keithbusch conflicts are huge so resubmitted with below mentioned pull id, and your review comments addressed!.
https://github.com/linux-nvme/nvme-cli/pull/864

PR #877, opened 2021-01-26
RFC: add "nvme monitor" subcommand

mwilck:
This patch set adds a new subcommand **nvme monitor**. In this mode, **nvme-cli** runs continuously, monitors events (currently, uevents) relevant for discovery, and optionally autoconnects to newly discovered subsystems.

The monitor mode is suitable to be run in a systemd service. An appropriate unit file is provided. As such, **nvme monitor** can be used as an alternative to the current auto-connection mechanism based on udev rules and systemd template units. If `--autoconnect` is active, **nvme monitor** masks the respective udev rules in order to prevent simultaneous connection attempts from udev and itself.

This method for discovery and autodetection has some advantages over the current udev-rule based approach:

 * By using the `--persistent` option, users can easily control whether persistent discovery controllers for discovered transport addresses should be created and monitored for AEN events. **nvme monitor** watches known transport addresses, creates discovery controllers as required, and re-uses existing ones if possible. It keeps track of persistent discovery controllers it created, and tears them down on exit. When run in `--persistent --autoconnect` mode *in the initial ramfs*, it will keep discovery controllers alive, so that a new instance started after switching root can simply re-use them.
 * In certain situations, the systemd-based approach may miss events due to race conditions. This can happen e.g. if an FC remote port is detected, but shortly after it's detection an FC relogin procedure is necessary e.g. due to an RSCN. In this case, an `fc_udev_device` uevent will be received on the first detection and handled by an `nvme connect-all` command run from `nvmf-connect@.service`. The connection attempt to the rport in question will fail with "no such device" because of the simultaneous FC relogin. `nvmf-connect@.service` may not terminate immediately, because it attempts to establish other connections listed in the Discovery Log page it retrieved. When the FC relogin eventually finishes, a new uevent will be received, and `nvmf-connect@` will be started again, but *this has no effect* if the previous `nvmf-connect@` service hasn't finished yet. This is the general semantics of systemd services, no easy workaround exists. 
**nvme monitor** doesn't suffer from this problem. If it sees an uevent for a transport address for which a discovery is already running, it will queue the handling of this event up and restart the discovery after it's finished.
 * Resource consumption for handling uevents is lower. Instead of running an udev worker, executing the rules, executing `systemctl start` from the worker, starting a systemd service, and starting a separate **nvme-cli** instance, only a single `fork()` operation is necessary. Of course, on the back side, the monitor itself consumes resources while it's running and waiting for events. On my system with 8 persistent discovery controllers, its RSS is ~3MB. CPU consumption is zero as long as no events occur. 
* **nvme monitor** could be easily extended to handle events for non-FC transports. 

I've tested `fc_udev_device` handling for NVMeoFC with an Ontap target, and AEN handling for RDMA using a Linux **nvmet** target.

### Implementation notes

I've tried to change the exisiting **nvme-cli** code as little as possible while reusing the code from `fabrics.c`. The majority of changes in the exisiting code exports formerly static functions and variables, so that they are usable from the monitor code. 

The main process just waits for events using `epoll()`. When an event is received that necessitates a new discovery, a child is forked. This makes it possible to fill in the configuration parameters for `do_discover()` without interfering with the main process or other  discovery tasks running in parallel. The program tracks *transport addresses* (called "connections" in the code) rather than NVMe controllers. In `--persistent` mode, it tries to maintain exactly one persistent discovery connection per transport address. 

Using `epoll()` may look over-engineered at this stage. I hope the better flexibility over `poll()` (in particular, the ability to add new event sources while waiting) will simplify future extensions and improvements.

### Todo

 * Referrals are not handled perfectly yet. They will be handled by `do_discover()` just as it would when called from **nvme connect-all**, but it would be better to pass referrals back to the main process to make it aware of the additional discovery controller rather than using recursion. The main process would e.g. know if a discovery is already running for the transport address in the referrral.
 *  When "add" uevents for nvme controller devices are received, the controller is consistently not in `live` state yet, and attempting to read the `subsysnqn` sysfs attribute returns `(efault)`. While this should arguably be fixed in the kernel, it could be worked around in user space by using timers or polling the `state` sysfs attribute for changes.
 * Parse and handle `discovery.conf` on startup.
 * Implement support for RDMA and TCP protocols.


mwilck:
Issues with the CI...  Sorry about that. It compiled warning-free for me on several different systems.

I'm working on it.

mwilck:
CI issues fixed. Most of it was `#include <time.h>`, and installing `libudev-dev`.

keithbusch:
I like this. There's a bit going on here, so it might take a little longer than usual to review. i'll send any comments on the mailing list once i get some spare time.

martin-belanger:
It's a nice design and I like the idea, but I'm wondering if `nvme-monitor` shouldn't be a separate executable instead of being rolled into `nvme`. This extra feature goes beyond the basic function of `nvme`, which can lead to software bloat. UNIX/Linux cardinal rule is for programs to "[Do one thing, and do it well](https://en.wikipedia.org/wiki/Unix_philosophy#:~:text=12%20References-,Origin,by%20adding%20new%20%22features%22.)". `nvme` already does a lot. Adding too much into a single executable makes it harder to maintain.

On a somehow related note, @keithbusch started the [libnvme](https://github.com/linux-nvme/libnvme) project. This library separates the low-level APIs from the `nvme` executable and would allow for separate projects, like `nvme-monitor`, to be build as a stand-alone application. I don't know how far along the `libnvme` project is, but maybe we've reached the point where such a library is now indispensable to build new applications (i.e. daemons).

My 2 cents.

P.S. @keithbusch - Would it make sense for `libnvme` to be under the same repo as `nvme-cli`?

mwilck:
> It's a nice design and I like the idea, but I'm wondering if `nvme-monitor` shouldn't be a separate executable instead of being rolled into `nvme`. This extra feature goes beyond the basic function of `nvme`, which can lead to software bloat.

It was deliberately designed this way, because it re-uses so much of nvme-cli's code. (FTR, nvme-cli already has 76 subcommands, so I'm not sure if it's currently well-described by the "Do one thing, and do it well" mantra - "monitor" just adds one more).

If other maintainers will see it like you, it wouldn't be much work move the monitor into a separate binary (still in the nvme-cli repo). It's debatable if that would be less bloat. The resulting package would be larger ...

> On a somehow related note, @keithbusch started the [libnvme](https://github.com/linux-nvme/libnvme) project.

... unless we build both nvme-cli and the monitor on top of that library.

keithbusch:
I am totally fine with including monitor in this repo. It fits in nicely with the project's goal: manage nvme devices on Linux.

I am completely behind on getting libnvme uptodate, but integrating that with any of our tools in this repo is on me to complete; no need to gate progress on that effort.

mwilck:
CI fails because it can't find libudev for the powerpc cross-compilation environment. Figuring out how to fix that exceeds my Debian know-how. Help would be appreciated.

mwilck:
I'made the CI work by separating the amd64 and power8 builds. This way libudev is available in the amd64 environment, but not on power8, which means the monitor code is only built on the former.

It turns out that I caused the failure myself by pullling in `libudev-dev` (for amd64)  in the build environment. If I hadn't done that, the monitor would just not have been compiled and the test would have succeeded. But I suppose this way is better. It'd be even better to figure out how to pull in the power8 libraries, but I haven't yet.

mwilck:
Sorry, I pushed an un-clean git history. This one is better, just one patch (for github workflow) on top of the v2 submission.

mwilck:
I've added a 3rd workflow now, which provides a full-featured ppc64 build in a container, including libudev and a build of the monitor.

mwilck:
I've re-pushed this, because it needs a rebase, my testing found a bug in the event handling code.

This time I've used **git subtree** to pull in commit d6c705ccec64 from my [minivent](https://github.com/mwilck/minivent) repository directly (a7e168d replacing 595b4c7).

Let me know if this approach is ok for you guys.

mwilck:
Rebased, and used submodule instead of subtree for the event handling part.

This is mainly for cleanup. If anyone seriously considers merging this, I should go through another round of testing first. 

The CI doesn't seem to work with the submodule. Not sure what to do about that.

mwilck:
Fixed the CI with submodules now. Sorry for the noise.

hreinecke:
Well. The whole point of the nvme-cli rebasing on top of libnvme was to make this one work on top of _libnvme_.
So if there's a rebase it should be on top of libnvme; whether it should be part of nvme-cli afterwards is open to debate.
One could easily envision it as part of libnvme.

And on the other hand, most of the rationale for this has been superseded by the python bindings for libnvme; the current plan is to go for persistent discovery connections and rely on discovery AENs to create new controllers.
For initial discovery we would look for TPAR 8009, which is under 30-day review now and should be ratified by the end of the month.

So I guess some more discussion is to be had here; sorry @mwilck .

mwilck:
> So I guess some more discussion is to be had here; sorry @mwilck .

I'd given up on this one a while a go already, realizing that it received little love from the maintainers.
I have currently no intention to pursue this work further.

Issue #784, opened 2020-07-27
fails to compile on ppc64el


anonymouse64:
On ubuntu 20.04 ppc64el, using master @ https://github.com/linux-nvme/nvme-cli/commit/8d39b9ba2ba815ea32a3d034a2f6201d43b49522 I see the following errors compiling:

```
+ make -j4
NVME_VERSION = 1.12.79.g8d39b
    CC nvme-print.o
    CC nvme-ioctl.o
    CC nvme-lightnvm.o
    CC fabrics.o
    CC nvme-models.o
    CC plugin.o
    CC nvme-status.o
    CC nvme-filters.o
    CC nvme-topology.o
    CC plugins/intel/intel-nvme.o
    CC plugins/lnvm/lnvm-nvme.o
    CC plugins/memblaze/memblaze-nvme.o
    CC plugins/wdc/wdc-nvme.o
    CC plugins/wdc/wdc-utils.o
    CC plugins/huawei/huawei-nvme.o
    CC plugins/netapp/netapp-nvme.o
    CC plugins/toshiba/toshiba-nvme.o
    CC plugins/micron/micron-nvme.o
    CC plugins/seagate/seagate-nvme.o
    CC plugins/virtium/virtium-nvme.o
    CC plugins/shannon/shannon-nvme.o
    CC plugins/dera/dera-nvme.o
    CC plugins/scaleflux/sfx-nvme.o
plugins/scaleflux/sfx-nvme.c: In function ‘bd_table_show’:
plugins/scaleflux/sfx-nvme.c:533:17: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘__u64’ {aka ‘long unsigned int’} [-Werror=format=]
  533 |   printf(" 0x%llx", *(bb_elem++));
      |              ~~~^   ~~~~~~~~~~~~
      |                 |   |
      |                 |   __u64 {aka long unsigned int}
      |                 long long unsigned int
      |              %lx
plugins/scaleflux/sfx-nvme.c:541:17: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘__u64’ {aka ‘long unsigned int’} [-Werror=format=]
  541 |   printf(" 0x%llx",*(bb_elem++));
      |              ~~~^  ~~~~~~~~~~~~
      |                 |  |
      |                 |  __u64 {aka long unsigned int}
      |                 long long unsigned int
      |              %lx
[...]

Issue #772, opened 2020-07-07
Compiling failed on ppc64le


yizhanglinux:
Hi

I found this compiling issue on ppc64le, and cannot reproduced on x86_64.

```
# make
    CC plugins/scaleflux/sfx-nvme.o
plugins/scaleflux/sfx-nvme.c: In function ‘bd_table_show’:
plugins/scaleflux/sfx-nvme.c:533:17: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘__u64’ {aka ‘long unsigned int’} [-Werror=format=]
   printf(" 0x%llx", *(bb_elem++));
              ~~~^   ~~~~~~~~~~~~
              %lx
plugins/scaleflux/sfx-nvme.c:541:17: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘__u64’ {aka ‘long unsigned int’} [-Werror=format=]
   printf(" 0x%llx",*(bb_elem++));
              ~~~^  ~~~~~~~~~~~~
              %lx
plugins/scaleflux/sfx-nvme.c: In function ‘show_cap_info’:
plugins/scaleflux/sfx-nvme.c:600:49: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘__u64’ {aka ‘long unsigned int’} [-Werror=format=]
  printf("logic            capacity:%5lluGB(0x%llx)\n",
                                              ~~~^
                                              %lx
    IDEMA_CAP2GB(ctx->user_space), ctx->user_space);
                                   ~~~~~~~~~~~~~~~
plugins/scaleflux/sfx-nvme.c:602:49: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘__u64’ {aka ‘long unsigned int’} [-Werror=format=]
  printf("provisioned      capacity:%5lluGB(0x%llx)\n",
                                              ~~~^
                                              %lx
    IDEMA_CAP2GB(ctx->phy_space), ctx->phy_space);
                                  ~~~~~~~~~~~~~~  
plugins/scaleflux/sfx-nvme.c:604:49: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘__u64’ {aka ‘long unsigned int’} [-Werror=format=]
  printf("free provisioned capacity:%5lluGB(0x%llx)\n",
                                              ~~~^
                                              %lx
[...]

PR #881, opened 2021-02-02
Trying to get a ppc64le build going

ericcurtin:
Fix warnings along with that

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -14,4 +14,5 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: make
-      run: make
+      run: sudo apt-get install gcc-10-powerpc* && make clean && make && make clean && make LD=powerpc64le-linux-gnu-ld CC=powerpc64le-linux-gnu-gcc-10 CFLAGS='-O2 -g -Wall -Wformat-security -Werror -m64 -mcpu=power8 -mtune=power8 -I -I/usr/powerpc64-linux-gnu/include/'
+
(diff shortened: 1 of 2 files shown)


ericcurtin:
There's demand from a couple of places:

https://bugzilla.redhat.com/show_bug.cgi?id=1854495
#772
#784

ericcurtin:
This compiles but doesn't link, we need the ppc64el version of libuuid to link. Not 100% sure if ubuntu has a mirror for that. We could add a compile-only option to the makefile for ppc64el variant or do this properly and get libuuid ppc64el variant installed on CI server.

ericcurtin:
Removed libuuid dependency, we can cross-compile for ppc64le no problem now!

ericcurtin:
It solves a problem also, for example, ubuntu 20.04, don't build with libuuid and just sacrafice functionailty:

```
$ nvme gen-hostnqn
"gen-hostnqn" not supported. Install lib uuid and rebuild.
```

now for downstream packages this will always work as libuuid is now not required to generate a random uuid, theres a feature in the Linux kernel for that.

```
$ nvme gen-hostnqn
nqn.2014-08.org.nvmexpress:uuid:4df83088-40cf-434a-8b09-23ab9a2f9514
```

keithbusch:
If we can use libuuid, then let's use it. If you want to re-implement it for when it's not available, then fine, but let's not compile in the reinvention if we don't need to. It actually sounds odd that libuuid is not available on ppc, though. Is there a technical reason for that?

ericcurtin:
You are right LD=powerpc64le-linux-gnu-ld is all I needed.

Will push this and re-implement uuid generator from /proc file in another PR

ericcurtin:
Mind looking at this again @keithbusch ?

PR #884, opened 2021-02-02
If temperature is absolute zero -273 degrees

ericcurtin:
Print a dash in this case, this temperature reading violates Third Law of
Thermodynamics.


ericcurtin:
I gathered all the liquid nitrogen I could and I still struggled to test this 100%

keithbusch:
A cryogenic temp reported is likely a bug in the device's implementation. Maybe they forgot it's in Kelvin? But if it's '0', that probably means that the f/w isn't reporting a temperature at all, in which case I think -273 is the appropriate alarming output the user should see.

On the patch itself, there are some style issues that we need to try to consistently maintain. Please see https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces for the specific issues in this patch.

But again, I'm not so sure we want to take this as '-' doesn't quite stand out as much as '-273' when you're looking at a temperature field.

ericcurtin:
It's in the case of and emulated nvme device, sorry I should have explained that:

https://bugzilla.redhat.com/show_bug.cgi?id=1660631

`smartctl --all` displays it in this case as `-`

ericcurtin:
I think the styling is ok now.

birkelund:
> It's in the case of and emulated nvme device, sorry I should have explained that:
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=1660631
> 
> `smartctl --all` displays it in this case as `-`

Then this is a bug that needs to be fixed in the emulated device. Is it the nvme device in QEMU?

PR #886, opened 2021-02-03
nvme: add latest opcodes for command supported and effects log

ganaidu707:


keithbusch:
Just FYI, the ZNS opcodes are specific to the ZNS command set indicator, so the opcode-to-string may not necessarily work correctly if a future command set spec re-uses the same opcodes.

ganaidu707:
Thanks Keith, can you revert it?, I will exclude the ZNS commands specified, I thought of it and again added!

keithbusch:
could you just sent a new patch on top of it? it's just slightly easier than to revert and rebase a new patch.

ganaidu707:
Keith, sent a new PR, https://github.com/linux-nvme/nvme-cli/pull/887, thanks.

PR #888, opened 2021-02-04
nvme: fix formatNVM command

ganaidu707:
Remove unassigned variable "lbads" and change metadata
settings config parameter from "ms" to "mset" (as per
NVM Express)."ms" is reference for metadata size in general.

### DIFF
--- a/Documentation/nvme-format.1
+++ b/Documentation/nvme-format.1
@@ -1,13 +1,13 @@
 '\" t
 .\"     Title: nvme-format
-.\"    Author: [FIXME: author] [see http://www.docbook.org/tdg5/en/html/author]
-.\" Generator: DocBook XSL Stylesheets vsnapshot <http://docbook.sf.net/>
-.\"      Date: 10/20/2020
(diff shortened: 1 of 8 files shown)


minwooim:
I'm afraid that it might be late to update the parameter name of format command.

Format NVM commands are widely being taken to format namespaces in various systems already.  `--ms` might be confusing, of course, because there is Metadata Size(MS) field in LBA Format data structure, but I am not sure this is a good time to change this param name.

ganaidu707:
Thanks @minwooim for the comment, understand the concern.
Lets take Keith's opinion and decide on this.

ganaidu707:
closing this PR, as @minwooim mentioned its very late to change to the parameter.

MatiasBjorling:
The shorthand has historically been (MS) in the specification. Given that it is used by many scripts, it would cause a lot of grief to update to the MSET shorthand in the later revisions.

keithbusch:
Any change in parameter names at this point will need to work out how to create alias's to the previous name.

PR #892, opened 2021-02-04
nvme: add uuid index field in get and set features

ganaidu707:


ganaidu707:
Rebased it over the latest commits.

PR #893, opened 2021-02-04
Generate random uuid from /proc/sys/kernel/random/uuid file

ericcurtin:
For people who don't link in libuuid, so they don't lose functionality


ericcurtin:
Ubuntu 20,04 an example of a distro that doesn't link libuuid

PR #897, opened 2021-02-06
Lf to f

ericcurtin:
When printing long double, zero is shown as nan or not a number after a
patch in glibc to fix CVE-2020-29573, printing doubles does not suffer
from this problem although you have lowered precision.


ericcurtin:
Another option is to store as a long double but print as a double? If we'd like to keep the stored precision

PR #899, opened 2021-02-08
nvme: remove unused functions nvme_{read, write, compare}

ganaidu707:
Functions nvme_write, nvme_read, nvme_compare are
not being called in place. Removing these unused
functions.


minwooim:
Looks good to me.  Nice cleanup!

PR #902, opened 2021-02-10
nvme-ioctl: remove unused function nvme_passthru_io

minwooim:
This function has not been called from anywhere, so remove it.


hreinecke:
With the move to libnvme this pull request doesn't make sense anymore; please check with libnvme if it's still required.

PR #908, opened 2021-02-12
nvme-print: align returns for better readability

ganaidu707:

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -48,29 +48,29 @@ static const char *nvme_cmd_to_string(int admin, __u8 opcode)
 {
 	if (admin) {
 		switch (opcode) {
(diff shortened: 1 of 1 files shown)


keithbusch:
What editor settings are you using where these don't align? Applying this patch makes it not align for me, and I'm using the same settings I use for kernel devel.

ganaidu707:
Oh sorry, I am using VS Code editor and tab size changed to 4 chars, didn't noticed that, so, changed to 8 it is fine, closing this PRs.

ganaidu707:
this https://github.com/linux-nvme/nvme-cli/pull/910 PR is fine? similar to this one,` switch-case` coding style fix?

PR #911, opened 2021-02-14
nvme-print: fix fetching status code type and status data type

ganaidu707:


keithbusch:
The status is only 15 bits, though, so why make it a u32 parameter? The current mask is wrong, though, so thanks for fixing that part.

keithbusch:
Oh sorry, I read the diff backwards! You changed it to u16, not to u32...

PR #915, opened 2021-02-16
Adding "Ignore Existing Key" parameter to rescap, and some minor re-ordering

AffeHMT:
Hello,

Here's couple of small details that I fixed while working with this tool. There was some other mods that I made also, but it seems that those were already handled in the most recent version.

BR, Affe


keithbusch:
No need to print out empty "reserved" lines, otherwise looks fine. Please rebase without the print, and I can merge it in.

AffeHMT:
Empty printouts removed.

PR #916, opened 2021-02-16
nvme-discover: add json output

hreinecke:
Add a '-j'/'--json' option to generate JSON output.


keithbusch:
Can we use the same format that we have for all other json emitting commands, which is "--output-format json" or "-o json"?

hreinecke:
> Can we use the same format that we have for all other json emitting commands, which is "--output-format json" or "-o json"?

Sure. Will be reposting the patch.

PR #917, opened 2021-02-16
nvme-print: remove unused argument passing to json output format print

ganaidu707:

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -127,7 +127,7 @@ static const char *get_sanitize_log_sstat_status_str(__u16 status)
 	}
 }
 
-static void json_nvme_id_ns(struct nvme_id_ns *ns, unsigned int mode)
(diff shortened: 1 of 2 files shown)


nvme-print.c
```diff
@@ -208,8 +208,8 @@ static void json_nvme_id_ns(struct nvme_id_ns *ns, unsigned int mode)
 	json_free_object(root);
 }
 
-static void json_nvme_id_ctrl(struct nvme_id_ctrl *ctrl, unsigned int mode,
-			void (*vs)(__u8 *vs, struct json_object *root))
+static void json_nvme_id_ctrl(struct nvme_id_ctrl *ctrl, void (*vs)(__u8 *vs,
```
keithbusch, line 211: This is not a very good way to break up this line. Otherwise, looks fine.

ganaidu707: modified that line and pushed again!

PR #918, opened 2021-02-17
Fixup loop discovery

hreinecke:
Commit 68bc86922296c817d385fd21d02862148b9bbb9d introduced a regression for loop devices, as 'ctrl_loss_tmo' is invalid for loop devices and with that commit discovery on loop devices will fail.


hreinecke:
Sigh. I only wanted to push the loop discovery fix, not the json one.

PR #931, opened 2021-02-23
nvme: add support for lba status log page

ganaidu707:
This log page is used to provide information about
subsequent actions the host may take to discover which
logical blocks, in namespaces that are attached to
the controller, may no longer be recoverable when read.
For more details see NVM Express 1.4 Spec. Section
5.14.1.14("LBA Status Information (Log Identifier 0Eh)")

Co-Authored-By: Karthik Balan karthik.b82@samsung.com

### DIFF
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -713,6 +713,29 @@ struct nvme_effects_log {
 	__u8   resv[2048];
 };
 
+struct nvme_lba_range_desc {
+	__le64	rslba;
+	__le32	rnlb;
+	__u8	rsvd12[4];
+};
+
+struct nvme_lba_status_ns_element {
+	__le32	neid;
+	__le32	nlrd;
+	__u8	ratype;
+	__u8	rsvd9[7];
+	struct nvme_lba_range_desc descs[];
+};
+
+struct nvme_lba_status_log_page {
+	__le32	lslplen;
+	__le32	nlslne;
+	__le32	estulb;
+	__u8	rsvd12[2];
+	__le16	lsgc;
+	struct nvme_lba_status_ns_element elements[];
+};
+
 enum nvme_ana_state {
 	NVME_ANA_OPTIMIZED		= 0x01,
 	NVME_ANA_NONOPTIMIZED		= 0x02,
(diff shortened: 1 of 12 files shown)


linux/nvme.h
```diff
@@ -713,6 +713,29 @@ struct nvme_effects_log {
...
+	struct nvme_lba_range_desc desc[];
+};
+
+struct nvme_lba_status_log_page {
+	__le32	lslplen;
+	__le32	nlslne;
+	__le32	estulb;
+	__u8	rsvd12[2];
+	__le16	lsgc;
+	struct nvme_lba_status_ns_element element;
```
keithbusch, line 736: This is not a single element. It is a variable length array of size nlslne.

ganaidu707: This has been addressed and rebased it!

nvme-print.c
```diff
@@ -1614,6 +1614,103 @@ void nvme_show_endurance_group_event_agg_log(
...
+		le32_to_cpu(lba_status->lslplen));
+        num_elements = le32_to_cpu(lba_status->nlslne);
+	json_object_add_value_uint(root, "nlslne",
+		num_elements);
+	json_object_add_value_uint(root, "estulb",
+		le32_to_cpu(lba_status->estulb));
+	json_object_add_value_uint(root, "lsgc",
+		le16_to_cpu(lba_status->lsgc));
+
+        elements_list = json_create_array();
```
keithbusch, line 1637: Tabs, not spaces please.

nvme-print.c
```diff
@@ -1614,6 +1614,103 @@ void nvme_show_endurance_group_event_agg_log(
...
+        for (int ele = 0; ele < num_elements; ele++) {
+                element = json_create_object();
+                json_object_add_value_uint(element, "neid",
+                        le32_to_cpu(lba_status->elements[ele].neid));
```
keithbusch, line 1641: I'm terribly confused how this even compiles. We have a variable length array who's elements are variable in size. How could the compiler possibly know the address of the element index?

birkelund: You need -Wpedantic to complain about this. I'm not sure if it is undefined behavior.

keithbusch: The behavior assumes the elements are fixed length `sizeof(lba_status->elements[0])`, so accessing anything except for the first index is going to yield the wrong address.

ganaidu707: I have addressed all the review comments and raised it as separate PR https://github.com/linux-nvme/nvme-cli/pull/933
will close this PR.
Thanks!

PR #933, opened 2021-02-23
nvme: add support for lba status log page

ganaidu707:
This log page is used to provide information about
subsequent actions the host may take to discover which
logical blocks, in namespaces that are attached to
the controller, may no longer be recoverable when read.
For more details see NVM Express 1.4 Spec. Section
5.14.1.14("LBA Status Information (Log Identifier 0Eh)")

Co-Authored-By: Karthik Balan <karthik.b82@samsung.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1097,6 +1097,73 @@ ret:
 	return nvme_status_to_errno(err, false);
 }
 
+static int get_lba_status_log(int argc, char **argv,
+		struct command *cmd, struct plugin *plugin)
+{
+	const char *desc = "Retrieve Get LBA Status Info Log " \
+			"and prints it, for the given device in either " \
+			"decoded format(default),json or binary.";
+	const char *rae = "Retain an Asynchronous Event";
+	void *lab_status;
+	enum nvme_print_flags flags;
+	int err, fd;
+	__u32 lslplen;
+
+	struct config {
+		bool rae;
+		char *output_format;
+	};
+
+	struct config cfg = {
+		.rae = false,
+		.output_format = "normal",
+	};
+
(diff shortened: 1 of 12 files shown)


nvme.c
```diff
@@ -1097,6 +1097,74 @@ static int get_endurance_event_agg_log(int argc, char **argv,
...
+
+	err = fd = parse_and_open(argc, argv, desc, opts);
+	if (fd < 0)
+		goto ret;
+
+	err = flags = validate_output_format(cfg.output_format);
+	if (flags < 0)
+		goto close_fd;
+
+	err = nvme_lba_status_log(fd, &lslplen, cfg.rae, sizeof(__u32));
```
keithbusch, line 1136: You should use 'true' for RAE when just getting the length of the log. The user's RAE selection should only apply to the final log request.

nvme.c
```diff
@@ -1097,6 +1097,74 @@ static int get_endurance_event_agg_log(int argc, char **argv,
...
+		perror("lba status log page");
+		goto close_fd;
+	} else if (err) {
+		nvme_show_status(err);
+		goto close_fd;
+	}
+
+	lab_status = calloc(lslplen, 1);
+	if (!lab_status) {
+		fprintf(stderr, "could not alloc buffer for lba status " \
```
keithbusch, line 1147: Please use `perror` insetead of fprintf for notifying system errors that already set errno.

nvme.c
ganaidu707: yeah, correct, taken care.

nvme.c
ganaidu707: changed,

PR #935, opened 2021-02-24
Add "json" option to discover command

martin-belanger:
The `discover` command now has a `--json` option. When the option is enabled, the output will be printed in JSON format. This makes it easier for other programs (like a Python script) to parse the output.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -88,6 +88,7 @@ static struct config {
 	bool persistent;
 	bool quiet;
 	bool matching_only;
+	bool json;
 } cfg = { .ctrl_loss_tmo = -1 };
 
 struct connect_args {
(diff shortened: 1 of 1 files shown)


keithbusch:
There's a json library included in the source code to handle pretty-printing these. Could you use that instead of manually formatting it?

martin-belanger:
Ok. I will see if I can use it.

martin-belanger:
Hi @keithbusch 
I made the changes you recommended.

keithbusch:
Looks like Hannes sent a pull request for the same feature, https://github.com/linux-nvme/nvme-cli/pull/937/commits/d22d5bdd9288fac06ee6fca9dec5654b5bd466ff. I'll pull that one since it is a bit more complete with the manpage updates and other minor details.

PR #942, opened 2021-02-26
nvme-discover: lookup existing persistent controllers

hreinecke:
If persistent controller connections are present they should be
preferred even if no --device option is given on the commandline.
To avoid selecting a temporary non-persistent controller the
'kato' attribute is checked; any controller which does not have
this attribute or for which the attribute is '0' will be skipped.

I have submitted a patch for adding the 'kato' sysfs controller attribute to linux-nvme.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -320,6 +320,27 @@ static bool ctrl_matches_connectargs(char *name, struct connect_args *args)
 	cargs.trsvcid = parse_conn_arg(addr, ' ', conarg_trsvcid);
 	cargs.host_traddr = parse_conn_arg(addr, ' ', conarg_host_traddr);
 
+	if (!strcmp(cargs.subsysnqn, NVME_DISC_SUBSYS_NAME)) {
+		char *kato_str = nvme_get_ctrl_attr(path, "kato"), *p;
+		unsigned int kato = 0;
+
+		/*
+		 * When looking up discovery controllers we have to skip
+		 * any non-persistent controllers (ie those with a zero
+		 * kato value). Otherwise the controller will vanish from
+		 * underneath us as they are owned by another program.
+		 *
+		 * The 'kato' attribute might not be present; assume a
+		 * non-persistent controller for these installations.
+		 */
+		if (kato_str) {
+			kato = strtoul(kato_str, &p, 0);
+			if (p == kato_str)
+				kato = 0;
+		}
+		if (kato == 0)
+			return found;
+	}
 	if (!strcmp(cargs.subsysnqn, args->subsysnqn) &&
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -320,6 +320,27 @@ static bool ctrl_matches_connectargs(char *name, struct connect_args *args)
 	cargs.trsvcid = parse_conn_arg(addr, ' ', conarg_trsvcid);
 	cargs.host_traddr = parse_conn_arg(addr, ' ', conarg_host_traddr);
 
+	if (!strcmp(cargs.subsysnqn, NVME_DISC_SUBSYS_NAME)) {
```
mwilck, line 323: I suggest moving this test further down after the `!strcmp(cargs.subsysnqn, args->subsysnqn)`. We don't need to test this if we aren't looking for a discovery controller.

fabrics.c
```diff
@@ -320,6 +320,27 @@ static bool ctrl_matches_connectargs(char *name, struct connect_args *args)
...
+
+		/*
+		 * When looking up discovery controllers we have to skip
+		 * any non-persistent controllers (ie those with a zero
+		 * kato value). Otherwise the controller will vanish from
+		 * underneath us as they are owned by another program.
+		 *
+		 * The 'kato' attribute might not be present; assume a
+		 * non-persistent controller for these installations.
+		 */
```
mwilck, line 335: If the device string was set by the user on the command line and there is no `kato` attribute, we should assume that the controller is usable (i.e. will persist at least until we try to retrieve the discovery page). Otherwise you basically disable the `--device` options on kernels that don't support `kato`.

So, this logic only makes sense to me if this function is called from `find_ctrl_with_connectargs()`. But OTOH, why not generally assume that the controller is persistent if there is no `kato` attribute? The worst thing that can happen is an error attempting to connect to it, and this error will be transient by definition.

mwilck:
There's another issue with this patch: referrals. Checking a given `cfg.device` against the connect args makes sense if `do_discover()` is called recursively from `connect_ctrl()`. If we don't do this, the discovery is almost guaranteed to fail.

I'm going to send you a PR to your branch.

mwilck:
I wanted to send a PR to your branch, but github won't let me do that. So please just have a look at the two topmost commit on https://github.com/mwilck/nvme-cli/tree/persistent-discovery

mwilck:
#992 should supersede this PR. @hreinecke, please double-check.

hreinecke:
> #992 should supersede this PR. @hreinecke, please double-check.

Yes, it does. And if not I'll rebase it on top of the (now-merged) #992.

hreinecke:
Superseded by #992

PR #944, opened 2021-02-26
nvme-print: split pmrmsc into pmrmscl and pmrmscu

ganaidu707:
Split the PMR Memory Space Control register 64 bits into
PMR Memory Space Control Lower and Upper 32 bits each as
per NVM Express Spec. 1.4b changes.


keithbusch:
Fortunately we always split this to a lo-hi read, so I don't think the misalignment in the spec caused us trouble here.

PR #945, opened 2021-02-26
nvme-print: fix register size for raw format print

ganaidu707:


ganaidu707:
@keithbusch  I have rebased it over the latest commits and now CI build passed all the checks.

PR #957, opened 2021-03-03
nvme: refactor show nvme controller bar registers

ganaidu707:
NVMe bar registers print currently have redundant code, having
same register fields declartions and defnitions for both noraml
and json format seperately, to avoid this used a nvme_bar_reg
structure.


ganaidu707:
Normal Format:
reg_normal

Json Format:
reg_json

[reg_human.txt](https://github.com/linux-nvme/nvme-cli/files/6076237/reg_human.txt)

ganaidu707:
@keithbusch would you mind look at this patch again?

PR #962, opened 2021-03-06
Making security-send command more flexible

### LINKED ISSUE #951: Allow stdin for security-send, and zero-padded input.
Some commands, such as admin-passthru, default to use of stdin if no filename is specified. This is not the case for security-send, however. I think this would be a good consistency improvement, and allow chaining of data through pipes from the command line.

It would also be good, like admin-passthru, to automatically zero-pad the data being sent if transfer length is greater than the size of the input file (or stdin).

fixthething:
by allowing data from stdin, and assuming transfer length to match input size if not specified.

resolve #951

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -3681,7 +3681,7 @@ static int sec_send(int argc, char **argv, struct command *cmd, struct plugin *p
 		goto close_sec_fd;
 	}
 
-	memset(sec_buf, 0, cfg.tl); // ensure zero fill if buf_size > sec_size
+	memset(sec_buf, 0, cfg.tl); // ensure zero fill if cfg.tl > sec_size
 
 	err = read(sec_fd, sec_buf, sec_size);
 	if (err < 0) {
@@ -3692,7 +3692,7 @@ static int sec_send(int argc, char **argv, struct command *cmd, struct plugin *p
 	}
 
 	err = nvme_sec_send(fd, cfg.namespace_id, cfg.nssf, cfg.spsp, cfg.secp,
-			cfg.tl, cfg.tl, sec_buf);
+			cfg.tl, sec_buf);
 	if (err < 0)
 		perror("security-send");
 	else if (err != 0)
--- a/nvme.c
+++ b/nvme.c
@@ -3614,7 +3614,8 @@ static int sec_send(int argc, char **argv, struct command *cmd, struct plugin *p
 	const char *nssf = "NVMe Security Specific Field";
 	int err, fd, sec_fd = -1;
 	void *sec_buf;
-	unsigned int sec_size;
+	unsigned int sec_size, buf_size;
+	int nread = 0;
 
 	struct config {
 		__u32 namespace_id;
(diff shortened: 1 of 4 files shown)


fixthething:
Changes allow easier/more flexible command line usage with piping data to stdin, and fixes some oddities regarding transfer buffer size when not matching the file size. Examples:
1) echo -n "hello" | nvme security-send -p 0xff --spsp 0x1234 /dev/nvme0
This will send 8 bytes in 2 dwords ("hello\0\0\0") auto-sized based on the waiting stdin buffer without need to explicitly set the transfer length
2) echo -n "hello" | nvme security-send -p 0xff --spsp 0x1234 /dev/nvme0 -t 0x200
This will send "hello" as before, but will zero-pad to 0x200 bytes
3) nvme security-send -p 0xff --spsp 0x1234 -f giant_file.bin -t 0x200 /dev/nvme0
This will send the first 0x200 bytes of the giant_file.bin, as before. In the original code, this would actually allocate the entire size of giant_file and send it to the ioctl even though the drive would only accept the first 0x200 because of the transfer length. I'm not sure of the consequences but that didn't seem right.
4) nvme security-send -p 0xff --spsp 0x1234 -f undersizefile.bin -t  0x4000 /dev/nvme0
I think I actually fixed a bug in this case assuming undersizefile.bin is less than the transfer length. Before, it would have allocated a buffer only as big as the actual file, yet told the drive to read past the end of the buffer provided. This would be a DMA buffer overflow? Not sure what would actually happen but again it seemed wrong.

keithbusch:
Does this really need to be dword aligned? I'm not finding that in spec, but it does reference other specs that I don't have, so I'm not sure.

fixthething:
> 
> 
> Does this really need to be dword aligned? I'm not finding that in spec, but it does reference other specs that I don't have, so I'm not sure.

The spec doesn't mention it directly. However:
1) NVMe spec 1.4b, section 4.3 on PRP entries: bits 1:0 of all PRP entries are hardcoded to 0 and must be dword aligned. Fair to say this only applies to the start of the DMA and not the end, but it's an implication that DMA hardware only needs dword granularity
2) NVMe spec 1.4b, section 4.4 on SGL lists: SGL lists have an optional specifier that can require dword alignment and bits 1:0 clear for both address and length
3) Pretty much every other command in the spec specifies number of dwords in cdw10. My guess is this one only uses bytes to be more compatible with the SCSI SECURITY PROTOCOL OUT command it's derived from.

keithbusch:
I suppose so.

We should just use the command's length parameter rather than trying to figure it out, and zero pad data short of the 'tl' parameter.

fixthething:
> 
> 
> I suppose so.
> 
> We should just use the command's length parameter rather than trying to figure it out, and zero pad data short of the 'tl' parameter.

Yes, I believe that's how it should behave if the length parameter is specified. The intent was that if specified, it would overrule any inferred size and truncate or zero-fill as necessary.

In defense of inferred size when 'tl' is not specified: whether using an input pipe or a file is specified, the size should be known, and forcing the user to enter the transfer size in those cases is most likely just an opportunity to enter the wrong number (would have to look up the file size with 'ls', or count the characters in their input pipe). The previous behavior gave no warning if 'tl' was not specified, and assumed 0 bytes, which was almost certainly not the intent of anybody sending the command.

fixthething:
> 
> 
> I suppose so.
> 
> We should just use the command's length parameter rather than trying to figure it out, and zero pad data short of the 'tl' parameter.

Re-reading your comment on zero padding, was this a suggestion to zero-pad only the data after sec_size? e.g. instead of memset of entire buffer, change to:
       if (sec_size < buf_size)
                memset(sec_buf + sec_size, 0, buf_size - sec_size);

Happy to make that change if that's what you meant.

keithbusch:
What I meant is that we enforce the 'tl' parameter be used, and we just allocate the buffer to that. If the input is smaller, then the rest is 0'ed. If the input is larger, then it gets truncated. The fact that the buffer size and tl field can be different looks like a bug in this implementation since the controller only knows of one of those values.

fixthething:
While I like the easier command line by omitting 'tl' when the size of the input is known, and the patch already respected cfg.tl over any assumed sizes if specified, I'll make tl required if you think the sending the input file size by default isn't a reasonable assumption. Please confirm if that's so, and what error message do you want displayed, and what return value, if not specified?

> The fact that the buffer size and tl field can be different looks like a bug in this implementation since the controller only knows of one of those values
Yes I saw that this bug existed in the current code (mentioned in use case number 4 from an earlier comment); where cfg.tl and sec_size (which could be totally unrelated values) were passed into nvme_sec_send. My patch fixes this by creating a buffer the same size as the transfer to be actually made, and passes buf_size for both parameters to nvme_sec_send. That buf_size should always be cfg.tl, if specified (plus dword alignment). I was trying to keep my diff small by only modifying nvme.c, but to solve the root of the issue, you might want to remove one of those parameters to nvme_sec_send, or else error in nvme_sec_send if the transfer length is greater than the buffer size which could cause the drive DMA engine to read past the end of the allocated buffer (probably a bad idea).

fixthething:
@keithbusch I added a third commit to my fork that I think does what you want with cfg.tl required.

I added a warning if the user chooses an unaligned size rather than adding it implicitly, since you didn't seem sure about it. I'm pretty sure at least some drives and configurations truncate unaligned data, but it's possible that some of them choose to round up instead of rounding down. If the drive rounds up, though, it will read past the end of the buffer we allocated/zeroed. Thus unaligned data size MIGHT work but it might also do something unintended, so I think a warning is warranted at the very least.

nvme.c
```diff
@@ -3692,7 +3692,7 @@ static int sec_send(int argc, char **argv, struct command *cmd, struct plugin *p
 	}
 
 	err = nvme_sec_send(fd, cfg.namespace_id, cfg.nssf, cfg.spsp, cfg.secp,
-			buf_size, buf_size, sec_buf);
+			cfg.tl, cfg.tl, sec_buf);
```
keithbusch, line 3695: At this point, every caller to this function now has the same value for the 6 and 7 parameters, they should be collapsed into the same one.

fixthething: > At this point, every caller to this function now has the same value for the 6 and 7 parameters, they should be collapsed into the same one.

I added a commit that removed the redundant parameter to nvme_sec_send. However, it raises another issue: the same issue applies to nvme_sec_recv, where "al" and "size" are redundant parameters passed from the user command line. I didn't address that since one of the parameters that affects the command-line user interface needs to be removed, and I thought the maintainers might have some preference about which one to remove/keep.

fixthething:
Is there anything remaining to be done on this request? I think I have addressed all the comments raised, but there has been no activity in a month. This is my first pull request so I'm still learning the process. Is there a merge window or approval process or something?

keithbusch:
manually merged.

PR #964, opened 2021-03-07
nvme-print: improve command support and effects log json format

ganaidu707:
Currently command support and effects log page on json format
printing all the opcodes irrespective of the command supported.
Most part of the log page is "Unkown" opcodes, fixed that. Also
added two json objects "acs" and "iocs" to print them seperately.


ganaidu707:
effects_log

PR #969, opened 2021-03-14
nvme: address 1.4 to 1.4b changes for Change NS event type

ganaidu707:
In NVMe 1.4 spec. Change Namespace Event Type (06h) of persistent
event log Data Format - NCAP field was assigned with 16 bytes. Now
in 1.4b spec. its modified to 8 bytes.


keithbusch:
Yes, looks like an accident when this was originally defined, so there shouldn't be any backward compatible issues. For the record, the spec fix is documented in the Technical Workgroup's NVMe 1.4 ECN 001 2020.03.09.

PR #974, opened 2021-03-18
Add nvme device to json output

martin-belanger:
Add the nvme device to the json output of the "nvme discover" command. Here's an example:
```
# nvme discover -o json -g -G -t tcp -s 8009 -a 127.0.0.1
{
  "device" : "nvme2",      ######## THIS IS THE NEW ITEM ########
  "genctr" : 6,
  "records" : [
    {
      "trtype" : "tcp",
      "adrfam" : "ipv4",
      "subtype" : "nvme subsystem",
      "treq" : "not specified, sq flow control disable supported",
      "portid" : 1,
      "trsvcid" : "8009",
      "subnqn" : "nqn.1988-11.com.dell:PowerSANxxx:01:20210225100113-454f73093ceb4847a7bdfc6e34ae8e28",
      "traddr" : "127.0.0.1",
      "sectype" : "none"
    }
  ]
}
```

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -57,6 +57,13 @@
 /* default to 600 seconds of reconnect attempts before giving up */
 #define NVMF_DEF_CTRL_LOSS_TMO		600
 
+#define STRLEN(x) (sizeof(""x"") - 1)
+#define DECIMAL_STR_MAX(type)                                           \
+        (2+(sizeof(type) <= 1 ? 3 :                                     \
+            sizeof(type) <= 2 ? 5 :                                     \
+            sizeof(type) <= 4 ? 10 :                                    \
+            sizeof(type) <= 8 ? 20 : sizeof(int[-2*(sizeof(type) > 8)])))
+
 const char *conarg_nqn = "nqn";
 const char *conarg_transport = "transport";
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -1379,12 +1397,8 @@ static int do_discover(char *argstr, bool connect, enum nvme_print_flags flags)
 	if (instance < 0)
 		return instance;
 
-	if (asprintf(&dev_name, "/dev/nvme%d", instance) < 0)
```
keithbusch, line 1382: Something wrong with asprintf()?

martin-belanger: Nothing wrong with asprintf(), but since we're printing string that has a fixed and known max length (i.e. we're printing an int with %d, and MAX_INT=2147483647, which requires 11 char + 1 nul), we can simply allocate a buffer from the stack instead of the heap. This also eliminate the need to free() the memory. I understand that nvme-cli typically runs quickly to completion so we need not worry about memory leaks. However, there are cases where we can daemonize nvme-cli (e.g. nvme-monitor). I think it's just good practice to avoid malloc'ating memory when all we need is a small temporary buffer.

PR #979, opened 2021-03-24
Read system UUID from DMI and merge hostnqn generation functions

hreinecke:
Both fabrics.c and the 'gen-hostnqn' command contain functions
to generate a default host NQN. Needless to say, both are different.
_And_ we have the script 'nvme-gen-hostnqn' to read the host nqn
from the system UUID, implement yet another way.

This patch merges all of these functions, and implements a hierarchy
in generating host NQNs:
- Try reading system UUID from /sys/firmware/dmi/entries
- Try reading systemd-generated UUID
- Generate a random UUID

### DIFF
--- a/nvme-topology.c
+++ b/nvme-topology.c
@@ -696,3 +696,76 @@ void *mmap_registers(const char *dev)
 	return membase;
 }
 
+#define PATH_DMI_ENTRIES	"/sys/firmware/dmi/entries"
+
+int uuid_from_dmi(char *system_uuid)
+{
+	int f;
+	DIR *d;
+	struct dirent *de;
+	char buf[512];
+
+	system_uuid[0] = '\0';
+	d = opendir(PATH_DMI_ENTRIES);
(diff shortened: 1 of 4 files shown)


nvme-topology.c
```diff
@@ -696,3 +696,76 @@ void *mmap_registers(const char *dev)
...
+			continue;
+		sprintf(filename, "%s/%s/type", PATH_DMI_ENTRIES, de->d_name);
+		f = open(filename, O_RDONLY);
+		if (f < 0)
+			continue;
+		len = read(f, buf, 512);
+		close(f);
+		if (len < 0)
+			continue;
+		if (sscanf(buf, "%d", &type) != 1)
```
keithbusch, line 726: I think we can avoid the need for this check based on how the kernel names the sysfs entries prefixed with the type. For example, we can filter the returned directories with 'scandir' instead of 'readdir' to return entries that match "1-*".

PR #980, opened 2021-03-24
Modify lat-stats cmd to cope with optane bucket format

franciscomunoz:
Optane lat-stat thresholds are 8 bytes per bucket. Also
the amout of buckets is smaller when compared to nand ssds.
This change allows to query both types of ssds transparently.

Also added the command line option to print lat-stats in json

Signed-off-by: Francisco Munoz <francisco.munoz.ruiz@intel.com>

### DIFF
--- a/plugins/intel/intel-nvme.c
+++ b/plugins/intel/intel-nvme.c
@@ -492,6 +492,28 @@ struct intel_lat_stats {
 	__u32 data[1216];
 };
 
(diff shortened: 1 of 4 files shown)


keithbusch:
Patch looks fine, but just need to fix the printf formats.

franciscomunoz:
@keithbusch please let me know if somethign else is needed.

PR #981, opened 2021-03-25
nvme-print: print phase tag field separately in error-log fields

ganaidu707:
Status Field in error-log 16 bits, 15 bits from q to 15 are represents
status field and 0th bit is phase tag, print phase tag seperately, in
1.4b spec its been clearly called out.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -370,7 +370,9 @@ static void json_error_log(struct nvme_error_log_page *err_log, int entries)
 		json_object_add_value_int(error, "cmdid",
 			le16_to_cpu(err_log[i].cmdid));
 		json_object_add_value_int(error, "status_field",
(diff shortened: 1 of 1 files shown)


keithbusch:
I don't know, altering the json output like this can break something that's parsing it.

ganaidu707:
Hi @keithbusch , I have tested this, it doesn't break json parsing.

error_log

keithbusch:
I'm saying the json output is often piped to another program. If that program is looking for existing keys, changing those keys in our output may cause a problem for them. I suppose this change is probably safe, though

ganaidu707:
@keithbusch is this patch could go in or due to compatibility issues for existing users (maybe, not sure), we will discard it for now? :)

keithbusch:
I guess we'll find out if anyone was using this particular output for anything useful.

PR #983, opened 2021-03-25
nvme: nvme get and set features buffer length clean up

ganaidu707:


keithbusch:
They all look fine, but I'm confused why where are multiple commits doing the same thing here.

ganaidu707:
yeah, I didn't notice this, I will close this and raise fresh PR.

PR #988, opened 2021-03-30
Identify Primary Controller Capabilities support

NateRoiger:
Adds support for Identify Primary Controller Capabilities command (CNS 14h)

For reference, the data structure is defined in 5.15.2.10 Primary Controller Capabilities data structure (CNS 14h) in the 1.4a specification. 

Signed-off-by: Nate Roiger <nate.roiger@hpe.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2413,6 +2413,60 @@ ret:
 	return nvme_status_to_errno(err, false);
 }
 
+static int primary_ctrl_caps(int argc, char **argv, struct command *cmd, struct plugin *plugin)
+{
+	const char *desc = "Send an Identify Primary Controller Capabilities "\
+		"command to the given device and report the information in a "\
+		"human-redable or binary format.";
+	const char *raw = "show capabilities in binary format";
+	const char *human_readable = "show capabilities in readable format";
+
+	enum nvme_print_flags flags = NORMAL;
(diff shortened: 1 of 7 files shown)


nvme.c
```diff
@@ -2413,6 +2413,60 @@ static int virtual_mgmt(int argc, char **argv, struct command *cmd, struct plugi
...
+		char *output_format;
+	};
+
+	struct config cfg = {
+		.output_format = "normal",
+	};
+
+	OPT_ARGS(opts) = {
+		OPT_FMT("output-format",   'o', &cfg.output_format,  output_format),
+		OPT_FLAG("raw-binary",     'b', &cfg.raw_binary,     raw),
```
keithbusch, line 2440: This is redundant with the output-format option. Older commands have the option only for backward compatibility.

nvme.c
```diff
@@ -2413,6 +2413,60 @@ static int virtual_mgmt(int argc, char **argv, struct command *cmd, struct plugi
...
+		int human_readable;
+		char *output_format;
+	};
+
+	struct config cfg = {
+		.output_format = "normal",
+	};
+
+	OPT_ARGS(opts) = {
+		OPT_FMT("output-format",   'o', &cfg.output_format,  output_format),
```
keithbusch, line 2439: There should be a json output to correspond with this option.

PR #990, opened 2021-03-31
ymtc plugin version 0.0.1 - get additional smart support

ymtc-jerry:
hi all,

    this is Jerry Liu from Yangtze Memory Technologies Co., Ltd, I'm very glad to add plugin for nvme-cli tool to support our SSDs, could you help review my code? if any questions, feel free to contact me.

thank you very much!

### DIFF
--- a/plugins/ymtc/ymtc-nvme.c
+++ b/plugins/ymtc/ymtc-nvme.c
@@ -1,0 +1,148 @@
+#include <fcntl.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "linux/nvme_ioctl.h"
+
+#include "nvme.h"
+#include "nvme-print.h"
+#include "nvme-ioctl.h"
+#include "plugin.h"
+
+#include "argconfig.h"
+#include "suffix.h"
+
+#define CREATE_CMD
+#include "ymtc-nvme.h"
+#include "ymtc-utils.h"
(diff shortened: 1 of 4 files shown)


plugins/ymtc/ymtc-nvme.c
```diff
@@ -0,0 +1,192 @@
...
+#include "plugin.h"
+
+#include "argconfig.h"
+#include "suffix.h"
+
+#define CREATE_CMD
+#include "ymtc-nvme.h"
+#include "ymtc-utils.h"
+
+#define STRN2_01    "Additional Smart Log for NVME device"
```
keithbusch, line 21: I am confused by this formatting decision. Why not just put these string constants where you're printing them?

ymtc-jerry: hi @keithbusch 

thank you for reviewing my code, mainly for 2 reasons:

1. easy to management, e.g. STRN2_01 and STRN2_02 belong to one printf.
2. every printf looks similiar and tidy, one printf can be viewed in one line.

of course, it also has some drawback which looks counterintuitive.
I can change this format if you still feel confused.

thanks.

keithbusch: Except it's neither easier nor tidy. It's absolutely obfuscated through non-obvious symbols what is going to print.

keithbusch:
Thanks, manually applied.

PR #991, opened 2021-03-31
add fuse field for write and compare commands

ganaidu707:
This PR does the necessary changes to have the `CDW0 bits 8:15 which represented as flags` and adds `fuse` field  in `submit_io`
as a user cli input.


keithbusch:
Why? The kernel doesn't allow it.

keithbusch:
Please take some time to test new features. :)

ganaidu707:
yeah sure, missed to test this, in general before submission I test them, thought this will work, now I tested it has thrown me error,
when give `nvme write /dev/nvme0n1 -s 0 -c 10 -z 512 -d data.txt -F 1` -> `Inavalid argment`
Thanks :)

PR #992, opened 2021-04-01
Some minor fixes/additions for nvme-cli

mwilck:
This corresponds to the patch series by the same name that I've posted to linux-nvme on Mar 30th. It's part of my larger PR #877. I'm posting it here knowing that #877 is harder to review and more controversial, in the hope to get these small fixes merged.

## cover letter from linux-nvme:

This series collects patches from my former submission "RFC: add "nvme
monitor" subcommand" which aren't directly related to the monitor
functionality, but required to make the monitor in its current form
compile. A v3 of the monitor code is about to follow.

The main part is the "generic logging functionality".

Changes wrt the RFC: add "nvme monitor" subcommand series: many.

 - rebased to current upstream code base e9d43dd ("[nvme-cli] Fix Max/Min 
   User data erase counts displayed in 0xC2 Log Page"))
 - renamed log() to msg() in order to prevent name clash with <math.h>
 - put logging and cleanup files in "util/" subdirectory
 - converted the bulk of the logging code from a macro into a function
   (Hannes)
 - added functionality to optionally log the PID (default off).
 - added Hannes' patch he recently submitted via github
   (https://github.com/linux-nvme/nvme-cli/pull/942), with two minor
   fixes on top, rationale in the commit messages
 - various additional fixes for memory handling

Both cleanup and logging functions / macros could be used in more
places with positive effects, but my focus is currently on preparing
the monitor functionality.

Changes v1 -> v2:

 - rebased on current upstream
 - squashed previous patches 1 and 2 into one, and simplified the logic
   (Sagi) - if the kernel has no "kato" attribute, we now simply assume
   that existing discovery controllers we encounter are persistent.
 - added Reviewed-by: trailers where appropriate

### DIFF
--- a/Documentation/nvme-connect-all.1
+++ b/Documentation/nvme-connect-all.1
@@ -2,12 +2,12 @@
 .\"     Title: nvme-connect-all
 .\"    Author: [FIXME: author] [see http://www.docbook.org/tdg5/en/html/author]
 .\" Generator: DocBook XSL Stylesheets vsnapshot <http://docbook.sf.net/>
-.\"      Date: 10/20/2020
+.\"      Date: 01/20/2021
 .\"    Manual: NVMe Manual
 .\"    Source: NVMe
 .\"  Language: English
 .\"
-.TH "NVME\-CONNECT\-ALL" "1" "10/20/2020" "NVMe" "NVMe Manual"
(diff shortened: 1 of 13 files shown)


keithbusch:
Yes, these all look good

keithbusch:
Manually applied due to a few trivial conflicts with some recent merges

PR #1026, opened 2021-05-06
completions: fix syntax typo in zsh completions

RPigott:


birkelund:
I noticed this as well, fix is good!

Issue #395, opened 2018-07-26
'id-ctrl -H' Field Definitions


krishnanadh:
Output fields of 'id-ctrl' like 'elpe', 'npss', 'tnvmcap', 'unvmcap' among others need definitions with the '-H' command option.

keithbusch:
Those are all scalar values, and -H was meant for decoding bit-fields. What would you want -H to do for those?

krishnanadh:
I see, understood. I was hoping -H could append for example the string "Unallocated NVM Capacity" for unvmcap.

igaw:
No progress in this bug report for a long time. If you think it's still relevant please reopen.

PR #1028, opened 2021-05-07
nvme-print: Changes for id-ctrl and degree celsius prints

ikegami-t:


ikegami-t:
It seems that the issue #395 can be resolved by the change 7bec0e7.

PR #1029, opened 2021-05-07
nvme: get-feature: Fix to print human readable THSEL and TMPSEL values

ikegami-t:


ikegami-t:
Sorry these changes are not correct. (Probably the drive behavior is not correct as not returned command Dword 11 as completion queue Dword 0.)

PR #1036, opened 2021-05-14
Fix scoped IPv6 IP addresses in TRADDR

martin-belanger:
[RFC4007](https://datatracker.ietf.org/doc/html/rfc4007) describes a way to specify IPv6 addresses with a scope as follows:

```
<ipv6-address>%<scope>

E.g.  fe80:aaaa:1807:5b9e:6be7:f38d%enp0s3
```

The '`%`' delimiter is used to separate the address from the scope. And the scope is used to identify an interface (e.g. `eth1`, `enp0s3`, etc). Linux supports this syntax. The `nvme-tcp` driver does too. Unfortunately, `nvme-cli` strips the scope away from `TRADDR`. This pull requests fixes the issue and allows passing the scope to the `nvme-tcp` driver.

Note that the fix is inspired from code found in the `nvme-tcp` driver. Namely, the use of the API `inet_pton_with_scope()` to determine whether an IPv6 is valid. The current code uses `inet_pton()`, which does not support scoped IPv6 addresses.

I tested the changes to make sure that both IPv4 and IPv6 work as expected. I tested IPv6 with and without the scope to make sure that both cases are working.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -38,6 +38,8 @@
 #include <sys/types.h>
 #include <arpa/inet.h>
 #include <netdb.h>
+#include <net/if.h>
+#include <limits.h>
 
 #include "util/parser.h"
 #include "nvme-ioctl.h"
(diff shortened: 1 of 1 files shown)


hreinecke:
Can you rebase it for the libnvme merge? It looks as if most functionality needs to be moved to libnvme.

PR #1041, opened 2021-05-16
nvme: add json output format for list_ns

ganaidu707:


keithbusch:
This conflicts with a previous pull that fixed the csi type

PR #1044, opened 2021-05-17
nvme: add fuse option for compare and write commands

ikegami-t:


keithbusch:
I don't see how you could have possibly tested this.

ikegami-t:
> I don't see how you could have possibly tested this.

Sorry for the flags parameter the errno EINVAL is returned from the driver. (I though as it was returned from the drive not supported the fused operation but not correct.)
By the way can we change the driver behavior also to handle the flags option with the error returned?

keithbusch:
Fused operations require each part submit to consecutive slots in the same queue. You can't enforce that from user space, and the block layer has no idea we need that either. There was an attempt to make that work a few months ago, but it is too invasive on the fast path.

ikegami-t:
> Fused operations require each part submit to consecutive slots in the same queue. You can't enforce that from user space, and the block layer has no idea we need that either. There was an attempt to make that work a few months ago, but it is too invasive on the fast path.

Thanks for your reply. Noted. It seems that it is better to be deleted the io-passthru command flags parameter also as same with the write and compare commands not supported the fuse option.

keithbusch:
We could mark it reserved, but we can't delete it from the structure since we have to preserve ABI compatibility. The initial idea was to match up with NVMe's CDW0 second byte, but those flags are really too low level for user space to safely use. The other example is SGL vs PRP, which is specific to DMA capabilities and physical memory, so these settings should be owned by the driver.

I have an idea to repurpose these flags as a user-space control option that has nothing to do with anything defined by the nvme spec. Specifically, I would like to be able to select types of blk-mq contexts to dispatch the request, and this field could be repurposed for that. I just need to convince the other kernel developers that it's a good idea.

ikegami-t:
I could understand well. Also looks good about the idea to repurpose. Thank you.

PR #1047, opened 2021-05-18
zns: fix get_zdes_bytes return value in failed cases

ganaidu707:
get_zdes_bytes shall return constant value in failed cases in
order to catch in called function

Check if the get_zdes_bytes returned proper data length or not.


ganaidu707:
rebased and resolved conflicts!

PR #1050, opened 2021-05-22
nvme-cli: Add lsi option for get-log command

ikegami-t:
Previously lsi is set as 0 as hard coded value by the get-log command.
To set this parameter by the command add lsi option for the command.


ikegami-t:
This request is for same changes with the request #823 but rebased and changed the branch to ikegami-t:get-log-lsi from ikegami-t:master.

PR #1053, opened 2021-05-25
Add --host-iface option

martin-belanger:
The option `--host-iface` is used to specify the network interface used on the host to connect to the Controller (e.g. IP eth1, enp2s0). This forces the connection to be made on a specific interface instead of letting the system decide.

Corresponding changes have been made to the linux-nvme nvme-tcp kernel module and merged to kernel 5.14 (see link below). 

Linux-nvme kernel 5.14 commit: http://git.infradead.org/nvme.git/commit/fb2a2b1064b367fce252045da00666d33aa8c2d3


martin-belanger:
Hi @keithbusch 

Just checking if there's anything I can do to help with this pull request. As mentioned above, changes were made to the kernel module that will be available in kernel 5.14 and it would be nice if nvme-cli had the corresponding changes. 

Thanks.

keithbusch:
Sorry, I missed that the kernel side is already staged. This all looks good to go.

PR #1054, opened 2021-05-26
ZN-350 - wdc vs-fw-activate-history command bug fix

AkhileshRN:


plugins/wdc/wdc-nvme.c
```diff
@@ -4021,17 +4021,12 @@ static void wdc_print_fw_act_history_log_normal(__u8 *data, int num_entries, __u
...
 	memset((void *)time_str, 0, 11);
 
+	printf("  Firmware Activate History Log \n");
+	printf("           Power on Hour       Power Cycle     Previous    New                               \n");
+	printf("  Entry      hh:mm:ss             Count        Firmware    Firmware    Slot   Action  Result \n");
+	printf("  -----  -----------------  -----------------  ---------   ---------   -----  ------  -------\n");
+
 	if (data[0] == WDC_NVME_GET_FW_ACT_HISTORY_C2_LOG_ID) {
-		printf("  Firmware Activate History Log \n");
-		if (cust_id == WDC_CUSTOMER_ID_0x1005) {
```
jeffreyalien, line 4026: There are certain FW versions that return Power On Hour in the hh:mm:ss format and others that return it in a timestamp format.  So I believe the cust_id check needs to remain here.

plugins/wdc/wdc-nvme.c
```diff
@@ -4071,10 +4066,14 @@ static void wdc_print_fw_act_history_log_normal(__u8 *data, int num_entries, __u
 				printf("%s", time_str);
 				printf("     ");
 			} else {
-				printf("   ");
```
jeffreyalien, line 4074: Same comment as above - need to leave this code in to support FW that uses the Timestamp instead of Power on Hour.

jeffreyalien (changes requested):
This pull request should not be merged as coded.  It will break the vs-fw-activate-history command for certain versions of WD drives.

keithbusch:
Thanks Jeff, I'll wait for your review confirmation on this one before pulling.

plugins/wdc/wdc-nvme.c
AkhileshRN: Addressed.

plugins/wdc/wdc-nvme.c
AkhileshRN: Addressed.

jeffreyalien (approved):
The updated changes look good to me.  Keith feel free to merge this PR if you also agree.

PR #1065, opened 2021-05-31
fabrics: skip connect if transport type doesn't match

martin-gpy:
Discovery log page data may include records belonging to different
transport types. If a connect is attempted to a record that doesn't
match the appropriate transport type, it could end up in a connect
failure for that record. For e.g. one would see the below error if
a connect is attempted to a tcp record using the 'fc' transport
type and its associated parameters:

nvme_tcp: malformed src address passed: nn-0xXXXX:pn-0xYYYY

Fix this by proceeding with the connect only if the appropriate
transport type matches for a given record during a nvme connect-all.


martin-gpy:
Need further edits to the patch. Will resubmit. Closing this pull request.

PR #1084, opened 2021-06-17
nvme: add missing endianness annotations

ganaidu707:

### DIFF
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -674,17 +674,17 @@ struct nvme_fw_slot_info_log {
 };
 
 struct nvme_lba_status_desc {
-	__u64 dslba;
-	__u32 nlb;
-	__u8 rsvd_12;
-	__u8 status;
-	__u8 rsvd_15_14[2];
+	__le64 dslba;
+	__le32 nlb;
+	__u8   rsvd_12;
+	__u8   status;
+	__u8   rsvd_15_14[2];
 };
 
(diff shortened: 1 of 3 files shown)


keithbusch:
You have them annotated correctly now, but didn't add any of the missing le*_to_cpu() in the accessors. I wonder if we can get incorporate something like the kernel's "sparse" make analysis to find these kinds of incorrect usages...

linux/nvme.h
```diff
@@ -1458,21 +1458,21 @@ struct nvme_error_log_page {
 struct nvme_firmware_log_page {
 	__u8	afi;
 	__u8	resv[7];
-	__u64	frs[7];
+	__le64	frs[7];
```
keithbusch, line 1461: this type may have incorrectly been carried over from tools implementing pre-releases of the spec. The field is an ASCII array, so it should be a __u8[8].

ganaidu707: yeah, I have fixed and added a PR https://github.com/linux-nvme/nvme-cli/pull/1088
Please have a look!
Thanks

ganaidu707:
> You have them annotated correctly now, but didn't add any of the missing le*_to_cpu() in the accessors. I wonder if we can incorporate something like the kernel's "sparse" make analysis to find these kinds of incorrect usages...

Added missing le*_to_cpu() in the accessors.
Thanks!

PR #1085, opened 2021-06-17
Add Timeout Parameter for Zone Management Commands

bjpaupor:
- Enabled the option to set the timeout value used for each of the zone management commands, specifically to allow for longer timeouts with the zone-reset command. 
- The default timeout for zone reset is also increased to 90 seconds.

### DIFF
--- a/plugins/zns/zns.c
+++ b/plugins/zns/zns.c
@@ -130,12 +130,13 @@ close_fd:
 	return nvme_status_to_errno(err, false);
 }
 
-static int __zns_mgmt_send(int fd, __u32 namespace_id, __u64 zslba,
-	bool select_all, enum nvme_zns_send_action zsa, __u32 data_len, void *buf)
+static int __zns_mgmt_send(int fd, __u32 namespace_id, __u64 zslba, 
+	bool select_all, __u32 timeout, enum nvme_zns_send_action zsa, 
+	__u32 data_len, void *buf)
 {
 	int err;
 
-	err = nvme_zns_mgmt_send(fd, namespace_id, zslba, select_all, zsa,
+	err = nvme_zns_mgmt_send(fd, namespace_id, zslba, select_all, timeout, zsa,
 			data_len, buf);
 	close(fd);
 	return err;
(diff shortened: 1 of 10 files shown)


plugins/zns/zns.c
```diff
@@ -154,14 +156,19 @@ static int zns_mgmt_send(int argc, char **argv, struct command *cmd, struct plug
 		__u64	zslba;
 		__u32	namespace_id;
 		bool	select_all;
+		__u32	timeout;
 	};
 
-	struct config cfg = {};
+	struct config cfg = {
+		/* Longer (90 second) default timeout for zone_reset */
```
MatiasBjorling, line 163: Hi,

It may be better to use the default timeout value. Otherwise it may confuse people using the tool that some commands allows a longer timeout, while others use the default timeout value (30 secs?)

bjpaupor: Howdy @MatiasBjorling,

The default value here is set similarly to how 'nvme format' handles the timeout parameter (90 seconds here vs. 60 seconds for format). Zone-reset seems to be the only zone management command that will regularly require a longer timeout than the default value, particularly with the -a option and a large namespace. 

Because of this, I've set the default to a higher value in that case to save the user some headache. With or without the extended default, users would have to find out about any timeout issue and set it manually with the new parameter.

keithbusch: If you are finding that the system's default time is insufficient, it's fine with me to let the tooling pick one that is more appropriate. That said, even the default 30 seconds seems like an awfully long time for a zone reset operation. Is that normal?

MatiasBjorling:
I think that if a full reset takes much longer than a second, there is a
bug somewhere.


fre. 18. jun. 2021 18.04 skrev Keith Busch ***@***.***>:

bjpaupor:
Some folks working on firmware shared that resetting all zones on a large capacity device with every zone full could result in expected times > 60 seconds, I can verify with them if that seems like an issue.

MatiasBjorling:
Ok, sounds good.

fre. 18. jun. 2021 18.49 skrev Brandon Paupore ***@***.***>:

bjpaupor:
Pushed a new version of the patch without the default being set, @keithbusch it seems like there was an unrelated issue in the build for CI?

keithbusch:
> Pushed a new version of the patch without the default being set, @keithbusch it seems like there was an unrelated issue in the build for CI?

Yeah, it looks to me the CI experienced a problem unrelated to your change.

PR #1087, opened 2021-06-18
Add capability for unique plugin version

jeffreyalien:


MatiasBjorling:
Hi Jeff,

It would help Keith to review the patch if you provide the motivation with the PR and also discuss the reason for the patch in the patch description.

Otherwise it will hard for people to know why this patch was needed.

MatiasBjorling:
Remember a signed off tag ad well.

jeffreyalien:
Matias, 
I can add a description and sign off tag, but not sure of the process to do so.  I don't see a way to do that without closing the PR and re-opening a new one.  Is that the only way?   

Here's the motivation.  We've had several users (both internal and external) who have requested a unique plugin version could be used to easily identify nvme cli versions with specific function.   Since the main version tag is not updated very often and the git commit id is cumbersome to search for and use as a level identifier, it was requested we provide a simple, unique plugin version.  This patch provides a way for the plugin vendors to uniquely identify the level of the plugin code included within the nvme cli code.

PR #1105, opened 2021-07-10
nvme: support ns generic device to submit_io

minwooim:
Retrieve `logical_block_size` sysfs attribute when I/O subcommand(e.g.,
read, write, ...) is given instead requesting BLKSSZGET ioctl to the
generic device which is not supported.

### DIFF
--- a/nvme-topology.c
+++ b/nvme-topology.c
@@ -667,6 +667,30 @@ char *nvme_char_from_block(char *dev)
 	return path;
 }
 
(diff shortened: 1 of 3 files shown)


minwooim:
@keithbusch ,

Has this commit been merged to `master`?  I don't think this commit is able to be seen in the master branch.
Maybe updated forcibly?

minwooim:
Oh I can see the `standalone` branch, so is this branch going to be managed two different ways like with `libnvme` and `standalone`?

Issue #1095, opened 2021-06-30
nvme format does not check holders


bdrung:
The `nvme format` command is dangerous: We use a RAID 1 with two NVMe. Sometimes a NVMe fails with a controller fatal status condition. Then we remove the faulty NVMe from the RAID, format the faulty NVMe, and add at back to the RAID for a full resync. Early in the morning this happened again, but this time I forgot to remove the faulty NVMe before formatting. Sadly this causes the filesystem on top of the RAID to choke on the formatted NVMe in the RAID. It required us to stop the consumers and do a filesystem check to get back to a clean state.

To prevent this mistake to happen again, I need a tool that checks that the NVMe is not in use before formatting. My preferred solution is that `nvme format` is enhanced instead of writing a wrapper script for checking it. I propose following change for `nvme format`:

* fail if there are holders for the nvme
* if called with `--force`, only a warning is printed
* `--ignore-holders` will not check if there are holders

Do you agree to this proposed change? Then I will start implementing it.

keithbusch:
What method do you have in mind to verify there are no holders?

bdrung:
Check the content of `/sys/block/$nvme/holders` for content. If there is content, there are holders. Otherwise it is safe to proceed.

keithbusch:
```
# mount | grep nvme
/dev/nvme0n1p1 on /boot/efi type vfat (rw,relatime,fmask=0077,dmask=0077,codepage=437,iocharset=ascii,shortname=winnt,errors=remount-ro)
# ls /sys/class/block/nvme0n1p1/holders
```

My mounted partition lists no holders, and would pass this check. That doesn't sound like what you'd want to happen. It looks like holders only lists stacking block drivers: device-mapper and md are the only subscribers to that interface, so that's only a partial solution to verifying if a namespace is in use.

The nvme device the command is typically sent to the character device rather than the block device or a partition of one. It should be possible to locate all the namespace block devices and their partitions associated with the character device's subsystem, but it's a bit messy.

But, the real problem is that multipathing these among multiple hosts. We can't really query the other hosts to see if the namespace is in use or not.

keithbusch:
Ignoring the multipath problem, I *think* just opening the affected namespace with O_EXCL should be able to verify no one is using it on the local machine. I haven't tested it yet.

bdrung:
Very good point. Let's test if opening the affected namespace with O_EXCL covers the two use cases:
* nvme used in raid
* nvme mounted

PR #1111, opened 2021-07-15
Fail on busy namespace

olase23:
To avoid formarting an in use namespace accidently i added a new function called `parse_and_open_exclusive()`. This opens 
the device file in exclusive mode `O_EXCL.` If this fails with the error EBUSY an error messages hints to the `--force` command line option. By calling `nvme format` with  the `--force` option it will continue formating the namespace.

This patch fixes issue: https://github.com/linux-nvme/nvme-cli/issues/1095

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -156,12 +156,12 @@ static bool is_blkdev(void)
 	return S_ISBLK(nvme_stat.st_mode);
 }
 
-static int open_dev(char *dev)
+static int open_dev(char *dev, int flags)
 {
 	int err, fd;
 
 	devicename = basename(dev);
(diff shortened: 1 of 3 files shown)


keithbusch:
Instead of having a special function to get the specific command line argument by name, I would really prefer just enumerating it in the OPTS structure. This will have it print out in the "--help" and keep it consistent with the self documenting code.

hreinecke:
This is far too convoluted. Patches updating the nvme-cli implementation to v2.0 are a good point for sure, but really do belong into a separate merge request. Similar for compilation issues.
Please separate out the various topics into separate merge requests.

PR #1119, opened 2021-07-24
Update nvme.c

jinhua2021:
$date +"%s%3N" >timestamp.log
$cat timestamp.log
1627131384309

$./nvme set-feature /dev/nvme0 -f 0x0e -v 1627131384309 
it works well, because it reads the value correctly.

$./nvme set-feature /dev/nvme0 -f 0x0e -d timestamp.log
It will read the timestamp file as data len 8, but actual it is 13. It only memcpy 16271313 instead of 1627131384309.

It will make set-feature, get-feature wrong behave for timestamp.


franciscomunoz:
@jinhua2021 I would like to add that I've been testing activation history and FID 0x0e is working as is.

jinhua2021:
@franciscomunoz @keithbusch
Thanks for your reply.

I would like to make it clear as follow:

jinhua2021:
I updated the PR and the description.

keithbusch:
Are you trying to write a string into this value instead of the integer representation?

keithbusch:
eh, this particular feature is implemented so badly...

Your patch won't work because the read size exceeds the buffer size.

I'll fix this up.

keithbusch:
I'm just going to remove the string conversion for now since trying to support it is a mess at the moment. Just use the integer representation.

Also, this is not something a user should be setting anyway. The driver sets this during device probe, and overriding it can cause some unusual issues on the device side.

PR #1121, opened 2021-07-28
Update bash completions

michael-gebis:
Bash completions were quite out of date; this brings them into sync with the last few years of command line options.

This also adds completions for plugin command line arguments, which didn't exist before.  This is the bulk of the diff.  It should be easy for new plugin authors to add extensions in the future.

Also fixed a typo in one .c file, and a formatting issue in another.  (I noticed the formatting when I was trying to extract the list of all new command line options).

I did not touch the zsh completions, which need a similar update.


keithbusch:
Hand applied to resolve merge conflict.

PR #1123, opened 2021-08-03
nvme: Bug fix default data_len used if no argument given

sc108-lee:
current logic ignore given data by user
so make priority to use user argument first


keithbusch:
I found this same problem almost at the same time. Thanks for the fix.

PR #1131, opened 2021-08-11
nvme: use libnvme for printing status

hreinecke:
With commit fe62ba7 ("util: Add nvme_status_to_string()") libnvme
now has a 'nvme_status_to_string()' function, which clashes with
the one provided here.
So drop the function in favour of the libnvme one, and convert
all plugins to use nvme_show_status().

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -2603,8 +2603,8 @@ void d_raw(unsigned char *buf, unsigned len)
 
 void nvme_show_status(__u16 status)
 {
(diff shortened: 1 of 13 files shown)


hreinecke:
And of course, this requires to update the libnvme submodule to at least commit fe62ba7. But I guess you've figured that out already.

hreinecke:
Retract in favour of the 'workflow' branch which has the same commit.

PR #1132, opened 2021-08-11
Update workflow to checkout libnvme

hreinecke:


keithbusch:
I think one of the useful parts of compiling for PPC was that it sometimes produced different size types from what x86_64 was producing, and sometimes caught unexpected size and print specifier differences. But I'd rather make the libnvme integration working first, then I can work back in the missing features from the older branch, so thank you for fixing up the CI.

PR #1133, opened 2021-08-12
Discovery fixes

hreinecke:
Heya,

this is the corresponding patchset to the libnvme pull request 'discovery-fixes'.
It fixes the handling of persistent (and non-persistent :-) discovery controllers, and also some crashes which I've encountered on the way.

Once both patchsets are merged one should reset the submodule branch to pointing to the upstream libnvme repository; I'll see if the CI will accept a patch for that even without the merge.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "libnvme"]
 	path = libnvme
-	url = https://github.com/hreinecke/libnvme.git
-	branch = nvme-cli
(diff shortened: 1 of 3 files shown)


hreinecke:
And indeed, that worked. So the last commit will switch to upstream libnvme; clearly one should merge this one only after the corresponding pull request has been merged into libnvme.

keithbusch:
Thanks!

PR #1134, opened 2021-08-13
Enable telemetry data area 4 in base and plugin

bjpaupor:
Allow for telemetry data area 4 to be gathered using the telemetry-log command and wdc-vs-internal-log. Left the default only gathering areas 1-3 because data area 4 may not be supported by the device.


bjpaupor:
@keithbusch Would you be able to take a look at this? Thanks for your time

plugins/wdc/wdc-nvme.c
```diff
@@ -523,6 +523,16 @@ typedef enum
     EOL_RRER                = 108,	/* Raw Read Error Rate */
 } EOL_LOG_PAGE_C0_OFFSETS;
 
+static __u16 nvme_wdc_feat_buf_len[0x100] = {
```
jeffreyalien, line 526: Could we use the version of this structure defined in nvme.c instead of defining our own version and having dual maintenance?  The nvme_feat_buf_len would need to be moved to a common include.

plugins/wdc/wdc-nvme.c
```diff
@@ -2060,6 +2071,46 @@ static int wdc_do_cap_telemetry_log(int fd, char *file, __u32 bs, int type, int
...
+		memset(buf, 0, nvme_wdc_feat_buf_len[NVME_FEAT_HOST_BEHAVIOR]);
+
+		err = nvme_get_feature(fd, NVME_NSID_ALL, NVME_FEAT_HOST_BEHAVIOR, 0, 0,
+				0, nvme_wdc_feat_buf_len[NVME_FEAT_HOST_BEHAVIOR], buf, &result);
+		if (err > 0) {
+			nvme_show_status(err);
+		} else if (err < 0) {
+			perror("get-feature");
+		} else {
+			if ((ctrl.lpa & 64)) {
```
jeffreyalien, line 2096: Does it make more sense to use 0x40 instead of 64 here and above in the base code?   To me, it seems more obvious which bit you're checking with 0x40.

plugins/wdc/wdc-nvme.c
bjpaupor: I agree that would be clearer, but putting it in any of the common includes makes the compiler complain about the defined but not used value in any file that includes it.
I could set up a special header file containing just that nvme_feat_buf_len and only include it in the files that use it, but that might be adding more complexity than simply duplicating the values.

jeffreyalien: Yea, I don't think it makes to create a special include for this.  Just thought it would nice have one a common struct vs having to maintain 2 versions of it.

bjpaupor:
@jeffreyalien I put in a new function to grab the feature buffer lengths to avoid duplicating the structure, and I also made that change you suggested of using 0x40 instead of 64 for clarity.
@keithbusch Any thoughts on these changes, or do they seem good to merge?

PR #1141, opened 2021-08-23
nvme-plugins: add project info for smart-log-add

jinhua2021:
add project info for smart-log-add command to get the correct log page format


jinhua2021:
Hello @keithbusch,
Would you please help check our patch for plugins? Thank you.

PR #1150, opened 2021-08-31
Fix the memory leak for nvme discover command

wubo009:
Fix the memory leak for nvme discover command

Signed-off-by: Wu Bo <wubo.linux@gmail.com>


keithbusch:
Just need to squash the commits to a single one for this, and should be okay.

PR #1161, opened 2021-09-09
Add formatting of ZNS I/O Commands when printing Effects Log Page

metaspace:

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -110,6 +110,9 @@ const char *nvme_cmd_to_string(int admin, __u8 opcode)
 		case nvme_cmd_resv_release:	return "Reservation Release";
 		case nvme_cmd_verify:		return "Verify";
 		case nvme_cmd_copy:		return "Copy";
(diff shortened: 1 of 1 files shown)


nvme-print.c
```diff
@@ -110,7 +110,10 @@ const char *nvme_cmd_to_string(int admin, __u8 opcode)
 		case nvme_cmd_resv_release:	return "Reservation Release";
 		case nvme_cmd_verify:		return "Verify";
 		case nvme_cmd_copy:		return "Copy";
-		}
+		case nvme_zns_cmd_mgmt_send: return "Zone Management Send";
+		case nvme_zns_cmd_mgmt_recv: return "Zone Management Receive";
```
MatiasBjorling, line 114: Could you add a header that tells the user that these commands are part of the ZNS Command Set. Similar to what is printed for the NVM Command Set?

metaspace: Pushed a new version with headers for ZNS commands, Vendor commands.

metaspace:
Closing to rewrite, asking for log pages using CSI to select command set.

PR #1163, opened 2021-09-14
Update handling of ZNS Command Set in `effects-log`

metaspace:
This patch set contains three patches that updates `zns effects-log` to work better with ZNS devices. See commit messages for details.


keithbusch:
I'm not sure about patch 2 changing all the printf's to use a FILE stream instead. What was wrong with the existing method?

metaspace:
@keithbusch I did not want to print the header if there are no entries for the header. I considered running through the entries twice, once for determining if the header should print, and once for printing the entries. However, I think it is claner instead to print to a memory buffer, check if anything was written to the buffer and if so, print the header to `stdout` before dumping the buffer contents to `stdout`.

metaspace:
@keithbusch would it be better without the vendor specific command header(s) ? In that case i can just revert that part of the patch.

PR #1164, opened 2021-09-15
zns: Add missing Zone Attributes Information on zone descriptor

sc108-lee:
NVMe-Zoned-Namespace-Command-Set-Specification-1.1a-2021.07.26-Ratified
Figure 37: Zone Descriptor Data Structure - Zone Attributes Information

### DIFF
--- a/linux/nvme.h
+++ b/linux/nvme.h
@@ -1928,7 +1928,8 @@ struct nvme_zns_desc {
 	__u8	zt;
 	__u8	zs;
 	__u8	za;
-	__u8	rsvd3[5];
+	__u8	zai;
+	__u8	rsvd4[4];
 	__le64	zcap;
 	__le64	zslba;
 	__le64	wp;
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -4498,7 +4498,7 @@ void nvme_show_zns_changed(struct nvme_zns_changed_zone_log *log,
 		printf("zid %03d: %"PRIu64"\n", i, (uint64_t)le64_to_cpu(log->zid[i]));
 }
 
-char *zone_type_to_string(__u8 cond)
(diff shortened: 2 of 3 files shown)


MatiasBjorling:
It'll most likely be very hard to read when multiple flags are set. Would it be worth considering to make an "extended option" - which would print out each of the separate fields in a sane format.

The same holds for the current attribute field. That is hard to read as well - but that has already been there for a long time, and may be hard to remove at this point.

sc108-lee:
@MatiasBjorling yes I worried about same thing.
But follow the way current za print.

if change is acceptable for this,
How about this, add human readable.
with human readable : zt : string, zs : string, za : string, zai : string
without human readable : zt : hex, zs : hex, za : hex, zai : hex

keithbusch:
Could you call the option "verbose" in order to expand the hex fields to strings?

sc108-lee:
option name human-readable is more common in nvme-cli.
eg) nvme.c  "human-readable" : 15, "verbose" : 4
so I thought human-readable is better.
Is there any specific reason that you chose verbose??

keithbusch:
I never liked the option name, and it doesn't fit with what it does. The default is already human readable, and the "human-readable" option just displays it more verbosely, hence i'd like to migrate to verbose options which aligns with other tools like lspci.

sc108-lee:
@keithbusch  thanks for the explanation.

I just checked that, report_zones which call this function already has human-readable option
https://github.com/linux-nvme/nvme-cli/blob/nvme-cli-monolithic/plugins/zns/zns.c#L598

But currently there is no functionality for that.
so It might be available to be changed to verbose

sc108-lee:
rebased after report zones json print is merged :)

PR #1165, opened 2021-09-15
Fix error handling in `list-ctrl`

metaspace:
This patch fixes a bug in `list-ctrl` where an error code was misinterpreted.


keithbusch:
Wow, that's a very silly oversight, I guess this has been an uncommon command.

PR #1166, opened 2021-09-15
Remove unused command line argument for `nvme list`

metaspace:
Remove the `--directory` argument for `nvme list`. The option has been without
effect since nvme-cli moved to libnvme.


keithbusch:
I believe I'll need to bring the option back at some point (was this for SPDK?), but since it's unused here, I'll merge this request.

PR #1168, opened 2021-09-16
Binary output support for list-ctrl command

NateRoiger:
Simple addition to handling the BINARY flag option.

Example
```
# nvme list-ctrl --namespace-id=1 --output-format=binary /dev/nvme1 | hexdump
0000000 0001 0001 0000 0000 0000 0000 0000 0000
0000010 0000 0000 0000 0000 0000 0000 0000 0000
*
0001000
```

Signed-off-by: Nate Roiger <nate.roiger@hpe.com>


keithbusch:
Thanks for sending this. All commands that return a known sized nvme protocol data structure should have this option.

PR #1175, opened 2021-09-23
Intel latstats: Optane clears latency stats evey time the log page is pulled, so we only query once

franciscomunoz:
Now we only perfom a single query and place the result in a byte
array. Then,  we determine the right log page layout in terms of
the media version.

Signed-off-by: Francisco Munoz <francisco.munoz.ruiz@intel.com>


keithbusch:
I'll take your word for it. :)

PR #1177, opened 2021-09-23
add json support for zns report zones

bjpaupor:
Generate json output when appropriate option is specified for zns report-zones.


keithbusch:
Looks good, though I'm going to have to look more into the ext data print for a potential future modification: I'm not sure if what this conveys is useful information.

PR #1181, opened 2021-09-29
[nvme-cli] ZNS report zones command fix for large number of zones

jeffreyalien:
When there are a large number of zones, the buffer required to contain the data becomes extremely large and the linux os is unable to handle that on all systems.   This change will break up the transfer of the zone data into smaller chunks - 128 zone entries at a time.   

Signed-off-by: Jeff Lien jeff.lien@wdc.com


plugins/zns/zns.c
```diff
@@ -627,23 +637,54 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
...
+	}
+	else {
+		nvme_show_status(err);
+		goto close_fd;
 	}
 
-	report_size = sizeof(struct nvme_zone_report) + cfg.num_descs *
-		(sizeof(struct nvme_zns_desc) + cfg.num_descs * zdes);
+	log_len = sizeof(struct nvme_zone_report);
+	struct nvme_zone_report *buff;
```
keithbusch, line 658: Please don't mix declarations in the code.

plugins/zns/zns.c
```diff
@@ -652,17 +693,37 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
...
-		nvme_show_zns_report_zones(report, cfg.num_descs, zdes,
-			report_size, flags);
-	else if (err > 0)
-		nvme_show_status(err);
-	else
-		perror("zns report-zones");
+	offset = cfg.zslba;
+	printf("nr_zones: %"PRIu64"\n", (uint64_t)le64_to_cpu(total_nr_zones));
+
+    while (nr_zones_retrieved < nr_zones) {
```
keithbusch, line 699: Should be a tab instead of spaces

plugins/zns/zns.c
```diff
@@ -652,17 +693,37 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
...
+
+		err = nvme_zns_report_zones(fd, cfg.namespace_id, offset,
+			cfg.extended, cfg.state, cfg.partial, log_len, report);
+		if (err > 0) {
+			nvme_show_status(err);
+			break;
+		}
+
+		if (!err)
+			 nvme_show_zns_report_zones(report, nr_zones_chunks, zdes,
```
keithbusch, line 716: This is going to change the output because it will print out the "nr_zones" header multiple times. Maybe that's okay, though, I'll see how it looks.

jeffreyalien: I moved the print of "nr_zones from the nvme_show_zns_report_zones function into this function so that shouldn't be a problem.

plugins/zns/zns.c
jeffreyalien: Fixed in latest commit.

plugins/zns/zns.c
jeffreyalien: Fixed in latest commit.

plugins/zns/zns.c
```diff
@@ -572,6 +572,18 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
 	__u32 report_size;
 	void *report;
 	bool huge = false;
+	struct nvme_zone_report *buff;
+
+	unsigned int nr_zones_chunks = 128,
```
MatiasBjorling, line 577: Have you considered using the MDTS value (in the Identify Controller data structure) to determine the maximum transfer size?

If not, I would recommend using that as the transfer size or maybe something like 32-64KB (good tradeoff wrt to I/O submission/completion, and data to be sent/represented)

MatiasBjorling (changes requested):
Recommend to increase the size from 128 to either follow MDTS or a much larger value, such as 16KB or 32KB to optimize for I/O transfers.

plugins/zns/zns.c
jeffreyalien: @MatiasBjorling  I can increase the transfer size to 64KB.  Using the MDTS value could still fail with some drives.  My experience with linux is that it starts hitting intermittent buffer mapping and allocation failures with buffers larger then 500k.   64K seems to be a good compromise.

keithbusch:
Since so many environments don't enable hugetlbfs and the kernel is unpredictable at mapping >=512k, I'm okay with anything < 512k.

plugins/zns/zns.c
jeffreyalien: I increased the buffer/transfer size to 64K (1000 entries at 64 bytes per entry).

MatiasBjorling: You'll need 1024 entries if you want to hit 64 KiB :)

bjpaupor (approved):
If 64KiB transfer length works for folks, then these changes seem good to me.

plugins/zns/zns.c
jeffreyalien: @MatiasBjorling You're correct and my comment above is wrong.  I did use 1024 in the code change so we should be good there.

PR #1184, opened 2021-10-05
nvme: allow fw commit action values 0x4 and 0x5

ganaidu707:
Allow FW Commit action values from 0 to 7 including reserved fields 0x4
and 0x5, device shall throw appropriate error status code and status type.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -3511,7 +3511,7 @@ static int fw_commit(int argc, char **argv, struct command *cmd, struct plugin *
 		err = -1;
 		goto close_fd;
 	}
-	if (cfg.action > 7 || cfg.action == 4 || cfg.action == 5) {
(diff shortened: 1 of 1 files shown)


igaw:
I am a bit confused by the commit message. Shouldn't this read 'disallow'? 0x4 and 0x5 are listed as reserved in the spec.

hreinecke:
Indeed, I have to concur with Daniel. Where is the point in allowing invalid values to be sent?
_Especially_ for firmware modification commands; I'd rather be extra careful that we're not doing anything wrong.

igaw:
BTW, also check if the master branch is correct.

igaw:
Closing due to inactivity. Please reopen if you have an update on the requested changes.

PR #1197, opened 2021-10-18
nvme: fail on busy namespace

olase23:
To avoid formarting a busy namespace accidently a new function called `open_exclusive()` was added. 
This function opens a namespace in exclusive mode `(O_EXCL)`. If this fails with error `EBUSY` an log
messages hints to the `--force` command line option. By calling `nvme format` with the `--force` option it
will continue formating the namespace.

This patch fixes issue: #1095

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -149,12 +149,12 @@ static bool is_blkdev(void)
 	return S_ISBLK(nvme_stat.st_mode);
 }
 
-static int open_dev(char *dev)
+static int open_dev(char *dev, int flags)
 {
 	int err, fd;
 
 	devicename = basename(dev);
-	err = open(dev, O_RDONLY);
+	err = open(dev, flags);
 	if (err < 0)
 		goto perror;
 	fd = err;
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -3498,6 +3508,24 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 		OPT_END()
 	};
 
+	err = argconfig_parse(argc, argv, desc, opts);
+	if (err)
+		goto ret;
+
+	err = fd = open_exclusive(argc, argv, cfg.force);
```
hreinecke, line 3515: I wonder if we should restrict the exclusive open to 'format'; other functions like 'sanitize' or 'write' will suffer from the same problem.
Care to add the '--force' option to 'sanitize', 'write', 'write-zeroes', and 'write-uncor'?

olase23: First i wanted to fix issue #1095 only. Once this is accepted my goal was to search for all other functions to add the --force option for them too.

PR #1198, opened 2021-10-19
nvme: fix wrong parameter passed to nvme_resv_report

suz-yang:
Fix wrong parameter passed to nvme_resv_report.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4802,7 +4802,7 @@ static int resv_report(int argc, char **argv, struct command *cmd, struct plugin
 	}
 	memset(status, 0, size);
 
-	err = nvme_resv_report(fd, cfg.namespace_id, cfg.numd, cfg.cdw11, status);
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -4802,7 +4802,7 @@ static int resv_report(int argc, char **argv, struct command *cmd, struct plugin
 	}
 	memset(status, 0, size);
 
-	err = nvme_resv_report(fd, cfg.namespace_id, cfg.numd, cfg.cdw11, status);
+	err = nvme_resv_report(fd, cfg.namespace_id, cfg.cdw11, size, status);
```
igaw, line 4805: nvme_resv_report() takes a bool as third argument (bool eds: Request extended Data Structure). 

Given this shouldn't the function be updated accordingly? Currently it takes 

		OPT_UINT("numd",          'd', &cfg.numd,           numd),
		OPT_UINT("cdw11",         'c', &cfg.cdw11,          cdw11),

IMO, a more user friendly option would be just to have the 'eds' command line option.

PR #1204, opened 2021-10-21
meson,Makefile: fix build

birkelund:
Since libnvme commit 2e3616eaf0ed, the endian helpers are no longer
exported. Instead, use ccan/endian and add it to the build systems.

Meson will always checkout head of libnvme, but for the make-based
building, this requires a bump of libnvme for the submodule.


birkelund:
This includes a bump of libnvme.

PR #1206, opened 2021-10-21
Revert use array

igaw:
Revert the change from #1188 as we have ccan now back in the nvme-cli, there is no point in not having the dynamic version for the effects log parser feature.

### DIFF
--- a/ccan/ccan/check_type/LICENSE
+++ b/ccan/ccan/check_type/LICENSE
@@ -1,0 +1,1 @@
+../../licenses/CC0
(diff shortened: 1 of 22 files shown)


igaw:
Revert the change from #1188 as we have ccan now back in the nvme-cli, there is no point in not having the dynamic version for the effects log parser feature.

birkelund:
I'm surprised that its not required to specify the build of the ccan c-files? For the endian include its just a header, but these dependencies have source files as well.

igaw:
Good point. Let me check what those *.c files do and what we have in libnvme in this case.

PR #1215, opened 2021-10-25
nvme: open namespace exclusive by default

olase23:
All data modifying functions open the namespace exclusive. With the --force or -f
option this can be bypassed.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -242,27 +242,30 @@ static int get_dev(int argc, char **argv, int flags)
 int parse_and_open(int argc, char **argv, const char *desc,
 	const struct argconfig_commandline_options *opts)
 {
+	const struct argconfig_commandline_options *s;
+	int flags = O_RDONLY;
 	int ret;
 
 	ret = argconfig_parse(argc, argv, desc, opts);
 	if (ret)
 		return ret;
 
-	ret = get_dev(argc, argv, O_RDONLY);
-	if (ret < 0)
-		argconfig_print_help(desc, opts);
+	for (s = opts; (s->option != NULL) && (s != NULL); s++) {
+		if (!strcmp(s->option, "force")) {
+			if (! *(int *) s->default_value)
+				flags |= O_EXCL;
+			break;
+		}
+	}
 
+	ret = get_dev(argc, argv, flags);
+	if (ret < 0) {
+		if (errno == EBUSY)
+			nvme_show_busy_namespace(basename(argv[optind]));
+		else
+			argconfig_print_help(desc, opts);
+	}
 	return ret;
-}
-
-int open_exclusive(int argc, char **argv, int force)
-{
-    int flags = O_RDONLY;
-
-	if (!force)
-		flags |= O_EXCL;
-
-    return get_dev(argc, argv, flags);
 }
 
 enum nvme_print_flags validate_output_format(const char *format)
@@ -3716,6 +3719,7 @@ static int sanitize(int argc, char **argv, struct command *cmd, struct plugin *p
 	const char *ause_desc = "Allow unrestricted sanitize exit.";
 	const char *sanact_desc = "Sanitize action.";
 	const char *ovrpat_desc = "Overwrite pattern.";
+	const char *force_desc = "The \"I know what I'm doing\" flag, skip confirmation before sending command";
 
 	int fd, ret;
 
@@ -3726,6 +3730,7 @@ static int sanitize(int argc, char **argv, struct command *cmd, struct plugin *p
 		int    ause;
 		__u8   sanact;
 		__u32  ovrpat;
+		int    force;
 	};
 
 	struct config cfg = {
@@ -3735,6 +3740,7 @@ static int sanitize(int argc, char **argv, struct command *cmd, struct plugin *p
 		.ause = 0,
 		.sanact = 0,
 		.ovrpat = 0,
+		.force = 0,
 	};
 
 	OPT_ARGS(opts) = {
@@ -3744,6 +3750,7 @@ static int sanitize(int argc, char **argv, struct command *cmd, struct plugin *p
 		OPT_FLAG("ause",       'u', &cfg.ause,       ause_desc),
 		OPT_BYTE("sanact",     'a', &cfg.sanact,     sanact_desc),
 		OPT_UINT("ovrpat",     'p', &cfg.ovrpat,     ovrpat_desc),
+		OPT_FLAG("force",      'f', &cfg.force,      force_desc),
 		OPT_END()
 	};
 
@@ -4106,24 +4113,6 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 		OPT_SUFFIX("block-size", 'b', &cfg.bs,           bs),
 		OPT_END()
 	};
-
-	err = argconfig_parse(argc, argv, desc, opts);
-	if (err)
-		goto ret;
-
-	err = fd = open_exclusive(argc, argv, cfg.force);
-	if (fd < 0) {
-		if (errno == EBUSY) {
-			fprintf(stderr, "Failed to open %s.\n",
-                basename(argv[optind]));
-			fprintf(stderr,
-				"Namespace is currently busy.\n"
-				"Use the force [--force|-f] option to ignore that.\n");
-		} else {
-			argconfig_print_help(desc, opts);
-		}
-		goto ret;
-	}
 
 	err = fd = parse_and_open(argc, argv, desc, opts);
 	if (fd < 0)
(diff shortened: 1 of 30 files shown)


nvme.c
```diff
@@ -3121,6 +3121,8 @@ static int sanitize(int argc, char **argv, struct command *cmd, struct plugin *p
 	const char *ause_desc = "Allow unrestricted sanitize exit.";
 	const char *sanact_desc = "Sanitize action.";
 	const char *ovrpat_desc = "Overwrite pattern.";
+	const char *force_desc = "The \"I know what I'm doing\" flag, skip confirmation before sending command";
+
```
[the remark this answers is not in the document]
olase23: ack

nvme.c
```diff
@@ -4120,29 +4135,48 @@ static int write_uncor(int argc, char **argv, struct command *cmd, struct plugin
...
 		__u16 block_count;
+		int force;
 	};
 
 	struct config cfg = {
 		.start_block     = 0,
 		.namespace_id    = 0,
 		.block_count     = 0,
+		.force        	 = 0,
+
```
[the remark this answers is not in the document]
olase23: ack

nvme.c
```diff
@@ -4220,9 +4255,23 @@ static int write_zeroes(int argc, char **argv, struct command *cmd, struct plugi
...
+	if (fd < 0) {
+		if (errno == EBUSY) {
+			fprintf(stderr, "Failed to open %s.\n",
+                basename(argv[optind]));
+			fprintf(stderr,
+				"Namespace is currently busy.\n"
+				"Use the force [--force|-f] option to ignore that.\n");
+		} else {
+			argconfig_print_help(desc, opts);
+		}
```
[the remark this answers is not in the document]
olase23: ack

olase23:
Why does the meson build fails here?

igaw:
> Why does the meson build fails here?

We just added Python tests to libnvme  which seems to fail here in this build. I'll look into it. For this PR we can ignore at this point.

We are still ramping up the meson build, thus some fallouts are to be expected. That's why still have the Makefiles around
in this transition phase.

nvme.c
```diff
@@ -3222,12 +3225,22 @@ static int sanitize(int argc, char **argv, struct command *cmd, struct plugin *p
...
+	ret = argconfig_parse(argc, argv, desc, opts);
+	if (ret)
+		goto ret;
+
+	ret = fd = open_exclusive(argc, argv, cfg.force);
+	if (fd < 0) {
+		if (errno == EBUSY)
+			nvme_show_busy_namespace(basename(argv[optind]));
+		else
+			argconfig_print_help(desc, opts);
```
igaw, line 3241: My comment on the repeating code was on the whole snippet, not just the printf output. Just introducing
a s printf helper looks a bit silly. Can't  you move this part here into a helper?

```
+       ret = argconfig_parse(argc, argv, desc, opts);
+       if (ret)
+               goto ret;
+
+       ret = fd = open_exclusive(argc, argv, cfg.force);
+       if (fd < 0) {
+               if (errno == EBUSY)
+                       nvme_show_busy_namespace(basename(argv[optind]));
+               else
+                       argconfig_print_help(desc, opts);
```
Basically a spin on parse_and_open(), e.g. parse_and_open_exclusive()?

igaw (changes requested):

nvme.c
olase23: Well, i had allready this approach see: https://github.com/linux-nvme/nvme-cli/pull/1111/commits/77911c6ea933b4779a3ff84df841f60915229ee6

But since the config structure is not available on this layer and a the option helper function was rejected i implemented it on this silly bases.

igaw: Hmm, lost a bit the overview to be frank. I just starred at this change and though it has a lot of code duplication.

Was it me who requested the change? I really don't see why we should not have a parse_and_open_exclusive() where needed. Let me dig into the history. Sorry for the confusion. 

BTW, I looked at the current open_exclusive() user and there seems to be a bug, see #1221

igaw:
> We are still ramping up the meson build, thus some fallouts are to be expected. That's why still have the Makefiles around in this transition phase.

I've decided to disable the test for now. There too many problems right now.

nvme.c
olase23: Bug #1221 would be fixed by this patch here.

igaw: Ah nice :)

Are you referring to this comment: https://github.com/linux-nvme/nvme-cli/pull/1111#issuecomment-894511945

I think Keith has a good point. Why not doing something like
```


commit 37634c3e09a5862f44d28329a87b6ff5265c6275
Author: Daniel Wagner <dwagner@suse.de>
Date:   Wed Nov 3 09:49:32 2021 +0100

    nvme: Use O_EXCL in parse_and_open when options is present
    
    Introduce a generic way to open files with O_EXCL. When the 'force'
    option is present in the command line options array, add a O_EXCL.
    
diff --git a/nvme.c b/nvme.c
index f40b47e69438..0987cb41bbb8 100644
--- a/nvme.c
+++ b/nvme.c
@@ -199,29 +199,28 @@ static int get_dev(int argc, char **argv, int flags)
 int parse_and_open(int argc, char **argv, const char *desc,
 	const struct argconfig_commandline_options *opts)
 {
+	const struct argconfig_commandline_options *s;
+	int flags = O_RDONLY;
 	int ret;
 
 	ret = argconfig_parse(argc, argv, desc, opts);
 	if (ret)
 		return ret;
 
-	ret = get_dev(argc, argv, O_RDONLY);
+	for (s = opts; (s->option != NULL) && (s != NULL); s++) {
+		if (!strcmp(s->option, "force")) {
+			flags |= O_EXCL;
+			break;
+		}
+	}
+
+	ret = get_dev(argc, argv, flags);
 	if (ret < 0)
 		argconfig_print_help(desc, opts);
 
 	return ret;
 }
 
-int open_exclusive(int argc, char **argv, int force)
-{
-    int flags = O_RDONLY;
-
-	if (!force)
-		flags |= O_EXCL;
-
-    return get_dev(argc, argv, flags);
-}
-
 enum nvme_print_flags validate_output_format(const char *format)
 {
 	if (!format)
@@ -3510,24 +3509,6 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 		OPT_END()
 	};
 
-	err = argconfig_parse(argc, argv, desc, opts);
-	if (err)
-		goto ret;
-
-	err = fd = open_exclusive(argc, argv, cfg.force);
-	if (fd < 0) {
-		if (errno == EBUSY) {
-			fprintf(stderr, "Failed to open %s.\n",
-                basename(argv[optind]));
-			fprintf(stderr,
-				"Namespace is currently busy.\n"
-				"Use the force [--force|-f] option to ignore that.\n");
-		} else {
-			argconfig_print_help(desc, opts);
-		}
-		goto ret;
-	}
-
 	err = fd = parse_and_open(argc, argv, desc, opts);
 	if (fd < 0)
 		goto ret;

```

Only compile tested. Also adding the warnings from your version would probably make good sense

igaw: BTW, when we expect that 'force' is always an option, you could add it as default to all argconfig_commandline_options. With this we don't have to do this stupid for loop with strcmp, instead you could just access 'cfg.force'.

olase23: Ok, this sounds good.

olase23: To add the 'force' option to the argconfig_commandline_option structure in each sub-command is not this easy since the short option '-f' may be used by different purpose. See 'fw_download' for example.

igaw: TBH, I am not a big fan of the short '-f' option for something like this. Introducing only the long version --force would be fine in my books. Would this work? 

(Sorry, I am not looking more closer right now. Pretty busy porting stuff from the monolithic branch to master)

olase23: Ok, i'll start soon to check if this works.

igaw: Thanks!

olase23: Well, it is possible to define the long option only for each sub-command which opens a namespace. If we do it this way the '--force' option must be used even when a sub-command like 'error-log' or 'persistent-event-log' is used which just reads data only.

igaw: Hmm, I think I don't understand your point. My understanding is that the currently open logic is used per default. So nothing should change in absence of '--force'. If the user decided to use '--force' the O_EXCL flag is added to the open call.

olase23: Sorry, may bad. I had the wrong assumtion you want to introduce the long option in each sub-command. Even though it only reads data. 
But what if the short option '-f' is already used for a different purpose? It would create a kind of inconsistency.

igaw: My take is to drop support for the short option completely and only have --force left. Sure breaks compatibility, but I personally prefer to have consistency and --force is 'I know what I am doing, leave me alone' doesn't need a short version in my books :)

olase23: got it :-)

olase23: > _If the user decided to use '--force' the O_EXCL flag is added to the open call._
Well, i guess you meant the other way arround, right? To avoid accidently overwriting we open the device by default with O_RDONLY and O_EXCL which will fail when the device is already in use. Once the '--force' option is used we open only with the flag  O_RDONLY.

olase23: > _If the user decided to use '--force' the O_EXCL flag is added to the open call._
Well, i guess you meant the other way arround, right? To avoid accidently overwriting we open the device by default with O_RDONLY and O_EXCL which will fail when the device is already in use. Once the '--force' option is used we open only with the flag  O_RDONLY.

olase23: @igaw please read my last comment above

igaw:
- I've rebased the branch to the current master and updated some minor coding style nitpicks.
- Also I dropped the short '-f' version in all commands where it is used for '--force'. 
- Updated the docs.

Can you please review my changes?

igaw (approved):

igaw:
I've rebased it the current master and gave it quick test. Looks good. Let's merge it now.

PR #1223, opened 2021-11-05
[WIP] Merge nvme cli monolithic

igaw:
This is a first rough version of my attempt to merge the missing patches from the monolithic branch into master.
I used git merge-base to figure out which patches are missing. During the merge I realized it didn't really
identify all patches. So it's likely there are more patches missing.

$ git merge-base  upstream/nvme-cli-monolithic HEAD
77422f443b0a18412f1b9e800f1bdc60f387dd88

Here the notes on which patches I didn't merge with some notes. I'll add the git hasches later.

- 0004-Removing-redundant-parameter-from-nvme_sec_send.patch
  nvme_sec_send interface, superseeded by vme_security_send

- 0005-Add-NVMe-MI-Features-Controller-Metadata-0x7E-and-Na.patch
  needs to be splitted between nvme-cli and libnvme. The library bits
  need a complete rewrite

- 0006-Print-device-name-on-connect.patch
  rewrite needed

- 0009-nvme-ioctl-return-1-on-failure-from-nvme_get_nsid.patch
  ported/adapted to libnvme

- 0010-nvme-topology-scan-all-controllers-in-scan_subsystem.patch
  ported/adapted to libnvme

- 0012-nvme-add-get-log-page-2.0-spec-fields.patch
  offset type API is missing in libnvme's nvme_get_log

- 0013-nvme-add-the-status-codes-from-2.0-spec.patch
  libnvme has the bits but the nvme.c print needs to updated accordingly

- 0014-nvme-extend-existing-id-ctrl-parameters-as-per-2.0.patch
  function nvme_show_id_ctrl_ofcs() has differs a bit, not clear why.

- 0016-Add-Timeout-Parameter-for-Zone-Management-Commands.patch
  conflicts plugins/zns/zns.c

- 0017-NVMe-CLI-Add-capability-for-unique-plugin-version.patch
  already in

- 0018-nvme-style-fix.patch
  changes for libnvme

- 0026-nvme-topology-fix-controller-check-in-scan_subsystem.patch
  needs to redevolped.

- 0027-fix-code-style.patch
  style fixed no needed

- 0039-nvme-support-ns-generic-device-to-submit_io.patch
  no users, propably should go into libnvme

- 0042-nvme-print-fix-nvme-list-segfault-if-controller-is-u.patch
  not clear if we need it (does not apply)

- 0043-fabrics-add-default-port-number-for-NVMe-TCP-I-O-con.patch
  is correctly handled in libnvme, not needed

- 0046-fix-whitespace-damage.patch
  fixes 0005-Add-NVMe-MI-Features-Controller-Metadata-0x7E-and-Na.patch

- 0063-nvme-cli-Add-support-for-new-SN650-device.patch
   already in

- 0065-Add-function-to-get-specific-feature-buffer-lengths.patch
   merged into previous patch

- 0068-nvme-cli-fix-the-data-mdata-_len-param.patch
   libnvme already supports data buffer and metadata buffer size parameter

- 0074-fabrics-fix-return-value-of-hostnqn_read_file.patch
   libnvme has a different implementation which seems to be correct


igaw:
Rebased and udpated due to changes in libnvme (see https://github.com/linux-nvme/libnvme/pull/94#issuecomment-965101897)

igaw:
It looks like the build failes because the CI doesn't checkout the right revision. 

The correct HEAD is dc14ce184c34 ("Rename aps to apws for nvme_id_psd") but the build job uses f3fd3f3aa20396459cb894517cb3d04f10e5afff

But it's the later ref doesn't exist. I am confused

PR #1229, opened 2021-11-08
nvme-cli: ctrl-loss-tmo should accept -1 as value

maurizio-lombardi:
When passing the parameter --ctrl-loss-tmo=-1
nvme-cli silently drops it and replaces its value with the
default one (600) despite the fact that -1 is valid
and accepted by the kernel.
ctrl_loss_tmo < 0 means that it will try to reconnect forever

This is due to the fact that -1 is treated like "variable unset"
by nvme-cli and the add_int_argument() function ignores
all arguments with such a value if allow_zero is true (the latter seems
to be completely unintended).

This patch fixes the bug by:

1) removing the "arg == -1" condition from add_int_argument() and letting
   the caller take care of checking if -1 is an acceptable value
   for a particular argument.

2) initializing ctrl_loss_tmo to its default value instead of -1.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -69,7 +69,7 @@ const char *conarg_host_traddr = "host_traddr";
 const char *conarg_host_iface = "host_iface";
 
 struct fabrics_config fabrics_cfg = {
-	.ctrl_loss_tmo = -1,
+	.ctrl_loss_tmo = NVMF_DEF_CTRL_LOSS_TMO,
 	.fast_io_fail_tmo = -1,
 	.output_format = "normal",
 };
@@ -965,7 +965,7 @@ add_int_argument(char **argstr, int *max_len, char *arg_str, int arg,
 {
 	int len;
 
-	if ((arg && !allow_zero) || (arg != -1 && allow_zero)) {
+	if (arg || allow_zero) {
 		len = snprintf(*argstr, *max_len, ",%s=%d", arg_str, arg);
 		if (len < 0)
 			return -EINVAL;
@@ -1006,9 +1006,6 @@ int build_options(char *argstr, int max_len, bool discover)
 			msg(LOG_ERR, "need a address (-a) argument\n");
 			return -EINVAL;
 		}
-		/* Use the default ctrl loss timeout if unset */
-		if (fabrics_cfg.ctrl_loss_tmo == -1)
-			fabrics_cfg.ctrl_loss_tmo = NVMF_DEF_CTRL_LOSS_TMO;
 	}
 
 	/* always specify nqn as first arg - this will init the string */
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -1044,10 +1041,12 @@ int build_options(char *argstr, int max_len, bool discover)
 	    (strncmp(fabrics_cfg.transport, "loop", 4) &&
 	     add_int_argument(&argstr, &max_len, "ctrl_loss_tmo",
 			      fabrics_cfg.ctrl_loss_tmo, true)) ||
+	    (fabrics_cfg.fast_io_fail_tmo != -1 &&
```
igaw, line 1044: As far I understand the kernel code, fast_io_fail_tmo = -1 is used to disable this feature. Which means it -1 is a valid value.

fabrics.c
```diff
@@ -1044,10 +1041,12 @@ int build_options(char *argstr, int max_len, bool discover)
...
 	    add_int_argument(&argstr, &max_len, "fast_io_fail_tmo",
-				fabrics_cfg.fast_io_fail_tmo, true) ||
+				fabrics_cfg.fast_io_fail_tmo, true)) ||
+	    (fabrics_cfg.tos != -1 &&
```
igaw, line 1047: Not sure about this either. So I see that RDMA and TCP are only considering a value >=0, though a comment says

`fabrics.c:573:  opts->tos = -1; /* < 0 == use transport default */  `

From this I'd say -1 is a valid value though the current kernel code filters it out.

igaw (commented):
Rest looks good. I am just a bit confused about -1 checks though

fabrics.c
maurizio-lombardi: if you pass tos=-1 to the kernel the latter will print an error and close the connection

fabrics.c
maurizio-lombardi: It was filtered out by add_int_argument() before my patch, so I did the same to avoid introducing regressions

fabrics.c
igaw: Yeah, but things don't seem to add up if I read it correctly. tos is initialized with -1 and when user space doesn't set it, the tos value will be ignored. But when we set it via userspace we can only set it in the range of [0,255] and hence we can't disable it later. So my point/question is should -1 not be allowed by the kernel as 'off' value?

fabrics.c
igaw: Got it, but I think we should figure out how it supposed to work correctly.

maurizio-lombardi:
> Rest looks good. I am just a bit confused about -1 checks though

This is why I added those two -1 checks:

http://bsdbackstore.it/misc/nvme_cli_bug_report.txt

igaw:
> > Rest looks good. I am just a bit confused about -1 checks though
> 
> This is why I added those two -1 checks:

thanks for double checking. Could also send a PR for master please? Thanks!

fabrics.c
maurizio-lombardi: > Got it, but I think we should figure out how it supposed to work correctly.

I will investigate and eventually submit another pull request

maurizio-lombardi:
> thanks for double checking. Could also send a PR for master please? Thanks!

Ok

fabrics.c
maurizio-lombardi: Ok, got it, will investigate this too

maurizio-lombardi:
> thanks for double checking. Could also send a PR for master please? Thanks!

I checked libnvme and it has been fixed there already, no PR for master needed

PR #1231, opened 2021-11-11
nvme: update description for fw-commit

MatiasBjorling:
I've continuously had to look up the meaning of the firmware commit values in the specification, and I think that this command is used by many, and therefore probably could need a slightly better description. I've therefore added what each value of commit action does, and also added that if no slot is selected, the controller selects one for the user.


MatiasBjorling:
Part of the man documentation. Not needed.

keithbusch:
The default --help can be a bit terse, but we do want it to easily fit on the screen. I've considered adding a more verbose help option for details like this, since sometimes that's just more convenient than getting the man page.

PR #1236, opened 2021-11-15
fabrics: fix 'nvme connect' segfault if transport type is omitted

martin-gpy:
Check if the transport type is available before dereferencing
it in discovery_trsvcid().

Fixes: 362c90f ("fabrics: add default port number for NVMe/TCP I/O
controllers")


igaw:
Looks like we need to maintain the nvme-cli-monolithic branch for a bit longer.

PR #1237, opened 2021-11-15
Update and document configuration file access

hreinecke:
This patchset adds a new option 'dump-config' to print out the resulting JSON configuration to stdout, and also documents the 'config-file' commandline parameter.

### DIFF
--- a/Documentation/nvme-connect-all.txt
+++ b/Documentation/nvme-connect-all.txt
@@ -29,6 +29,7 @@ SYNOPSIS
 		[--matching               | -m]
 		[--persistent             | -p]
 		[--quiet                  | -S]
+		[--dump-config            | -O]
 
 DESCRIPTION
 -----------
@@ -165,6 +166,10 @@ OPTIONS
 -S::
 --quiet::
 	Suppress error messages.
+
+-O::
+--dump-config::
+	Print out resulting JSON configuration file to stdout.
 
 
 EXAMPLES
--- a/Documentation/nvme-connect-all.txt
+++ b/Documentation/nvme-connect-all.txt
@@ -17,6 +17,7 @@ SYNOPSIS
 		[--hostnqn=<hostnqn>      | -q <hostnqn>]
 		[--hostid=<hostid>        | -I <hostid>]
 		[--raw=<filename>         | -r <filename>]
+		[--cfg-file=<cfg>         | -C <cfg>]
 		[--keep-alive-tmo=<#>     | -k <#>]
 		[--reconnect-delay=<#>    | -c <#>]
 		[--ctrl-loss-tmo=<#>	  | -l <#>]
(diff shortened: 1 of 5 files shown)


Documentation/nvme-connect-all.txt
```diff
@@ -109,6 +110,12 @@ OPTIONS
 	and dump it to a raw binary file. By default 'nvme connect-all' will
 	dump the output to stdout.
 
+-C <cfg>::
+--config-file=<cfg>::
+	Use the specified JSON configuration file instead of the
+	default /etc/nvme/config.json file or 'none' to not read in
```
sagi-lb, line 116: @hreinecke couple of comments:
1. We need this file to be added as part of nvme-cli installation - so pkg-mgmt can upgrade it correctly
2. We need to have an example on how this file looks like - preferably in the documentation
    The reason is that people will want to populate it without the dump command.

fabrics.c
```diff
@@ -634,6 +641,8 @@ int nvmf_connect(const char *desc, int argc, char **argv)
 		free(hnqn);
 	if (hid)
 		free(hid);
+	if (dump_config)
+		nvme_dump_config(r);
 	nvme_free_tree(r);
 	return errno;
 }
```
sagi-lb, line 648: This looks good.

Documentation/nvme-connect-all.txt
hreinecke: Well, the config file is optional, so I deliberately did not add it as part of the installation.
If a user wants to use a config file he can generate one; and as it's user-generated we shouldn't update it.
Doing so will create an unholy mess of trying to parse and _update_ user-generated input.

hreinecke: But yeah, I already figured we'd need a json schema for the config file.

hreinecke: I have submitted a pull request for a json config schema  in https://github.com/linux-nvme/libnvme/pull/102; please check if that's what you had in mind.

sagi-lb:
btw Hannes, it would be great to have a connect/connect-all that will dump a json entry to stdout (and combined with --dry-run you can simply use this command to create a json entry to be added to the config file).

Just a thought.

Documentation/nvme-connect-all.txt
sagi-lb: I still think that it needs to be installed as an empty file and and be removed with the pkg. Primarily because its a default path nvme-cli/libnvme assume and make use of...
The upgrade should do the standard of .rpmsave and/or .rpmnew

hreinecke: So, added the 'config.json' file as a '%ghost' file to the rpm template.

hreinecke:
While I fully agree with the '--dry-run' thingie (and, in fact, had implemented it myself some time back) I'd rather open another PR for that, as it's arguably a new functionality..

PR #1238, opened 2021-11-16
meson build and configuration updates

igaw:
- Install configuration files, udev rules, documentation etc.
When asciidoctor is installed, the documentation will  generated when '-Ddocs=[all,html,man]' is used.
- Add support back to build against json-c 0.13.
- workflows cleanups
- fix version naming


igaw:
Rebase to the current master.

igaw:
I think we can fix the Makefile build. Currently we still don't have a hard dependency on json-c. The wrapper is still in place. But is it worth?

igaw:
There is still a problem:

> [49/82] Compiling C object 'subprojects/libnvme/test/8d37954@@test-zns@exe/zns.c.o'.
> In file included from ../subprojects/libnvme/ccan/ccan/endian/endian.h:5,
>                  from ../subprojects/libnvme/test/zns.c:19:
> subprojects/libnvme/config.h:9: warning: "CONFIG_JSONC" redefined
>     9 | #define CONFIG_JSONC
>       | 
> In file included from <command-line>:
> ./config.h:9: note: this is the location of the previous definition
>     9 | #define CONFIG_JSONC 1
>

hreinecke:
> There is still a problem:
> 
> > [49/82] Compiling C object 'subprojects/libnvme/test/8d37954@@test-zns@exe/zns.c.o'.
> > In file included from ../subprojects/libnvme/ccan/ccan/endian/endian.h:5,
> > from ../subprojects/libnvme/test/zns.c:19:
> > subprojects/libnvme/config.h:9: warning: "CONFIG_JSONC" redefined
> > 9 | #define CONFIG_JSONC
> > |
> > In file included from :
> > ./config.h:9: note: this is the location of the previous definition
> > 9 | #define CONFIG_JSONC 1

Why do we compile the tests from the submodule? After all, we're not ever testing the submodule here, we're just using it to compile the main thing (which does have its own testsuite).
Can't we introduce a compilation target for libnvme which just compiles the library but _not_ the tests?

igaw:
I haven't looked into why tests are also build in the subpackage. The problem I am pointing out is the problem with the config.h file.  Both nvme-cli and libnvme create a config.h file which both get included. I thought I figured this already out. Time to read the meson documentation on this topic again.

igaw:
So I figured out the part with the config file. libnvme should not place a config.h into the include search path. This will collide with the one produced by nvme-cli. I've fixed by renaming libnvme's config file as recommended by the official meson documentation.

PR #1240, opened 2021-11-16
libnvme: git submodule update

igaw:
Update to latest libnvme version.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -13,8 +13,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: install libraries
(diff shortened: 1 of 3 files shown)


igaw:
Looks like the old json-c is causing the fallout

igaw:
I think the simplest way forward is to use the wrap feature from meson. I've send out a PR to update the json-c version
provided in meson's wrapdb: https://github.com/mesonbuild/wrapdb/pull/229

The other option I see is trying to get somehow a newer json-c library installed in the ubuntu-latest container for the CI.

igaw:
The meson build is working again. But there isn't much I can do for the Makefile build. I think it's time to rip it out.

PR #1243, opened 2021-11-17
Compability for older json-c releases

hreinecke:
commit 0b63a8b7cfb7 ("Check json-c libversion, recover commit d127537")
broke building on older releases when using 'make'.
Add a compability code to enable building on older json-c releases.


igaw:
I found a fix for this which is better than reverting this one. See #1238

hreinecke:
And this is even better.
We really only have an issue if we try to print out numbers larger than UINT_MAX on older json-c releases.
So here I've separated the 64bit value from the others, and added a warning if we've built against older json-c releases _and_ an overflow occurs.
Far better than requiring to update json-c, which will be challenging if you need to build on older releases.

igaw:
Sure there is this argument. I though the 0.13 release was pretty old but it's from 2018. 0.14 and 0.15 are both from mid 2020.
Okay, I'll update the meson builds accordingly.

PR #1245, opened 2021-11-17
TP8013 & 8014 support

hreinecke:
Hi all,

here's now the companion patch to support TP8013 & TP8014 for nvme-cli.
The biggest difference here is that we now have the '--nqn' argument for 'discover' and 'connect-all', allowing to connect to a unique discovery controller NQN. And of course we now decode the new fields and values in the discovery log page entry.


hreinecke:
Sigh. Merge error.

PR #1248, opened 2021-11-18
nvme-print: print generic device in list command

minwooim:
Generic namespace device is one of I/O paths that needs to be listed up
in the `nvme list` command along with the blkdev.  Print it out in the
simple list command.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -6926,8 +6926,8 @@ void nvme_show_list_item(nvme_ns_t n)
 	snprintf(format, sizeof(format), "%3.0f %2sB + %2d B", (double)lba,
 		l_suffix, nvme_ns_get_meta_size(n));
 
-	printf("%-21s %-20s %-40s %-9d %-26s %-16s %-8s\n",
-		nvme_ns_get_name(n), nvme_ns_get_serial(n),
+	printf("%-21s %-21s %-20s %-40s %-9d %-26s %-16s %-8s\n",
+		nvme_ns_get_name(n), nvme_ns_get_generic_name(n), nvme_ns_get_serial(n),
 		nvme_ns_get_model(n), nvme_ns_get_nsid(n), usage, format,
 		nvme_ns_get_firmware(n));
 }
@@ -6939,10 +6939,10 @@ static void nvme_show_simple_list(nvme_root_t r)
 	nvme_ctrl_t c;
 	nvme_ns_t n;
 
-	printf("%-21s %-20s %-40s %-9s %-26s %-16s %-8s\n",
-	    "Node", "SN", "Model", "Namespace", "Usage", "Format", "FW Rev");
-	printf("%-.21s %-.20s %-.40s %-.9s %-.26s %-.16s %-.8s\n", dash, dash,
-		dash, dash, dash, dash, dash);
+	printf("%-21s %-21s %-20s %-40s %-9s %-26s %-16s %-8s\n",
+	    "Node", "Generic", "SN", "Model", "Namespace", "Usage", "Format", "FW Rev");
+	printf("%-.21s %-.21s %-.20s %-.40s %-.9s %-.26s %-.16s %-.8s\n",
+		dash, dash, dash, dash, dash, dash, dash, dash);
 
 	nvme_for_each_host(r, h) {
 		nvme_for_each_subsystem(h, s) {


nvme-print.c
```diff
@@ -6346,8 +6346,8 @@ void nvme_show_list_item(nvme_ns_t n)
 	snprintf(format, sizeof(format), "%3.0f %2sB + %2d B", (double)lba,
 		l_suffix, nvme_ns_get_meta_size(n));
 
-	printf("%-21s %-20s %-40s %-9d %-26s %-16s %-8s\n",
-		nvme_ns_get_name(n), nvme_ns_get_serial(n),
+	printf("%-21s %-21s %-20s %-40s %-9d %-26s %-16s %-8s\n",
+		nvme_ns_get_name(n), nvme_ns_get_generic_name(n), nvme_ns_get_serial(n),
```
hreinecke, line 6350: Well, that will change the output as a new field is added in between existing ones. Which will break any parser there is.
I'm not that happy about it; our system management folks have the habit of parsing outputs, and do get terribly upset by these things.
Keith?

nvme-print.c
```diff
@@ -6359,10 +6359,10 @@ static void nvme_show_simple_list(nvme_root_t r)
...
 	nvme_ns_t n;
 
-	printf("%-21s %-20s %-40s %-9s %-26s %-16s %-8s\n",
-	    "Node", "SN", "Model", "Namespace", "Usage", "Format", "FW Rev");
-	printf("%-.21s %-.20s %-.40s %-.9s %-.26s %-.16s %-.8s\n", dash, dash,
-		dash, dash, dash, dash, dash);
+	printf("%-21s %-21s %-20s %-40s %-9s %-26s %-16s %-8s\n",
+	    "Node", "Generic", "SN", "Model", "Namespace", "Usage", "Format", "FW Rev");
+	printf("%-.21s %-.21s %-.20s %-.40s %-.9s %-.26s %-.16s %-.8s\n",
+		dash, dash, dash, dash, dash, dash, dash, dash);
```
hreinecke, line 6365: Same story here. I would be okay with adding new fields at the end of the line, but slap-bang in the middle ...

nvme-print.c
minwooim: Agreed on that point. but I think namespace generic device pretty much deserves to print out here along with other fields. Just like the --verbose mode, can we print it out on normal mode also ?  It would be great to print out where the I/O should   go through  (blkdev or chardev) for namespaces.

minwooim: And of course, it can be parsed well with a fixed format, but we have `--output-format=json` for the parsers.
(I missed to add generic device to json format, but we have this format for parsers)

nvme-print.c
javigon: @hreinecke Would you be ok with presenting this info if we just move the generic device to the end of the line?

minwooim:
@hreinecke ,

Then, Can we have this generic device printed in the last column in the `nvme list` ?
Adding chardev list below the existing list could be a duplication.

igaw:
Well, I've ported this  from the monolithic branch (see a296e439f61f ("nvme-print: print generic ns chardev in verbose mode")). The original commit is 7169d78c9ccc ("nvme-print: print generic ns chardev in verbose mode") and got added in Sept 2021. It got released as v1.16 and so far no seems to complain. 

And we I would argue that screen scraping is just wrong for this tool as we have the 'nvme list -o json' output format which should be for parsing.

minwooim:
@igaw ,

This PR was to print out generic device in `nvme list` simple list command, not only just for `verbose` mode which had already been merged in monolithic branch as you mentioned. Can we have generic device printed in normal mode ?

minwooim:
> And we I would argue that screen scraping is just wrong for this tool as we have the 'nvme list -o json' output format which should be for parsing.

And I agree on this.

igaw:
True, I got that wrong. Can you explain why the generic name is so interesting that it needs to pop up in 'nvme list'?

minwooim:
@igaw ,

Sorry for the late response. I've been got into somewhere else.
The reason why generic device should be printed in the normal mode of `nvme list` is that 
when block device initialization fails from the kernel side during initialization due to some reasons
like unsupported PI, zone size is not power of 2, and something like these, use cannot see
I/O channel (generic device) from `nvme list` and may think that no NVMe device at all in the system.

So, generic device is a I/O channel just like the block device, so I think it sould be printed out in the normal
mode also.

What do you think?

igaw:
Thanks for explaining. Sounds reasonable to me, though I agree with @hreinecke it think we should at it to the end of the line as even humans will complain if we change the well established format of 'nvme list'. But then we are releasing an major version, if we are going to change the output format it would be now. I haven't really made up my mind. @keithbusch, @hreinecke any preferences?

minwooim:
How about this ?

```
Node                  Generic               SN                   Model                                    Namespace Usage                      Format           FW Rev  
--------------------- --------------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
/dev/nvme0n1          /dev/ng0n1        ....
(none)                /dev/ng0n2        ....
```

A situation that /dev/nvme0n2 was failed to initialize from block layer.  If we put the generic list below the previous existing list,
it would be too long for machines that have hundreds of NVMe devices.

igaw:
The thing I like about your proposal, it makes 'nvme list' and 'nvme list -v' output the same. Though @hreinecke's argument
is that there are many scripts out there parsing the output of 'nvme list'. Any change in the format is likely to be seen a regression, see #1422 for example. 

Personally, I think it is okay to break the 'nvme list' format for the 2.0 release. After all it's an major version update and we should be allowed to break a few things.

igaw:
Just discussed this with Hannes and he is okay with changing it now. The major version update is covering the break in the output format.

igaw:
FYI, just rebased on the current HEAD

PR #1249, opened 2021-11-18
zns: fix report-zone slba offset along with libnvme

minwooim:
nvme_zns_report_zones in libnvme has argument:

int nvme_zns_report_zones(int fd, __u32 nsid, __u64 slba, __u32 timeout,
                          bool extended, enum nvme_zns_report_options opts,
                          bool partial, __u32 data_len, void *data,
                          __u32 *result)

The SLBA is the third argument and nvme-cli should pass the `offset` to
the third not fourth which is timeout.

In case more than one zone_mgmt_recv commands needed, zones will be
reported from the 0 SLBA over and over and over again.

### DIFF
--- a/plugins/zns/zns.c
+++ b/plugins/zns/zns.c
@@ -807,8 +807,8 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
 			log_len = sizeof(struct nvme_zone_report) + ((sizeof(struct nvme_zns_desc) * nr_zones_chunks) + (nr_zones_chunks * zdes));
 		}
 
-		err = nvme_zns_report_zones(fd, cfg.namespace_id,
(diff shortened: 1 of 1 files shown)


igaw:
I think it would make sense to stream line those zns functions with the rest of the API. That is the second last argument is the timeout and the last one is results. see Hannes' fixes https://github.com/linux-nvme/libnvme/commit/38d5f958a9c557499060f326d68fca47b8da362d

hreinecke:
Yeah, that was my thought, too.

hreinecke:
I have created a new PR #1251 to sanitize the zns calling convention, such that they fall in line with the other commands and have the 'timeout' as second-to-last argument, and 'result' as the last argument.
That should take care of this issue.

PR #1253, opened 2021-11-22
Fix/build errors

sc108-lee:


Makefile
```diff
@@ -52,6 +52,13 @@ ifeq ($(LIBOPENSSL),0)
 	override LIB_DEPENDS += openssl
 endif
 
+GLIBC_MAJOR_VERSION := $(shell ldd --version | awk '/ldd/{print $NF}' | cut -f1 -d.)
+GLIBC_MINOR_VERSION := $(shell ldd --version | awk '/ldd/{print $NF}' | cut -f2 -d.)
+GLIBC_GE_2_25 := $(shell [ $(GLIBC_MAJOR_VERSION) -gt 2 -o \( $(GLIBC_MAJOR_VERSION) -eq 2 -a $(GLIBC_MINOR_VERSION) -ge 25 \) ] && echo true)
+ifeq ($(RH_GT_5_3),true)
+	CFLAGS += -DGLIBC_GE_2_25
```
hreinecke, line 59: Whoa. Seems pretty RH specific, no?
And as such we probably don't need a 'glibc' define, but rather an 'HAVE_SYS_RANDOM'.
Can you use that instead of 'GLIBC_GE_2_25'?

nvme.c
```diff
@@ -6757,7 +6765,8 @@ static int gen_dhchap_key(int argc, char **argv, struct command *command, struct
 	if (!raw_secret)
 		return -ENOMEM;
 	if (!cfg.secret) {
-		if (getrandom(raw_secret, cfg.key_len, GRND_NONBLOCK) < 0)
+		// if (getrandom(raw_secret, cfg.key_len, GRND_NONBLOCK) < 0)
+		if (syscall(SYS_getrandom, raw_secret, cfg.key_len, GRND_NONBLOCK) < 0)
```
hreinecke, line 6769: Hmm? You first introduce a wrapper, only to call the syscall directly?
Please, no. Use the wrapper.

hreinecke (changes requested):

Makefile
sc108-lee: oh... my mistake, Thanks for checking, It didnt saved well

nvme.c
sc108-lee: oh... my mistake, Thanks for checking, It didnt saved well

Makefile
```diff
@@ -52,6 +52,13 @@ ifeq ($(LIBOPENSSL),0)
 	override LIB_DEPENDS += openssl
 endif
 
+GLIBC_MAJOR_VERSION := $(shell ldd --version | awk '/ldd/{print $NF}' | cut -f1 -d.)
+GLIBC_MINOR_VERSION := $(shell ldd --version | awk '/ldd/{print $NF}' | cut -f2 -d.)
+GLIBC_GE_2_25 := $(shell [ $(GLIBC_MAJOR_VERSION) -gt 2 -o \( $(GLIBC_MAJOR_VERSION) -eq 2 -a $(GLIBC_MINOR_VERSION) -ge 25 \) ] && echo true)
+ifeq ($(GLIBC_GE_2_25),true)
+	CFLAGS += -DHAVE_SYS_RANDOM
```
hreinecke, line 59: Sorry to pester you again, but wouldn't it be better to directly check for <sys/random.h> instead of checking the various glibc versions (which will fail anyway for anyone _not_ using glibc).
Can you update ccan/tools/configurator/configurator.c to check for sys/random.h, about time that it does something useful.
And we'll need a similar test for meson ...

nvme.c
```diff
@@ -55,7 +55,15 @@
...
 #include <sys/stat.h>
-#include <sys/random.h>
+
+#ifdef HAVE_SYS_RANDOM
+	#include <sys/random.h>
+#else
+	#include <sys/syscall.h>
+	#include <linux/random.h>
+	#define getrandom(buf, buflen, flags) \
+		syscall(SYS_getrandom, buf, buflen, flags)
```
hreinecke, line 65: Hmm. Caling a syscall directly is always a bit awkward. Wouldn't it be better to read from /dev/urandom (or /dev/random if /dev/urandom doesn't exist)?
Then we would avoid having to use the syscall directly (which always has a chance of failing as there is no guarantee that 'SYS_getrandom' is defined) and would at least _attempt_ to be portable.

hreinecke (changes requested):

Makefile
sc108-lee: nice suggestion added both, ccan & meson :)

nvme.c
sc108-lee: I added reading /dev/urandom & random

sc108-lee:
rebased on master

nvme.c
```diff
@@ -159,6 +162,39 @@ void *nvme_alloc(size_t len, bool *huge)
...
 
+static int getrandom_bytes(void *buf, size_t buflen)
+{
+#if HAVE_SYS_RANDOM
+	return getrandom(buf, buflen, GRND_NONBLOCK);
+#else
+	int fd = open("/dev/urandom", O_RDONLY);
+	int result = 0, read_len;
+	if (fd < 0)
+	{
```
igaw, line 173: The brakets placement is not really following the normal coding style.

nvme.c
```diff
@@ -159,6 +162,39 @@ void *nvme_alloc(size_t len, bool *huge)
...
+		fd = open("/dev/random", O_RDONLY);
```
igaw, line 174: I thought '/dev/random' is a legacy interface and you don't need to read from it on Linux. So no need
for supporting the read from blocking interface such as random (IIRC it doesn't block either on modern kernels
either.)

https://www.2uo.de/myths-about-urandom/
https://unix.stackexchange.com/questions/324209/when-to-use-dev-random-vs-dev-urandom

meson.build
```diff
@@ -50,7 +50,7 @@ conf.set('LIBHUGETLBFS', libhugetlbfs.found(), description: 'Is libhugetlbfs req
 libz_dep = dependency('zlib', required: true)
 
 # Check for open-ssl availability
-libopenssl = dependency('openssl', required: false)
+libopenssl = dependency('openssl', version: '>=1.1.0', required: false)
```
igaw, line 53: Move this to the next commit.

igaw (changes requested):
Almost there! Thanks for fixing it!

sc108-lee:
@igaw thanks for the comments, updated & rebased :)

nvme.c
```diff
@@ -159,6 +162,22 @@ void *nvme_alloc(size_t len, bool *huge)
...
+	return getrandom(buf, buflen, GRND_NONBLOCK);
+#else
+	int result = 0;
+	int fd = open("/dev/urandom", O_RDONLY);
+	if (fd < 0)
+		return fd;
+	else
+		result = read(fd, buf, buflen);
+	close(fd);
+	return result;
```
igaw, line 177: If you don't mind I would change the above to something like this here:
```
	int fd, err = 0;

	fd = open("/dev/urandom", O_RDONLY);
	if (fd < 0)
		return fd;
	if (read(fd, buf, buflen) < 0)
		err = errno;
	close(fd);
	errno = err;
	return err? -1 : 0;
```
The close will overwrite the errno from the read. Hence you need to cache it.

nvme.c
```diff
@@ -6757,7 +6776,7 @@ static int gen_dhchap_key(int argc, char **argv, struct command *command, struct
 	if (!raw_secret)
 		return -ENOMEM;
 	if (!cfg.secret) {
-		if (getrandom(raw_secret, cfg.key_len, GRND_NONBLOCK) < 0)
+		if (getrandom_bytes(raw_secret, cfg.key_len) < 0)
 			return errno;
```
igaw, line 6780: I think we need to do a 'return -errno' here.

igaw (commented):
If you agree with my review I'll update your patch and merge it. Let me know what you think.

nvme.c
sc108-lee: I would like to remain my code,
it should be works same with exist sys/random.h , so return value should be read length.

getrandom 
 On success, getrandom() returns the number of bytes
 On error, -1 is returned, and errno is set to indicate the error.

read
 On success, the number of bytes read is returned 
 On error, -1 is returned, and errno is set to indicate the error.

nvme.c
sc108-lee: hum I didnt notice that, I think you are right, since we are returning like above ( return -ENOMEM; ) for errors

nvme.c
igaw: Good point. But let me still add the errno caching.

igaw: In this case let's get also the return type correct :)

```
static ssize_t getrandom_bytes(void *buf, size_t buflen)
{
#if HAVE_SYS_RANDOM
	return getrandom(buf, buflen, GRND_NONBLOCK);
#else
	ssize_t result;
	int fd, err = 0;

	fd = open("/dev/urandom", O_RDONLY);
	if (fd < 0)
		return fd;
	result = read(fd, buf, buflen);
	if (result < 0)
		err = errno;
	close(fd);
	errno = err;
	return result;
#endif
}```

Would that work for you?

sc108-lee: @igaw great :)

igaw:
I've merge the change in with the command line.

PR #1261, opened 2021-11-26
Makefile: Add missing define (LIBJSONC_14)

sc108-lee:


hreinecke:
Looks good.

PR #1264, opened 2021-12-01
Pel fix nvme cli

wxiong38:


igaw:
Merged via command line after rebase to the current master. Also updated the libnvme ref accordingly. While at it I took the liberty to set the author correctly.

Issue #140, opened 2016-11-16
NVME List Output JSON Segmentation Fault


Teumer:
environment
1. nvme-cli version 1.0
2. ubuntu 14.04.5 kernel 4.1.10

steps to reproduce
1. make && make install
2. nvme list --output-format=json
3. returns "Segmentation fault"
> dmesg
[ 2947.885599] nvme[26026]: segfault at 0 ip 00007f74d1274ebc sp 00007ffd8ffd6820 error 4 in libc-2.19.so[7f74d1206000+1ba000]
[ 2974.819305] nvme[27464]: segfault at 0 ip 00007f31ac4b9ebc sp 00007ffd5c36a410 error 4 in libc-2.19.so[7f31ac44b000+1ba000]
[ 3054.771953] nvme[29480]: segfault at 0 ip 00007f1df528debc sp 00007fff24d03850 error 4 in libc-2.19.so[7f1df521f000+1ba000]

Note other commands formatted w/ json output work w/o issue (only nvme list has the fault)
1. nvme id-ctrl /dev/nvme0 --output-format=json
2. nvme fw-log /dev/nvme0n1 -o json

keithbusch:
hm, this is testing successfully here. How many nvme controllers do you have?

Teumer:
6 total controllers

keithbusch:
Okay, 6 appears to be the tipping point. Will have a fix by today.

keithbusch:
oh, 6 was just a red herring on my machine. the real problem is the product name pulled from the models file, which may not be installed on every environment.

keithbusch:
Fix pushed.

PR #1266, opened 2021-12-08
MAINT: Fix clang-format indentation errors

ikegami-t:
Corrected by the clang-format tool with the linux kernel style.
Also .clang-format file desribed by the following page.
  https://clang.llvm.org/docs/ClangFormatStyleOptions.html#examples

Note: The tool executed as below command.
  find . -iname "*.h" -o -iname "*.c" | xargs clang-format -i


ikegami-t:
Now trying to fix the build errors by the libnvme pull request change #140.

igaw:
The final result is certainly consistent but it uglifies a lot of code, IMO. Also it ruins 'git blame' considerable.

@hreinecke do you think we should do a massive whitespace cleanup like this?

ikegami-t:
> The final result is certainly consistent but it uglifies a lot of code, IMO. Also it ruins 'git blame' considerable.
> 
> @hreinecke do you think we should do a massive whitespace cleanup like this?

Thanks for your comment. Let me reply to you as below.
1. It looks the issue can be resolved by the git blame -w option below.
```
-w
Ignore whitespace when comparing the parent’s version and the child’s to find where the lines came from.
```
2. The .clang-format used in the linux kernel seems better to reuse than the clang format style example in the PR.
  <https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/.clang-format>
3. Seems better to be checked by the CI build to prevnet the coding sytle erorr after the format at first.

By the way the CI builds error can be resovled by adding the option below into the .clang-format file.
```
SortIncludes:false 
```

igaw:
While the 'git blame -w' might work (haven't tried it out) you completely ignored my first argument. The resulting code looks pretty bad. I am fine with having coding style fixes but not in this way. Sorry.

PR #1268, opened 2021-12-09
Fixes

sc108-lee:
zns: Bug fix(wrong api call from changed_zone_list)
zns: remove unnecessary label, fix wrong goto

### DIFF
--- a/util/json.h
+++ b/util/json.h
@@ -54,6 +54,7 @@ int json_object_add_value_type(struct json_object *obj, const char *name, int ty
 	json_object_add_value_type((obj), name, JSON_TYPE_INTEGER, (long long) (val))
 #define json_object_add_value_uint(obj, name, val) \
 	json_object_add_value_type((obj), name, JSON_TYPE_UINT, (unsigned long long) (val))
+#define json_object_add_value_uint64 json_object_add_value_uint
 #define json_object_add_value_float(obj, name, val) \
 	json_object_add_value_type((obj), name, JSON_TYPE_FLOAT, (val))
 #define json_object_add_value_string(obj, name, val) \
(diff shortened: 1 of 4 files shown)


sc108-lee:
[zns: Bug fix(wrong api call from changed_zone_list) ]
it's required update libnvme to be passed for meson build 
(https://github.com/linux-nvme/libnvme/pull/142)

util/json.h
```diff
@@ -54,6 +54,7 @@ int json_object_add_value_type(struct json_object *obj, const char *name, int ty
 	json_object_add_value_type((obj), name, JSON_TYPE_INTEGER, (long long) (val))
 #define json_object_add_value_uint(obj, name, val) \
 	json_object_add_value_type((obj), name, JSON_TYPE_UINT, (unsigned long long) (val))
+#define json_object_add_value_uint64 json_object_add_value_uint
```
igaw, line 57: We wanted set hard dependency on json-c and libuuid. Do you still need the built-in json-c support? 
Anyway, I don't mind adding it now but we really wanted to drop the built-in json parser.

sc108-lee: when make build is replaced with meson (as your mention on #1272)
meson bring that package and build with static, then there is no problem to be removed.
since static build with json-c higher version, older system user can use with same build as well.

igaw: Yes, that was one of my main motivation to get the meson build working. It really addresses the dependency issues at least when developing or when it is shipped for an really old distribution.

Though so far meson's wrapdb ships only json-c at this point. I was playing with the idea to at least add also libuuid. This shouldn't be too difficult.

I just realized meson ships openssl, so we can really try to have all external decencies covered by wrapdb which would allow to build a completely static build of nvme-cli. I am sure this is interesting for small devices and even cloud environments.

PR #1277, opened 2021-12-13
nvme: Add Command Set Idendifier and Offset Type to get_log

igaw:
Add Command Set Identifier (CSI) and Offset Type (OT) support to
get_log as per the Spec 2.0.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1541,6 +1541,8 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 	const char *rae = "retain an asynchronous event";
 	const char *raw = "output in raw format";
 	const char *uuid_index = "UUID index";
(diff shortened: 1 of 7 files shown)


nvme.c
```diff
@@ -1553,6 +1555,8 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 		__u64 lpo;
 		__u8  lsp;
 		__u8  uuid_index;
+		__u8  csi;
+		__u8  ot;
```
igaw, line 1559: Maybe this should just be an int as 307142b ("nvme: add get log page 2.0 spec fields") and define 'ot' just as bool, the same way we do for 'rae'

igaw: I changed the type of the 'ol' to bool.

PR #1280, opened 2021-12-14
nvme-topology: fix memory leaks

maurizio-lombardi:
Fix memory leaks in nvme_logical_block_size_from_ns_char()

Free the allocated strings before returning from the function.

### DIFF
--- a/nvme-topology.c
+++ b/nvme-topology.c
@@ -686,10 +686,13 @@ int nvme_logical_block_size_from_ns_char(const char *dev)
 		return -EINVAL;
 
 	s = nvme_get_ctrl_attr(path, "logical_block_size");
+	free(path);
 	if (!s)
 		return -EINVAL;
 
-	return atoi(s);
+	ret = atoi(s);
+	free(s);
+	return ret;
 }
 
 void *mmap_registers(const char *dev)


nvme-topology.c
```diff
@@ -686,10 +686,16 @@ int nvme_logical_block_size_from_ns_char(const char *dev)
 		return -EINVAL;
 
 	s = nvme_get_ctrl_attr(path, "logical_block_size");
-	if (!s)
+	if (!s) {
+		free(path);
```
igaw, line 690: The free could be placed before the 'if, which avoids to add the free twice.

maurizio-lombardi: You're right, I updated the PR

maurizio-lombardi:
> Thanks!
> 
> BTW, do we need the same treatment in master?

No, this applies only to the monolithic branch

keithbusch:
Yes, this is a correct fix for this branch only, but I'm not sure it matters. It's just a short-lived, one-shot program: everything allocated is released upon exit, so leaks exist for extremely short times. We do want correctness because we re-use much from here in libraries, daemons, and long-lived applications, but this branch is deprecated.

maurizio-lombardi:
> Yes, this is a correct fix for this branch only, but I'm not sure it matters. It's just a short-lived, one-shot program: everything allocated is released upon exit, so leaks exist for extremely short times. We do want correctness because we re-use much from here in libraries, daemons, and long-lived applications, but this branch is deprecated.

I know, I submitted this patch just to silence a code analyzer that constantly complained about this leak when submitting package updates for RHEL

keithbusch:
Yeah, fair enough.

Issue #1221, opened 2021-11-02
open_exclusive leaks filedescriptor


igaw:
https://github.com/linux-nvme/nvme-cli/blob/07c13a28dfa85374e0debd5ed2a41ff9c9550110/nvme.c#L3588

open_exclusive() calls open for a namespace and right after this function parse_and_open() is called which opens the very same device again. If I am not completely mistaken, this leaks a filedescriptor.

When at it fix also the weird intention for the basename() call.

igaw:
PR merged.

PR #1283, opened 2021-12-15
nvme-cli: fix redundant open call

olase23:
fix open_exclusive leaks filedescriptor #1221

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4117,10 +4117,6 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 		goto ret;
 	}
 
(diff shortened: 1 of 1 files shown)


igaw:
Is this supposed to go on top of our other PR #1215 ?

olase23:
Well, since PR #1215 was not commented for a while i thought i can fix the redundant open() call with this change here.

igaw:
Alright, I think we are almost there with #1215. If you can review my changes on top of your commit, I'll apply #1215 which renders this PR unnecessary.

igaw:
Not needed with PR #1215 anymore.

PR #1284, opened 2021-12-15
Fixes

sc108-lee:
! was for test reason putted, but merged somehow.
lid to string using spec naming.

should we move this lid to string into libnvme? or remaining as current?

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -164,7 +164,7 @@ void *nvme_alloc(size_t len, bool *huge)
 
 static ssize_t getrandom_bytes(void *buf, size_t buflen)
 {
-#if !HAVE_SYS_RANDOM
+#if HAVE_SYS_RANDOM
 	return getrandom(buf, buflen, GRND_NONBLOCK);
 #else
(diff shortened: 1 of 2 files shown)


nvme.c
```diff
@@ -164,7 +164,7 @@ void *nvme_alloc(size_t len, bool *huge)
 
 static ssize_t getrandom_bytes(void *buf, size_t buflen)
 {
-#if !HAVE_SYS_RANDOM
```
igaw, line 167: Ooh, my bad. This was from my testing.

sc108-lee:
@igaw how do you think that "nvme_log_to_string(__u8 lid)" move to libnvme?

nvme.c
sc108-lee: since we find it fast, doesn't matter :)

igaw:
Personally I think these print function should part of the library, so not just nvme_log_to_string(). But this is major API surgery I don't think we should do now. This is something for 3.x IMO :)

PR #1285, opened 2021-12-15
build: wrapdb for OpenSSL and zlib, detect libhugetblfs

igaw:
meson wrapdb has support for OpenSSL and zlib. Let's make use of it. This allows building/deploying nvme-cli on systems which have either outdated or no support for those libraries.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -47,7 +47,7 @@ libhugetlbfs = dependency('hugetlbfs', required: false)
 conf.set('LIBHUGETLBFS', libhugetlbfs.found(), description: 'Is libhugetlbfs required?')
 
 # Check for zlib availability
-libz_dep = dependency('zlib', required: true)
+libz_dep = dependency('zlib', fallback : ['zlib', 'zlib_dep'])
 
 # Check for OpenSSL availability
 want_openssl = get_option('openssl')
--- a/meson.build
+++ b/meson.build
@@ -42,9 +42,16 @@ if json_c.found()
   conf.set('LIBJSONC_14', json_c.version().version_compare('>=0.14'), description: 'Is json-c at least 0.14?')
 endif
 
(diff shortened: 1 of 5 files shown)


igaw:
libhugetblfs was not correctly detected. Added a fix for this. 
While at it also updated the help text in config.h.

meson.build
```diff
@@ -33,25 +33,41 @@ libnvme_dep = dependency('libnvme', fallback : ['libnvme', 'libnvme_dep'])
...
-libz_dep = dependency('zlib', required: true)
-
-# Check for open-ssl availability
-libopenssl = dependency('openssl', version: '>=1.1.0', required: false)
-conf.set('OPENSSL', libopenssl.found(), description: 'Is open-ssl required?')
+libz_dep = dependency('zlib', fallback : ['zlib', 'zlib_dep'])
+
+# Check for OpenSSL availability
+want_openssl = get_option('openssl')
+if want_openssl != 'false'
```
eli-schwartz, line 61: I would have suggested using a Feature option: https://mesonbuild.com/Build-options.html#features

You can simply pass a feature option directly to the `required: ` kwarg and it automatically handles:
- disabled? return a not-found dependency, and log to the effect of "skipped dependency openssl because feature openssl is disabled"
- enabled? This is equivalent to your code for "true"
- auto? This is equivalent to your code for "auto"


The neat thing is that by doing this you get:
- better informational logging
- a few less lines, no need to if/else anything and openssl_dep.found() returns `false` when the option is disabled.

igaw: Thanks for the tip! I suspected that I was doing something wrong. The Feature option is very handy! 

See #1291

eli-schwartz: Great, happy to help. ;) That change LGTM.

PR #1286, opened 2021-12-17
Add ctrl metadata and ns metadata

igaw:


igaw:
This depens on https://github.com/linux-nvme/libnvme/pull/146

PR #1292, opened 2021-12-21
memblaze-nvme: remove power loss protection from smart-log-add

jinhua2021:
Signed-off-by: jinhua2021 <jinhua.huang@memblaze.com>

### DIFF
--- a/plugins/memblaze/memblaze-nvme.c
+++ b/plugins/memblaze/memblaze-nvme.c
@@ -78,9 +78,14 @@ static int compare_fw_version(const char *fw1, const char *fw2)
 int getlogpage_format_type(char *model_name)
 {
     int logpage_format_type = INTEL_FORMAT;
-    const char *boundary_model_name = "P5920"; // Use INTEL_FORMAT from Raisin P5920.
-    if (strncmp(model_name, boundary_model_name, strlen(boundary_model_name)) < 0) {
-        logpage_format_type = MEMBLAZE_FORMAT;
+    const char *boundary_model_name1 = "P"; // MEMBLAZE P7936DT0640M00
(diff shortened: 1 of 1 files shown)


plugins/memblaze/memblaze-nvme.c
```diff
@@ -220,8 +225,8 @@ static void show_memblaze_smart_log_new(struct nvme_memblaze_smart_log *s,
     printf("%-32s: %3d%%       %s%u%s%u%s%u\n", STR17_01,  *nm, STR17_03, *raw,
         STR17_04, *(raw+2), STR17_05, *(raw+4));
     /* 18 RAISIN_SI_VD_POWER_LOSS_PROTECTION */
-    get_memblaze_new_smart_info(smart, RAISIN_SI_VD_POWER_LOSS_PROTECTION, nm, raw);
-    printf("%-32s: %3d%%       %"PRIu64"\n", STR18_01, *nm, int48_to_long(raw));
+    // get_memblaze_new_smart_info(smart, RAISIN_SI_VD_POWER_LOSS_PROTECTION, nm, raw);
+    // printf("%-32s: %3d%%       %"PRIu64"\n", STR18_01, *nm, int48_to_long(raw));
```
igaw, line 229: I suggest you remove the code instead just commenting it out. 

Anyway, looking through the memblaze-nvme.c plugin I noticed that is breaking coding style rules in a lot of places.
The result is that the code is harder to read and to maintain IMO. Would you mind to cleanup the file?

jinhua2021: OK, I will delete these useless code.
For the coding style, there are much work to cleanup. We will create a new pr for that.

igaw: Excellent, sounds like a plan! Let's do this.

Issue #1244, opened 2021-11-17
Port missing commits from nvme-cli-monolithic


igaw:
When we merged the nvme-cli-monolithic branch into master in #1223, a few commits didn't make it. Track them here:

Changes which are missing in master:

- [x] b8a403b ("Add NVMe MI Features: Controller Metadata (0x7E) and Namespace Metadata (0x7F).")
Needs to be splitted between nvme-cli and libnvme.
Rewrite needed
include 2ae875d ("fix whitespace damage")
- [x] 0ad49ad ("Print device name on connect")
Rewrite needed
- [x] bace574 ("nvme-topology: fix controller check in scan_subsystem()")
Rewrite needed
- [x] 307142b ("nvme: add get log page 2.0 spec fields")
Offset type API is missing in libnvme's nvme_get_log
- [x] df159a7 ("nvme: add the status codes from 2.0 spec")
libnvme has the bits but the nvme.c print needs to updated accordingly
- [x] fa97500 ("Add Timeout Parameter for Zone Management Commands")
Conflicts plugins/zns/zns.c
Changes we might need in master (either nvme-cli or libnvme)"
- [x] 59e7477 ("nvme-print: fix 'nvme list' segfault if controller is unavailable")
not clear if we need it (does not apply)
- [x]  fc28cdc ("nvme: extend existing id ctrl parameters as per 2.0")
nvme_show_id_ctrl_ofcs() has differs a bit, not clear why.
- [x] daa6f93 ("nvme: support ns generic device to submit_io")
~~no users, propably should go into libnvme~~
- ~~a58556cc3022 ("connect-all: add -m/--matching option")
and edf0998f5a66 ("fabrics: skip connect if transport type doesn't match")
b2509be70946 ("fabrics: fix infinite loop is discovery recursion contains a loop") ->[ 02489f250655 ("fabrics: fix endless loop in connect-all for NVME_NQN_CURR")](https://github.com/linux-nvme/libnvme/commit/02489f250655)~~

sc108-lee:
fc28cdc ofcs different reason is from missing commit 782086cc4fd590f1df171c927ebf9a2debc0a052 
so I manually recover the commits

sc108-lee:
I've reviewed df159a7 ("nvme: add the status codes from 2.0 spec")
no need any change, it's already applied well.
print also works fine via nvme_status_to_string in libnvme/src/nvme/utils.c

igaw:
[...]

PR #1294, opened 2021-12-21
Fabrics matching

igaw:
Missing commits from the monolithic branch

See also #1244


igaw:
Alright, so you say we don't need these from the monolithic branch as libnvme needs handle this?

igaw:
Closing this PR.

Issue #1295, opened 2021-12-21
error: ‘HMAC_CTX_free’ is deprecated: Since OpenSSL 3.0


johnmeneghini:
Looks like the recent OPENSSL changes were developed with a deprecated Open SSL library.

```
rhel-storage-08:nvme-cli(master) > make nvme
NVME_VERSION = 1.14.331.gf550
    CC ccan/tools/configurator/configurator
ccan/tools/configurator/configurator > ccan/config.h
    CC nvme.o
nvme.c: In function ‘gen_dhchap_key’:
nvme.c:6814:17: error: ‘HMAC_CTX_new’ is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
 6814 |                 HMAC_CTX *hmac_ctx = HMAC_CTX_new();
      |                 ^~~~~~~~
In file included from nvme.c:50:
/usr/include/openssl/hmac.h:33:33: note: declared here
   33 | OSSL_DEPRECATEDIN_3_0 HMAC_CTX *HMAC_CTX_new(void);
      |                                 ^~~~~~~~~~~~
nvme.c:6818:17: error: ‘ENGINE_load_builtin_engines’ is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
 6818 |                 ENGINE_load_builtin_engines();
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from nvme.c:48:
/usr/include/openssl/engine.h:358:28: note: declared here
  358 | OSSL_DEPRECATEDIN_3_0 void ENGINE_load_builtin_engines(void);
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
nvme.c:6819:17: error: ‘ENGINE_register_all_complete’ is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
 6819 |                 ENGINE_register_all_complete();
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from nvme.c:48:
/usr/include/openssl/engine.h:415:27: note: declared here
  415 | OSSL_DEPRECATEDIN_3_0 int ENGINE_register_all_complete(void);
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
nvme.c:6821:17: error: ‘HMAC_Init_ex’ is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
[...]

PR #1296, opened 2021-12-21
Fix openssl library dependencies

johnmeneghini:
The new openssl code has a dependency on  legacy APIs from version 1.1.1 or less.

See: https://github.com/linux-nvme/nvme-cli/issues/1295

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ HAVE_SYSTEMD = $(shell pkg-config --exists libsystemd  --atleast-version=242; ec
 LIBJSONC_14 = $(shell pkg-config --atleast-version=0.14 json-c; echo $$?)
 LIBJSONC = $(shell pkg-config json-c; echo $$?)
 LIBZ = $(shell $(LD) -o /dev/null -lz >/dev/null 2>&1; echo $$?)
-LIBOPENSSL = $(shell pkg-config --atleast-version=1.1.0 openssl; echo $$?)
+LIBOPENSSL = $(shell pkg-config --max-version=1.1.0 openssl; echo $$?)
 NVME = nvme
 INSTALL ?= install
 DESTDIR =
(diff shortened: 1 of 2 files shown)


johnmeneghini:
Note that this patch really only supports version 1.1.1k.  Don't ask my why... I think pkg-config is broken.

```
jmeneghi:nvme-cli(fix_hmac) > pkg-config --version
1.4.2
jmeneghi:nvme-cli(fix_hmac) > pkg-config --modversion openssl
1.1.1k
jmeneghi:nvme-cli(fix_hmac) > pkg-config --max-version=1.0.0 openssl
jmeneghi:nvme-cli(fix_hmac) > echo $?
1
jmeneghi:nvme-cli(fix_hmac) > pkg-config --exact-version=1.0.0 openssl
jmeneghi:nvme-cli(fix_hmac) > echo $?
1
jmeneghi:nvme-cli(fix_hmac) > pkg-config --exact-version=3.0.0 openssl
jmeneghi:nvme-cli(fix_hmac) > echo $?
1
jmeneghi:nvme-cli(fix_hmac) > pkg-config --exact-version=1.1.1 openssl
jmeneghi:nvme-cli(fix_hmac) > echo $?
1
jmeneghi:nvme-cli(fix_hmac) > pkg-config --exact-version=1.1.1k openssl
jmeneghi:nvme-cli(fix_hmac) > echo $?
0
```
Looks like I should use --exact-version instead.

igaw:
> Note that this patch really only supports version 1.1.1k. Don't ask my why... I think pkg-config is broken.

Would it possible to change our code so it works with either version of the libraries? Limiting to the 1.x version of the library is asking for troubles in the long run I think. I'll check the docs if there is a recommended way to handle this situation.

(and yes all versioning tracking tools are somehow broken...)

igaw:
Deprecated API: https://www.openssl.org/docs/manmaster/man3/HMAC_Init.html
Replacement API: https://www.openssl.org/docs/manmaster/man3/EVP_MAC_CTX_new.html

The new replacement API doesn't seem to be present in the 1.x library. Our only code snipped depending on these function is https://github.com/linux-nvme/nvme-cli/blob/f5508e03882c332f3b3f40cb75d7eafd9d014a4e/nvme.c#L6813
I think it would be better to add just a second implementation for the 3.x API.

igaw:
Let's a support for OpenSSL instead. see #1297.

PR #1300, opened 2021-12-24
Meson fixups

eli-schwartz:
Fix some errors revealed when trying to build using https://sr.ht/~lattis/muon


igaw:
Thanks a lot! Highly appreciated!

PR #1305, opened 2022-01-07
nvme-print: modify flbas rsvd bits to be used

sc108-lee:
flbas Bits 6:5 indicate the most significant 2 bits of
the Format Index of the supported LBA Format indicated
in this data structure that was used to format the namespace


sc108-lee:
will update with https://github.com/linux-nvme/libnvme/pull/164

PR #1307, opened 2022-01-07
nvme: modify flbas rsvd bits to be used

sc108-lee:
flbas Bits 6:5 indicate the most significant 2 bits of
the Format Index of the supported LBA Format indicated
in this data structure that was used to format the namespace

using libnvme util function nvme_id_ns_flbas_to_lbaf_inuse
to extract currently used lba index

linux-nvme/libnvme#164


igaw:
Can you update this so it matches with the version is now in libnvme? Thanks!

sc108-lee:
@igaw  rebased and updated :)

igaw:
Thanks!

PR #1312, opened 2022-01-10
For upstream/tp4095

sc108-lee:
Based on libnvme PR https://github.com/linux-nvme/libnvme/pull/166

I was add these command into original id-ns, zns id-ns ...
But this command does not need NSID.

So I split this command as a new command.
if have any opinion for this, please let me know :)

### DIFF
--- a/Documentation/nvme-id-ns-lba-format.1
+++ b/Documentation/nvme-id-ns-lba-format.1
@@ -1,0 +1,172 @@
+'\" t
+.\"     Title: nvme-id-ns-lba-format
+.\"    Author: [FIXME: author] [see http://docbook.sf.net/el/author]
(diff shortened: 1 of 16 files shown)


igaw:
I think the CI build is able to build against the matching libnvme commit. You just need to specify the matching ref in subprojects/libnvme.wrap.

igaw:
@MatiasBjorling you want probably to review this first :)

sc108-lee:
rebased :)

sc108-lee:
rebased and added one commit
https://github.com/linux-nvme/nvme-cli/pull/1312/commits/30703696f0440ab146503d98c2010ac964053bfb

igaw:
Just reordered the commits.

PR #1314, opened 2022-01-11
nvme: Adapt to API rework

igaw:
libnvme changed the calling API, which uses an an argument structure
instead of passing all arguments one by one. Update all call sites
accordingly.


hreinecke:
Looks good.

Reviewed-by: Hannes Reinecke <hare@suse.de>

igaw:
I've done some testing and all looks so far good. I think we should go ahead with this and sort any problems in the tree.

PR #1315, opened 2022-01-13
nvme: Move timeout and result argument

igaw:
libnvme reorganized the layout of the struct with the goal to have all
holes filled. Along the hole filling also the common arguments have
been grouped together. Hence move the timeout and result argument to
the common area of the struct.

depends on https://github.com/linux-nvme/libnvme/issues/169

### DIFF
--- a/nvme-rpmb.c
+++ b/nvme-rpmb.c
@@ -272,6 +272,8 @@ static int send_rpmb_req(int fd, unsigned char tgt, int size,
 	struct nvme_security_send_args args = {
 		.args_size	= sizeof(args),
 		.fd		= fd,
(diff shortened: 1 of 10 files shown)


hreinecke:
Do we really need this? I thought that with using named initializers it becomes moot...

igaw:
The compiler was really complaining in the ioctl.h header file (the inline functions) but it doesn't give a warning when compiling the *.c files in libnvme. Maybe this was just a problem for the inline functions...

igaw:
Nope, doesn't seem to be a problem. No warning. Let's ditch this PR .

PR #1318, opened 2022-01-14
Update to libnvme host_traddr changes

hreinecke:
Update fabrics.c for the recent libnvme host_traddr changes.


igaw:
Tested-by: Daniel Wagner <dwagner@suse.de>

PR #1329, opened 2022-01-20
list-ctrl should support CNS values of 12h and 13h

samiWaheed:
list-ctrl should support CNS values of 12h and 13h.  The support for CNS=13h was dropped from 2.0 version


igaw:
Thanks!

PR #1332, opened 2022-01-21
nvme-print: check namespace in nvme_ctrl_for_each_path()

hreinecke:
We should check if the namespace is available when scanning all
paths in nvme_ctrl_for_each_path(); a controller might be resetting
and might not be able to access the namespaces here.


igaw:
Yes this fixes my issue.

Issue #1323, opened 2022-01-19
build warning in nvme.c:get_telemetry_log


igaw:
For debug builds ` meson setup --buildtype=debug --optimization=g .build` with gcc 11.2.1:

```
../nvme.c: In function ‘get_telemetry_log’:
../nvme.c:547:12: warning: ‘total_size’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  547 |         if (err != total_size) {
      |            ^
```

I've starred at this for a while and I think the code is correct. I suspect that gcc is 
getting a bit confused. Maybe this could be refactored also to improve the
readability.

igaw:
The fix is merged. Closing.

PR #1338, opened 2022-01-21
nvme: MAINT: Refactor telemetry-log command for build warning #1323

ikegami-t:


ikegami-t:
Sorry fixed the commit message subject error and also separated additionally the function.

common.h
```diff
@@ -33,4 +33,6 @@ static inline uint64_t mmio_read64(void *addr)
 	return ((uint64_t) high << 32) | low;
 }
 
+#define fallthrough do {} while (0)
```
igaw, line 36: All relative modern [compilers](https://gcc.gnu.org/onlinedocs/gcc/Statement-Attributes.html) know about fallthrough attribute/statement. Hence you should test if the compiler supports it.

The Linux kernel used to do something like this
```
#if __has_attribute(__fallthrough__)
# define fallthrough                    __attribute__((__fallthrough__))
#else
# define fallthrough                    do {} while (0)  /* fallthrough */
#endif
```
but IIRC it was now dropped as all supported gcc/clang versions know about fallthrough

nvme.c
```diff
@@ -407,20 +407,110 @@ static int get_ana_log(int argc, char **argv, struct command *cmd,
 	return nvme_status_to_errno(err, false);
 }
 
-static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct plugin *plugin)
+static int get_telemetry_log_total_size_dalb4(int fd,
+					      struct nvme_telemetry_log *log,
+					      unsigned *total_size)
```
igaw, line 412: Could you also change from `unsigned` to `unsigned int` when touching this code?

nvme.c
```diff
@@ -469,84 +559,19 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd, struct
...
-	}
-
-	err = write(output, (void *)log, total_size);
-	if (err != total_size) {
-		fprintf(stderr, "Failed to flush all data to file!\n");
-		goto close_output;
+	err = get_telemetry_log_total_size(fd, cfg.data_area, log, &total_size);
+	if (total_size) {
+		err = write(output, (void *)log, total_size);
+		if (err != total_size)
```
igaw, line 573: Since err and total_size have two different types, it's probably better to do:
`if (err < 0 || err != total_size)`

igaw (changes requested):

nvme.c
ikegami-t: Fixed as mentioned.

nvme.c
ikegami-t: Fixed this also.

ikegami-t:
Thasnk for your comments. Fixed the code as you mentioned so please review again.

ikegami-t:
Additionally changed the code as below.
1. Changed "unsigned" still remained to "unsigned  int".
2. Changed "long long unsigned int" to "unsigned long long".

igaw:
Looks good. I just splited the type changes out of the initial patch.

igaw:
Thanks!

PR #1339, opened 2022-01-21
fabrics: Stop using nvme_msg from libnvme

igaw:
nvme_msg() is logging function from libnvme. The cli should not use
the logging infrastructure from the library instead it should use it's
own infrastructure. At this point nvme-cli doesn't have one, so just
use the default stderr for all error messages.

This is a preperation step to unexport the nvme_msg() from libnvme.


igaw:
see also 

- https://github.com/linux-nvme/libnvme/issues/187
- https://github.com/linux-nvme/libnvme/issues/168

PR #1345, opened 2022-01-28
Handle libnvme error codes and return proper error code when termiating nvme-cli

### LINKED ISSUE #697: About nvme command return code
image

Hi all, I use the latest nvme-cli version-V1.10.1 I found some different with nvme-cli-V1.5.x
I hope can got the real nvme return code after command by `echo $?` but the V1.10.1 I got the return code is remapping from errno.h.
For this change, Is there any way to get the return code more convenient? Or I only can get it from the parser the stdout message. Thanks~

igaw:
Fixes #697


igaw:
I add some more nvme_strerror(). Currently we do not make use of this feature for anything but the disconnect path.

hreinecke (approved):
Reviewed-by: Hannes Reinecke <hare@suse.de>

PR #1346, opened 2022-01-31
nvme: Adapt to logging changes in libnvme

igaw:
libnvme allows to define which file descriptor should be used for
logging. This is only for the fabric related interface relevant hence
this is attached to the nvme_root_r object.

Use common code for this into a helper which maps the verbose command
line levels to the correct log levels and setup nvme_root_r
accordingly.

See
- https://github.com/linux-nvme/libnvme/pull/202
- https://github.com/linux-nvme/libnvme/pull/207

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -322,8 +322,9 @@ static int __discover(nvme_ctrl_t c, const struct nvme_fabrics_config *defcfg,
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


igaw:
No need to move the include libnvme into the nvme.h header file with the latest version.

PR #1349, opened 2022-02-01
[nvme-cli] Add support for data area 4 to get-telemetry-log

jeffreyalien:
Added size pointer and data area parms to the libnvme functions:  nvme_get_ctrl_telemetry, nvme_get_host_telemetry, and nvme_get_new_host_telemetry.   The size will be calculated by those functions based on the passed data area parm and they returned to the calling function via the size pointer.

Since additional parameters were added, this pull request is dependent on the libnvme pull request (https://github.com/linux-nvme/libnvme/pull/213) that contains the changes to the  nvme_get_ctrl_telemetry, nvme_get_host_telemetry, and nvme_get_new_host_telemetry functions.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -437,94 +437,6 @@ ret:
 	return err;
 }
 
-static int get_telemetry_log_total_size_dalb4(int fd,
-					      struct nvme_telemetry_log *log,
-					      unsigned int *total_size)
-{
-	const size_t bs = 512;
-	int err;
-	int block_size = NVME_LOG_TELEM_BLOCK_SIZE;
-	__u32 result, len;
-	void *buf = NULL;
-	struct nvme_id_ctrl ctrl;
-	struct nvme_get_features_args args = {
-		.args_size	= sizeof(args),
-		.fd		= fd,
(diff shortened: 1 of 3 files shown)


jeffreyalien:
I see this is failing because of the dependency on my libnvme changes in https://github.com/linux-nvme/libnvme/pull/213.  I wasn't sure how to handle this case but it appears the libnvme PR needs to be merged first???

igaw:
> I see this is failing because of the dependency on my libnvme changes in [linux-nvme/libnvme#213](https://github.com/linux-nvme/libnvme/pull/213). I wasn't sure how to handle this case but it appears the libnvme PR needs to be merged first???

The trick is to update the 'revision' in subprojects/libnvme.wrap to the ref of  linux-nvme/libnvme#213. Though I recommend to do this as the first commit in this PR. In this case if someone later wants to pick a commit for a backport or something like this it doesn't always conflict with libnvme.wrap.

jeffreyalien:
Ok, I can update the libnvme.wrap file once https://github.com/linux-nvme/libnvme/pull/213 has been merged.

igaw:
@jeffreyalien: I'll take care for the libnvme wrap update.

PR #1351, opened 2022-02-04
meson: fix default dracut directory based on dracut manual

sc108-lee:


igaw:
Thanks!

PR #1352, opened 2022-02-04
nvme: Add Enhanced Controller Meta Data(FID: 0x7D)

Arunpandian15:


igaw:
I was too quick and fixed the build warning first on master. Thus rebased this one

Issue #1356, opened 2022-02-08
Regression due to "nvme: open namespace exclusive by default"


igaw:
Since commit 2100609ddefb ("nvme: open namespace exclusive by default") the device is always opened with O_EXCL. The original commit claims that --forces will workaround this issue.

This change introduce regressions: https://bugzilla.suse.com/show_bug.cgi?id=1195637

So the first thing I don't like at the current code is that  O_EXCL should be used *when* is used. If a certain command wants to set the default of always use force it could change the force flag.

In particular in the id-ns case, the force flag is overloaded. It's not just O_EXCL it's also switching between nvme_identify_allocated_ns() and nvme_identify_ns().

igaw:
@olase23

PR #1357, opened 2022-02-08
nvme: Fix --force flag inversion

igaw:
argparse sets the default_value member when the argument has been
provided by the user. But in parse_and_open we set O_EXCL when --force
is not used.

It also introduced a regression in the 'id-ns' where the --force flag
is used to determine which function is used to identify the
namespace (nvme_identify_allocated_ns() vs nvme_identify_ns()).

Fixes: 2100609 ("nvme: open namespace exclusive by default")

Fixes: #1356


igaw:
@olase23 I fear we have to the default thing is not working correctly.

PR #1358, opened 2022-02-08
nvmf: Remove --matching from systemd service file

igaw:
'--matching' is not a great concept. The whole point of a discovery
log is to return different port addresses from the one we're using.

Thus the '--matching' logic was dropped when switching over the
libnvme but we forgot to remove it from the systemd service file.


igaw:
See also https://github.com/linux-nvme/nvme-cli/pull/1294

PR #1359, opened 2022-02-08
README: Mention meson subprojects update

igaw:
New meson users are not aware of this rather nifty update
mechanism. Mention it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ run following commands
 	$ meson .build
 	$ ninja -C .build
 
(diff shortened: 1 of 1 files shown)


igaw:
While at it also update the install instruction, e.g. mention dependency on json-c and libuuid.

igaw:
Dependencies for openSUSE: git-core meson gcc gcc-c++ libuuid-devel libopenssl-devel zlib-devel libjson-c-devel libhugetlbfs-devel

PR #1360, opened 2022-02-08
nvme connect-all: remove references to --matching option

mwilck:
Support for --matching was removed in 18de3a6d ("Convert to libnvme").
Using this option will cause errors. Remove all references to it
in nvmf-connect@.service, the bash completions, and the documentation.


mwilck:
Dupllicate of #1358

PR #1364, opened 2022-02-08
Update README.md

Bobarshad:


igaw:
Please add some descriptive commit subject titles and commit messages. Just 'Update XXX' is not good enough. 

And sorry for my ignorance. It looks like you are basing your effort on the Intel plugin. I assume this has nothing to do with Intel, right? So please explain what it actually is for? 

While at it, please run checkpatch from the Linux tree and fix the Error message.

igaw:
This seems to be something in the development phase. Not ready to be merged at all.

Issue #1347, opened 2022-02-01
Inreasing log level doesn't work


igaw:
Increasing the log level is supported only a few commands but this is not working. The code handling OPT_INCR is not working as expected.

`argconfig_parse` is considering OPT_INCR as flag and never goes into the CFG_INCREMENT code snipped. But even with this simple patch here it doesn't work as expected. 

```patch
@@ -219,7 +219,8 @@ int argconfig_parse(int argc, char *argv[], const char *program_desc,
                        }
                        if (option_index == options_count)
                                continue;
-                       if (long_opts[option_index].flag) {
+                       if (long_opts[option_index].flag &&
+                           options[option_index].config_type != CFG_INCREMENT) {
                                *(uint8_t *)(long_opts[option_index].flag) = 1;
                                continue;
                        }

```

More debugging needed.

ikegami-t:
Hi,

It seems that the current implementation and behavior work as expected as below.
1. The defult log level is set to `notice` if not defined but possible to be not set to `notice` if defined already.
```
#ifndef DEFAULT_LOGLEVEL
#  define DEFAULT_LOGLEVEL LOG_NOTICE
#endif
```
2. The verbose argument is set to `1` at first.
```
			if (s->argument_type == no_argument
			    && s->default_value != NULL) {
				value_addr = (void *)(char *)s->default_value;

				long_opts[option_index].flag = value_addr;
				long_opts[option_index].val = 1;
			} else {
				long_opts[option_index].flag = NULL;
				long_opts[option_index].val = 0;
			}
```
3. Then incremented to `2` but the comment description seems not correct.
```
		} else if (s->config_type == CFG_INCREMENT) {
			/*
			 * Extreme getopt_long fiddling.
			 *
			 * As per man page:
			 * If flag is non-NULL, getopt_long() returns 0,
			 * and flag  points to a variable which is set to
			 * val if the option is found.
			 *
			 * So we need to increase 'val', not 'value_addr'.
			 */
			long_opts[option_index].val++;
```
4. At last the log level chnaged to `info` from `notice` but seems the log level changed to `warning` if the option not set and also seems not changed the log level to `debug` actually.
```
	switch (verbose) {
	case 0:
		nvme_log_level = LOG_WARNING;
		break;
	case 1:
		nvme_log_level = LOG_NOTICE;
		break;
	case 2:
		nvme_log_level = LOG_INFO;
		break;
	default:
		nvme_log_level = LOG_DEBUG;
[...]

PR #1369, opened 2022-02-10
argconfig: Increase the flag value

igaw:
If the flag is provided several time increase the flag value. This is
a pretty ugly fix for how getopt_long_only() handles '--verbose' vs
'-v'. With this we make sure we increase the value also when multiple
'-v' are given by the user.

Fixes: #1347


igaw:
Not nice as we handle the CFG_INCREMENT differently depending if the long option is used or not. But I haven't found out the exact reason but this seems to be a bug in getlong. Anyway this does the trick.

PR #1373, opened 2022-02-11
nvme: fix dsm, dspec wrong usage

sc108-lee:
dsm is 8bits on Command Dword13 [07:00]
dspec is 16bits on Command Dword13 [31:16], and should issued on nvme_io_args.dspec
libnvme updated that dspec changed as 16bits


sc108-lee:
+ Remove getting nsid twice from io command

sc108-lee:
libnvme has been updated https://github.com/linux-nvme/libnvme/pull/230
so modified subproject based on that commit hash

subprojects/libnvme.wrap
```diff
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = ee30b6824f1209417e314d2c5b2e8d22b8707b6e
+revision = a849f9752edee2341feb52efe3d1bd9034f8ad38
```
igaw, line 3: Please do a libnvme.wrap update as single commit first in the series. I'd like to avoid mixing code fixes with CI build changes. This makes backporting of changes simpler.

Documentation/nvme-write.html
```diff
@@ -992,9 +992,9 @@ <h2 id="_options">OPTIONS</h2>
 </dt>
 <dd>
 <p>
-        The optional data set management attributes for this command. The
-        argument for this is the lower 16 bits of the DSM field in a write
-        command; the upper 16 bits of the field come from the directive
+        The optional data set management attributes for this command. The argument
```
igaw, line 995: Also here, just update the source and remove the generated changes from the commit. I am going to regenerated all the precompiled documentation right before the release.

igaw (changes requested):

subprojects/libnvme.wrap
sc108-lee: splitted into a new commit for that !
as you commented

Documentation/nvme-write.html
sc108-lee: removed :)

igaw:
Looks all good now. Thanks!

PR #1375, opened 2022-02-11
[nvme-cli]: Fix incorrect access after memory is already freed up.

indraneel-m:
Fix incorrect access after memory is already freed up. Also couple of minor fixes to use correct cpu endianness while accessing data read from drive.

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -1750,10 +1750,15 @@ static bool get_dev_mgment_cbs_data(nvme_root_t r, int fd, __u8 log_id, void **c
 	length = sizeof(struct wdc_c2_log_page_header);
 	hdr_ptr = (struct wdc_c2_log_page_header *)data;
 	sph = (struct wdc_c2_log_subpage_header *)(data + length);
-	found = wdc_get_dev_mng_log_entry(hdr_ptr->length, log_id, hdr_ptr, &sph);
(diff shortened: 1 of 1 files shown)


igaw:
@jeffreyalien can you review this? thanks!

jeffreyalien:
The changes look fine to me but I found a few more changes that should be added.  There are several more calls to the get_dev_mgment_cbs_data that don't call free on the allocated buffer.   In 5 cases, the call to get_dev_mgment_cbs_data is used to get the fw customer id so I created a new function - wdc_get_fw_cust_id to do this and free the buffer when done.  There was one other call to get_dev_mgment_cbs_data that needed a call to free the data.  Here's a link to a commit of my changes in my forked repo along with the changes from this PR for you to review -  https://github.com/jeffreyalien/nvme-cli/commit/ce10fbd0f4ab53081785c74e8d43f15b2d775f49

igaw:
@jeffreyalien so should I merged it or do you send a new PR with all changes? Let me know what you prefer. Please mention @indraneel-m contribution in this case though.

jeffreyalien:
@igaw I'll create a new pull request and mention @indraneel-m.  
@indraneel-m  Do you have any comments on my changes?

jeffreyalien:
@indraneel-m @igaw I created https://github.com/linux-nvme/nvme-cli/pull/1382 that contains indraneel-m's fixes as well as my additional changes.  Please review.  Also, I didn't know indraneel-m's email so if you send it to me, I can update the commit message.

igaw:
Let's close this one in favor of #1382.

PR #1378, opened 2022-02-14
Update nvme-print.c:  Only print FW slots that contain  a valid FW image revision number

samiWaheed:
Existing 2.0 branch prints all 7 fw slots including the ones with null.   The check for a valid FW revision number is no longer working due to the change of the definition of frs in  the nvme_firmware_slot struct in the latest libnvme.   Current definition uses "char    frs[7][8]" instead of old definition of  "__u64      frs[7]".

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -523,7 +523,7 @@ static void json_fw_log(struct nvme_firmware_slot *fw_log, const char *devname)
 	json_object_add_value_int(fwsi, "Active Firmware Slot (afi)",
 		fw_log->afi);
 	for (i = 0; i < 7; i++) {
-		if (fw_log->frs[i]) {
+		if (fw_log->frs[i][0]) {
 			snprintf(fmt, sizeof(fmt), "Firmware Rev Slot %d",
 				i + 1);
 			frs = (__le64 *)&fw_log->frs[i];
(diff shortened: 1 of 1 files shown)


samiWaheed:
Hi Daniel,

I am not too familiar with meson.  The libnvme.wrap file has a certain
revision/tag number in it which does not appear to be the latest libnvme.
In order to get the latest libnvme, I had to change "revision = head".
Should the nvme-cli build be based on the latest revision of libnvme, or is
there a reason to force it to a certain tag.

igaw:
Hi Sami,

nvme-cli should build a fixed version because when we change something in the API in libnvme, every PR will break for nvme-cli. We had 'revision = head' in the beginning for getting things working but it proofed to be unusable.  Too many build breaks and not buildable master branch.

So if I want to change something in libnvme and nvme-cli I create a PR for libnvme and change libnvme.wrap with the ref from the my PR as first commit. This way you build a PR for nvme-cli which depends on the new API.

Later when we merge PR, only the libnvme.wrap should result in a conflict which can easily sorted out. Also note it's important to have it as separate commit to allow simple backports. 

It's not perfect, I know. But I think it balances the tradeoffs in a good way. Maybe there are better ways to do it? I am open for better/improved workflows.

PR #1382, opened 2022-02-15
[nvme-cli] Additional changes for PR 1375

jeffreyalien:
Created a new function - wdc_get_fw_cust_id that will simplfy the
  code and handle the free of the allocated data

  Original message from PR-1375:
  Fix incorrect access after memory is already freed up. Also couple
  of minor fixes to use correct cpu endianness while accessing data
  read from drive.

Co-Authored-by: Indraneel-M

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -167,6 +167,7 @@
 #define WDC_CUSTOMER_ID_0x1004				0x1004
 #define WDC_CUSTOMER_ID_0x1008				0x1008
 #define WDC_CUSTOMER_ID_0x1304				0x1304
+#define WDC_INVALID_CUSTOMER_ID				-1
 
 #define WDC_ALL_PAGE_MASK                   0xFFFF
 #define WDC_C0_PAGE_MASK                    0x0001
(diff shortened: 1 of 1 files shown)


igaw:
I've folded all changes into one commit. It doesn't really make sense to have merge conflict fixes for such a small project as this one. Just rebase everything on the current master and fix up the conflicts in the new patch. This gives a us one single commit which can also be easily backported if needed.

Also I've updated  the commit message. Generally, they should be self contained. Referencing github issues etc is not really useful outside of github IMO. 

Please also note you need to have new line between commit subject and body. It makes all it look right when reading the git history.

jeffreyalien (commented):
@igaw Looks good to me.  Thanks for cleaning up the commits.  Still trying to figure out how to do that correctly in git.  I usually make it worse though.

igaw:
Thanks for the review. git is a difficult tool. I think in your case this sequence here could help:
```

  git checkout master
  git remote add upstream https://github.com/linux-nvme/libnvme.git
  git fetch --all
  git reset --hard upstream/master
  git checkout $branch
  git rebase HEAD~ master
```
This assumes you have one commit and it lives in a branch. If not do
```
  git checkout $branch
```
Just make a backup first. You will loose data otherwise :)

PR #1383, opened 2022-02-16
netapp-nvme: fix nvme ns desc uuid handling for ontapdevices

martin-gpy:
The nvme ns desc uuid handling in the NetApp plugin for
ontapdevices is broken after the recent upgrade to libnvme. The
uuid handling logic here was flawed to begin with, so rework this
code to ensure the respective uuid descriptor itself is stored in
the ontapdevice_info structure and processed appropriately.


hreinecke (approved):
Looks good.

PR #1389, opened 2022-02-21
container_of: Streamline documentation

Arunpandian15:
Make the documentation consistent with the other header files.
fixed whitespace issues and intentions levels.


igaw:
Thanks for your effort to get the docs in shape. I really appreciate it!.

Thought, these are the ccan files which we should not modify at all as they are verbatim copy from the ccan archive. So I'd rather not touch them.

PR #1395, opened 2022-02-25
fabrics: ensure zero kato for non-persistent controllers

igaw:
We are missing a67d001330ae ("fabrics: ensure zero kato for non-persistent controllers") from the monolithic branch.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -47,6 +47,8 @@
 #define MAX_DISC_ARGS		32
 #define MAX_DISC_RETRIES	10
 
+#define NVMF_DEF_DISC_TMO	30
+
 /* Name of file to output log pages in their raw format */
 static char *raw;
 static bool persistent;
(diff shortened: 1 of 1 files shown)


hreinecke (approved):

fabrics.c
```diff
@@ -115,6 +115,16 @@ static void space_strip_len(int max, char *str)
 	}
 }
 
+static void set_discovery_kato(struct nvme_fabrics_config *cfg)
+{
+	/* Set kato to NVMF_DEF_DISC_TMO for persistent controllers */
+	if (persistent && !cfg->keep_alive_tmo)
```
hreinecke, line 121: Where do you get the 'persistent' value from? Shouldn't it be rather an argument to the function?

igaw: The persistent variable is a global bool. It's a bit ugly I agree. struct nvme_fabrics_config doesn't have it. We could add it there (libnvme) and also move this set_discovery_kato could move then there. Though I don't know if this makes sense.

igaw:
I've decide to squash the two patches.

fabrics.c
igaw: I've added the patch in the meantime as it fixes a regression. We can still cleanup ontop of this change

PR #1400, opened 2022-03-02
cmds-main: Add all the admin, NVM commands in main commands list

Arunpandian15:
Updated all the missing admin, NVM commands in cmds-main.txt


Arunpandian15:
@igaw Shall we include plugin commands list also ?

igaw:
Yes, sounds reasonable to me. I would probably have them their own section (plugins/vendor extensions or something like that). clearly separating them between the generic commands and the plugins.

PR #1403, opened 2022-03-03
cmds-main: Add intel plugin commands

Arunpandian15:
Added intel plugin commands in list


igaw:
What about adding in nvme.txt a new section called 'Plugins/Vendor extension commands' and do a 'include cmd-plugins.txt' and have these intel commands added in this new file?

Arunpandian15:
Thanks @igaw 
I have added intel plugin commands in new file cmd-plugins.txt and created new section Plugins/Vendor extension commands, added include cmd-plugins.txt there.

igaw:
Thanks!

PR #1405, opened 2022-03-03
Add new OCP plugin

arthurshau:
New OCP plugin that consolidates some of the functionality from the other vendor plugins. For now, implements getting of C0 and C3 log pages.

Signed-off-by: Arthur Shau <arthurshau@fb.com>
Signed-off-by: Wei Zhang <wzhang@fb.com>
Signed-off-by: Venkat Ramesh <venkatraghavan@fb.com>

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -1,0 +1,801 @@
+/*
+ * Copyright (C) 2022 Meta Platforms, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors: Arthur Shau <arthurshau@fb.com>,
+ *          Wei Zhang <wzhang@fb.com>,
+ *   	    Venkat Ramesh <venkatraghavan@fb.com>
+ */
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <inttypes.h>
+#include <errno.h>
+#include <limits.h>
+#include <fcntl.h>
+#include <unistd.h>
+
+#include "common.h"
+#include "nvme.h"
+#include "libnvme.h"
+#include "plugin.h"
+#include "linux/types.h"
+#include "nvme-print.h"
+
+#define CREATE_CMD
+#include "ocp-nvme.h"
+
+/* C0 SCAO Log Page */
+#define C0_SMART_CLOUD_ATTR_LEN             0x200
+#define C0_SMART_CLOUD_ATTR_OPCODE          0xC0
+#define C0_GUID_LENGTH                      16
+#define C0_ACTIVE_BUCKET_TIMER_INCREMENT    5
+#define C0_ACTIVE_THRESHOLD_INCREMENT       5
+#define C0_MINIMUM_WINDOW_INCREMENT         100
+
+static __u8 scao_guid[C0_GUID_LENGTH]    = { 0xC5, 0xAF, 0x10, 0x28, 0xEA, 0xBF,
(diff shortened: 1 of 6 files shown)


plugins/ocp/ocp-nvme.c
```diff
@@ -0,0 +1,672 @@
+#include <stdio.h>
```
igaw, line 1: Could you add the SPDIX license identifier to this file and the header file as well?

plugins/ocp/ocp-nvme.c
```diff
@@ -0,0 +1,672 @@
...
+    SCAO_SVN                = 144,	/* Security Version Number */
+    SCAO_NUSE               = 152,	/* NUSE - Namespace utilization */
+    SCAO_PSC                = 160,	/* PLP start count */
+    SCAO_EEST               = 176,	/* Endurance estimate */
+    SCAO_PLRC               = 192,  /* PCIe Link Retraining Count */
+    SCAO_LPV                = 494,	/* Log page version */
+    SCAO_LPG                = 496,	/* Log page GUID */
+} SMART_CLOUD_ATTRIBUTE_OFFSETS;
+
+struct __attribute__((__packed__)) ssd_latency_monitor_log {
```
igaw, line 77: FYI, __attribute__((__packed__)) will align all access to sizeof(1) so you might get compiler warnings for certain operation when this type is used. E.g when the size of the struct is not a multiple of the largest member and you 
alloc an array. Usually you want to add an aliment attribute along with the packed attribute.

See also https://stackoverflow.com/questions/8568432/is-gccs-attribute-packed-pragma-pack-unsafe

plugins/ocp/ocp-nvme.c
```diff
@@ -0,0 +1,672 @@
...
+    for (i = 0; i < 16; i++) {
+        result *= 256;
+        result += data[15 - i];
+    }
+    return result;
+}
+
+static int convert_ts(time_t time, char *ts_buf)
+{
+
```
igaw, line 128: one newline too much

igaw (changes requested):
some nitpicking and one big request please use Kernel Coding style, especially the indention. I really would like to keep the formatting consistent in the project.

plugins/ocp/ocp-nvme.c
igaw: In case the struct doesn't have holes due to natural alignment of the members, you don't gain anything from packed. Instead, it drops alignment requirements ('inner and outer') and the compiler will complain for certain things (passing
a member as argument in 'call by reference' situation IIRC) . As I think you don't have any holes I don't see a reason to have the packed. That was just an observation. If want to keep the packed I don't mind. My comment was triggered because we just had a bit of a discussion on this topic and I now I see this topic popping up everywhere.

plugins/ocp/ocp-nvme.c
```diff
@@ -0,0 +1,672 @@
...
+                (__u8)log_data[SCAO_MAVF]);
+        printf("  NVMe Errata Version				%d\n",
+                (__u8)log_data[SCAO_NEV]);
+        printf("  PCIe Link Retraining Count			%"PRIu64"\n",
+            (uint64_t)le64_to_cpu(*(uint64_t *)&log_data[SCAO_PLRC]));
+    }
+    printf("\n");
+}
+
+static void ocp_print_C0_log_json(void *data)
```
ikegami-t, line 227: Seems that this can be combined with the ocp_print_C0_log_normal() since almost same code as duplicated.

venkatrag1: Could we maybe go the other way and nest the groups in json such as -

ikegami-t: I see. Just considered both output functions needed to maintain as same in future.

igaw:
Are you going to update the PR soon? If not I am going to close it.

arthurshau:
> Are you going to update the PR soon? If not I am going to close it.

Yes, I was off for a couple of weeks. I’ll update it next week

plugins/ocp/ocp-nvme.c
arthurshau: In this case, I think we can just keep packed, just so we stay consistent with the rest of the plugins, since a lot of them are also using it for their structs.

arthurshau:
Amend: changed formatting to be consistent with the Kernel Coding Style, grouped some json objects in the outputs of the C0 and C3 log pages, and broke the Active Latency Config in the C3 log page into bitfields to be displayed in the same Read/Write/Trim table format.

arthurshau:
rebased, added SPDX identifier, and added documentation for the 2 commands

igaw (approved):

igaw:
Thanks. All looks reasonable, if there are things to fixup we can do that in tree.

PR #1409, opened 2022-03-07
nvme: nvme write bug fix (no parse for option)

sc108-lee:
occured from 0adb8c2c79df953bbef9e08ce684239a5650638f
write command always print below error
data size not provided


igaw:
Ooops, sorry about that.

PR #1416, opened 2022-03-10
installation path modification

sc108-lee:
revert as 1.16
1. bash-completion
    AS IS : DATADIR/bash-completion/completions/nvme/bash-nvme-completion.sh
    TO BE : DATADIR/bash-completion/completions/nvme
2. sbindir include prefix
3. sysconfdir do not use prefix

fix rpm macro wrong use

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -248,7 +248,7 @@ executable(
                   libhugetlbfs_dep ],
   include_directories: incdir,
   install: true,
-  install_dir: get_option('sbindir')
+  install_dir: sbindir
 )
 
 ################################################################################
--- a/meson.build
+++ b/meson.build
@@ -253,7 +253,8 @@ executable(
 
 ################################################################################
 install_data('completions/bash-nvme-completion.sh',
-             install_dir: datadir + '/bash-completion/completions/nvme')
+             rename: 'nvme',
+             install_dir: datadir + '/bash-completion/completions')
 install_data('completions/_nvme',
              install_dir: datadir + '/zsh/site-functions')
 
--- a/nvme.spec.in
+++ b/nvme.spec.in
@@ -16,8 +16,8 @@ cli rpm installs core management tools with minimal dependencies.
 
 %install
 meson install --destdir %{buildroot} --skip-subprojects
-touch %{buildroot}/%{_sysconfdir}/nvme/hostnqn
-touch %{buildroot}/%{_sysconfdir}/nvme/hostid
+touch %{buildroot}%{_prefix}%{_sysconfdir}/nvme/hostnqn
+touch %{buildroot}%{_prefix}%{_sysconfdir}/nvme/hostid
 
 %files
 %defattr(-,root,root)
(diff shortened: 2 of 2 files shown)


meson.build
```diff
@@ -228,4 +226,4 @@ install_data(['nvmf-autoconnect/udev-rules/70-nvmf-autoconnect.rules',
               'nvmf-autoconnect/udev-rules/71-nvmf-iopolicy-netapp.rules'],
              install_dir: udevrulesdir)
 install_data(disc,
-             install_dir: join_paths(sysconfdir, 'nvme'))
+             install_dir: join_paths(get_option('sysconfdir'), 'nvme'))
```
igaw, line 229: Don't think this is correct. The path install path needs to be prefixed. This would install it hard under /etc/nvme/discovery.conf. For example if the user uses --prefix=/tmp/ the discovery.conf has to go to /tmp/etc/nvme/discovery.conf.

meson.build
```diff
@@ -205,7 +203,7 @@ executable(
                   libhugetlbfs_dep ],
   include_directories: incdir,
   install: true,
-  install_dir: get_option('sbindir')
+  install_dir: join_paths(prefixdir, get_option('sbindir'))
```
igaw, line 206: For consistency sake I suggest you introduce a new variable and apply the prefix as we have for mandir, bindir etc.

igaw (changes requested):

meson.build
sc108-lee: This is the reason I did remove prefix from it.

1. documentation & source file defined without prefix
https://github.com/linux-nvme/nvme-cli/blob/master/fabrics.c#L45
https://github.com/linux-nvme/nvme-cli/blob/master/Documentation/nvme-discover.txt#L43
https://github.com/linux-nvme/nvme-cli/blob/master/Documentation/nvme-connect-all.txt#L41
https://github.com/linux-nvme/nvme-cli/blob/master/nvmf-autoconnect/systemd/nvmf-autoconnect.service#L3

2. Released packages add files on /etc (not /usr/etc)

3. rpm macros does not include prefix on sysconfdir by default.

meson.build
sc108-lee: addressed :)

meson.build
igaw: Good points. First thing, The hard coded values are wrong. We need to make the paths configurable. Next thing is to figure out the correct location (see prefix vs destdir discussion, e.g. https://unix.stackexchange.com/questions/444211/what-is-the-diffference-between-the-prefix-and-destdir-params-for-dh-auto-instal). On point 3. let me read up on this one. I wouldn't be surprised if there is a catch to it (e.g. distros are not agreeing on this...)

I'll send patches for point 1.

sc108-lee: I revert that and add prefix on rpmbuild :)

sc108-lee: I create issue before your patches committed #1427

igaw:
Looks good, though there is a merge conflict with bindir. My latest changes will depend on bindir. Could you rebase your changes to the current HEAD? Thanks!

sc108-lee:
> Looks good, though there is a merge conflict with bindir. My latest changes will depend on bindir. Could you rebase your changes to the current HEAD? Thanks!

That was fast 👍 .

igaw (approved):

PR #1418, opened 2022-03-10
tests: Add tests for copy, verify, dsm commands

Arunpandian15:
Tests has been added for copy, verify, DSM
commands and respective tests updated in meson.build
file.

### DIFF
--- a/tests/nvme_copy_test.py
+++ b/tests/nvme_copy_test.py
@@ -1,0 +1,60 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# This file is part of nvme-cli
+#
+# Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# Author: Arunpandian J <apj.arun@samsung.com>
+
+"""
+NVMe Copy Testcase:-
+
+    1. Issue copy command on set of block; shall pass.
+
+"""
+
+from nose.tools import assert_equal
+from nvme_test import TestNVMe
+
+
+class TestNVMeCopy(TestNVMe):
+
+    """
+    Represents NVMe Verify testcase.
+        - Attributes:
+              - start_block :   starting block of to verify operation.
+              - range :         Range of blocks for DSM operation.
+              - slbs :          64-bit addr of first block per range
+              - test_log_dir :  directory for logs, temp files.
+    """
+
+    def __init__(self):
+        """ Pre Section for TestNVMeCopy """
+        TestNVMe.__init__(self)
+        self.start_block = 0
+        self.range = 1
+        self.slbs = 1
+        self.namespace = 1
+        self.setup_log_dir(self.__class__.__name__)
+
+    def __del__(self):
+        """ Post Section for TestNVMeCopy """
+        TestNVMe.__del__(self)
+
+    def copy(self):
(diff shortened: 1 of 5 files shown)


tests/nvme_copy_test.py
```diff
@@ -0,0 +1,71 @@
+# Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
```
igaw, line 2: I think we should start using SPDX licensee identifier for all files and remove the license text. Would you mind submitting these new files with a SPDX tag?

The main motivation is to make any tooling around tool chain verification simpler.

igaw (changes requested):
Looks good, except my request on adding SPDX tags and removing the license boilerplate text.

tests/nvme_copy_test.py
Arunpandian15: @igaw you mean to say by adding "SPDX-License-Identifier: LGPL-2.1-or-later" this text and remove "Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved." this one right ?

tests/nvme_copy_test.py
```diff
@@ -1,17 +1,19 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
igaw, line 1: nvme-cli is license under GPL-2.0-only, libnvme is LGPL-2.1-or-later. I don't mind if you license this file under LGPL as it is compatible.

tests/nvme_copy_test.py
```diff
@@ -1,17 +1,19 @@
...
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
+# GNU Lesser General Public License for more details.
 #
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
```
igaw, line 16: With the SPDX in place I'd say we also remove this redundant license text. For example the Linux kernel uses this form for Python code:

```python
#!/usr/bin/env python3
# SPDX-License-Identifier: GPL-2.0
"""
tdc_multibatch.py - a thin wrapper over tdc_batch.py to generate multiple batch
files

Copyright (C) 2019 Vlad Buslov <vladbu@mellanox.com>
"""

```

or

```python
#!/usr/bin/env python3
# SPDX-License-Identifier: GPL-2.0
#
# A thin wrapper on top of the KUnit Kernel
#
# Copyright (C) 2019, Google LLC.
# Author: Felix Guo <felixguoxiuping@gmail.com>
# Author: Brendan Higgins <brendanhiggins@google.com>


```

igaw (changes requested):

igaw:
FYI, I've squashed the changes together. This make the commit history a bit better to read.

All looks good. Thanks for updating the License info, excellent work.

igaw (approved):

Arunpandian15:
Thank You :)

PR #1419, opened 2022-03-10
Initialize create_ns config values

bjpaupor:
If left unspecified by the user default to not using protection
information (dps=0), not being multipath (nmic=0), and using the NVM
command set (csi=0).

Also zero out the nsze and ncap values to avoid using garbage data if
the user forgot to specify one or both.

Fixes issues using uninitialized data.


sc108-lee:
looks like duplicated with #1414

igaw:
Yes, that should be fixed by #1414, which I merged now. Please let us know if it fixed your problem. Thanks.

PR #1425, opened 2022-03-14
connect: Set errno to zero on nvmf_add_ctrl() success

tbzatek:
The libnvme's nvmf_add_ctrl() might return zero (success) while
having errno set to a non-zero value due to lots of calls inside.
And since nvmf_connect() returns errno primarily, make sure
it's zeroed on success.

See related https://github.com/linux-nvme/libnvme/pull/292

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -688,6 +688,7 @@ int nvmf_connect(const char *desc, int argc, char **argv)
 		fprintf(stderr, "no controller found: %s\n",
 			nvme_strerror(errno));
 	else {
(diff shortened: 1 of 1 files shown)


tbzatek:
Found when debugging https://github.com/linux-nvme/libnvme/issues/291 by `nvme connect --verbose --transport=loop --hostnqn=test_hostnqn --nqn=test_subnqn` always returning `2` with no output as it actually succeeded.

hreinecke (approved):
Looks good.

Issue #1427, opened 2022-03-15
hard coded sysconf dir


sc108-lee:
documentation & source file defined without prefix
https://github.com/linux-nvme/nvme-cli/blob/master/fabrics.c#L45
https://github.com/linux-nvme/nvme-cli/blob/master/Documentation/nvme-discover.txt#L43
https://github.com/linux-nvme/nvme-cli/blob/master/Documentation/nvme-connect-all.txt#L41
https://github.com/linux-nvme/nvme-cli/blob/master/nvmf-autoconnect/systemd/nvmf-autoconnect.service#L3

_Originally posted by @sc108-lee in https://github.com/linux-nvme/nvme-cli/pull/1416#discussion_r825549240_

sc108-lee:
@igaw working on this point

PR #1431, opened 2022-03-15
Honor sysconfdir, sbindir and bindir

igaw:
The user might decide to put files to unusual places. Replace all hard coded paths with the one provided by the user.

Fixes: #1427

@mwilck: Not sure about the systemctl paths. Do you think we should use the same sbindir there or should be introduce a config variable just for systemctl?


igaw:
The `rootprefixdir` logic is non trivial especially with the `split_usr` variable. Sure we can introduce `rootbindir` but does this really cover all cases for the other programs we are calling, e.g. modprobe, sh and echo? Or are they all covered by the POSIX standard and we can hardcode these? 

If so we could just introduce `systemctlbin` which defaults to `/usr/bin/systemctl`, which is way simpler IMO. For those distros with usr splits could then set it to `/bin/systemctl`.

igaw:
POSIX says not to '#!/bin/sh" 
  https://pubs.opengroup.org/onlinepubs/009695399/utilities/sh.html 
In theory we can't hard code this path either. Though in practice I think we can get away with it.

Anyway, I'll just introduce a build option for the systemctl path as this seems the only real world problem in these automation files.

mwilck:
> … does this really cover all cases for the other programs we are calling, e.g. modprobe, sh and echo?

No. It's not possible to cover every case. Cover a reasonable subset, and let the rest either patch the sources or file an issue.

PR #1437, opened 2022-03-16
nvme-print: Make JSON keys consistent with nvme-cli 1.x

### LINKED ISSUE #1424: JSON output key naming difference between 1.x and 2.0 series
_(this is rather a question, might be an issue for somebody)_

Just noticed the key naming differences in verbose JSON ouput, compare:
https://github.com/linux-nvme/nvme-cli/blob/7ff90b5f5ff718f760b266c06eecdad359d86a28/nvme-print.c#L7102-L7115
vs.
https://github.com/linux-nvme/nvme-cli/blob/1264c6323937c4a0342174fdd9be5a66ab1eaf24/nvme-print.c#L6899-L6919

This might a concern to some consumers as they need to adapt their parsers. Differences between major versions are expected, though difficult to spot as the `nvme` commands and arguments are mostly consistent otherwise.

Sorry if this has been discussed previously, too lazy to search...

tbzatek:
This should roughly make the JSON output consistent with the nvme-cli-monolithic branch.

Closes #1424

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -7091,28 +7091,28 @@ static void json_detail_list(nvme_root_t r)
 	nvme_for_each_host(r, h) {
 		struct json_object *hss = json_create_object();
 
-		json_object_add_value_string(hss, "hostnqn", nvme_host_get_hostnqn(h));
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #1448, opened 2022-03-19
nvme-print: print discovery async event support

martin-gpy:
Print the discovery log page change notices async event support
in nvme_feature_show_fields().


nvme-print.c
```diff
@@ -6772,6 +6772,8 @@ void nvme_feature_show_fields(enum nvme_features_id fid, unsigned int result, un
 		printf("\tDisable Normal (DN): %s\n", (result & 0x00000001) ? "True":"False");
 		break;
 	case NVME_FEAT_FID_ASYNC_EVENT:
+		printf("\tDiscovery Log Page Change Notices: %s\n",
+			((result & 0x80000000) >> 31) ? "Send async event":"Do not send async event");
```
igaw, line 6776: I think we should use the defines like NVME_FEATURE_AENCFG_NOTICE_DISCOVERY_CHANGE eventually.

PR #1450, opened 2022-03-22
nvme_smart_log_test: call smart log for all ns based on LPA 

Arunpandian15:
bug fix - added check condition in nvme_smart_log_test.py to check smart log support per namespace
and call accordingly.
added supp_check_id_ctrl wrapper in nvme_test.py to check support
based on passed key value to make it generic.

### DIFF
--- a/tests/nvme_smart_log_test.py
+++ b/tests/nvme_smart_log_test.py
@@ -83,4 +83,6 @@ class TestNVMeSmartLogCmd(TestNVMe):
     def test_smart_log(self):
         """ Testcase main """
         assert_equal(self.get_smart_log_ctrl(), 0)
-        assert_equal(self.get_smart_log_all_ns(), 0)
+        smlp = self.supp_check_id_ctrl("lpa")
+        if smlp & 0x1 == True:
+            assert_equal(self.get_smart_log_all_ns(), 0)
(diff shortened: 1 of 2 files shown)


igaw:
The commits are in the wrong order. Please introduce first the function and then the callside.

tests/nvme_smart_log_test.py
```diff
@@ -83,4 +83,6 @@ def get_smart_log_all_ns(self):
     def test_smart_log(self):
         """ Testcase main """
         assert_equal(self.get_smart_log_ctrl(), 0)
-        assert_equal(self.get_smart_log_all_ns(), 0)
+        smlp = self.supp_check_id_ctrl("lpa")
```
igaw, line 86: Please introduce supp_check_id_ctrl first in the patch series.

igaw (changes requested):
reorder the patches please

tests/nvme_smart_log_test.py
Arunpandian15: Reorder the patch series as per the ask. Thanks @igaw

Arunpandian15:
> reorder the patches please

Reordered the patches. Thanks @igaw

igaw:
Thanks!

PR #1451, opened 2022-03-22
nvme list --output-format=json --verbose fixes

tbzatek:
Found while debugging https://github.com/linux-nvme/libnvme/issues/291

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -2318,8 +2318,9 @@ static void json_print_nvme_subsystem_list(nvme_root_t r, unsigned int nsid)
 		host_attrs = json_create_object();
 		json_object_add_value_string(host_attrs, "HostNQN",
 					     nvme_host_get_hostnqn(h));
-		json_object_add_value_string(host_attrs, "HostID",
-					     nvme_host_get_hostid(h));
+		if (nvme_host_get_hostid(h))
+			json_object_add_value_string(host_attrs, "HostID",
+						     nvme_host_get_hostid(h));
 		subsystems = json_create_array();
 		nvme_for_each_subsystem(h, s) {
 			nvme_ctrl_t c;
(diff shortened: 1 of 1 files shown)


nvme-print.c
```diff
@@ -2316,10 +2316,12 @@ static void json_print_nvme_subsystem_list(nvme_root_t r, unsigned int nsid)
 		nvme_subsystem_t s;
 
 		host_attrs = json_create_object();
-		json_object_add_value_string(host_attrs, "HostNQN",
-					     nvme_host_get_hostnqn(h));
-		json_object_add_value_string(host_attrs, "HostID",
-					     nvme_host_get_hostid(h));
+		if (nvme_host_get_hostnqn(h))
+			json_object_add_value_string(host_attrs, "HostNQN",
```
igaw, line 2320: Alright, this makes it behave in the same way as 1.x.  But in 1x branch I see that also `Controller`, `Transport`, `Address` and `State` are handled this way. Would you mind updating these as well?

igaw (changes requested):
Please also update  `Controller`,  `Transport`, `Address` and `State` too

nvme-print.c
igaw: Yep, this is a bit of a whac-o-mole situation. Maybe it would be better to teach json_object_add_value_string() to handle NULL pointers.

igaw:
Even if this is the a case that /etc/nvme/hostnqn was missing due the bug, it's perfectly fine that there is no /etc/nvme/hostnqn?

hreinecke:
Sure. But then we should be reading the UUID from dmi, shouldn't we?

tbzatek:
> Errm. A NULL hostnqn? I'm pretty sure I never considered this situation (if /etc/nvme/hostnqn is not present we should read from DMI)...

Correct, not sure how did I end up in such situation that I needed that check. Looking at the libnvme code I see no chance for hostnqn being NULL during topology scanning. Hostid looks optional though.

nvme-print.c
hreinecke: This really is a bug a in the code. With 2.x we _always_ should fill out 'hostnqn' and 'hostid', as this it the primary search key for the 'nvme_host_t' structure, and without the 'hostnqn' we won't be able to find it in the tree.
So I'm not sure if it's worth fixing that in the json code.

igaw: I agree. If we want to protect against these kind errors, it would be better to make json_object_add_value_string() & friends more robust.

igaw:
Since we all agree it should not be possible to end up in the output function with NULL pointers for those attributes, let's close this one. If we hit this we should fix the source.

PR #1452, opened 2022-03-23
Add lbstm option to create-ns

bjpaupor:
Depends on changes in libnvme commit f054255, including removing nvme_init_id_ns and instead just setting fields directly in create_ns.

### DIFF
--- a/Documentation/nvme-create-ns.txt
+++ b/Documentation/nvme-create-ns.txt
@@ -16,6 +16,7 @@ SYNOPSIS
 			[--anagrp-id=<anagrpid> | -a <anagrpid>]
 			[--nvmset-id=<nvmsetid> | -i <nvmsetid>]
 			[--csi=<command_set_identifier> | -y <command_set_identifier>]
+			[--lbstm=<lbstm> | -l <lbstm>]
 			[--block-size=<block-size> | -b <block-size>]
 			[--timeout=<timeout> | -t <timeout>]
 DESCRIPTION
@@ -63,6 +64,10 @@ OPTIONS
 --csi=<command_set_identifier>::
 	This field specifies the identifier of command set.
 	if not issued, NVM Command Set will be selected.
+
+-l <lbstm>::
+--lbstm=<lbstm>::
+	Logical Block Storage Tag Mask for end-to-end protection.
 
 -b::
 --block-size::
--- a/nvme.c
+++ b/nvme.c
@@ -2304,6 +2304,7 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 	const char *anagrpid = "ANA Group Identifier (ANAGRPID)";
 	const char *nvmsetid = "NVM Set Identifier (NVMSETID)";
 	const char *csi = "command set identifier (CSI)";
+	const char *lbstm = "logical block storage tag mask (LBSTM)";
 	const char *timeout = "timeout value, in milliseconds";
 	const char *bs = "target block size, specify only if \'FLBAS\' "\
 		"value not entered";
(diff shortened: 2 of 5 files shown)


nvme.c
```diff
@@ -2400,8 +2404,16 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 		goto close_fd;
 	}
 
-	nvme_init_id_ns(&ns, cfg.nsze, cfg.ncap, cfg.flbas, cfg.dps, cfg.nmic,
-			 cfg.anagrpid, cfg.nvmsetid);
+	memset(&ns, 0, sizeof(ns));
```
igaw, line 2407: Let's keep it consistent with the rest. For this kind of initialization we use following pattern in the rest of the file:

```
	struct nvme_io_args args = {
		.args_size	= sizeof(args),
		.fd		= fd,
		.nsid		= cfg.namespace_id,
		.slba		= cfg.start_block,
		.nlb		= cfg.block_count,
		.control	= control,
		.dsm		= cfg.dsmgmt,
		.dspec		= cfg.dspec,
		.reftag		= cfg.ref_tag,
		.apptag		= cfg.app_tag,
		.appmask	= cfg.app_tag_mask,
		.storage_tag	= cfg.storage_tag,
		.data_len	= buffer_size,
		.data		= buffer,
		.metadata_len	= cfg.metadata_size,
		.metadata	= mbuffer,
		.timeout	= NVME_DEFAULT_IOCTL_TIMEOUT,
		.result		= NULL,
	};
```

igaw (changes requested):
Looks good, just a style issue :)

Documentation/nvme-create-ns.txt
```diff
@@ -16,6 +16,7 @@ SYNOPSIS
 			[--anagrp-id=<anagrpid> | -a <anagrpid>]
 			[--nvmset-id=<nvmsetid> | -i <nvmsetid>]
 			[--csi=<command_set_identifier> | -y <command_set_identifier>]
+			[--lbstm=<lbstm> | -l <lbstm>]
```
igaw, line 19: btw, it's okay just to update the *.txt files. I'll recreate the other documentation before each release.

nvme.c
bjpaupor: I thought that pattern only works for initialization, but the memset to zero all other fields would need to be first here? Though the memset might not be necessary, it seems that structs initialized that way have all non-specified fields zeroed out anyways by the standard.

bjpaupor: Hmm, the ns struct is used earlier in the function for the identify command so initializing it here would require having a separate nvme_id_ns. Seems better to just leave it as-is to me, thoughts?

Documentation/nvme-create-ns.txt
bjpaupor: Good to know, I can remove the .html and .1 file changes if that'd be best?

nvme.c
igaw: We are already exploiting the fact that we switched to C99 and hence we can declare new variable mixed in the code, although only in few places. The general rule is: define all variables at the beginning of the block. But for things like this it's just fine and makes the code more readable.

So just introduce a new `nvme_id_ns` variable right before the `nvme_ns_mgmt_create` call. 

e.g.
```
	struct nvme_io_args args = {
		.args_size	= sizeof(args),
		.fd		= fd,
		.nsid		= cfg.namespace_id,
		.slba		= cfg.start_block,
		.nlb		= cfg.block_count,
		.control	= control,
		.reftag		= cfg.ref_tag,
		.apptag		= cfg.app_tag,
		.appmask	= cfg.app_tag_mask,
		.storage_tag	= cfg.storage_tag,
		.timeout	= NVME_DEFAULT_IOCTL_TIMEOUT,
		.result		= NULL,
	};
	err = nvme_verify(&args);
```

bjpaupor: Sure, change has been implemented.

igaw:
Thanks!

PR #1453, opened 2022-03-25
fabrics: Introduce force flag to overwrite persistence logic

igaw:
The persistence flag will be overwritten by the built in logic. Allow
the user to overwrite this by providing a force command line option.

The code is a bit hard to follow, hence linearize it using forward
jumps. This allows use to centralize the overwrite logic and reduces
my head to hurt.

There is the wish to bring back a feature from 1.x branch which allows
the user always to create new discovery connection. This was done
in 1.x by looking at the KATO value. But this seems a bit fragile.
So I think it's better to introduce a explicit --force flag to overwrite
the logic.
See also:  https://github.com/linux-nvme/nvme-cli/commit/14d0e0f7e0c1b420da65783881c064e49ac7d6d7

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -524,6 +524,7 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 					"ignoring ctrl device %s, "
 					"command-line options do not match\n",
 					device);
+				nvme_free_ctrl(c);
 				c = NULL;
 				persistent = false;
 			} else {
--- a/fabrics.c
+++ b/fabrics.c
@@ -334,6 +334,68 @@ static int __discover(nvme_ctrl_t c, const struct nvme_fabrics_config *defcfg,
 	return 0;
 }
 
+/*
+ * Compare two C strings and handle NULL pointers gracefully.
+ * If either of the two strings is NULL, return 0
+ * to let caller ignore the compare.
+ */
+static inline int strcmp0(const char *s1, const char *s2)
+{
+	if (!s1 || !s2)
+		return 0;
+	return strcmp(s1, s2);
+}
+
+/*
+ * Compare two C strings and handle NULL pointers gracefully.
+ * If either of the two strings is NULL, return 0
+ * to let caller ignore the compare.
+ */
+static inline int strcasecmp0(const char *s1, const char *s2)
+{
+	if (!s1 || !s2)
+		return 0;
+	return strcasecmp(s1, s2);
+}
+
+static bool ctrl_config_match(nvme_ctrl_t c, char *transport,
+			      char *traddr, char *host_traddr,
+			      char *host_iface, char *trsvcid)
+{
+	if (!strcmp0(nvme_ctrl_get_transport(c), transport) &&
+	    !strcasecmp0(nvme_ctrl_get_traddr(c), traddr) &&
+	    !strcmp0(nvme_ctrl_get_trsvcid(c), trsvcid) &&
+	    !strcmp0(nvme_ctrl_get_host_traddr(c), host_traddr) &&
+	    !strcmp0(nvme_ctrl_get_host_iface(c), host_iface))
+		return true;
+
+	return false;
+}
+
+static nvme_ctrl_t lookup_discover_ctrl(nvme_root_t r, char *transport,
+					char *traddr, char *host_traddr,
+					char *host_iface, char *trsvcid)
+{
+	nvme_host_t h;
+	nvme_subsystem_t s;
+	nvme_ctrl_t c;
+
+	nvme_for_each_host(r, h) {
+		nvme_for_each_subsystem(h, s) {
+			nvme_subsystem_for_each_ctrl(s, c) {
+				if (!nvme_ctrl_is_discovery_ctrl(c))
+					continue;
+				if (ctrl_config_match(c, transport, traddr,
+						      host_traddr, host_iface,
+						      trsvcid))
+					return c;
+			}
+		}
+	}
+
+	return NULL;
+}
+
 static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
 				   const char *desc, bool connect,
 				   const struct nvme_fabrics_config *defcfg)
@@ -348,6 +410,7 @@ static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
 	enum nvme_print_flags flags;
 	char *format = "normal";
 	struct nvme_fabrics_config cfg;
+	bool force = false;
 
 	OPT_ARGS(opts) = {
 		NVMF_OPTS(cfg),
@@ -356,6 +419,7 @@ static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
 		OPT_FLAG("persistent",   'p', &persistent,    "persistent discovery connection"),
 		OPT_FLAG("quiet",        'S', &quiet,         "suppress already connected errors"),
 		OPT_INCR("verbose",      'v', &verbose,       "Increase logging verbosity"),
+		OPT_FLAG("force",          0, &force,         "Force persistent discovery controller creation"),
 		OPT_END()
 	};
 
@@ -401,6 +465,17 @@ static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
 
 		set_discovery_kato(&cfg);
 
+		if (!force) {
+			c = lookup_discover_ctrl(r, transport, traddr,
+						 cfg.host_traddr, cfg.host_iface,
+						 trsvcid);
+			if (c) {
+				__discover(c, &cfg, raw, connect,
+					   true, flags);
+				goto next;
+			}
+		}
+
 		c = nvme_create_ctrl(r, subsysnqn, transport, traddr,
 				     cfg.host_traddr, cfg.host_iface, trsvcid);
 		if (!c)
@@ -440,6 +515,7 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 	char *format = "normal";
 	struct nvme_fabrics_config cfg;
 	char *device = NULL;
+	bool force = false;
 
 	OPT_ARGS(opts) = {
 		OPT_STRING("device",   'd', "DEV", &device, "use existing discovery controller device"),
@@ -451,6 +527,7 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 		OPT_STRING("config",     'J', "FILE", &config_file, nvmf_config_file),
 		OPT_INCR("verbose",      'v', &verbose,       "Increase logging verbosity"),
 		OPT_FLAG("dump-config",  'O', &dump_config,   "Dump configuration file to stdout"),
+		OPT_FLAG("force",          0, &force,         "Force persistent discovery controller creation"),
 		OPT_END()
 	};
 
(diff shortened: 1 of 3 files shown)


igaw:
@martin-gpy please give this a test if possible. I played around with it and at least for me it makes sense how it behaves.

martin-gpy:
> @martin-gpy please give this a test if possible. I played around with it and at least for me it makes sense how it behaves.

I applied this patch and gave it a try. Unfortunately, it doesn't create a new PDC (Persistent Discovery Controller) with the -p option even when no PDCs were present on the host i.e. it simply ignored the -p option here:

# nvme list-subsys
#
# nvme discover -t fc -w nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 -a nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6 -p
ignoring --persistent
Discovery Log Number of Records 2, Generation counter 96
=====Discovery Log Entry 0======
trtype:  fc
adrfam:  fibre-channel
subtype: current discovery subsystem
treq:    not specified
portid:  0
trsvcid: none
subnqn:  nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:discovery
traddr:  nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags:  explicit discovery connections, duplicate discovery information
=====Discovery Log Entry 1======
trtype:  fc
adrfam:  fibre-channel
subtype: nvme subsystem
treq:    not specified
portid:  0
trsvcid: none
subnqn:  nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:subsystem.subsys_119
traddr:  nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags:  not specified
# nvme list-subsys
#

martin-gpy:
> @martin-gpy please give this a test if possible. I played around with it and at least for me it makes sense how it behaves.

I applied this patch and gave it a try. Unfortunately, it failed to create a new PDC (Persistent Discovery Controller) with the -p option even when no PDCs were present on the host i.e. it simply ignored the -p option here:

$  nvme list-subsys
$
$  nvme discover -t fc -w nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 -a nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6 -p

ignoring --persistent
Discovery Log Number of Records 2, Generation counter 96
=====Discovery Log Entry 0======
trtype: fc
adrfam: fibre-channel
subtype: current discovery subsystem
treq: not specified
portid: 0
trsvcid: none
subnqn: nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:discovery
traddr: nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags: explicit discovery connections, duplicate discovery information
=====Discovery Log Entry 1======
trtype: fc
adrfam: fibre-channel
subtype: nvme subsystem
treq: not specified
portid: 0
trsvcid: none
subnqn: nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:subsystem.subsys_119
traddr: nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags: not specified

$  nvme list-subsys
$

fabrics.c
```diff
@@ -505,63 +508,77 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
...
+	     strcmp(nvme_ctrl_get_trsvcid(c), trsvcid))) {
+		fprintf(stderr,
+			"ignoring ctrl device %s, "
+			"command-line options do not match\n",
+			device);
+		nvme_free_ctrl(c);
+	} else {
+		device_found = true;
+		goto discover;
 	}
```
hreinecke, line 545: Why so complicated?
Can't we just the 'device = NULL' when the --force option is given?

fabrics.c
```diff
@@ -505,63 +508,77 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
...
+		 */
+		persistent = true;
+	} else if (persistent) {
+		/*
+		 * No controller found, fall back to create one.
+		 * But that controller cannot be persistent.
+		 */
+		persistent = false;
+		fprintf(stderr,	"ignoring --persistent");
 	}
```
hreinecke, line 582: Similar here; if we were just setting 'device = NULL' when --force is given we would fallback to the 'normal' code path and create a new device.

hreinecke (changes requested):

hreinecke:
IE something like this:

~~~
@@ -467,6 +469,9 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
        if (!strcmp(config_file, "none"))
                config_file = NULL;
 
+       if (force)
+               device = NULL;
+
        r = nvme_create_root(stderr, map_log_level(verbose, quiet));
        if (!r) {
                fprintf(stderr, "Failed to create topology root: %s\n",
~~~

fabrics.c
```diff
@@ -505,63 +508,77 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
...
+		goto create_ctrl;
+
+	/* Use provided device as ctrl */
+	c = nvme_scan_ctrl(r, device);
+	if (!c) {
+		ret = errno;
+		goto out_free;
+	}
+
+	/* Check if device matches command-line options */
```
martin-belanger, line 525: @hreinecke - I had a discussion with @igaw regarding the `--device` option and we wanted your opinion. As you know, the `--device` option allows reusing a persistent connection. However, when using this option one must also provide all the other options that would have been used to create the connection (-t, -a, -s, etc). And I believe it's because of this code right here that tries to match all the command-line options.

When using `--device`, wouldn't it make sense to just allow it w/o requiring that all other parameters be specified as well? Here's an example: 
```
nvme discover --device nvme0
```
Currently, this is what one has to do this to reuse a persistent device:
```
nvme discover -t tcp -a 212.10.1.199 -s 8009 -n nqn.1988-11.com.dell:powerstore:00:829bb3af984dBD581792 --device nvme0
```
Any thoughts?

hreinecke: Oh well. Over-specification _again_.
I really hate this kind of making-absolutely-sure-and-add-more-failure-cases style of programming.
If we have enough information to uniquely identify something  by all means _use it_ and don't add more parameters.

This goes especially for the persistent discovery controller case; here we have a) the parameters (which in itself are enough to identify the device) and b) the device node (which again is enough to identify the .. well .. device).
So why do I have to specify _both_?
Where's the point in that, other than adding more possible failure cases?

And then there's the '-p' option, which serves to confuse the matters even more, as it really only signals 'create a persistent discovery controller', but remains silent if there already _is_ a persistent discovery controller.

fabrics.c
igaw: sure. just updated the patch

fabrics.c
hreinecke: Anyway. What we have been agreed upon is to keep the '-p' option to indicate 'Create a persistent discovery controller'.
And if a persistent discovery controller is present '-p' will be a no-op.
(That alone should be a patch, and if only to update the manpage reflecting this scenario.)

hreinecke: Additionally there will be a new option '--force' to force the creation of a persistent discovery controller, even if one exists.
(Which is what this patch is about.)

hreinecke: And the final question is what you outlined above:
Is 'nvme discover -d <devname>' a valid command?

In my opinion, yes, that should be sufficient. I completely fail to see what we would gain if we require the user to specify all connect options, other than introducing yet more possible failure cases.

igaw:
> I applied this patch and gave it a try. Unfortunately, it failed to create a new PDC (Persistent Discovery Controller) with the -p  option even when no PDCs were present on the host i.e. it simply ignored the -p option here:

It behaves exactly as designed. Without providing the --device option the --persistence option has no effect.

fabrics.c
igaw: okay let me try to get the 'nvme discover -d' case working

igaw: What about just adding a warning and and just use the '-d' discovery controller? Basically we do this for the --persistent  option already. We just ignore it when it doesn't make sense.

martin-gpy:
Applied above 2 patches to the latest master branches of the nvme-cli & libnvme git repos:

fabrics: Introduce force flag to overwrite persistence logic
fabrcis: Free non-matching controller during discovery

Now I'm able to create the new PDC using the -p option. But when attempting to create another PDC, it doesn't end up as a no-op, but instead creates a duplicate PDC again:

$ nvme list-subsys
$
$ nvme discover -t fc -w nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 -a nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6 -p

Discovery Log Number of Records 2, Generation counter 100
=====Discovery Log Entry 0======
trtype:  fc
adrfam:  fibre-channel
subtype: current discovery subsystem
treq:    not specified
portid:  0
trsvcid: none
subnqn:  nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:discovery
traddr:  nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags:  explicit discovery connections, duplicate discovery information
=====Discovery Log Entry 1======
trtype:  fc
adrfam:  fibre-channel
subtype: nvme subsystem
treq:    not specified
portid:  0
trsvcid: none
subnqn:  nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:subsystem.subsys_119
traddr:  nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags:  not specified

$ nvme list-subsys
nvme-subsys0 - NQN=nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:discovery
\
 +- nvme0 fc traddr=nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6,host_traddr=nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 live

$ nvme discover -t fc -w nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 -a nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6 -p

Discovery Log Number of Records 2, Generation counter 100
=====Discovery Log Entry 0======
trtype:  fc
adrfam:  fibre-channel
subtype: current discovery subsystem
treq:    not specified
portid:  0
trsvcid: none
subnqn:  nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:discovery
traddr:  nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags:  explicit discovery connections, duplicate discovery information
=====Discovery Log Entry 1======
trtype:  fc
adrfam:  fibre-channel
subtype: nvme subsystem
treq:    not specified
portid:  0
trsvcid: none
subnqn:  nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:subsystem.subsys_119
traddr:  nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
eflags:  not specified

$ nvme list-subsys
nvme-subsys0 - NQN=nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:discovery
\
 +- nvme1 fc traddr=nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6,host_traddr=nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 live
 +- nvme0 fc traddr=nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6,host_traddr=nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 live
$

igaw:
This version is ignores the command line options if `--device` is used. The `persistent` option works as defined. If the device is found it wont allocate a new discover controller unless `--force` is used. 

Though there is still a bit of a downside in this version. For example 'nvme discover -t tcp --traddr 10.161.8.24 --trsvcid=4420  -p' is used, it will create a new discovery controller for every invocation.

igaw:
This new version unifies the different code paths in regards how `--persistent` and `--force` are treated. For this I introduced a new function `lookup_discover_ctrl` which is used to retrieve any existing discovery controller. If it finds it, nvme-cli will use it. If not a new is created and made persistent if asked. When `--persistent` and `--force` is used nvme-cli always creates a new persistent discovery controller.

igaw:
FYI, @martin-gpy has done some testing and reported in our downstream bugzilla that all is good with the last version. @hreinecke any objection on merging this?

fabrics.c
```diff
@@ -505,25 +559,29 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
...
-				    cfg.host_iface)) ||
-			    (trsvcid && nvme_ctrl_get_trsvcid(c) &&
-			     strcmp(nvme_ctrl_get_trsvcid(c), trsvcid))) {
+			if (strcmp0(nvme_ctrl_get_transport(c), transport) ||
+			    strcasecmp0(nvme_ctrl_get_traddr(c), traddr) ||
+			    strcmp0(nvme_ctrl_get_host_traddr(c),
+				    cfg.host_traddr) ||
+			    strcmp0(nvme_ctrl_get_host_iface(c),
+				    cfg.host_iface) ||
+			    strcmp0(nvme_ctrl_get_trsvcid(c), trsvcid)) {
```
hreinecke, line 572: Do make this a common function (eg nvmf_trargs_matchI()) which then can be shared with the discovery lookup.
Having two instances of doing the same thing always causes trouble later on when the match needs to be modified due to errors or spec updates.

igaw: Good point. I've introduce a `ctrl_config_match()` helper for this.

igaw:
- This version moves the common config compare logic into one helper.
- Added a bit more information to the commit message
- Dropped the non used subsysnqn argument

fabrics.c
```diff
@@ -334,6 +334,50 @@ static int __discover(nvme_ctrl_t c, const struct nvme_fabrics_config *defcfg,
...
+
+static nvme_ctrl_t lookup_discover_ctrl(nvme_root_t r, char *transport,
+					char *traddr, char *host_traddr,
+					char *host_iface, char *trsvcid)
+{
+	nvme_host_t h;
+	nvme_subsystem_t s;
+	nvme_ctrl_t c;
+
+	if (nvme_scan_topology(r, NULL) < 0) {
```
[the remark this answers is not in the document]
igaw: Looking closer, we don't need this as `nvme_scan_topology()` is done right after `nvme_create_root()`.

igaw:
- Removed the scan topology call in the lookup function
- Ignore the NULL strings in `strcmp0` and `strcasecmp0` instead sorting them. Before that 'nvme discover -traddr=xxx -p' would always create a new pdc as trsvcid is NULL.

hreinecke (approved):
Looks good.

PR #1463, opened 2022-03-30
fabrics: Correctly stringify discovery.conf and config.json paths

hreinecke:
Port 'fabrics: Correctly stringify default hostnqn and hostid paths' from libnvme.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -42,8 +42,8 @@
 #include "libnvme.h"
 #include "nvme-print.h"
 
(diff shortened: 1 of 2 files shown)


igaw:
We need to introduce a non quoted version of SYSCONFDIR for the nvme.spec.in and the documentation. I'll take care of it.

igaw:
Forget it, I got confused. We already have this covered via the additional configuration set `substs`

PR #1468, opened 2022-04-01
fabrics: Add DIM command

martin-belanger:
TP8010 defines the DIM command (Discovery Information Management)
as a way to Register-with or Deregister-from a Discovery
Controller. This is used for NVMe-over-Fabrics when connecting to
Central Discovery Controllers (CDC).


igaw:
Thanks!

Issue #1469, opened 2022-04-01
nvme config fails to write config file if none exists


hreinecke:
If the config file doesn't exist, 'nvme config' will fail to create one.

Steps to reproduce:
~~~
rm /etc/nvme/config.json
nvme config --scan --update
</etc/nvme/config.json still missing>
~~~

igaw:
Fixed by PR #329

PR #1470, opened 2022-04-01
build: Update libnvme.wrap

hreinecke:
Get "tree: always allocate config file in nvme_read_config()" from upstream.


hreinecke:
Fixes issue #1469 .

Issue #1456, opened 2022-03-25
Just a question: Assembled json nodes are not printed properly in the latest snapshot.


franciscomunoz:
I noticed a that I'm not able to see the last  json node via json_print_object when compiling the latest version of the intel plugin

With SHA 53086d6e0c9792935e4, my tree looks like the one below (22 Sep 2021)

{
  "latstats" : {
    "type" : "read",
    "values" : {
      "bucket" : {
        "id" : 0,
        "start" : "0.00us",
        "end" : "9.00us",
        "value" : 0
      },
      "bucket" : {
        "id" : 1,
        "start" : "10.00us",
        "end" : "19.00us",
        "value" : 0
      },
      "bucket" : {
        "id" : 2,
        "start" : "20.00us",
        "end" : "39.00us",
        "value" : 0
      },
      "bucket" : {
        "id" : 3,
        "start" : "40.00us",
        "end" : "79.00us",
        "value" : 0
      },
      "bucket" : {
        "id" : 4,
        "start" : "80.00us",
        "end" : "159.00us",
        "value" : 0
      },
      "bucket" : {
        "id" : 5,
        "start" : "160.00us",
        "end" : "319.00us",
        "value" : 0
      },
      "bucket" : {
        "id" : 6,
        "start" : "320.00us",
        "end" : "639.00us",
        "value" : 0
      },
      "bucket" : {
        "id" : 7,
 "start" : "640.00us",
        "end" : "+INF",
        "value" : 0
      }
    }
  },
  "Average latency since last reset" : {
    "value in us" : 0
  }

, but i'm  just seeing

{
  "latstats":{
    "type":"read",
    "values":{
      "bucket":{
        "id":7,
        "start":"0.00us",
        "end":"+INF",
        "value":0
      }
    }
  },
  "Average latency since last reset":{
    "value in us":0
  }
}

Not sure if this is an issue. I already tried in an Ubuntu/RHEL hosts.

igaw:
Let me double check if d10feb0b1b95 ("nvme: Adapt to API rework") didn't introduce this regression.

igaw:
Unfortunately, I was not able to figure out where we broke things and I can't really test it (no hardware). From you bug report it's a Optane device, correct? 

There was no change in the JSON bits, so it's something with either the port from the monolithic branch to master branch or the API libnvme changes. Can't really say.

@franciscomunoz would it possible for you to figure out what's going on?

igaw:
[...]

PR #1474, opened 2022-04-04
plugins/intel: make 'buckets' a json array

hreinecke:
The 'buckets' elements in the statistics output are an array, not
a list of (identically named) objects.


hreinecke:
Should fix PR #1456 .

franciscomunoz:
I tried with an optate disk and it worked.

igaw:
Thanks for testing!

hreinecke:
... I tried it with an optane disk and managed to fry it.
But that was a pre-production thingie, so one might say it's expected :->

Issue #1473, opened 2022-04-02
nvme connect will be failed if no /etc/nvme/hostnqn


yizhanglinux:
Hello
With the latest nvme-cli, the nvme connect will be failed when there is no /etc/nvme/hostnqn, is that expected?
It works with the nvme-cli 1.x version.

https://github.com/linux-nvme/nvme-cli/blob/27eb620d76c2de4520d15636e811bd61be5173ce/fabrics.c#L750
```

$ git diff
diff --git a/fabrics.c b/fabrics.c
index ef80ec0..1f5460f 100644
--- a/fabrics.c
+++ b/fabrics.c
@@ -749,10 +749,14 @@ int nvmf_connect(const char *desc, int argc, char **argv)
 
        if (!hostnqn)
                hostnqn = hnqn = nvmf_hostnqn_from_file();
+       fprintf(stderr, "hnqn: %s\n", hnqn);
+       fprintf(stderr, "hostnqn: %s\n", hostnqn);
        if (!hostid)
                hostid = hid = nvmf_hostid_from_file();
        h = nvme_lookup_host(r, hostnqn, hostid);
        if (!h) {
+               fprintf(stderr, "nvme_lookup_host: failed: hnqn: %s\n", hnqn);
+               fprintf(stderr, "nvme_lookup_host: failed: hostnqn: %s\n", hostnqn);
                errno = ENOMEM;
                goto out_free;
        }
$ .build/nvme connect -t loop -n blktests-subsystem-1 
hnqn: (null)
hostnqn: (null)
nvme_lookup_host: failed: hnqn: (null)
nvme_lookup_host: failed: hostnqn: (null)
```

igaw:
I think the 1.x version called `nvmf_hostnqn_file() -> hostnqn_read()` which tried to read the hosnqn from `/etc/nvme/hostnqn` and if this failed it tried `hostnqn_read_dmi` and eventually `hostnqn_generate_systemd` as fallback. I think we could do the same approach here.

PR #1477, opened 2022-04-04
fabrics: Support connect even when no /etc/nvme/hostnqn file exists

igaw:
The connect call will fail if there is no /etc/nvme/hostnqn file
available. The 1.x version did have a fallback mechanisme in place
when the config file was missing.

Let's add this feature back by calling nvmf_hostnqn_generate() when
there is no /etc/nvme/hostnqn file.

Fixes: #1473

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -504,7 +504,6 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 	char *subsysnqn = NVME_DISC_SUBSYS_NAME;
 	char *hostnqn = NULL, *hostid = NULL, *hostkey = NULL, *ctrlkey = NULL;
 	char *transport = NULL, *traddr = NULL, *trsvcid = NULL;
-	char *hnqn = NULL, *hid = NULL;
 	char *config_file = PATH_NVMF_CONFIG;
 	enum nvme_print_flags flags;
 	nvme_root_t r;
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -560,9 +559,11 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 	set_discovery_kato(&cfg);
 
 	if (!hostnqn)
-		hostnqn = hnqn = nvmf_hostnqn_from_file();
+		hostnqn = nvmf_hostnqn_from_file();
+	if (!hostnqn)
+		hostnqn = nvmf_hostnqn_generate();
 	if (!hostid)
-		hostid = hid = nvmf_hostid_from_file();
+		hostid = nvmf_hostid_from_file();
```
hreinecke, line 566: Watch out. This patch is actually required, as it differentiate between 'hostid' being a command-line argument (which is static, and can't be freed) and read or generated (which is dynamic, and needs to be freed).

fabrics.c
```diff
@@ -657,10 +658,8 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 out_free_ctrl:
 	nvme_free_ctrl(c);
 out_free:
-	if (hnqn)
-		free(hnqn);
-	if (hid)
-		free(hid);
+	free(hostnqn);
+	free(hostid);
```
hreinecke, line 662: This is wrong. 'hostnqn' and/or 'hostid' might point to a command-line argument, in which case it can't be freed.

fabrics.c
igaw: `nvmf_hostid_from_file()` calls `nvmf_read_file()` which returns a new string via `strndup()`. I can't see where we return a static string (which would be a very confusing API...)

fabrics.c
igaw: Oh, I see it's `argconfig_parse()`. I really start to develop a very strong opinion on this parser.

Issue #1476, opened 2022-04-04
Wrong KATO value assigned to I/O controllers during nvme connect-all with -p option


martin-gpy:
Noticed that whenever nvme connect-all is run with the -p option, it assigns the same 30s KATO value to both the persistent discovery controller as well as the I/O controller. Again this problem is seen only when the -p option is passed to the nvme connect-all, and not otherwise:

$ nvme list-subsys
$

$ nvme connect-all -t fc -w nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 -a nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
$

$ nvme list-subsys
nvme-subsys1 - NQN=nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:subsystem.subsys_119
\
 +- nvme1 fc traddr=nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6,host_traddr=nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 live

$ cat /sys/class/nvme-subsystem/nvme-subsys1/nvme1/kato
5

$ nvme disconnect-all

$ nvme list-subsys
$

$ nvme connect-all -t fc -w nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 -a nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6 -p

$ nvme list-subsys
nvme-subsys1 - NQN=nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:subsystem.subsys_119
\
 +- nvme1 fc traddr=nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6,host_traddr=nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 live
nvme-subsys0 - NQN=nqn.1992-08.com.netapp:sn.e6f0479b886211eca69d00a098fd4ca6:discovery
\
 +- nvme0 fc traddr=nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6,host_traddr=nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 live

$ cat /sys/class/nvme-subsystem/nvme-subsys0/nvme0/kato
30

$ cat /sys/class/nvme-subsystem/nvme-subsys1/nvme1/kato
30

martin-gpy:
[...]

PR #1479, opened 2022-04-04
 fabrics: Only update KATO for discovery controllers

igaw:
The keep alive timeout value should only be updated on discovery
controllers. __discovery(..., connect=true, ...) is also called for
normal controllers, hence we can't set the nvme_fabrics_config on top
level. Instead we have to move set_discover_kato down in the call
chain right before we connect it. Also we have to make sure not update
the default nvme_fabrics_config.

Fixes: #1476

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -269,7 +269,7 @@ static void json_connect_msg(nvme_ctrl_t c)
 	json_free_object(root);
 }
 
-static int __discover(nvme_ctrl_t c, const struct nvme_fabrics_config *defcfg,
+static int __discover(nvme_ctrl_t c, struct nvme_fabrics_config *defcfg,
 		      char *raw, bool connect, bool persistent,
 		      enum nvme_print_flags flags)
 {
@@ -299,12 +299,19 @@ static int __discover(nvme_ctrl_t c, const struct nvme_fabrics_config *defcfg,
 		else
 			print_discovery_log(log, numrec);
 	} else if (connect) {
+		int keep_alive_tmo = defcfg->keep_alive_tmo;
 		int i;
 
 		for (i = 0; i < numrec; i++) {
 			struct nvmf_disc_log_entry *e = &log->entries[i];
 			bool discover = false;
 			nvme_ctrl_t child;
+
+			if (e->subtype == NVME_NQN_DISC ||
+			    e->subtype == NVME_NQN_CURR)
+				set_discovery_kato(defcfg);
+			else
+				defcfg->keep_alive_tmo = keep_alive_tmo;
 
 			errno = 0;
 			child = nvmf_connect_disc_entry(h, e, defcfg,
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -299,13 +299,20 @@ static int __discover(nvme_ctrl_t c, const struct nvme_fabrics_config *defcfg,
...
 		int i;
 
 		for (i = 0; i < numrec; i++) {
 			struct nvmf_disc_log_entry *e = &log->entries[i];
 			bool discover = false;
 			nvme_ctrl_t child;
 
+			if (e->subtype == NVME_NQN_DISC ||
+			    e->subtype == NVME_NQN_CURR)
+				set_discovery_kato(defcfg);
```
martin-gpy, line 312: Checking again, there still seems to be a problem here since set_discovery_kato() is still invoked in nvmf_discover() - so the I/O controllers will continue to get the wrong KATO values.

And we can't remove set_discovery_kato() from nvmf_discover() since the KATO values for the discovery controllers will no longer be set for the 'nvme discover' command. So this is a little tricky now...

igaw: I think we just can drop the additional `set_discovery_kato()` in `nvmf_discover()`. `nvmf_discover()` calls eventually `__discover()` where we have the above code. Let me try to recreate this.

igaw: Okay, got an idea how to address this. So we basically have to introduce setter function for the controller for these kind of attributes which are applied directly.

igaw: Scratch this idea. We just need to make sure everytime we create a new discovery controller we just set the kato. This works well for all discovery paths, but not not for 'nvme connect'. Still trying to figure out what we could do there.

igaw:
The latest version should take care setting all controllers the correct KATO value (discovery vs I/O controller) if  'nvme discover'  is used.

For 'nvme connect' it's currently not possible to figure out if the controller we are setting up on the host is going to be a discovery controller or not. I can't really say if this is a problem or not. A possible solution to this problem is to try discover the controller before connecting it (basically doing the thing we do now for 'nvme discover'). But I don't know if this is correct or even allowed via the spec.  Need to read up on this part. Unless someone can tell me if this is okay to do or not.

igaw:
Superseded by #1489

fabrics.c
```diff
@@ -299,13 +299,19 @@ static int __discover(nvme_ctrl_t c, const struct nvme_fabrics_config *defcfg,
...
 	} else if (connect) {
+		int keep_alive_tmo = defcfg->keep_alive_tmo;
 		int i;
 
 		for (i = 0; i < numrec; i++) {
 			struct nvmf_disc_log_entry *e = &log->entries[i];
 			bool discover = false;
 			nvme_ctrl_t child;
 
+			if (e->subtype == NVME_NQN_DISC)
```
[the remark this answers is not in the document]
igaw: NVME_NQN_CURR can be ignored. Check the nvmf_connect_disc_entry() code.

PR #1481, opened 2022-04-06
fabrics: Update filter functions and add local ana state getter

igaw:
The filters API has changed in libnvme update the callsides in
nvme-cli. Also implement the nvme_ctrl_get_ana_state() in nvme-cli as
the library has now all APIs to retrieve this information.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -2373,6 +2373,24 @@ void nvme_show_supported_cap_config_log(
 	}
 }
 
+static const char *nvme_ctrl_get_ana_state(nvme_ctrl_t c, __u32 nsid)
+{
+        nvme_path_t p;
+	nvme_ns_t n;
+
+        if (nsid == NVME_NSID_ALL)
+		return NULL;
+
+	nvme_ctrl_for_each_path(c, p) {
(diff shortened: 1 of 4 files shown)


nvme-print.c
```diff
@@ -2373,6 +2373,24 @@ void nvme_show_supported_cap_config_log(
...
+	nvme_ctrl_for_each_path(c, p) {
+		n = nvme_path_get_ns(p);
+		if (!n || nvme_ns_get_nsid(n) != nsid)
+			continue;
+		return nvme_path_get_ana_state(p);
+	}
+
+	return NULL;
+}
+
```
hreinecke, line 2393: Errm. But this is precisely why I _did_ do the rework.
The filter function is supposed to prune the tree from all unrelated entries, and we just have to print the tree itself.
Let me push my changes here.

hreinecke (changes requested):

hreinecke:
Check PR #1482.

nvme-print.c
igaw: The pruning works too good in my case...

igaw:
Got replaced by #1485 (which is base on #1482)

PR #1483, opened 2022-04-06
nvme-print: Fix json array object instantiation

tbzatek:
Forgotten commit from #1451 that was unrelated to the NULL-values discussed there, still a pretty important fix for the JSON builder. Resubmitting.

This caused memory corruption in case of multiple host records.


igaw:
Looks good, though I wouldn't mind with a bit more added to the commit message :)

igaw:
NV. Added the snippet myself.

tbzatek:
Thanks!

PR #1487, opened 2022-04-08
fabrics: create a persistent controller using unique subsystem nqn

martin-gpy:
One should ideally create a persistent discovery controller
using the unique subsystem nqn if supported, rather than
the well-known subsystem nqn.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -505,6 +505,31 @@ static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
 		errno = 0;
 		ret = nvmf_add_ctrl(h, c, &cfg);
 		if (!ret) {
+			if (persistent) {
+				struct nvme_id_ctrl id = { 0 };
+				ret = nvme_ctrl_identify(c, &id);
+				if (ret)
+					goto next;
+				/* Recreate the controller using the
+				 * unique subsystem nqn, if supported
+				 */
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -505,6 +505,31 @@ static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
...
+							trsvcid);
+					if (!c)
+						goto next;
+					nvme_ctrl_set_discovery_ctrl(c, true);
+					errno = 0;
+					ret = nvmf_add_ctrl(h, c, &cfg);
+					if (ret)
+						goto next;
+				}
+			}
```
igaw, line 532: I think this should go into a function. The code below looks almost identically.

igaw (changes requested):

fabrics.c
martin-gpy: > I think this should go into a function. The code below looks almost identically.

You mean a common helper function? 

I did think about that initially, but then one would have to pass all the nvme_create_ctrl() & nvmf_add_ctrl() params to that helper function, which would make it look quite ugly and unwieldy. So dropped that idea...

igaw: I don't. Having almost identically code but does the same thing is a nightmare to maintain.

igaw:
Merged #1490 instead. Closing this one.

PR #1490, opened 2022-04-08
Create persistent controller using unique subsystem NQN

igaw:
Prefer unique subsystem NQN discovery controllers over well-known subsystem NQN when a persistent controller is created.

This is series is inspired by #1487, but avoiding code duplication.

Note: this is based on #1489


igaw:
Splitted the last patch into a refactoring patch and the strict feature adding patch. Also I decided to avoid the recursive call to `create_discover_ctrl()` just to be on the save side.

PR #1492, opened 2022-04-08
WDC Plugin Changes for SN_810-2 Drive

jeffreyalien:


plugins/wdc/wdc-nvme.c
```diff
@@ -96,8 +96,7 @@
 #define WDC_NVME_ZN350_DEV_ID				0x5010
 #define WDC_NVME_ZN350_DEV_ID_1 			0x5018
 #define WDC_NVME_SN810_DEV_ID				0x5011
-// #define WDC_NVME_SN810_DEV_ID_2			0x5037  jal not sure if the official ID is 5037 or 5030
-#define WDC_NVME_SN810_DEV_ID_2 			0x5030
+#define WDC_NVME_SN810_DEV_ID_2		        0x5037
```
igaw, line 99: These changes should be merged with the first patch introducing these definitions. The subject of this patch  talks about documentation. 

Also I would merge the meson.build changes directly into the patch when you introduce the new documents.

BTW, you could use just the 'doc: ...' prefix instead of 'documentation:...' to safe some space in the title.

igaw (changes requested):

plugins/wdc/wdc-nvme.c
igaw: You don't have to close this PR. Just cleanup your branch (keep the same name) and then do a 'git push --force'. This will overwrite the old commits in this branch. This keeps this discussion/review in on PR.

jeffreyalien:
Closed - replaced with PR-1504

jeffreyalien:
@igaw ok I'll reopen it and see if I can get all the commits cleaned up.  Ignore PR-1504 for now.

jeffreyalien:
@igaw Updated the commits - 1 for the wdc plugin code changes and 1 for the documentation changes.  Please re-review.  Thanks.

igaw:
Thanks a lot for the cleanup! Way better!

Just one last thing: I updated the commit message in the first patch. The commit title and the body need to be separated by a new line. If this is missing, everything is considered as subject title.

igaw (approved):

PR #1493, opened 2022-04-09
meson: Install udev rules to /usr/lib 

Foxboron:
Signed-off-by: Morten Linderud <morten@linderud.pw>


Foxboron:
Okay this doesn't work and I'm very confused as to why :/

igaw:
FWIW, the sysconfig was used to match it up with the nvme-cli 1.x

PR #1495, opened 2022-04-11
fabrics: avoid segfault when nvme discover fails with the -p option

martin-gpy:
If the initial discovery controller creation fails while passing
the -p option to the nvme discover (or connect-all) command, it
could lead to the below segfault:

nvme discover -t rdma -w 192.168.1.114 -a 192.168.1.113 -p
Failed to write to /dev/nvme-fabrics: Connection reset by peer

Avoid this segfault by checking if the controller created in
__create_discover_ctrl() is valid before proceeding.


igaw:
Just a style update.

igaw:
Thanks.

PR #1497, opened 2022-04-11
fabrics: error message for nvme discover/connect-all with no params

martin-gpy:
Currently when nvme discover or connect-all is called without
passing any params, it simply blanks out without any meaningful
error. So add an appropriate error message for such scenarios,
similar to what's done in the erstwhile nvme-cli-monolithic branch.


igaw:
But shouldn't be the message be something like `printf("Couldn't open %s: %s\n", strerror(errno))` instead? "No params given ..." seems a bit strange on an `fopen` failure.

igaw:
After looking again, I see why 'no params' is there...

PR #1498, opened 2022-04-12
nvme: Fix error when passthru does not match opcode based flags

sc108-lee:
when write option is set(but opcode is not), file should open with read flag
same for read
it occur 'Bad file descriptor' error when access the file

Reported-by: Kyungsik Shin <ks1012.shin@samsung.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -7313,8 +7313,22 @@ static int passthru(int argc, char **argv, bool admin,
 	if (fd < 0)
 		goto ret;
 
-	flags = cfg.opcode & 1 ? O_RDONLY : O_WRONLY | O_CREAT;
-	dfd = mfd = cfg.opcode & 1 ? STDIN_FILENO : STDOUT_FILENO;
+	if (cfg.opcode & 0x01)
+		cfg.write = true;
+
+	if (cfg.opcode & 0x02)
+		cfg.read = true;
+
+	if (cfg.write) {
+		flags = O_RDONLY;
+		dfd = mfd = STDIN_FILENO;
+	}
+
+	if (cfg.read) {
+		flags = O_WRONLY | O_CREAT;
+		dfd = mfd = STDOUT_FILENO;
+	}
+
 	if (strlen(cfg.input_file)) {
 		dfd = open(cfg.input_file, flags, mode);
 		if (dfd < 0) {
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -7315,6 +7315,18 @@ static int passthru(int argc, char **argv, bool admin,
 
 	flags = cfg.opcode & 1 ? O_RDONLY : O_WRONLY | O_CREAT;
 	dfd = mfd = cfg.opcode & 1 ? STDIN_FILENO : STDOUT_FILENO;
+	if (cfg.write && !(cfg.opcode & 0x01)) {
```
igaw, line 7318: Could we not make this a bit more readable when we touch these lines? It's a bit silly first to set the flags and file descriptor and then set them again after checking the consistency between `cfg.write` and `cfg.opcode`. This logic is making my head hurt...

sc108-lee: yes I agreed, could we just remove opcode checking and using write & read flags?

sc108-lee: spec mentioned below
 3. Indicates the data transfer direction of the command. All options to the command shall transfer data as specified or transfer no data. All commands, including vendor specific commands, shall follow this convention: 00b = no data transfer; 01b = host to controller; 10b = controller to host; 11b = bidirectional.

So using opcode mask to check direction is right.
But I assume that we are using write, read flags to make it work even if its not satisfying spec's Data Transfer field.
For User usage, so current commits has cfg.wirte & cfg.read.

But Do we need it?
As Spec mentioned above **shall follow this convention**.
That's why I'm consider removing  cfg.wirte & cfg.read.

If you agree on this, I might remove those flags, and maybe modify like below
eg ) 
static inline __u16 nvme_data_transfer(__u8 opcode)
{
	return NVME_GET(opcode, DATA_TRANSFER);
}

if (nvme_data_transfer(opcode) == TRANSFER_HOST_TO_CONTROLLER)

OR using
NVME_GET(opcode, TRANSFER_HOST_TO_CONTROLLER)

igaw: I am fine with removing the cfg.write and cfg.read logic but I probably would leave the command line options and just ignore them for backwards compatibility.

sc108-lee: humm. In that case to leave the options, I think It would be better to using as current commit.

override flags by opcodes mask

if opcode shows write flag:
   cfg.write = true

if opcode shows read flag:
   cfg.read= true

using cfg.write, cfg.read flags to determine the way of file handling.

igaw:
Thanks. I think the resulting code is way better to understand!

PR #1499, opened 2022-04-12
nvme: Add git ref to the binary

igaw:
In order to be able to figure out which binary is in use (for example
in debugging situation) it's really helpful to have the 'git describe'
ref added to the binary.

      $ .build/nvme --version
      nvme version 2.0 (git 2.0-2-gcbd3f8d+)

      $ .build/nvme wdc --version
      nvme wdc version 1.16.4 (git 2.0-2-gcbd3f8d+)

And with the linvme introduced nvme_get_version() the output is:

      $ nvme --version
      nvme version 2.0 (git 2.0-2-g037372a+)
      libnvme 1.0 (git 1.0-2-gfb628fe)


wrapper.c
```diff
@@ -0,0 +1,22 @@
...
+
+#include <dlfcn.h>
+
+#include <libnvme.h>
+
+const char * __attribute__((weak)) nvme_get_version(enum nvme_version type)
+{
+	const char *(*libnvme_get_version)(enum nvme_version type);
+
+	libnvme_get_version = dlsym(RTLD_NEXT, "nvme_get_version");
```
igaw, line 16: I am still pondering if this is a good idea. For one, building nvme-cli statically isn't possible with this anymore. At least not out of the box.

igaw: After a bit of experimenting this isn't an issue. So even if we have this code in place, when the linker kicks in (for the shared version or the static version) it will always drop this weak function in favor of the one in libnvme if found. 

So we don't have to any fancy stuff here.

meson.build
```diff
@@ -256,6 +257,7 @@ executable(
   sources,
   dependencies: [ libnvme_dep, libuuid_dep, json_c_dep, libz_dep,
                   libhugetlbfs_dep ],
+  link_args: '-ldl',
```
igaw, line 260: btw, link_args is only used when building shared binaries. So nothing to worry about static builds.

igaw:
Some more testing on this feature.

Running against a library which has `nvme_get_version()`:
```shell
$ LD_LIBRARY_PATH=/tmp/libnvme-version/lib64 .build/nvme --version   
nvme version 2.0 (git 2.0-18-g5a8623a)
libnvme 1.0 (git 1.0-2-gfb628fe)
```

Running against a library which has *no* `nvme_get_version()`:
```shell
$ LD_LIBRARY_PATH=/tmp/libnvme-no-version/lib64 .build/nvme --version
nvme version 2.0 (git 2.0-18-g5a8623a)
libnvme n/a (git n/a)
```

I am pretty happy with this result. The code doesn't look too cluttered and I think it's a good idea trying to be not forward and backward compatible if it is possible in a maintainable way.

igaw:
Make the output consistent:

```shell
$ .build/nvme --version
nvme version 2.0 (git 2.0-18-g5a8623a)
libnvme version 1.0 (git 1.0-2-gfb628fe)
```

PR #1503, opened 2022-04-13
fabrics: skip connect if the transport types don't match

martin-gpy:
Discovery log page data may include records belonging to different
transport types. If during a nvme connect-all, a connect is attempted
on a record that doesn't match the transport type passed here, it
would end up in a connect failure for that record. For e.g. one would
see the following error if a connect is attempted on a tcp record,
but the transport type passed here is 'fc' and its associated params:

nvme_tcp: malformed src address passed: nn-0xXXXX:pn-0xYYYY

Fix this by proceeding with the connect only if the transport types
match for a given record during the connect-all.

[marting: backport from monolithic branch]

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -386,6 +386,10 @@ static int __discover(nvme_ctrl_t c, struct nvme_fabrics_config *defcfg,
 			nvme_ctrl_t child;
 			int tmo = defcfg->keep_alive_tmo;
 
+			/* Skip connect if the transport types don't match */
+			if (strcmp(nvme_ctrl_get_transport(c), nvmf_trtype_str(e->trtype)))
+				continue;
+
 			if (e->subtype == NVME_NQN_DISC)
 				set_discovery_kato(defcfg);
 


igaw:
This patch is missing on purpose. @hreinecke was strictly against porting this patch. See https://github.com/linux-nvme/nvme-cli/pull/1294

martin-gpy:
> This patch is missing on purpose. @hreinecke was strictly against porting this patch. See #1294

I don't understand why.

It is normal to see records with different trtypes in the discovery log page inventory. And if the transport types don't match as explained above, the connect to that specific record will obviously end up in an error. So why even bother attempt to connect to it in the first place? You might as well skip that record and proceed to the next, during the connect-all.

Avoiding such connect errors will also help reduce the noise in the logs....

martin-gpy:
> This patch is missing on purpose. @hreinecke was strictly against porting this patch. See #1294

Not sure what the objection is all about. Hannes seems to be talking about port addresses in #1294, which has nothing to do with this patch. All port addresses for a given transport type are returned in the discovery log page, and we should connect to every one of them - no confusion there. 

But this patch is about trtypes, and not port addresses. It is normal to see records with different trtypes in the discovery log page inventory. And if the trtypes don't match as explained above (irrespective of the port addresses), the connect to that specific record will always end up in an error. So why even bother attempt to connect to it in the first place? You might as well skip that record and proceed to the next, during the connect-all. And that's what this patch is trying to do.

Avoiding such connect errors will also help reduce the noise in the logs....

igaw:
IIRC, it's not possible to decide from discovery arguments what the connection arguments are supposed to be.

martin-gpy:
> IIRC, it's not possible to decide from discovery arguments what the connection arguments are supposed to be.

Could you elaborate? We are only concerned about the trtype specified by the user here (i.e. -t option or read from the discovery.conf) for this connect-all command. And that trtype applies to both the discovery and the connect to each individual record in the discovery log page data, as part of this connect-all. Right?

igaw:
I can't remember the exact argument from @hreinecke but IIRC it is exactly this statement

> And that trtype applies to both the discovery and the connect to each individual record in the discovery log page data, as part of this connect-all. Right?" 

which is not true. The transport you use to connect to the discovery controller might not be the same transport which the I/O controller is reachable. Again, if my memory doesn't fail me, the example was if you connect to the discovery controller via tcp and it list connection to tcp and fc targets why should your host not connect via nvme-tcp and nvme-fc if it's possible?

martin-gpy:
> I can't remember the exact argument from @hreinecke but IIRC it is exactly this statement
> 
> > And that trtype applies to both the discovery and the connect to each individual record in the discovery log page data, as part of this connect-all. Right?"
> 
> which is not true. The transport you use to connect to the discovery controller might not be the same transport which the I/O controller is reachable. Again, if my memory doesn't fail me, the example was if you connect to the discovery controller via tcp and it list connection to tcp and fc targets why should your host not try connect via nvme-tcp and nvme-fc?

So how do we accomplish this (i.e. connect to the discovery & I/O subsystems through different transports) today with the current connect-all command? Do we need a separate config file to do this? Could you please provide an example?

The only intent of my patch was to avoid connecting from say a tcp port to a fc port, or a fc port to a rdma port, etc. That obviously would error out, hence just avoid attempting such a connect in the first place.

PR #1507, opened 2022-04-18
fabrics: avoid segfault if transport type is omitted

martin-gpy:
A nvme discover or connect-all would currently end up in a
segfault if the transport type is omitted as shown below:

nvme discover -w 192.168.1.114 -a 192.168.1.160

Avoid this segfault by checking if the transport type is
available before proceeding in get_default_trsvcid().


hreinecke (approved):
Looks good.

PR #1509, opened 2022-04-19
sfx cmd: fixed byte aligned issue; add additional smart log items

AndyHouHit:
expend  "nvme sfx smart-log-add" info

### DIFF
--- a/plugins/scaleflux/sfx-nvme.c
+++ b/plugins/scaleflux/sfx-nvme.c
@@ -66,25 +66,26 @@ struct nvme_capacity_info {
 	__u64 used_space;
 	__u64 free_space;
 };
-struct	__attribute__((packed)) nvme_additional_smart_log_item {
-	uint8_t			   key;
-	uint8_t			   _kp[2];
-	uint8_t			   norm;
-	uint8_t			   _np;
-	union {
-		uint8_t		   raw[6];
-		struct wear_level {
-			uint16_t	min;
-			uint16_t	max;
-			uint16_t	avg;
-		} wear_level ;
-		struct thermal_throttle {
-			uint8_t    pct;
-			uint32_t	count;
+
+struct  __attribute__((packed)) nvme_additional_smart_log_item {
+	__u8			key;
+	__u8			_kp[2];
+	__u8			norm;
+	__u8			_np;
+	union __attribute__((packed)) {
+		__u8		raw[6];
+		struct __attribute__((packed))  wear_level {
+			__le16	min;
+			__le16	max;
+			__le16	avg;
+		} wear_level;
+		struct __attribute__((packed)) thermal_throttle {
+			__u8	pct;
+			__u32	count;
 		} thermal_throttle;
-	};
-	uint8_t			   _rp;
-};
+	} ;
+	__u8			_rp;
+} ;
 
 struct nvme_additional_smart_log {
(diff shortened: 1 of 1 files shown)


AndyHouHit:
@keithbusch, I am engineer Wei Hou from Scaleflux. Could you help to approve the running workflows for this MR?

AndyHouHit:
@keithbusch , check pass, plz help to approve MR.

plugins/scaleflux/sfx-nvme.c
```diff
@@ -105,6 +106,11 @@ struct nvme_additional_smart_log {
 	struct nvme_additional_smart_log_item	 erase_timeout_cnt;
 	struct nvme_additional_smart_log_item	 read_timeout_cnt;
 	struct nvme_additional_smart_log_item	 read_ecc_cnt;//retry cnt
+	struct nvme_additional_smart_log_item    non_media_crc_err_cnt;
+	struct nvme_additional_smart_log_item    compression_path_err_cnt;
+	struct nvme_additional_smart_log_item    out_of_space_flag;
+	struct nvme_additional_smart_log_item    physical_usage_ratio;
+	struct nvme_additional_smart_log_item    grown_bb; //grown bad block
```
hreinecke, line 113: Where are these additional values used? Shouldn't you have a patch for printing them out or something?

plugins/scaleflux/sfx-nvme.c
```diff
@@ -66,25 +66,26 @@ struct nvme_capacity_info {
...
+	union __attribute__((packed)) {
+		__u8		raw[6];
+		struct __attribute__((packed))  wear_level {
+			__le16	min;
+			__le16	max;
+			__le16	avg;
+		} wear_level;
+		struct __attribute__((packed)) thermal_throttle {
+			__u8	pct;
+			__u32	count;
```
hreinecke, line 84: One does wonder why this structure is one byte _less_ that all other elements in the union.
Is this correct?

hreinecke (changes requested):

plugins/scaleflux/sfx-nvme.c
AndyHouHit: Good Eye. Yes, we missed the output for MR. Will update soon.

plugins/scaleflux/sfx-nvme.c
AndyHouHit: yes, it is correct. Main purpose of this MR is fixing the compatibility issue for older device.

AndyHouHit: Or we can use "#pragma pack(push,1)" and "#pragma pack(pop)"  before and after the structure nvme_additional_smart_log_item definition to make it more clear? How do you think?

AndyHouHit:
@keithbusch , plz help to approval t he workflow again.

AndyHouHit (commented):
@hreinecke , any comment about the change?

AndyHouHit:
@keithbusch  @hreinecke ,  Do you have any comment about the change? If not, could you help to MR the change?

plugins/scaleflux/sfx-nvme.c
AndyHouHit: shown part was updated.

igaw:
Merged via #1522

PR #1511, opened 2022-04-21
fabrics: fix help string for connect traddr option

vincentkfu:
This is just a small change to fix up an error message.


hreinecke (approved):
Looks good.

PR #1513, opened 2022-04-27
meson_options: Include system sub directory

Foxboron:
It's unclear to me if we want `system` included in `meson_options.txt` or if it should be implicitly appended in `meson.build`?


igaw:
Makes sense to use `lib/systemd/system` for systemddir as default. The reason I didn't see the problem is, because our build system sets it always. `-Dsystemddir=/usr/lib/systemd/system`

PR #1525, opened 2022-05-13
ci: add static library compilation

glimchb:
Now possible since we can use uuid for fallback

Signed-off-by: Boris Glimcher <Boris.Glimcher@emc.com>


igaw:
I think we need to bump the `libnvme.wrap` too.  But so far I am not really getting good results even with the latest version of libnvme.

Here some experiments:

```
  Subprojects
    json-c         : YES
    libnvme        : YES
    openssl        : YES
    uuid           : YES 1 warnings
    zlib           : YES

  User defined options
    default_library: static
    wrap_mode      : forcefallback

$ ldd .build/nvme 
        linux-vdso.so.1 (0x00007fff4c8ba000)
        libuuid.so.1 => /home/wagi/work/nvme-cli-upstream/.build/subprojects/uuid/libuuid/libuuid.so.1 (0x00007f0bae408000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f0bae1bc000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f0bae413000)

```

another thing which is strange is when looking at the libnvme static build the libuuid symbols from libuuid.a are not linked into the final libnvme.a.

```
nm -o src/libnvme.a | grep uuid
src/libnvme.a:nvme_fabrics.c.o:0000000000000250 t uuid_from_dmi_entries
src/libnvme.a:nvme_fabrics.c.o:                 U uuid_generate_random
src/libnvme.a:nvme_fabrics.c.o:                 U uuid_parse
src/libnvme.a:nvme_fabrics.c.o:                 U uuid_unparse_lower
src/libnvme.a:nvme_tree.c.o:0000000000002110 T nvme_ns_get_uuid
src/libnvme.a:nvme_tree.c.o:                 U uuid_copy
```

glimchb:
I added a commit to bump `libnvme.wrap` now

testing now with 
```
git clean -ffdx  && meson .build --wrap-mode=forcefallback  --default-library=static --debug && meson compile -C .build
```
seeing same as you...
```
$ ldd .build/nvme
        linux-vdso.so.1 (0x00007fff571f2000)
        libuuid.so.1 => /home/glimchb/nvme-cli/.build/subprojects/uuid/libuuid/libuuid.so.1 (0x00007f7779dcf000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f7779bcb000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f77799ab000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f77795e8000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f7779fd6000)
```

eli-schwartz:
> another thing which is strange is when looking at the libnvme static build the libuuid symbols from libuuid.a are not linked into the final libnvme.a.

I don't think that's ever something one would expect. Static libraries cannot have anything "linked" into them... they just require you to link to the transitive dependencies as well.

You could extract all the *.o files from libuuid.a and add them as additional sources for libnvme.a, but crossing project boundaries like this feels a bit uncomfortable...

Meson simply installs *both* projects (libnvme and util-linux) and includes instructions in the libnvme pkg-config file that you must link to uuid at the same time.

As far as linking to the shared library instead of the static library goes... the util-linux build system seems to be forcing `both_libraries()` which decreases user choice.

igaw:
> > another thing which is strange is when looking at the libnvme static build the libuuid symbols from libuuid.a are not linked into the final libnvme.a.
> 
> I don't think that's ever something one would expect. Static libraries cannot have anything "linked" into them... they just require you to link to the transitive dependencies as well.

Got it. I've never really looked into all these details :)

> You could extract all the *.o files from libuuid.a and add them as additional sources for libnvme.a, but crossing project boundaries like this feels a bit uncomfortable...

Don't think we need to be special. Just wondering what the normal usage pattern is.

> As far as linking to the shared library instead of the static library goes... the util-linux build system seems to be forcing `both_libraries()` which decreases user choice.

Understood, so we get this libuuid.so dependency via this path.

igaw:
Not sure if we can do the `--werror` with the subprojects:

```
 ../subprojects/uuid/libmount/src/tab_parse.c: In function ‘mnt_table_parse_stream’:
../subprojects/uuid/libmount/src/tab_parse.c:406:11: error: ‘num’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  406 |  fs->size = num;
      |  ~~~~~~~~~^~~~~
../subprojects/uuid/libmount/src/tab_parse.c:372:11: note: ‘num’ was declared here
  372 |  uint64_t num;
      |           ^~~
```

If I read this https://github.com/mesonbuild/meson/issues/8636 right, we can't define warning levels per subprojects.

I suggest to drop the --werror for this static build for the time being. We cover this in the other builds.

eli-schwartz:
> If I read this mesonbuild/meson#8636 right, we can't define warning levels per subprojects

That issue is saying that at the time of writing, it was impossible to set `c_args` per subproject.

It also is saying that warning_level *can* be set per subproject, and suggesting the addition of another warning level "-1".

igaw:
I thought because it is till marked as open, the report is still valid.  In this case we should just lower the warning level for the subprojects.

This would also make sense for the other builds. I don't think we are really interested to get these during our build I suppose.

eli-schwartz:
> I thought because it is till marked as open, the report is still valid.

It's reasonable to assume so.

But the issue simply has nothing to do with warning_level.

And what you actually want isn't either one, you want to pass `'werror=false'` to the subproject

> Not sure if we can do the `--werror` with the subprojects:

You can, in fact, do werror for nvme-cli and not for util-linux. You could for a long time.

glimchb:
from https://mesonbuild.com/Builtin-options.html#specifying-options-per-subproject
```
Since 0.54.0 default_library and werror built-in options can be defined per subproject. This is useful for example when building shared libraries in the main project, but static link a subproject, or when the main project must build with no warnings but some subprojects cannot.
```
and
```
Since 0.56.0 warning_level can also be defined per subproject.
```

So something like this should work:
```
-Duuid:werror=false
```

right ?

igaw:
Thanks @glimchb and @eli-schwartz!

PR #1535, opened 2022-05-24
fabrics: decode 'current' discovery subsystem

hreinecke:
TP8014 added a new discovery log page entry type '3' for the
'current discovery subsystem'; decode it accordingly.


igaw:
Looks good.

igaw (approved):

PR #1537, opened 2022-05-24
Added Vendor Unique Solidigm Garbage Collect Log.

lgdacunha:


plugins/meson.build
```diff
@@ -10,7 +10,7 @@ sources += [
   'plugins/scaleflux/sfx-nvme.c',
   'plugins/seagate/seagate-nvme.c',
   'plugins/shannon/shannon-nvme.c',
-  'plugins/solidigm/solidigm-smart.c',
+  'plugins/solidigm/solidigm-nvme.c',
```
igaw, line 13: If you add a meson.build in the solidigm subdir, just move this also there.

plugins/solidigm/solidigm-garbage-collection.c
```diff
@@ -0,0 +1,132 @@
+/*
```
igaw, line 1: Could also add the SPDX identifier to all files.

plugins/solidigm/solidigm-garbage-collection.c
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA  02110-1301, USA.
```
igaw, line 17: If you add the SPDX identifier this header license text can be dropped. I'd like to follow here the Linux kernel way [1] and just have the SPDX identifier and the copyright statement, e.g. [2].


[1] https://www.kernel.org/doc/html/latest/process/license-rules.html
[2] https://elixir.bootlin.com/linux/v5.11/source/kernel/bpf/bpf_inode_storage.c

igaw (changes requested):
a few nitpicks and missing SPDX identifiers

plugins/meson.build
lgdacunha: My first implementation was how you are suggesting but it had a side effect on the help. The Solidigm plugin would show up as the last in the plug-in list, instead of showing up in alphabetical order.
Are you ok with breaking the alphabetical order pattern?

igaw: Ah sorry about that. I was under the impression there will be only one file. 

About the ordering: When I added the initial meson.build file I was just to lazy to add to each subdir plugin dir one. I'd like to keep the alphabetical order. I think adding per plugin dir meson.build file would be the cleanest solution.

plugins/solidigm/solidigm-garbage-collection.c
lgdacunha: Done!

plugins/solidigm/solidigm-garbage-collection.c
lgdacunha: Done!

plugins/solidigm/solidigm-nvme.c
```diff
@@ -0,0 +1,38 @@
...
+
+static int get_additional_smart_log(int argc, char **argv, struct command *cmd, struct plugin *plugin)
+{
+	return solidigm_get_additional_smart_log(argc, argv, cmd, plugin);
+}
+
+static int get_garbage_collection_log(int argc, char **argv, struct command *cmd, struct plugin *plugin)
+{
+	return solidigm_get_garbage_collection_log(argc, argv, cmd, plugin);
+}
```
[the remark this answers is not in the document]
lgdacunha: Done!

plugins/solidigm/solidigm-garbage-collection.c
igaw: Thanks! If you are not against it, I would change from old (deprecated since 2.0-rc2 https://spdx.org/licenses/)  `GPL-2.0+` to new  `GPL-2.0-or-later`.

lgdacunha: Done!

igaw (approved):

Issue #1505, opened 2022-04-14
nvme connect-all reports report 'already connected' anymore


igaw:
If `nvme connect-all ` is executed more than once we get this error message:

```shell
$ .build/nvme connect-all -t tcp --traddr=10.161.8.24  --trsvcid=4421
Failed to write to /dev/nvme-fabrics: Operation already in progress
failed to connect controller, error 1006
```

The problem is that libnvme doesn't return` errno=EALREADY` anymore and the check 
for `EALREADY` in __discover() fails:

```c
			} else if (errno == EALREADY && !quiet) {
				char *traddr = log->entries[i].traddr;

				space_strip_len(NVMF_TRADDR_SIZE, traddr);
				fprintf(stderr,
					"traddr=%s is already connected\n",
					traddr);
			}
```

libnvme returns ENVME_CONNECT_WRITE instead.

igaw:
Though there is no change in behavior as we will ignore this errno completely.

GIven this, we need to make the libnvme logging less noisy and catch the in nvme-cli correct errno.

martin-gpy:
> If `nvme connect-all ` is executed more than once we get this error message:
> 
> ```shell
> $ .build/nvme connect-all -t tcp --traddr=10.161.8.24  --trsvcid=4421
> Failed to write to /dev/nvme-fabrics: Operation already in progress
> failed to connect controller, error 1006
> ```

The above error seen when 'nvme connect-all' is repeatedly invoked is actually consistent with the behavior in the nvme-cli-monolithic branch.

Problem now is after you have lowered the logging level with your commit at https://github.com/linux-nvme/libnvme/pull/351/commits/9282931a5b92892b4d3445c906062abf16899d7f, I don't see any errors at all now if 'nvme connect-all' is repeatedly invoked as shown below:

$ nvme connect-all -t fc -w nn-0x20000024ff19ab70:pn-0x21000024ff19ab70 -a nn-0x203400a098fd4ca6:pn-0x203500a098fd4ca6
$

Even the return value shows as 0 for the above (indicating no errors):

$ echo $?
0

And nothing specific logged in the messages file too for the above invocation:

...
kernel: [923480.309807][T30325] nvme nvme0: queue_size 128 > ctrl maxcmd 32, reducing to maxcmd
kernel: [923480.309838][T30325] nvme nvme0: NVME-FC{0}: controller connect complete
[...]

PR #1550, opened 2022-06-02
Log connect errors

igaw:
fabrics: Already connected uses a different error code

The libnvme library reports connection errors with it's own error
codes, e.g. ENVME_CONNECT_ALREADY stands for already connected. Update
the connect-all call so that it prints 'already connected' again.

Fixes: #1505

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -409,7 +409,7 @@ static int __discover(nvme_ctrl_t c, struct nvme_fabrics_config *defcfg,
 					nvme_disconnect_ctrl(child);
 					nvme_free_ctrl(child);
 				}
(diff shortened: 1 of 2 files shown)


igaw:
Not sure if we need to handle all error codes for `connect-all` but at least for `connect` it works out of the box already.

```
.build/nvme connect -t tcp -a 10.162.198.42 -w 10.161.16.48 -n nqn.1988-11.com.dell:powerstore:00:f03028e73ef7D032D81E -S 23fa
no controller found: invalid arguments/configuration
```

igaw:
rebase and wrap update

PR #1552, opened 2022-06-03
print: Update nvme id-ctrl ctratt fields and small cleanup

ankit-sam:


igaw:
Thanks, this was on my todo list for a while :)

PR #1561, opened 2022-06-09
libnvme wrap update, update power descriptor fields

ankit-sam:
1. update libnvme wrap
2. Add missing active power workload field
3. Use enums for power scale
4. Fix indentation

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -375,10 +375,12 @@ static void json_nvme_id_ctrl(struct nvme_id_ctrl *ctrl,
 		json_object_add_value_int(psd, "idle_power",
 			le16_to_cpu(ctrl->psd[i].idlp));
 		json_object_add_value_int(psd, "idle_scale",
-			ctrl->psd[i].ips);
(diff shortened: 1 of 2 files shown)


nvme-print.c
```diff
@@ -4553,6 +4576,8 @@ static void nvme_show_id_ctrl_power(struct nvme_id_ctrl *ctrl)
 		printf(" active_power:");
 		print_ps_power_and_scale(ctrl->psd[i].actp,
 				 nvme_psd_power_scale(ctrl->psd[i].apws));
+		printf("\n            active_power_workload:");
+		print_psd_workload(ctrl->psd[i].apws);
```
igaw, line 4580: Don't we need to update `json_nvme_id_ctrl()` too?

ankit-sam: Yes, it also needs to be updated. Will do thanks.

ankit-sam: @igaw please check now. I have also fixed idle scale which was not correct for json output.

PR #1574, opened 2022-06-17
Bash Completion fixes

Karlson2k:
Fixed:
* error when completion file is sourced for the second time
* typo resulted in broken completion for seagate plugin


igaw:
Thanks!

PR #1575, opened 2022-06-17
bash-nvme-completion.sh: a better fix for "readonly" with bash 5.x

Karlson2k:
As a follow-up for #1574

The error with one-line "readonly" with assignment on bash 5.x was fixed by 25fd8c707b106ca0763402566ad657ef710bf09e. This commit restores "readonly" functionality and adds some uniformity.


Karlson2k:
Thanks!

PR #1576, opened 2022-06-17
fabrics: Avoid out of bounds string chomping

igaw:
First move nvme_strip_chomp() to fabrics.c as it is the only
user. Also rename it to strchomp as there is nothing nvme specific.

While at it, fix the implementation of the function to handle
all the corner cases by using strnlen() to figure out the
max length of the string. After we terminate it properly,
we can start chopping off any trailing whitespace.

Also make sure that all call sides of strchomp to use the correct max
lenght of the string to avoid out of bounds access.

Also we don't need space_strip_len() to tell use what the max size of
the string is. We know it at compile time.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -108,6 +108,24 @@ static const match_table_t opt_tokens = {
 	{ OPT_ERR,		NULL		},
 };
 
+
+static inline void strchomp(char *s, int l)
+{
+	int j;
+
+	/*
+	 * make sure the string is not longer than l and
+	 * the last char is a '\0'.
+	 */
+	j = strnlen(s, l);
+	s[j] = '\0';
+
+	/* remove trailing whitespaces before the '\0'. */
+	l = j - 1;
+        while (l && s[l] == ' ')
+                s[l--] = '\0';
+}
+
 const char *arg_str(const char * const *strings,
 		size_t array_size, size_t idx)
 {
@@ -681,17 +699,6 @@ out:
 	return error;
 }
 
-static int space_strip_len(int max, const char *str)
-{
-	int i;
-
-	for (i = max - 1; i >= 0; i--)
-		if (str[i] != '\0' && str[i] != ' ')
-			break;
-
-	return i + 1;
-}
-
 static void strtolower(char *str)
 {
 	for ( ; *str; str++)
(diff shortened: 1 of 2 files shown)


igaw:
Test program:

```c
#include <stdio.h>
#include <string.h>

static inline void strchomp(char *s, int l)
{
	while (l >= 0 && (s[l] == '\0' || s[l] == ' '))
                s[l--] = '\0';
}

static inline void strchomp2(char *s, int l)
{
	int j;

	/* make sure the last char is a \0 */
	j = strnlen(s, l);
	s[j] = '\0';

	l = j - 1;
	while (l >= 0 && s[l] == ' ')
                s[l--] = '\0';
}

static int space_strip_len(int max, const char *str)
{
	int i;

        for (i = max - 1; i >= 0; i--)
		if (str[i] != '\0' && str[i] != ' ')
			break;

	return i + 1;
}


#define MAX_STRING 256
#define NVMF_TRSVCID_SIZE 8

struct test_data {
	char str[MAX_STRING];
} data[] = {
	{ .str = "abcdefg" },
	{ .str = "123456789" },
	{ .str = "" },
	{ .str = "12            \0              "},
	{  { 0,} } ,
};

{
	int i;

	for (i = 0; i < sizeof(data)/sizeof(struct test_data); i++) {
		char str[MAX_STRING];

		memcpy(str, data[i].str, sizeof(str));
		printf("- input: \'%s\'\n", str);
		printf("  strlen %ld\n\n", strlen(str));

		printf("  space_strip_len %d\n\n", space_strip_len(NVMF_TRSVCID_SIZE, str));

		strchomp(data[i].str, NVMF_TRSVCID_SIZE - 1);
		printf("  strchomp: strlen %ld\n", strlen(str));
		printf("  strchomp: \'%s\'\n\n", str);

		memcpy(str, data[i].str, sizeof(str));
		strchomp2(str, NVMF_TRSVCID_SIZE - 1);
		printf("  strchomp2: strlen %ld\n", strlen(str));
		printf("  strchomp2: \'%s\'\n\n", str);
	}
	return 0;
}
```

```
- input: 'abcdefg'
  strlen 7

  space_strip_len 7

  strchomp: strlen 7
  strchomp: 'abcdefg'

  strchomp2: strlen 7
  strchomp2: 'abcdefg'

- input: '123456789'
  strlen 9

  space_strip_len 8

  strchomp: strlen 9
  strchomp: '123456789'

  strchomp2: strlen 7
  strchomp2: '1234567'

- input: ''
  strlen 0

  space_strip_len 0

  strchomp: strlen 0
  strchomp: ''

  strchomp2: strlen 0
  strchomp2: ''

- input: '12            '
  strlen 14

  space_strip_len 2

  strchomp: strlen 14
  strchomp: '12            '

  strchomp2: strlen 2
  strchomp2: '12'

- input: ''
  strlen 0

  space_strip_len 0

  strchomp: strlen 0
  strchomp: ''

  strchomp2: strlen 0
  strchomp2: ''
```

fabrics.c
```diff
@@ -721,13 +728,9 @@ static void print_discovery_log(struct nvmf_disc_rsp_page_hdr *log, int numrec,
 		printf("subtype: %s\n", subtype_str(e->subtype));
 		printf("treq:    %s\n", treq_str(e->treq));
 		printf("portid:  %d\n", e->portid);
-		printf("trsvcid: %.*s\n",
-		       space_strip_len(NVMF_TRSVCID_SIZE, e->trsvcid),
-		       e->trsvcid);
+		printf("trsvcid: %.*s\n", NVMF_TRSVCID_SIZE - 1, e->trsvcid);
```
igaw, line 731: This is wrong. We should keep `space_strip_len()` but just make sure we don't do an out of bound access by using `NVMF_TRSVCID_SIZE - 1`.

igaw:
The issue we saw downstream is not caused by string functions. Instead it uncovered an unbreakable loop in the connect path whenever the kernel returns EINVAL. I'll open an issue on this.

PR #1578, opened 2022-06-21
docs: Document nvme-id-ns behavior when nsid is 0xFFFFFFFF

zwhfly:
as documented in NVM Command Set Specification Revision 1.0b,
chapter 4.1.5.1 - NVM Command Set Identify Namespace Data Structure (CNS 00h).


igaw:
Looks good, but could just update the `.txt` file. The other files are generated (for every release) from the `.txt` file.

igaw:
Thanks!

PR #1581, opened 2022-06-22
create-ns: fix create-ns fail when -b 4096 while id-ns return nlbaf=1…

AndyHouHit:
… and 0 for 512, 1 for 4096

Fail case:
sudo nvme create-ns /dev/nvme0 --nsze=1875366486   --ncap=937684566  --block-size=4096  --timeout=120000
   FLBAS corresponding to block size 4096 not found
   Please correct block size, or specify FLBAS directly

Signed-off-by: Wei Hou <wei.hou@scaleflux.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2438,7 +2438,7 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 			}
 			goto close_fd;
 		}
-		for (i = 0; i < ns.nlbaf; ++i) {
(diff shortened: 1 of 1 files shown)


AndyHouHit:
> Ah, this look way better. Thanks!
> 
> Just on last thing. The commit message format, it should be
> 
> ```
> Summary Line
> 
> Commit message
> ```
> 
> The empty line between the summary line and the commit message is important. Without it, `git log` will assume the complete message is the summary line.
> 

Done.

PR #1583, opened 2022-06-22
Added support to Solidigm Latency Tracking.

lgdacunha:


igaw:
I update the patch to be more consistent with the whole code base. No functional changes.
 - a bunch of small whitespace changed
 - function arguments alignment
 - variable declaration moved up to the basic block header.

PR #1584, opened 2022-06-23
plugins: Add Dell plugin with basic id-ctrl

glimchb:
Signed-off-by: Boris Glimcher <Boris.Glimcher@emc.com>


igaw:
So far it looks good

igaw:
I was already typing that it would be nice to have -H support but I see we don't pass this flag down.

Anyway, looks good.

PR #1586, opened 2022-06-28
admin_passthru cmd, only default set read/write for one-directional  opcode

AndyHouHit:
The nvme spec required the opcode of admin_passthru cmd should follow the rules: All commands, including vendor specific commands, shall follow this convention: 00b = no data transfer; 01b = host to controller; 10b = controller to host; 11b = bidirectional.

The question is about the 11b birectional, for this type of opcode, after changes of the commit:
    commit 3a7c07b2370810fe6ef04c5566345806da796014
      Author: Steven Seungcheol Lee sc108.lee@samsung.com
      Date:   Tue Apr 12 14:45:46 2022 +0900
      nvme: Fix error when passthru does not match opcode based flags

will force to do write then read operation for this bi-directional opcode.
However, the bidirectional should means do read or write. Some of the Vendor using the 11b type opcode for read or write VU operations.

Signed-off-by: Wei Hou <wei.hou@scaleflux.com>


keithbusch:
Fwiw, the kernel doesn't really support bidirectional commands.

sc108-lee:
1. As @keithbusch mentioned, there is no support bidirectional command.
2. There is no command that supports bidirectional.
3. Even if you wants support bidirectional command, the fix is not fully support command
    - file opened without flags initialization
    - input, output file will be same

AndyHouHit:
> 1. As @keithbusch mentioned, there is no support bidirectional command.
> 2. There is no command that supports bidirectional.
> 3. Even if you wants support bidirectional command, the fix is not fully support command

My propose is: If you use a  bidirectional op code, you can use it to read or write. Not do read & write at the same time.  But there is a limitation,  for bidirectional op code, you have to provides -b or -w option. 
This will make vendor defined VU more conveniently. The same op code provides read or write capability,

keithbusch:
I won't do much good considering the kernel does not comprehend bidirectional commands. The current code will dma map it as a data-out command, and a capable iommu would prevent data-in access, or bounce buffered payloads will be copied the wrong way.

sc108-lee:
@AndyHouHit I think you just want to use passthru command conveniently.
But that will be spec violation, if it's allowed.

Indicates the data transfer direction of the command. All options to the command shall transfer data as specified or transfer no data. **All commands, including vendor specific commands, shall follow this convention:** 00b = no data transfer; 01b = host to controller; 10b = controller to host; 11b = bidirectional

AndyHouHit:
> @AndyHouHit I think you just want to use passthru command conveniently. But that will be spec violation, if it's allowed.
> 
> Indicates the data transfer direction of the command. All options to the command shall transfer data as specified or transfer no data. **All commands, including vendor specific commands, shall follow this convention:** 00b = no data transfer; 01b = host to controller; 10b = controller to host; 11b = bidirectional

Yes, here is the argument point. How do you understand the bidirectional in spec?  It means must do write then read or means write or read what ever you want. How do you think? @sc108-lee  @keithbusch

AndyHouHit:
@keithbusch  @sc108-lee  @igaw , any more comment about this ?  If you do confirm the bidirectional means must to do write then read, I will discard this PR.

keithbusch:
It's impossible to conclude anything at this level because there are no defined bidirectional commands. This tool can't provide support for it because the kernel doesn't support it. The kernel doesn't support it because there are no defined use cases in spec or otherwise.

AndyHouHit:
> It's impossible to conclude anything at this level because there are no defined bidirectional commands. This tool can't provide support for it because the kernel doesn't support it. The kernel doesn't support it because there are no defined use cases in spec or otherwise.

So, can you accept this change ?  Otherwise, all opcode for format 11b not works because it force to do write then read now.

keithbusch:
The main problem is the kernel is not going to do what you want with such an opcode. It will treat it as a write, including dma mappings and bounce buffering. You might find some unusual scenario where you could make it do something you expect, but we can't deterministically make it work. No point in supporting it in tooling without kernel support, and kernel won't support it without a spec defined bi-directional command.

PR #1587, opened 2022-06-30
Added average latency field to solidigm latency-tracking-log.

lgdacunha:
Removed -b flag option from solidigm plug-in, in favor of "-o binary".


lgdacunha:
I see "Auto Request Review" failed for this pull request . I will try to resubmit it.

PR #1588, opened 2022-06-30
Added average latency field to solidigm latency-tracking-log.

lgdacunha:
Removed -b flag option from solidigm plug-in, in favor of "-o binary".


igaw:
It seems I got the github action wrong configured. I'll fix this. Just a sec.

igaw:
FYI, I've rebased it and massaged the commit message. 

Please keep in mind the commit message consistent of a summary line followed by an empty line, then the  commit message body, e.g.

```
solidigm: Add new feature foo

Add new festeature foo which supports bar use case.

Signed-off-by: Harry Hacker <hh@example.com>
```

igaw:
I see the auto review request bot hasn't been trigged with the push. I suppose will see if my fix works for the next PR.

PR #1590, opened 2022-07-01
rpmbuild: Do docs build to avoid error from missing .1, .html files

sc108-lee:
Program asciidoc found: YES (/usr/bin/asciidoc)
Documentation/meson.build:253:8: ERROR: File nvme-wdc-cloud-boot-SSD-version.1 does not exist.


igaw:
Don't worry about the auto request review build failure. Still got the reviewers.xml not right

Issue #1565, opened 2022-06-13
Use `opcode ==` checking if it's a read or write op instead of  `&1`


igaw:
Just use the `opcode ==` approach for testing if it's a read or write operation, instead
of

```c
	else if (opcode & 1)
		err = nvme_write(&args);
	else
		err = nvme_read(&args);
```

_Originally posted by @igaw in https://github.com/linux-nvme/nvme-cli/pull/1564#issuecomment-1153555455_

igaw:
The fix from @ikegami-t fixed this issue.

PR #1564, opened 2022-06-09
submit_io(): fix "nvme compare" to pass correct command

jlevon:
Since "18de3a6d Convert to libnvme", an "nvme compare" was being
converted incorrectly into an "nvme write" in submit_io().

Fixes #1562.

Signed-off-by: John Levon <john.levon@nutanix.com>

PR #1597, opened 2022-07-06
nvme: Fix submit_io to use nvme_io for #1565

ikegami-t:
It was fixed for the compare command by 9b6d3fca (#1564).
But still other I/O commands except write and read commands are not implemented
correctly to pass valid opcode value.
  Note: This is different from to use "opcode ==" checking mentioned by #1565
        but it should be worked correctly.


igaw:
FWIW, I've reworded the commit message. Thanks!

PR #1599, opened 2022-07-07
nvme: Check timestamp format result and increase string size

ikegami-t:
Since the default size is possible to be exceeded for a locale date and time.

### DIFF
--- a/Documentation/cmd-plugins.txt
+++ b/Documentation/cmd-plugins.txt
@@ -65,7 +65,7 @@ linknvme:nvme-transcend-badblock[1]::
 	Retrieve Transcend NVMe device's bad blocks
 
 linknvme:nvme-transcend-healthvalue[1]::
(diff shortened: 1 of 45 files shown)


igaw (changes requested):
Good catch. Though I think I would just make the default size of the buffer bigger (as in this patch to 320 bytes or something more aligned as 256) and test for the return code of `strftime` and set the buffer to `'-\0'`.  I think the realloc part is a bit over engineered.

ikegami-t:
> Good catch. Though I think I would just make the default size of the buffer bigger (as in this patch to 320 bytes or something more aligned as 256) and test for the return code of `strftime` and set the buffer to `'-\0'`. I think the realloc part is a bit over engineered.

Thanks for your review comment. Fix the patch as suggested.

igaw (approved):

igaw:
Nice cleanups! Thanks a lot!

PR #1600, opened 2022-07-08
    nvme: Use correct json serializing for long doubles

### LINKED ISSUE #1374: long double does not fit with json-c
https://github.com/linux-nvme/nvme-cli/blob/master/nvme-print.c#L188

https://github.com/linux-nvme/nvme-cli/blob/master/nvme.h#L47

128 bits long double is not fit in 64 bits double
https://github.com/json-c/json-c/blob/master/json_object.c#L1066)

igaw:
Fixup all places which use json_object_add_value_float() to serialize
long double types with json_object_add_value_double(). The float
version cuts off the upper 64 bits of the 128 bit long double type.

Fixes #1374


ids1024:
This is no longer conditionally compiling based on `CONFIG_JSONC` and `CONFIG_JSONC_14`. Which seems good, but `meson.build` should also be changed to make `json-c >= 0.14` a hard dependency.

igaw:
The 2.x version introduced the hard dependency on json-c. The `CONFIG_JSONC` defines are just a left over. There is no need to drop support for the 0.13 which is still present in long time support distros. It doesn't hurt.

ids1024:
It isn't a hard dependency as `meson.build` is written, I believe:

```meson
json_c_dep = dependency('json-c', version: '>=0.13', fallback : ['json-c', 'json_c_dep'])
if json_c_dep.found()
  conf.set('CONFIG_JSONC', true, description: 'Is json-c available?')
  if json_c_dep.version().version_compare('>=0.14')
    conf.set('CONFIG_JSONC_14', true, description: 'Is json-c at least 0.14?')
    requires = 'Requires: json-c >= 0.14'
  else
    requires = 'Requires: json-c >= 0.13'
  endif
endif
```

That should pass `required: true` to `dependency()`. And then the `if json_c_dep.found()` and `conf.set` are unneeded.

Though for `CONFIG_JSONC_14`, I misread the diff and see it's still used in the new version. So leaving support for json-c 0.13 is good, assuming it actually works.

igaw:
Just a sec, I add cleanup patch to this series. Makes more sense.

PR #1608, opened 2022-07-12
plugins: minor fixes

jk-ozlabs:
This series is a set of minor fixes for various plugins: a couple of fixes in error paths, plus removing some unused `fd` arguments for display functions.

[this is in preparation for a RFC change to the way we're passing the nvme devices around]

### DIFF
--- a/ccan/meson.build
+++ b/ccan/meson.build
@@ -7,6 +7,6 @@ sources += files([
 ])
 
 if get_option('buildtype') == 'debug'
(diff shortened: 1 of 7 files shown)


jk-ozlabs:
Added a couple of the `more-fixes` changes into this branch, should result in fewer PRs!

igaw:
Updated version of this PR is in #1634

PR #1609, opened 2022-07-12
RFC: Make device data non-global, contained within a `struct nvme_dev`

jk-ozlabs:
This draft change is a bit of an experiment on the path of using `nvme-cli` to also handle NVMe-MI transports.

At present, we have a few bits of data that pertain to the nvme device handle:

 * a global `devicename` string
 * local `fd`s to pass to the libnvme API, and/or perform direct `ioctl()`s
 * a static `struct stat`

This series attempts to tidy that up a little, by moving these three into a single new struct (`struct nvme_dev`), which we allocate on `parse_and_open()`. The intention will be to later make this flexible enough to use a MI-specific handle (`nvme_mi_ep_t`) alongside the `fd`. 

This patchset is quite large, as almost every function in the core and plugins assumes direct access to the `fd`. However, the changes are mostly mechanical - replacing `int fd` with `struct nvme_dev` + `dev->fd`.

### DIFF
--- a/ccan/meson.build
+++ b/ccan/meson.build
@@ -7,6 +7,6 @@ sources += files([
 ])
 
 if get_option('buildtype') == 'debug'
-    add_project_arguments('-DCCAN_LIST_DEBUG=1',  language : ['c', 'cpp'])
-    add_project_arguments('-DCCAN_STR_DEBUG=1',  language : ['c', 'cpp'])
+    add_project_arguments('-DCCAN_LIST_DEBUG=1',  language : ['c'])
+    add_project_arguments('-DCCAN_STR_DEBUG=1',  language : ['c'])
 endif
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 ################################################################################
 project(
-    'nvme-cli', ['c', 'cpp'],
+    'nvme-cli', ['c'],
     meson_version: '>= 0.47.0',
     license: 'GPL-2.0-only',
     version: '2.1.1',
--- a/nvme-rpmb.c
+++ b/nvme-rpmb.c
@@ -872,8 +872,9 @@ int rpmb_cmd_option(int argc, char **argv, struct command *cmd, struct plugin *p
 	unsigned char *msg_buf = NULL;
 	unsigned int msg_size = 0;
 	unsigned int key_size = 0;
-	int fd = -1, err = -1;
 	struct nvme_id_ctrl ctrl;
+	struct nvme_dev *dev;
+	int err = -1;
 
 	union ctrl_rpmbs_reg {
 		struct {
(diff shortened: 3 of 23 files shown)


igaw:
No objections from my side. Grouping these togehter makes sense. The only thing I am pondering on is if we should start to prefix the function in nvme-cli with `nvme_cli` to avoid name clash with the libnvme library. But this is something we can change later and on the complete code base.

jk-ozlabs:
@iagw: thanks for taking a look; this is a fairly intrusive change, so I'm cautious about going off on a direction that might not be acceptable later.

I'm happy to do the work on avoiding namespace clashes as a future change; as part of this, I've been messing with coccinelle a bit, which might make bulk renames easier.

In terms of future work on this series: the main design decision will be how we allow different transports. Because the libnvme-mi API is split:

```c
    int nvme_identify(int fd, ...);
```
vs:
```c
    int nvme_mi_identify(nvme_mi_ep_t ep, ...);
```
We would need to split the library calls in the nvme-cli code, along the lines of:
```c
    if (dev->type == NVME_DEV_DIRECT) {
        nvme_identify(dev->fd, ...);
    } else if (dev->type == NVME_DEV_MI) {
        nvme_mi_identify(dev->ep, ...);
    }
```
... but we could neaten this by adding suitable wrappers using the `nvme_cli_` namespace:
```c
    int nvme_cli_identify(struct nvme_dev *dev, ...)
```
which handles the dev types appropriately.

The alternative is to do the transport abstraction in libnvme instead, but as you've said elsewhere, that's probably a consideration for a future libnvme API, rather than doing now.

jk-ozlabs:
D'oh, sorry about that, and thanks for letting me know. I'll double-check mentions in future!

igaw:
If you get coccinelle working it is very handy :)

Anyway, I think as the current libnvme API distinguishes between libnvme and libnvme-mi it would be rather unfortunate to see the check which API is used in all calling places. I don't see another way than to introduce wrappers. We can't really change the libnvme API. This is clearly something we have to think about when doing a libnvme v2. 

BTW, how many wrappers are needed? I haven't dug into the MI spec yet, so I don't have a complete understanding what is even possible.

hreinecke:
While we're at it: the one problem we have with libnvme that we are pretty careless with filedescriptors. As soon as we're scanning a NVMe device we're opening a filedescriptor, irrespective on whether we need it or not.
What I'd like to see is to move that into an on-demand thing, ie opening the filedescriptor only when we're about to do an action on it (like calling ioctl() and the like).
Otherwise we're risking using up all available filedescriptors on large installations just by initialising the library.
Might be another patchset, but still something to keep in mind.

jk-ozlabs:
@igaw : I've just been working on doing some prototype wrappers today. Since the ioctl and MI APIs are very similar, I think it'll be manageable. 

In terms of numbers: the MI spec exposes all of the Admin commands, so (in theory) we'd need a wrapper for each of the Admin-command-related libnvme functions.

I'll push another RFC branch up soon, with partial coverage of the command set, so you can see what it looks like.

@hreinecke : ok, good to know. For a future (v2?) libnvme API, I'd propose an opaque structure to contain the transport details, which would be the caller's reference to the device, rather than the fd directly - a little bit similar to what we're doing here. Once that's done, it'll be fairly straightforward to open file descriptors on-demand too.

jk-ozlabs:
OK, prototype wrappers are up, but we should probably move discussion to something more general than this PR. So, I've filed https://github.com/linux-nvme/nvme-cli/issues/1613 to contain that discussion; this also has references to the proposed wrapper implementations.

We probably want to hold off on this PR until we're OK with #1613, so I'll keep it as a draft for now.

PR #1619, opened 2022-07-21
print: Fix nvme_id_uuid_list

jeffreyalien:

The previous code would skip the first entry in the list.  This fix will print all entries.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -5661,7 +5661,7 @@ void nvme_show_id_uuid_list(const struct nvme_id_uuid_list *uuid_list,
 
 	/* The 0th entry is reserved */
 	printf("NVME Identify UUID:\n");
-	for (i = 1; i < NVME_ID_UUID_LIST_MAX; i++) {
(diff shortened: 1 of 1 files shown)


nvme-print.c
```diff
@@ -5685,7 +5685,7 @@ void nvme_show_id_uuid_list(const struct nvme_id_uuid_list *uuid_list,
 				break;
 			}
 		}
-		printf(" Entry[%3d]\n", i);
+		printf(" Entry[%3d]\n", i+1);
```
igaw, line 5688: This is leads to an out of bound access. You need to update condition accordingly.

igaw (changes requested):

nvme-print.c
jeffreyalien: The i+1 is only used to print the entry number.  It is not used to access any storage nor increment the value of i.  I don't see the issue here.

igaw: Indeed. I saw something what wasn't there. I blame the weather :)

igaw (approved):

PR #1621, opened 2022-07-22
nvme: Set default rae value for nvme_get_nsid_log users

igaw:
libnvme 1.0 hard codes the rae value for all nvme_get_nsid_log to
false. The upcoming 1.1 release will pass the rae value through
to the kernel. This is causing a behavior change. To be on the safe
side set all rae=true to false.

At least for nvme_get_log_error and nvme_get_log_smart this seems also
what the spec recommends.

Reported-by: Jeremy Kerr <jk@codeconstruct.com.au>

See #https://github.com/linux-nvme/libnvme/issues/437

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -351,7 +351,7 @@ static int get_smart_log(int argc, char **argv, struct command *cmd, struct plug
 	if (cfg.human_readable)
 		flags |= VERBOSE;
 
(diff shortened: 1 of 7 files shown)


jk-ozlabs:
Looks good - now we can enable in the individual cases where needed.

Issue #1618, opened 2022-07-20
nvme connect-all fails to process dhchap_ctrl_key


martin-gpy:
Turns out nvme connect-all processes the dhchap_key, but not the dhchap_ctrl_key. For e.g. dhchap_key shows up properly below but dhchap_ctrl_key shows up as 'none' below:

```
# nvme connect-all -t tcp -a 10.228.36.89 -w 10.228.32.190 -S DHHC-1:00:QaxYTIQLMUZOE+\/dmIForZfP5Mf9OX0Hiy8+uhG8kdOjrVxN: -C DHHC-1:00:2hfbQDMe7VIZngfJM6DbQ+5DKNzgUOjymfxl9LYs2wcZ7Qym:

# nvme list
Node                  Generic               SN                   Model                                    Namespace Usage                      Format           FW Rev
--------------------- --------------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
/dev/nvme1n1          /dev/ng1n1            80A2rNT89y0FAAAAAAAB NetApp ONTAP Controller                  1         524.29  MB / 524.29  MB      4 KiB +  0 B   FFFFFFFF

# cat /sys/class/nvme-subsystem/nvme-subsys1/nvme1/dhchap_secret
DHHC-1:00:QaxYTIQLMUZOE+/dmIForZfP5Mf9OX0Hiy8+uhG8kdOjrVxN:

# cat /sys/class/nvme-subsystem/nvme-subsys1/nvme1/dhchap_ctrl_secret
none
```

martin-gpy:
But this works for nvme connect though. For e.g. 

```
# nvme connect -t tcp -a 10.228.36.89 -w 10.228.32.190 -n nqn.1992-08.com.netapp:sn.94d3e4c5f3ed11eca5871c666dbe2b2a:subsystem.ss_tcpnvme_2_1 -S DHHC-1:00:QaxYTIQLMUZOE+\/dmIForZfP5Mf9OX0Hiy8+uhG8kdOjrVxN: -C DHHC-1:00:2hfbQDMe7VIZngfJM6DbQ+5DKNzgUOjymfxl9LYs2wcZ7Qym:

# nvme list
Node                  Generic               SN                   Model                                    Namespace Usage                      Format           FW Rev
--------------------- --------------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
[...]

PR #1622, opened 2022-07-26
fabrics: Pass ctrl key to connect logic

igaw:
The dhchap control key needs to be passed into libnvme via the new nvmf_connect_disc_entry2 in order to support the -S and -C options in the connect-all case.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/1618


igaw:
Drop it. The controller host key via command line doesn't really make sense as we might have more than one controller with different keys. The command line is not flexible to handle this kind of configuration. We better remove the command line argument and defer to the json config approach here.

PR #1630, opened 2022-08-10
Add MI support to `nvme-cli`

jk-ozlabs:
This series adds MI channel support for `nvme-cli`, by adding a new format of device specifier:

    nvme id-ctrl mctp:1,9

We do this by introducing a new struct to contain the "transport" to the device. For the existing ioctl-based cases, this has the device fd. When using the MI transport, this contains the MI endpoint (`nvme_mi_ep_t`):

```c
enum nvme_dev_type {
	NVME_DEV_DIRECT,
	NVME_DEV_MI,
};

struct nvme_dev {
	enum nvme_dev_type type;
	union {
		struct {
			int fd;
			struct stat stat;
		} direct;
		struct {
			nvme_root_t root;
			nvme_mi_ep_t ep;
			nvme_mi_ctrl_t ctrl;
		} mi;
	};

	const char *name;
};
```

.. and then the majority of the changes involve passing this struct to each of the command implementations, rather than the fd.

This change is on top of my `fixes` PR: https://github.com/linux-nvme/nvme-cli/pull/1608 , so currently includes those commits too. I'd suggest merging that one first to keep the changes distinct.

Of course, any questions/comments/etc are more than welcome!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -46,6 +46,8 @@ conf.set('SYSCONFDIR', '"@0@"'.format(sysconfdir))
 # Check for libnvme availability
 libnvme_dep = dependency('libnvme', version: '>=1.1', required: true,
                          fallback : ['libnvme', 'libnvme_dep'])
+libnvme_mi_dep = dependency('libnvme-mi', required: true,
+                         fallback : ['libnvme', 'libnvme_mi_dep'])
 
 # Check for libuuid availability
 libuuid_dep = dependency('uuid', required: true,
@@ -258,7 +260,7 @@ subdir('Documentation')
 executable(
   'nvme',
   sources,
-  dependencies: [ libnvme_dep, libuuid_dep, json_c_dep, libz_dep,
+  dependencies: [ libnvme_dep, libnvme_mi_dep, libuuid_dep, json_c_dep, libz_dep,
                   libhugetlbfs_dep ],
   link_args: '-ldl',
   include_directories: incdir,
--- a/meson.build
+++ b/meson.build
@@ -247,6 +247,7 @@ sources = [
   'nvme-models.c',
   'nvme-print.c',
   'nvme-rpmb.c',
+  'nvme-wrap.c',
   'plugin.c',
   'wrapper.c',
 ]
--- a/nvme-rpmb.c
+++ b/nvme-rpmb.c
@@ -872,8 +872,9 @@ int rpmb_cmd_option(int argc, char **argv, struct command *cmd, struct plugin *p
 	unsigned char *msg_buf = NULL;
 	unsigned int msg_size = 0;
 	unsigned int key_size = 0;
-	int fd = -1, err = -1;
 	struct nvme_id_ctrl ctrl;
+	struct nvme_dev *dev;
+	int err = -1;
 
 	union ctrl_rpmbs_reg {
 		struct {
@@ -886,11 +887,11 @@ int rpmb_cmd_option(int argc, char **argv, struct command *cmd, struct plugin *p
 		unsigned int rpmbs;
 	} regs;
 	
-	if ((fd = parse_and_open(argc, argv, desc, opts)) < 0)
-		return fd;
+	if ((err = parse_and_open(&dev, argc, argv, desc, opts)) < 0)
+		return err;
 	
 	/* before parsing  commands, check if controller supports any RPMB targets */
-	err = nvme_identify_ctrl(fd, &ctrl);
+	err = nvme_identify_ctrl(dev->fd, &ctrl);
 	if (err)
 		goto out;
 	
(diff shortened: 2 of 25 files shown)


igaw:
@jk-ozlabs I assume you are going to rebase this now?

jk-ozlabs:
> @jk-ozlabs I assume you are going to rebase this now?

You beat me to it! Yep, just rebased and pushed.

igaw:
Looked through the commits. Looks in good shape. A few minor things

- There are some whitespace damaged spaces instead tabs. e.g in the first few changes in`Combine NVMe file descriptor into struct nvme_dev` this happens pretty often, then it stops. I suppose you edited first by hand and started to use tools :). You might be able to fix this by using sed on top of the patch. Maybe this helps https://superuser.com/questions/1405022/replace-leading-tabs-and-spaces-with-sed
If this is too complicated (I know you spend a lot of time to get this to this point) I might just fix those lines myself before merging.
- bikeshedding alarm: The name of the two macros `do_admin_args_op` and `do_admin_op` sound a bit off for me. They are just function wrappers, not sure what the admin part is about :)
- The last change introduces new files under `GPL-2.0-only`. Could I convince you provide these under `GPL-2.0-or-later`?

jk-ozlabs:
Thanks for taking a look! In reply to your points:

>     * There are some whitespace damaged spaces instead tabs. e.g in the first few changes in`Combine NVMe file descriptor into struct nvme_dev` this happens pretty often, then it stops. I suppose you edited first by hand and started to use tools :). You might be able to fix this by using sed on top of the patch. Maybe this helps https://superuser.com/questions/1405022/replace-leading-tabs-and-spaces-with-sed

The larger changes were part coccinelle + a bit of other tooling; I'll take a look and get these fixed up.
 
>     * bikeshedding alarm: The name of the two macros `do_admin_args_op` and `do_admin_op` sound a bit off for me. They are just function wrappers, not sure what the admin part is about :)

These are in reference to the Admin command-set; they might come in handy when we have the MI command-set implemented too. However, these are completely self-contained, so there's no issue with renaming later (when the MI commands are added) if necessary. I'm happy to switch to whatever you prefer.

>     * The last change introduces new files under `GPL-2.0-only`. Could I convince you provide these under `GPL-2.0-or-later`?

Yes, absolutely. Will fix up the license to match the rest of the project.

igaw:
> The larger changes were part coccinelle + a bit of other tooling; I'll take a look and get these fixed up.

I know, it's a bit silly but trying to keep things consistent pays out in the end. At least this is my experience. 

> These are in reference to the Admin command-set; they might come in handy when we have the MI command-set implemented too. However, these are completely self-contained, so there's no issue with renaming later (when the MI commands are added) if necessary. I'm happy to switch to whatever you prefer.

True, I didn't think about this too deeply. I just thought there is no 'admin' in the `nvme_identify` function name... Let's keep it then. As I said it's bikeshedding .

> Yes, absolutely. Will fix up the license to match the rest of the project.

Thanks!

jk-ozlabs:
> I know, it's a bit silly but trying to keep things consistent pays out in the end. At least this is my experience.

I am 100% on board with getting it right in the PR, rather than having to fix up!

I've just updated the branch with the whitespace fixes, plus the correct SPDX headers.

For the whitespace changes: Some of the plugins use spaces instead of tabs, and some with different indent levels. I've tried to keep my changes consistent on a per-plugin basis, but all of the core bits should be as you expect - let me know if not.

[it looks like there were more errors than just from the automated changes - so I can't blame it all on the tools!]

jk-ozlabs:
... and another small change to default to controller ID 0 (rather than 1), if it isn't specified in the device string.

igaw:
> > I know, it's a bit silly but trying to keep things consistent pays out in the end. At least this is my experience.
> 
> I am 100% on board with getting it right in the PR, rather than having to fix up!

:)
 
> I've just updated the branch with the whitespace fixes, plus the correct SPDX headers.

Thanks!

> For the whitespace changes: Some of the plugins use spaces instead of tabs, and some with different indent levels. I've tried to keep my changes consistent on a per-plugin basis, but all of the core bits should be as you expect - let me know if not.

Yes, that is also what I do. The plugins have all sorts for coding style, I think it would make sense to address this, eventually. But it this has a low prio right now. The only thing I try to enforce is that new plugins should follow the coding style.

> [it looks like there were more errors than just from the automated changes - so I can't blame it all on the tools!]

all good. let's get this PR in asap.

igaw:
I've added a prefix to all commit message, to make it consistent with the other commit messages. But I was not able to push the changes back here, so I went ahead and merged it directly. Hope this is okay.

Just to make it clear. The changes are merged into master.

Thanks!

igaw:
just to keep track of fallouts:
```
../nvme.c: In function ‘ns_rescan’:
../nvme.c:4395:19: warning: ‘dev’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  struct nvme_dev *dev;
                   ^~~
```

jk-ozlabs:
> ```
> ../nvme.c: In function ‘ns_rescan’:
> ../nvme.c:4395:19: warning: ‘dev’ may be used uninitialized in this function [-Wmaybe-uninitialized]
>   struct nvme_dev *dev;
>                    ^~~
> ```

Hm, not seeing that here, either with gcc or clang. Is this with any particular meson config?

igaw:
I got this with gcc-7 on freshly installed openSUSE Leap 15.4 box.... but I was using --buildtarget=debug --optimization=0

jk-ozlabs:
Still no dice with those options on gcc-9 to gcc-11, and I can't see any path where where rc is zero and dev is uninitialised.

Is this just in `ns_rescan()`? It's almost identical (in its handling of `dev`) to `reset()` right above.

I can initialise `dev` to `NULL` to silence the warning, but would be nice to figure out what's happening here...

igaw:
Thanks for trying out all these compilers. If they don't trigger let's not bother any longer.  gcc-7 is outdated.

PR #1632, opened 2022-08-10
json:  Fix json_object_add_value_uint64 wrapper

jeffreyalien:
The current wrapper does not correctly handle 64 bit  values.  It appears the int/uint32 functions and values were copied and then never updated for 64 bit values.  The function works fine for values under 0xffffffff and that's maybe why nobody noticed this issue before.

### DIFF
--- a/util/json.h
+++ b/util/json.h
@@ -16,14 +16,14 @@
 	json_object_object_add(o, k, json_object_new_int(v))
 #ifdef CONFIG_JSONC_14
 #define json_object_add_value_uint64(o, k, v) \
-	json_object_object_add(o, k, json_object_new_uint64(v))
+	json_object_object_add(o, k, json_object_new_int64(v))
 #else
 #define json_object_add_value_uint64(o, k, v) \
-	if ((v) > UINT_MAX) {						\
-		fprintf(stderr, "Value overflow in %s", k);		\
-		json_object_object_add(o, k, json_object_new_int(-1));	\
-	} else								\
-		json_object_object_add(o, k, json_object_new_int(v))
+	if ((v) > ULONG_MAX) { \
+		fprintf(stderr, "\nValue overflow in %s, v = %ld\n", k, (long)v); \
+		json_object_object_add(o, k, json_object_new_int64(-1)); \
+	} else \
+		json_object_object_add(o, k, json_object_new_int64(v))
 #endif
 #define json_object_add_value_double(o, k, v) \
 	json_object_object_add(o, k, util_json_object_new_double(v))


jeffreyalien:
Looks like there's a compile issue with the changes I made for this pull request.  I'll look into them and update the request once fixed.

igaw:
Could you also add the information you added here in this issue as commit message? Currently, it just says it fixes something without explanation. The git log should be as self contained as possible because we should not relay on the github platform to strongly.

util/json.h
```diff
@@ -16,14 +16,14 @@
 	json_object_object_add(o, k, json_object_new_int(v))
 #ifdef CONFIG_JSONC_14
 #define json_object_add_value_uint64(o, k, v) \
-	json_object_object_add(o, k, json_object_new_uint64(v))
+	json_object_object_add(o, k, json_object_new_int64(v))
```
igaw, line 19: This doesn't make sense at all. Why should we use the int64 version of the function?

util/json.h
```diff
@@ -16,14 +16,14 @@
...
 #else
 #define json_object_add_value_uint64(o, k, v) \
-	if ((v) > UINT_MAX) {						\
-		fprintf(stderr, "Value overflow in %s", k);		\
-		json_object_object_add(o, k, json_object_new_int(-1));	\
-	} else								\
-		json_object_object_add(o, k, json_object_new_int(v))
+	if ((v) > ULONG_MAX) { \
```
igaw, line 22: v can never be larger than ULONG_MAX unless we have 128 bit processors.

util/json.h
```diff
@@ -16,14 +16,14 @@
...
+		fprintf(stderr, "\nValue overflow in %s, v = %ld\n", k, (long)v); \
+		json_object_object_add(o, k, json_object_new_int64(-1)); \
```
igaw, line 24: And did you check if this code actually compiles against json-c v0.13? The very reason we have this workaround here is because older version of json-c do not have this API.

3c3b5920f77c ("add uint64 data to json-c")

```
$ git tag --contains 3c3b5920f77c751471cd46b92f9d3f5aa6ff2557                                                                                                                json-c-0.14-20200419
json-c-0.15-20200726
json-c-0.16-20220414
```

igaw:
Nope, this is completely wrong.

jeffreyalien:
@igaw Maybe there's a better way to fix but the current code is does not handle values greater then 0xFFFFFFFF.  Here's the data from the cap field of  the show-regs command from the current code at the master branch:
t630-rh8:root:05:37 ~/nvmeCliKB > .build/nvme show-regs /dev/nvme0
**cap     : 100030500303ff**
t630-rh8:root:05:37 ~/nvmeCliKB > .build/nvme show-regs /dev/nvme0 -o json
Value overflow in capValue overflow in asqValue overflow in acq{
  **"cap":-1,**

With the patch:
t630-rh8:root:05:36 .../include/json-c > nvme show-regs /dev/nvme0 -o json
{
  **"cap":4503807128175615,**  
which is  100030500303ff in hex

This is from a system with json version 0.13.1 so the change I made does work there.

igaw:
Well, one thing you could do is print it into a string. So first you would print the upper 32 bits and then the lower ones.

igaw:
See` util_json_object_new_double()` as example

igaw:
Thinking about it, you could test with meson if the function exists, if not you provide our own implementation. In this case we could even drop the test on the version of json-c.

jeffreyalien:
Ok, here's a new suggested fix:  
#ifdef CONFIG_JSONC_14
#define json_object_add_value_uint64(o, k, v) \
	json_object_object_add(o, k, json_object_new_uint64(v))
#else
#define json_object_add_value_uint64(o, k, v) \
		json_object_object_add(o, k, json_object_new_int64(v))
#endif

Basically getting rid of  if v > UINIT_MAX check and changing json_object_new_int(v) to json_object_new_int64(v)

I verified that json_object_new_int64 has been supported back to tag 0.10 tag.  It doesn't make sense to me why json_object_new_int was ever used to handle a uint64 value, but I could be missing something.  I'm not real familiar with the json c code.

json_object_new_uint64 was added in the 0.14 tag so it makes sense to use that call for version 14 and newer.

PR #1637, opened 2022-08-11
docs: Update formatting for sanact list

igaw:
asciidoc creates list with either a '*' or '-'. Without it this is
rendered as one paragraph which is hard to read.

While at it, also use hex values to match with example.


igaw:
Just realized 'nvme format' uses table for such things. This should  use the same format.

PR #1638, opened 2022-08-11
nvme: Add helper function to parse 16-bit comma separated list

Francis-Pravin:
In copy command, nlbs is declared as __u16. While parsing command line argument,
it is typecast to integer pointer. So nlbs[1], nlbs[3], nlbs[5], ... are stored
as zero after getting parsed. Therefore it is passing wrong value to controller.
Hence, adding a helper function to parse the 16-bit comma separated list

Signed-off-by: Francis Pravin Antony Michael Raj <francis.michael@solidigm.com>
Signed-off-by: Jonathan Derrick <jonathan.derrick@solidigm.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5795,7 +5795,7 @@ static int copy(int argc, char **argv, struct command *cmd, struct plugin *plugi
 		goto ret;
 	}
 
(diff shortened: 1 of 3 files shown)


igaw:
Thanks!

PR #1639, opened 2022-08-12
modify plugin function smart-log-add

innogritkenny:
modify smart-log-add for add items
      program_err_cnt
      erase_err_cnt
      flash_type
      hs_crc_err_cnt


igaw:
The first patch is already in the tree. You need to rebase your branch to the current HEAD of the master branch and drop the first patch.

PR #1641, opened 2022-08-16
json: Support uint64 types serialization for older json-c versions

igaw:
Notable json-c 0.13 is lacking support for serializing uint64_t
types. At this point of writing the 0.13 version is widely
deployed. Let's add our own version of the serialization function as
fallback.

Fixes #1632

### DIFF
--- a/util/json.c
+++ b/util/json.c
@@ -17,3 +17,18 @@ struct json_object *util_json_object_new_double(long double d)
 	return obj;
 
 }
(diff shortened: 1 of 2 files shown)


jeffreyalien:
The patch looks good to me.  I verified that it resolves the issues I was seeing.

igaw:
Thanks for testing!

PR #1642, opened 2022-08-17
MI: Add (and use) `nvme_cli_*` wrappers for more admin functions

jk-ozlabs:
Now that we have the MI transport & wrapper infrastructure merged, and the admin commands implemented in libnvme-mi, we can add more functionality to the wrapper code, and start using the `nvme_cli_` functions in nvme-cli.

As always, comments, queries, suggestions etc are most welcome.

### DIFF
--- a/nvme-wrap.c
+++ b/nvme-wrap.c
@@ -133,6 +133,196 @@ int nvme_cli_sanitize_nvm(struct nvme_dev *dev, struct nvme_sanitize_nvm_args *a
 	return do_admin_args_op(sanitize_nvm, dev, args);
 }
 
+int nvme_cli_get_log(struct nvme_dev *dev, struct nvme_get_log_args *args)
+{
+	return do_admin_args_op(get_log, dev, args);
+}
+
+int nvme_cli_get_nsid_log(struct nvme_dev *dev, bool rae,
+			  enum nvme_cmd_get_log_lid lid,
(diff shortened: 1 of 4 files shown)


jk-ozlabs:
We also need a bump to libnvme.wrap; updating.

igaw:
Looks good.

I am still trying to figure out if we should/need to provide dummy wrappers for the libnvme-mi parts to 'weaken the dependencies'. The idea of the `wrapper.c` (stupid name btw) that for each new function dependency we add an empty implementation. When there
is a library which supports the function we are using it if not we just error out. This is for the case where we have an older libnvme installed (not updated yet etc) and a newer nvme-cli binary (not source!).

Not sure if this is actually really working though.

jk-ozlabs:
We could add empty `__attribute__((weak))` symbols for any new MI functions we add here; this would mean `nvme-cli` could link against older libnvme-mi - both at build-time and runtime. Those functions would return some sensible error (`-ENOSYS`?)

In fact, those weak implementations don't have to be specific to the MI code; we could do that for the entire libnvme core interface too.

It does mean we would be duplicating the library API in nvme-cli though.

igaw:
This is basically what I started to do in `wrapper.c`. Any new function call which is not present in the libnvme 1.0 API should 
be added there. That was at least my plan to avoid having a nvme-cli binary which wants symbols from `@LIBNVME_1.1` or later

PR #1648, opened 2022-08-23
Added parsing for Solidigm telemetry observable data.

lgdacunha:


lgdacunha:
@igaw , I see the build-distro check failing because of missing function on older JSON-C version. Should I worry about  making new code compatible with the JSON-C version used on build-distro?

igaw:
No, it's not acceptable to break the CI build. The good news is that if you rebase all your changes onto of the current master it should be fixed (see e852f5b8c194 ("json: Support uint64 types serialization for older json-c versions"))

lgdacunha:
@igaw, I merged the master tip 2 days ago. Commit e852f5b is already part of this branch.
I hadn't noticed the change so I will push code changes to use the changes you pointed out.
There is still no json-c 0.13 compatibility code for json_object_get_uint64(), so I will change the code to use json_object_get_int() for now.

igaw:
Create a new topic branch on top of the current head which just contain your fix. Merging the upstream head into your downstream branch is making the resulting history completely broken.

Issue #1647, opened 2022-08-23
nvme-cli v2.x compile warning


martin-gpy:
Of late, I see the below compile warning when attempting to build the latest upstream nvme-cli v2.x (after successfully compiling & installing latest libnvme) on my host:

```
# ninja -C .build
ninja: Entering directory `.build'
[39/42] Compiling C object nvme.p/nvme.c.o
../nvme.c: In function ‘ns_rescan’:
../nvme.c:4389:19: warning: ‘dev’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  struct nvme_dev *dev;
                   ^~~
[42/42] Linking target nvme
```

This used to compile successfully with the earlier nvme-cli v2.x upstream builds. Any idea what broke now and how to get rid of this?

igaw:
What compiler version do you use? Any special configuration, e.g. debug build?

@jk-ozlabs

jk-ozlabs:
@martin-gpy I have a workaround for this (explicitly initialising `dev` to `NULL`), but I'd prefer to find the underlying cause here, as I'm not seeing any paths where `dev` can be used uninitialised. If I can replicate your setup here, that would definitely help.

Feel free to assign to me; I'll at least send a PR for the workaround if we can't find a specific cause soon.

jk-ozlabs:
@martin-gpy in the meantime, this is the workaround patch: https://github.com/linux-nvme/nvme-cli/commit/7f6362ae24e33ebf4dd9bcf5841d03385a0edca7

igaw:
IIRC, some distro compiler do enable additional checks/features to 'harden' the binaries. Maybe one of these options is getting the compiler confused.

 For example

  `CFLAGS=-Wmaybe-uninitialized meson .build --optimization=1`

results in 

```
[35/96] Compiling C object subprojects/libnvme/examples/mi-mctp.p/mi-mctp.c.o
../subprojects/libnvme/examples/mi-mctp.c: In function ‘do_action_endpoint’:
../subprojects/libnvme/examples/mi-mctp.c:665:16: warning: ‘rc’ may be used uninitialized [-Wmaybe-uninitialized]
  665 |         return rc;
      |                ^~
../subprojects/libnvme/examples/mi-mctp.c:637:13: note: ‘rc’ was declared here
  637 |         int rc;
      |             ^~
```

with my compiler: `gcc (SUSE Linux) 12.1.1 20220721`

The workaround is certainly an option if it is just these few places.

[...]

PR #1649, opened 2022-08-24
tree: fail on non-negative return values from parse_and_open

jk-ozlabs:
Currently, we see a warning with specific compilers:

```
  ../nvme/nvme.c: In function 'ns_rescan':
  ../nvme/nvme.c:4389:19: warning: 'dev' may be used uninitialized in this function [-Wmaybe-uninitialized]
```

This is due to a theoretical failure path in `parse_and_open()`, where
`argconfig_parse()` returns a positive value - this would require the
global `errno` to be negative.

Even though the unintialised use of dev is only possible with negative
error (which shouldn't happen), we should still consider any non-zero
return value of `parse_and_open` a failure. This change fixes the
instances of:

    err = parse_and_open(...);
    if (err < 0)

changing to:

    err = parse_and_open(...);
    if (err)

... and also applies this to the single use of `open_exclusive()`.

The positive return values from `parse_and_open()` were removed in
11542bbd; beforehand, `parse_and_open()` would return a fd.

Fixes: 11542bbd ("tree: Combine NVMe file descriptor into struct nvme_dev")
Fixes: https://github.com/linux-nvme/nvme-cli/issues/1647

Signed-off-by: Jeremy Kerr <jk@codeconstruct.com.au>

### DIFF
--- a/nvme-rpmb.c
+++ b/nvme-rpmb.c
@@ -887,7 +887,7 @@ int rpmb_cmd_option(int argc, char **argv, struct command *cmd, struct plugin *p
 		unsigned int rpmbs;
 	} regs;
 	
-	if ((err = parse_and_open(&dev, argc, argv, desc, opts)) < 0)
+	if ((err = parse_and_open(&dev, argc, argv, desc, opts)))
 		return err;
 	
 	/* before parsing  commands, check if controller supports any RPMB targets */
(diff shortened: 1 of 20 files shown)


igaw:
Nice catch! Hopefully this makes those old compilers happy!

PR #1652, opened 2022-08-26
fguid printed with wrong format

pierrelabat:
Signed-off-by: Pierre Labat <plabat@micron.com>

The fguid (__u8  fguid[16]) of the  Identify Controller Data Structure is made of non printable characters. From the nvme spec "This field uses the EUI-64 based 16-byte designator format". As a result when fguid is not zero, the printf output is garbled/spread on several lines because the format used is a string format (%-.*s).
fguid should be printed the same as other similar fields of the Identify Controller Data Structure (containing 16 non printable characters). So no garbling is occurring.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -4624,7 +4624,7 @@ void nvme_show_id_ctrl(struct nvme_id_ctrl *ctrl, enum nvme_print_flags flags,
 	printf("cntrltype : %d\n", ctrl->cntrltype);
 	if (human)
 		nvme_show_id_ctrl_cntrltype(ctrl->cntrltype);
-	printf("fguid     : %-.*s\n", (int)sizeof(ctrl->fguid), ctrl->fguid);
+    printf("fguid     : %.0Lf\n", int128_to_double(ctrl->fguid));
 	printf("crdt1     : %u\n", le16_to_cpu(ctrl->crdt1));
 	printf("crdt2     : %u\n", le16_to_cpu(ctrl->crdt2));
 	printf("crdt3     : %u\n", le16_to_cpu(ctrl->crdt3));
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -4624,7 +4624,7 @@ void nvme_show_id_ctrl(struct nvme_id_ctrl *ctrl, enum nvme_print_flags flags,
 	printf("cntrltype : %d\n", ctrl->cntrltype);
 	if (human)
 		nvme_show_id_ctrl_cntrltype(ctrl->cntrltype);
-    printf("fguid     : %.0Lf\n", int128_to_double(ctrl->fguid));
+	printf("fguid     : %s\n", nvme_uuid_to_string(ctrl->fguid));
 	printf("crdt1     : %u\n", le16_to_cpu(ctrl->crdt1));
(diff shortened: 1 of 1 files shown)


pierrelabat:
Well, it seems that the method (int128_to_double()) used by nvme-print.c to encode fields (fguid, maxdna, etc...) composed of 16 bytes not printable into a double float, is lossy.

That is, if one tries to decode the int128_to_double() output to get the original bytes (of the fguid for ex) one can't get the value of the original 16 bytes. I did some tests and one can get only the value of 8 bytes among the 16.

It is because it is not possible to store 128 bits of information (16 bytes)  in a double float encoded on 80 or 128 bits. Because some bits of the float are used for the exponent.

Dumb question, but for these 16 bytes fields, why don't we simply encode them as hexadecimal when printing?

So there is more to do beyond fixing the print of fguid to not use %s.

igaw:
> Dumb question, but for these 16 bytes fields, why don't we simply encode them as hexadecimal when printing?

I don't know, but sounds like a good idea to me. The spec says it has to be treated as IEEE EUI-64 16 bytes long. So I suppose "xx:yy:zz:..." would be the correct format?

pierrelabat:
> > Dumb question, but for these 16 bytes fields, why don't we simply encode them as hexadecimal when printing?
> 
> I don't know, but sounds like a good idea to me. The spec says it has to be treated as IEEE EUI-64 16 bytes long. So I suppose "xx:yy:zz:..." would be the correct format?

That would be one way to go. That would work. Then I looked if some other uid were printed by nvme-cli. To do it the same way to stay consistent. I see that there is a nvme_uuid_to_string(uuid_t uuid) used to print some uid. We could use that for fguid.
Seems appropriate to me.

If I don't get negative comment, I will change the patch in order to use nvme_uuid_to_string() to print the fguid. As a result a user  can get the original EUI from the print output (as opposed to using  int128_to_double() + floating point format).

pierrelabat:
I pushed a second commit that prints the fguid as a uuid_t (16 bytes). For example the 16 bytes EUI 0x000102030405060708090a0b0c0d0e0f
shows as:
fguid     : 00010203-0405-0607-0809-0a0b0c0d0e0f

igaw:
> For device not returning a fguid, that shows: fguid : 00000000-0000-0000-0000-000000000000 Before the change it was showing: fguid : 0

The new format seems a bit more consistent. As we are changing the output anyway, do we care if the '0' case changes as well?

pierrelabat:
> > For device not returning a fguid, that shows: fguid : 00000000-0000-0000-0000-000000000000 Before the change it was showing: fguid : 0
> 
> The new format seems a bit more consistent. As we are changing the output anyway, do we care if the '0' case changes as well?

Personally I don't care. I am fine with the new format for 0.

How does it work in regards to the PR https://github.com/linux-nvme/nvme-cli/pull/1661 that also touches nvme-print.c ?
Spencer would go first, then I would go next? Or Spencer pick up my change?

igaw:
I dropped the first patch and added the JSON output part. While at it I also added missing values in the JSON output format. Hence I couldn't push my changes back, I'll merged this by hand. 78b7ad235507 ("nvme-print: Print fguid as a UUID")

PR #1656, opened 2022-08-29
zns.c: report zones should be started after retrieved zone

sc108-lee:

### DIFF
--- a/plugins/zns/zns.c
+++ b/plugins/zns/zns.c
@@ -842,9 +842,8 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
 	int zdes = 0, err = -1;
 	struct nvme_dev *dev;
 	__u32 report_size;
-	void *report;
 	bool huge = false;
-	struct nvme_zone_report *buff;
+	struct nvme_zone_report *report, *buff;
 
 	unsigned int nr_zones_chunks = 1024,   /* 1024 entries * 64 bytes per entry = 64k byte transfer */
 			nr_zones_retrieved = 0,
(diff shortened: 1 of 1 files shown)


plugins/zns/zns.c
```diff
@@ -999,7 +999,8 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
 					zdes, log_len, flags, zone_list);
 
 		nr_zones_retrieved += nr_zones_chunks;
-		offset = (nr_zones_retrieved * zsze);
+		offset = (le64_to_cpu(((struct nvme_zone_report *)report)
```
igaw, line 1002: I think you could go one step further. Why not changing the `void*` type of report to the actual type?

```diff
--- a/plugins/zns/zns.c
+++ b/plugins/zns/zns.c
@@ -842,7 +842,7 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
        int zdes = 0, err = -1;
        struct nvme_dev *dev;
        __u32 report_size;
-       void *report;
+       struct nvme_zone_report *report;
        bool huge = false;
        struct nvme_zone_report *buff;
 
@@ -999,8 +999,8 @@ static int report_zones(int argc, char **argv, struct command *cmd, struct plugi
                                        zdes, log_len, flags, zone_list);
 
                nr_zones_retrieved += nr_zones_chunks;
-               offset = (le64_to_cpu(((struct nvme_zone_report *)report)
-                                       ->entries[nr_zones_chunks-1].zslba) + zsze);
+               offset = le64_to_cpu(report->entries[nr_zones_chunks-1].zslba)
+                        + zsze;
     }
 
        if (flags & JSON)
```

igaw (changes requested):
Could you test if my diff on top of yours is correct?

plugins/zns/zns.c
sc108-lee: nice catch ! fixed :)

igaw (approved):

PR #1657, opened 2022-08-29
Adding support for Seagate OCP compliant drives.

DebabrataSTX:
Adding support for Seagate OCP compliant drives.
Also added documentation for existing and new commands.


DebabrataSTX:
Adding support for Seagate OCP compliant drives.

plugins/seagate/seagate-nvme.c
```diff
@@ -1645,13 +1645,13 @@ static int vs_clr_pcie_correctable_errs(int argc, char **argv, struct command *c
 {
 	const char *desc = "Clear Seagate PCIe Correctable counters for the given device ";
 	const char *save = "specifies that the controller shall save the attribute";
-<<<<<<< HEAD
```
igaw: We don't want to have this in our git history. This looks like a left over from the development process.

igaw (changes requested):
This series needs to be cleanup. That means rebase all the work on the current HEAD. There should not be any merge commits for example (and especially no cleanup commits). Any commit should be able to build. Basically I expect a bunch clean commits.

igaw:
Please see here what I want from you: https://github.com/linux-nvme/nvme-cli/pull/1659/commits/bf0e8370d74ed2bdb824f98cf35f00d4e973d1a5

plugins/seagate/seagate-nvme.c
DebabrataSTX: Only clean way I can see is to discard this pull request, create a new fork and do a new pull request. Guess I have to do that.

igaw:
No, this is not what I tried to explain. Don't do `git merge`, instead do a `git rebase master` and fixup all fallouts.

DebabrataSTX:
Even with this it does not seems to resolve the issue. I will try a new fork and check in to that.

igaw:
If you do `git rebase` there will be conflicts which you need to resolve by hand. Note you should also look into `git rebase -i` which allows you to reorder and squash commits. And one thing I noticed, you seem to develop on the original master branch. This makes things complex if you are not an expert in git. I really recommend to branch from the master and do your work in a branch.

As you can see here with `git fetch --all && git rebase -i origin/master` you get this list. I assume you can squash most of the commits int the the previous one or drop them because they already in the tree:
image

PR #1660, opened 2022-08-31
add smart items for smart-log-add

innogritkenny:


igaw:
Patch applied directly after prefix the subject.

PR #1666, opened 2022-09-07
Add uint128 type

igaw:
This is a complete reworked version of #1655 

- reorganized/splitted the PR
- moved common type bits to new location util/types.h
- dropped the print_uint128 function. Just use uint128_t_to_string() directly

TODOs(?):
- The le128_to_cpu() function should probably be a macro like the other le*_to_cpu() helpers from ccan
- Naming scheme seems a bit off and a bit too verbose, e.g. `_to_string`.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -13,14 +13,8 @@
 #include "nvme-print.h"
 #include "nvme-models.h"
 #include "util/suffix.h"
+#include "util/types.h"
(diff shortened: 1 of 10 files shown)


igaw:
Let's go with this version as it really fixes the output. The TODO points I mentioned can still be fixed later. Ideally we the `le128_*` stuff would be living in ccan.

PR #1675, opened 2022-09-13
udev: Add HOST_IFACE to udev rule

martin-belanger:
The HOST_IFACE only applies to TCP connections. When provided, it forces TCP connections on a specific interface instead of letting the routing table decide. Some users may prefer to force connections on specific interfaces instead of configuring the routing table. This change makes sure that the HOST_IFACE used to set up the persistent connection to the discovery controller also gets applied to the I/O controller connections.

### DIFF
--- a/nvmf-autoconnect/udev-rules/70-nvmf-autoconnect.rules.in
+++ b/nvmf-autoconnect/udev-rules/70-nvmf-autoconnect.rules.in
@@ -4,16 +4,22 @@
 #   controller and connect to elements in the discovery log.
 #
 #
+ACTION!="change", GOTO="autoconnect_end"
+
+# For backwards compatibility. Make sure HOST_IFACE is not an empty string.
+ENV{NVME_HOST_IFACE}=="", ENV{NVME_HOST_IFACE}="none"
 
 # Events from persistent discovery controllers or nvme-fc transport events
 # NVME_AEN:
 #   type 0x2 (NOTICE) info 0xf0 (DISCOVERY_LOG_CHANGE) log-page-id 0x70 (DISCOVERY_LOG_PAGE)
 ACTION=="change", SUBSYSTEM=="nvme", ENV{NVME_AEN}=="0x70f002",\
   ENV{NVME_TRTYPE}=="*", ENV{NVME_TRADDR}=="*", \
-  ENV{NVME_TRSVCID}=="*", ENV{NVME_HOST_TRADDR}=="*", \
-  RUN+="@SYSTEMCTL@ --no-block start nvmf-connect@--device=$kernel\t--transport=$env{NVME_TRTYPE}\t--traddr=$env{NVME_TRADDR}\t--trsvcid=$env{NVME_TRSVCID}\t--host-traddr=$env{NVME_HOST_TRADDR}.service"
+  ENV{NVME_TRSVCID}=="*", ENV{NVME_HOST_TRADDR}=="*", ENV{NVME_HOST_IFACE}=="*", \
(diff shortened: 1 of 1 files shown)


igaw:
1. Is this backwards compatible? What happens when a kernel doesn't set the environment variable?
2. If no `host_iface` is used, the kernel will emit `NVME_HOST_IFACE=none`. Neither nvme-cli nor libvnme will do any processing on this value and pass it straight to the driver. There is no filtering done either as far I can tell, hence we end up trying to open a socket and trying to bind to an interface called 'none':

```c
	if (nctrl->opts->mask & NVMF_OPT_HOST_IFACE) {
		char *iface = nctrl->opts->host_iface;
		sockptr_t optval = KERNEL_SOCKPTR(iface);

		ret = sock_setsockopt(queue->sock, SOL_SOCKET, SO_BINDTODEVICE,
				      optval, strlen(iface));
```

martin-belanger:
I think that `none` is already handled by `libnvme` in `build_options()` and `add_argument()`. `add_argument()` filters out `none` so that it does not get passed to the kernel.
```
static int add_argument(char **argstr, const char *tok, const char *arg)
{
	char *nstr;

	if (!(arg && strcmp(arg, "none")))     <--- Ignore "none"
		return 0;
	if (asprintf(&nstr, "%s,%s=%s", *argstr, tok, arg) < 0) {
		errno = ENOMEM;
		return -1;
	}
	free(*argstr);
	*argstr = nstr;

	return 0;
}
```

BTW, I tested both cases (i.e. `HOST_IFACE=none` and `HOST_IFACE!=none`) and everything works as expected (i.e. the kernel does not get passed `none`).

igaw:
I've missed the `add_argument()` part. So this part works fine. What about backwards compatible? Does the udev rule still work on an kernel without the `HOST_IFACE` environment variable?

martin-belanger:
That's a good question. However, this udev rule is only invoked when there is a persistent connection to a Discovery Controller (DC) using "`nvme discover --persistent ..."`. If such a connection was established successfully with the `--host-iface` option, then we know that the kernel supports `HOST_IFACE`. And if the kernel does not support `HOST_IFACE`, then `nvme discover --persistent --host-iface ...` will not succeed. Therefore, on a system where the kernel does not support `HOST_IFACE` we can never have connections to DCs that were made with `--host-face`, which means `HOST_IFACE` will always be `none` on those systems (and as stated above `none` gets filtered out by `libnvme`).

martin-belanger:
@igaw - It took a while, but I created a VM with an older kernel version (5.8), and you are correct. There is an issue. What I see is that host-iface is used with an empty value instead of "none". And empty value does not get filtered by libnvme and that causes a problem. I think I can fix it in the udev rule itself to make sure an empty value gets translated to "none", but we should probably fix libnvme to filter out empty values as well.

igaw:
I know, I am bit cautions with the udev part, as the last update caused a bit of trouble. Thanks for double checking. 

Yes, let's add the empty filter to libnvme as well.

martin-belanger:
I found an easy way to make sure `NVME_HOST_IFACE` can never be an empty string. If it is, the udev rule will simply set it to `none`, which we know gets filtered out by libnvme.

I tested this on the following systems:
1. Fedora 33 (kernel 5.8): No kernel support for host-iface
2. Fedora 36 (kernel 5.19): With kernel support for host-iface. 

I ran the following tests:
1. I tried for connections made with `--host-iface` (only works on Fedora 36 as expected).
2. I tried for connections made without `--host-iface` (on both Fedora 33 and 36). 
3. I triggered udev "change" events by adding subsystems to a discovery controller which resulted in AENs being sent. 

In all cases, the udev rules responded properly and created the I/O controller connections.

igaw:
Thanks a lot!

PR #1678, opened 2022-09-17
nvme: Add wrappers for identify secondary ctrl

drakedog2008:
Signed-off-by: Hao Jiang <jianghao@google.com>

### DIFF
--- a/nvme-wrap.c
+++ b/nvme-wrap.c
@@ -95,6 +95,39 @@ int nvme_cli_identify_allocated_ns_list(struct nvme_dev *dev, __u32 nsid,
 	return do_admin_op(identify_allocated_ns_list, dev, nsid, list);
 }
 
+int nvme_cli_identify_secondary_ctrl_list(struct nvme_dev *dev, __u32 nsid, __u16 ctrl_id,
+										  struct nvme_secondary_ctrl_list *list)
+{
+
+	int rc;
+	if (dev->type == NVME_DEV_DIRECT)
+	{
+		rc = nvme_identify_secondary_ctrl_list(dev->direct.fd, nsid, ctrl_id, list);
+	}
+	else if (dev->type == NVME_DEV_MI)
+	{
+		struct nvme_identify_args args = {
+			.result = NULL,
+			.data = list,
+			.args_size = sizeof(args),
+			.cns = NVME_IDENTIFY_CNS_SECONDARY_CTRL_LIST,
+			.csi = NVME_CSI_NVM,
+			.nsid = nsid,
+			.cntid = ctrl_id,
+			.cns_specific_id = NVME_CNSSPECID_NONE,
+			.uuidx = NVME_UUID_NONE,
+		};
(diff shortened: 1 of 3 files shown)


nvme-wrap.c
```diff
@@ -95,6 +95,39 @@ int nvme_cli_identify_allocated_ns_list(struct nvme_dev *dev, __u32 nsid,
 	return do_admin_op(identify_allocated_ns_list, dev, nsid, list);
 }
 
+int nvme_cli_identify_secondary_ctrl_list(struct nvme_dev *dev, __u32 nsid, __u16 ctrl_id,
```
jk-ozlabs, line 98: Can you do this with a wrapper in the same style as the rest of this file? If the underlying call isn't available in libnvme, we should add the identify logic there instead.

nvme-wrap.h
```diff
@@ -24,6 +24,8 @@ int nvme_cli_identify_active_ns_list(struct nvme_dev *dev, __u32 nsid,
 				     struct nvme_ns_list *list);
 int nvme_cli_identify_allocated_ns_list(struct nvme_dev *dev, __u32 nsid,
 					struct nvme_ns_list *list);
+int nvme_cli_identify_secondary_ctrl_list(struct nvme_dev *dev, __u32 nsid,
+										  __u16 ctrl_id, struct nvme_secondary_ctrl_list *list);
```
jk-ozlabs, line 28: Some strange spacing happening here (and the next chunk). Do you have your tabs set to something other than 8 columns maybe?

jk-ozlabs (changes requested):
Hi Hao,

Thanks for the contribution! I think we'd need some changes though - we don't want to build up too much in the wrappers, they should be simple one-line functions whereever possible.

nvme-wrap.c
lightgoleft: The libnvme misses nvme_identify(fd, ...), and libnvme-mi misses nvme_mi_mi_admin_identify_secondary_ctrl_list(). 

We need either of these to simplify the logic here. 

But sure, we can make changes in the lib instead of cli.

lightgoleft:
> Hi Hao,
> 
> Thanks for the contribution! I think we'd need some changes though - we don't want to build up too much in the wrappers, they should be simple one-line functions whereever possible.

Hi Jeremy, 

Thanks for the review. I will reply the questions in place.

nvme-wrap.c
jk-ozlabs: I'm happy to add those to libnvme if you like!

lightgoleft: That would be great.

jk-ozlabs: OK, PR sent:

https://github.com/linux-nvme/libnvme/pull/480

I haven't tested this with nvme-cli, as that's what you're proposing here. Let me know how it goes!

igaw:
This PR is out of date and didn't' get any of the requested changes.

PR #1682, opened 2022-09-23
util/types: Add fallback for 128 bit types on 32bit platforms

igaw:
gcc provides the __uint128_t type on only on 64bit platforms. A common
workaround is to use the gmp library in case longer arithmetic types
are needed.

As we don't do any math with these values (just reading and printing)
we can workaround by introducing a pointer type. Thus, all endian
conversion have then to be done in a buffer. We already have a buffer
for creating the string to print (uint128_t_to_string), so nothing is
really lost as long these function are not called in a nested way.

While at it we change the output format to hex based because this
avoids the overflow handling with the ten based system.

Fixes: #https://github.com/linux-nvme/libnvme/issues/475


igaw:
If we go this fix we should also update the output to prefix the strings with '0x'

igaw:
Replaced by #1685

PR #1684, opened 2022-09-23
Update README.md

samvotter:
trivia change. Deleted some text from readme


igaw:
It looks like you are trying to figure out how github and git are working. The change doesn't really make sense, at least I don't know what `thon` means.

A good practice is to setup your own github repo and play around with this instance and not real project. Anyway, no worry, everyone gets things wrong at one point. The trick is not to repeat it :)

PR #1688, opened 2022-09-26
Add warning fixes and enable cross compilation targets in github workflows

jk-ozlabs:
This change adds a couple of cross-compilation targets for nvme-cli: armhf and powerpc64.

In order to do this though, we have to fix a couple of warnings for printf-specifiers and casts on those arches.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5982,7 +5982,7 @@ static int copy(int argc, char **argv, struct command *cmd, struct plugin *plugi
 	if (cfg.format == 0)
 		nrts = argconfig_parse_comma_sep_array(cfg.eilbrts, (int *)eilbrts.f0, ARRAY_SIZE(eilbrts.f0));
 	else if (cfg.format == 1)
-		nrts = argconfig_parse_comma_sep_array_long(cfg.eilbrts, (__u64 *)eilbrts.f1, ARRAY_SIZE(eilbrts.f1));
+		nrts = argconfig_parse_comma_sep_array_long(cfg.eilbrts, (unsigned long long *)eilbrts.f1, ARRAY_SIZE(eilbrts.f1));
 	else {
 		fprintf(stderr, "invalid format\n");
 		err = -EINVAL;
--- a/plugins/scaleflux/sfx-nvme.c
+++ b/plugins/scaleflux/sfx-nvme.c
@@ -611,7 +611,8 @@ static void show_lat_stats_myrtle(struct sfx_lat_stats_myrtle *stats, int write)
 	for (i = 0; i < 64; i++)
 		printf("Bucket %2d: %u\n", i, stats->bucket_19[i]);
 
-	printf("\nAverage latency statistics %lld\n", stats->average);
+	printf("\nAverage latency statistics %" PRIu64 "\n",
+	       (uint64_t)stats->average);
 }
 
 
--- a/.github/cross/ubuntu-armhf.txt
+++ b/.github/cross/ubuntu-armhf.txt
@@ -1,0 +1,17 @@
+[binaries]
+c = '/usr/bin/arm-linux-gnueabihf-gcc'
+ar = '/usr/arm-linux-gnueabihf/bin/ar'
+strip = '/usr/arm-linux-gnueabihf/bin/strip'
+pkgconfig = '/usr/bin/arm-linux-gnueabihf-pkg-config'
+ld = '/usr/bin/arm-linux/gnueabihf-ld'
+
+[properties]
+root = '/usr/arm-linux-gnueabihf'
+has_function_printf = true
+skip_sanity_check = true
+
+[host_machine]
+system = 'linux'
+cpu_family = 'arm'
+cpu = 'armv7'
+endian = 'little'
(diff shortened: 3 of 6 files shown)


plugins/scaleflux/sfx-nvme.c
```diff
@@ -611,7 +611,8 @@ static void show_lat_stats_myrtle(struct sfx_lat_stats_myrtle *stats, int write)
 	for (i = 0; i < 64; i++)
 		printf("Bucket %2d: %u\n", i, stats->bucket_19[i]);
 
-	printf("\nAverage latency statistics %lld\n", stats->average);
+	printf("\nAverage latency statistics %lld\n",
+	       (unsigned long long)stats->average);
```
igaw, line 615: Wouldn't the `__SANE_USERSPACE_TYPES__` trick not work here? https://github.com/linux-nvme/libnvme/pull/389

nvme.c
```diff
@@ -5982,7 +5982,7 @@ static int copy(int argc, char **argv, struct command *cmd, struct plugin *plugi
 	if (cfg.format == 0)
 		nrts = argconfig_parse_comma_sep_array(cfg.eilbrts, (int *)eilbrts.f0, ARRAY_SIZE(eilbrts.f0));
 	else if (cfg.format == 1)
-		nrts = argconfig_parse_comma_sep_array_long(cfg.eilbrts, (__u64 *)eilbrts.f1, ARRAY_SIZE(eilbrts.f1));
+		nrts = argconfig_parse_comma_sep_array_long(cfg.eilbrts, (unsigned long long *)eilbrts.f1, ARRAY_SIZE(eilbrts.f1));
```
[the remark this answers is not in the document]
jk-ozlabs: Up to you, but I assume the function name implies a `long` rather than a `u64`.

plugins/scaleflux/sfx-nvme.c
jk-ozlabs: No, that macro is only for powerpc - this is happening for me on amd64, if I attempt to use `PRId64` for this.

The issue seems to be that I'm getting the `__u64` defined through the int-ll64.h header, which has

```
__extension__ typedef __signed__ long long __s64;
__extension__ typedef unsigned long long __u64;
```

but the `PRI` macros are based entirely on the word size, rather than int-l64/int-ll64 configuration:

```
# if __WORDSIZE == 64
#  define __PRI64_PREFIX	"l"
#  define __PRIPTR_PREFIX	"l"
# else
#  define __PRI64_PREFIX	"ll"
#  define __PRIPTR_PREFIX
# endif
```

nvme.c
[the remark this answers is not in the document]
igaw: I am sitting on the fence as I don't like type cast as the usually are pretty good in hiding stuff from the compiler. Anyway, there are two users of argconfig_parse_comma_sep_array_long() and the first one also does the type case. I guess we can stick to this and clean it up if needed.

plugins/scaleflux/sfx-nvme.c
jk-ozlabs: (having an int-ll64 configuration here seems a bit strange, so I'll investigate a little further...)

jk-ozlabs: Yep, looks like a potential headers issue, possibly fixed by http://patchwork.ozlabs.org/project/glibc/patch/20220621090951.29911-1-metan@ucw.cz/

nvme.c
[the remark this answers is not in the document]
jk-ozlabs: I'm happy to incorporate the fix here if you prefer!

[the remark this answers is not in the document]
igaw: Thanks but I think your version here is good for the time being. Let's try to get the rest sorted out first. This is just a minor thing which can be still looked at later.

plugins/scaleflux/sfx-nvme.c
igaw: Oh, that explains it. The question is how we should deal with it.

jk-ozlabs: From a brief look, those existing `PRI{u,d}64` uses are where we're printing a `uint64_t`, not a `__u64` - the latter being the kernel-defined types from `int-{l,ll}64.h`.

In some cases there's a cast from the `__u64` to a `uint64_t` first, which might be more appropriate here.

jk-ozlabs: OK, pushed a new version with `uint64_t` instead. This still has the casts, which isn't ideal, but we should be casting to the equivalent type.

igaw:
Thanks!

plugins/scaleflux/sfx-nvme.c
jk-ozlabs: As for a longer-term fix - perhaps we should only be using the `__u{8,26,32,64}` types where they're defined in kernel interfaces and/or structured, and use the stdint `uint{8,16,32,64}_t` types everywhere else.

This would be a lot of churn though

PR #1691, opened 2022-09-26
nvme: Add wrappers for firmware commands

jk-ozlabs:
This change adds nvme_dev wrappers for the firmware download and firmware commit commands, allowing access over a MI channel.

We need a bump to libnvme for the underlying MI firmware calls.

Signed-off-by: Jeremy Kerr <jk@codeconstruct.com.au>

### DIFF
--- a/subprojects/libnvme.wrap
+++ b/subprojects/libnvme.wrap
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = 501347ffdf6321d42ceabf173792c95bd44dac0d
+revision = 677075a11e8c2755d5566174c8321812f886ab29
 
 [provide]
 libnvme = libnvme_dep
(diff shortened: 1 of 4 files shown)


subprojects/libnvme.wrap
```diff
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = 501347ffdf6321d42ceabf173792c95bd44dac0d
+revision = 677075a11e8c2755d5566174c8321812f886ab29
```
igaw, line 3: Could you split the libnvme.wrap update into a separate commit? So first update of the dependency and then do the change?

I'd like to avoid mixing 'CI' updates from code updates. Reason being for anyone doing backports it is a hassle when you need to pick a fix/change which also updates a file like libnvme.wrap. It will almost certainly always conflict in this case.

igaw (changes requested):

subprojects/libnvme.wrap
jk-ozlabs: ok, done!

igaw (approved):

Issue #1668, opened 2022-09-07
nvme connect-all still relies on discovery.conf for targets that are TP 8013/8014 compliant


martin-gpy:
Turns out the latest libnvme/nvme-cli still relies on the discovery.conf even for a NVMe target that is already TP 8013/8014 compliant. For e.g.

```
# nvme connect-all -J /etc/nvme/config.json
No params given and no /etc/nvme/discovery.conf
```

With TP 8013/8014, all discovery info (including those of current & other discovery subsystems) is available in the discovery log page itself. So the host should no longer rely on an explicit discovery.conf anymore for such 8013/8014 compliant NVMe targets.

Issue #1667, opened 2022-09-07
NVMe/FC auto-connect scripts should leverage the config JSON file for connect-all


martin-gpy:
The NVMe/FC auto-connect scripts should now have the capability to leverage the config JSON file for connect-all. If this config JSON file is missing, then it may fall back to using the standard default command line options itself to the connect-all.

PR #1694, opened 2022-09-29
fabrics: Honor JSON config file in connect-all command

igaw:
Also use the config.json config file when doing an auto connect.

Fixes: #1667
Fixes: #1668

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -1134,6 +1134,18 @@ int nvmf_config(const char *desc, int argc, char **argv)
 		nvme_subsystem_t s;
 		nvme_ctrl_t c;
 
+		if (!subsysnqn) {
+			fprintf(stderr,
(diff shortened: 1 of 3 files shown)


fabrics.c
```diff
@@ -665,7 +735,8 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 		nvme_free_tree(r);
 		return ret;
 	}
-	nvme_read_config(r, config_file);
+	if (!nvme_read_config(r, config_file))
+		json_config = true;
```
igaw, line 739: I am still pondering if the `json_config` makes sense. `nvme_read_config` returns 0 independent if the config file exist or not. But we just fail if `nvme_read_config()` returns an error I suppose.

igaw: I had a discussion offline with @hreinecke and @martin-gpy and we decided this approach is okay.

Issue #1681, opened 2022-09-21
Set feature for 0xD Host Memory Buffer, CLI expects buffer


1pandaman:
According to the NVMe spec FID 0xD uses a databuffer for get feature but not set feature, but a set feature on the cli will wait on user input if a data buffer is not provided.

igaw:
Can you post the commend you are using?

1pandaman:
sudo nvme set-feature /dev/nvme0 --feature-id=0xD --value=0

keithbusch:
Yeah, it needs a buffer, so if you don't provide a file or input stream, then you get to write it from stdin. What do you want to happen instead?

This feature though, and every feature that's owned by the kernel driver, seems like a bad idea to modify from user space unless part of some kind of disruptive test.

1pandaman:
I was expecting not to need a buffer as the nvme spec states it is doesn't require one. But if the cli requires it is sending a 0 filled buffer effectively equivalent to not sending a buffer?

keithbusch:
Oh, right, the set side for this feature doesn't require a buffer in the traditional sense. It's just the "get" side of this feature ID that needs a host buffer. I suppose we should disable the buffer check in that case, though if you're turning the feature on, it does take a descriptor list. That is a type of host buffer, but not the kind that user space can readily supply, so it's not really a safe feature to be messing with.

igaw:
The fix is in so closing the issue. Thanks!

PR #1699, opened 2022-10-13
nvme: Fix set feature command to get feature identifier 0Dh length as…

ikegami-t:
… zero

This fix is for the issue #1681.


jk-ozlabs:
Yes please - my intention for the `nvme-wrap.c` functions is to provide an API that takes a `struct nvme_dev`, so we can use for both iocal and MI transports.

ikegami-t:
The nvme_cli_get_feature_length() and nvme_cli_get_feature_length2() functions are just to retreive the command payload length for a specific feature identifier and the set and get features specifications are same for both nvme devices direct and mi. So seems not necessary to use struct nvme_dev or implemented as nvme-wrap function. By the way seems also the nvme_set_features() function is not implemented as a nvme-wrap function so current nvme set-feature itself deos not support for the nvme device mi currently.

jk-ozlabs:
Ah, my mistake - I'd interpreted the `int fid` argument as a file descriptor.

If these aren't using the ioctl interface, then there's no need for any abstraction via `struct nvme_dev`, so the functionality is fine as-is.

However, that means that `nvme_cli_get_feature_length2` may not belong in nvme-wrap.c, which was originally intended for the `struct nvme_dev` wrappers. If we want to change that intention, that's fine, but we should be doing so purposefully.

As a side note - there seem to be some unrelated changes in `libnvme-wrap.c`? And it's being renamed?

igaw:
> However, that means that `nvme_cli_get_feature_length2` may not belong in nvme-wrap.c, which was originally intended for the `struct nvme_dev` wrappers. If we want to change that intention, that's fine, but we should be doing so purposefully.

I was pondering where to put it. My idea was to group all `nvme_cli*` function in one place and nvme-wrap.c is pretty generic name. No strong preferences. So if you think it should go to a different file that is no issue.

> As a side note - there seem to be some unrelated changes in `libnvme-wrap.c`? And it's being renamed?

Not unrelated but surely not obvious. Let me split this into a separate patch in this case. Thanks for the review!

jk-ozlabs:
> My idea was to group all `nvme_cli*` function in one place and nvme-wrap.c is pretty generic name.

OK, that's a good a policy as any! Let's settle on that and keep the function there.

igaw:
@jk-ozlabs are you fine with the current version?

jk-ozlabs:
Yep, all looks good to me. The split makes sense now.

PR #1706, opened 2022-10-23
Add support for libvnme status types

jk-ozlabs:
WIP: needs upstream libnvme changes first

This series adds support for the status type encoding proposed in https://github.com/linux-nvme/libnvme/pull/502

We first add decoding using the nvme_status_get_type() helper, and then add support for MI-specific status values through nvme_mi_status_to_string().

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -3255,10 +3255,28 @@ void d_raw(unsigned char *buf, unsigned len)
 		putchar(*(buf+i));
 }
 
-void nvme_show_status(__u16 status)
+void nvme_show_status(int status)
 {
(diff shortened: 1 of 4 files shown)


igaw:
Any change to get this updated till tomorrow? I'd like to the release on Friday as I am out of office on Monday :)

jk-ozlabs:
Can do. It shouldn't need much of an update (just a re-CI), but I'll test first here, and push with a corresponding .wrap change.

jk-ozlabs:
Looks like the `.wrap` has already been updated, so it'll just be a plain rebase.

igaw:
Thanks!

PR #1715, opened 2022-10-30
build: fix

ajs124:
when we tried updating to 2.2 in nixpkgs (see https://github.com/NixOS/nixpkgs/pull/198471) we saw a handful of compiler errors about `uuid.h` still being included and stuff from `time.h` not being declared.

maybe we're just "holding it wrong", but these changes seem necessary to be able to build this release.

### DIFF
--- a/plugins/innogrit/innogrit-nvme.c
+++ b/plugins/innogrit/innogrit-nvme.c
@@ -5,6 +5,7 @@
 #include <errno.h>
 #include <unistd.h>
 #include <sys/stat.h>
+#include <time.h>
 
 #include "common.h"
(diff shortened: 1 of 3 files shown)


zseri:
@ajs124 thanks

igaw:
Your fix looks absolutely right. It looks like on our CI builds and also my local builds always found libuuid installed.

I might want to do a 2.2.1... Let me first check what else is gone wrong with 2.2. Maybe something else needs to be added into a 2.2.1.

igaw:
I've splitted the commit up and added commit messages and the I've applied the final result. Thanks!

PR #1718, opened 2022-10-31
Add Inspur plugin extension.

wangmengmeng02:
Add sub-command to retrieve Inspur device Vendor log.

### DIFF
--- a/Documentation/cmd-plugins.txt
+++ b/Documentation/cmd-plugins.txt
@@ -198,3 +198,6 @@ linknvme:nvme-zns-zone-mgmt-send[1]::
 
 linknvme:nvme-zns-zrwa-flush-zone[1]::
 	Flush LBAs associated with a ZRWA to a zone
+
+linknvme:nvme-inspur-nvme-vendor-log[1]::
+	NVMe Inspur Device Vendor log page request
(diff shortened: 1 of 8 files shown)


wangmengmeng02:
Add sub-command to retrieve Inspur device Vendor log.

Signed-off-by: wangmengmeng [wangmengmeng02@inspur.com]

igaw (changes requested):
A couple of small things:

- Could you drop the `.1` and `.html` documentation files. These will be auto generated for the next release. 
- The documentation for the data structs seems to be javadoc(?) but we are using kernel-doc comments. I also suggest not to mentioning the byte address in the comment as these get easy out of sync. It's better to us pahole to verify the position and alignment of the members.

Rest looks good.

wangmengmeng02:
> * Could you drop the `.1` and `.html` documentation files. These will be auto generated for the next release.

@igaw ,thanks for reviewing, this comment means drop all the modification in directory Documentation/* or only *.1 and *.html document?

igaw:
> @igaw ,thanks for reviewing, this comment means drop all the modification in directory Documentation/* or only *.1 and *.html document?

No just do not add the generated documents. In our case these are the *.1 and *html files. The *.txt file should stay as it is the source for the other files.

wangmengmeng02:
update for comments

igaw (approved):

igaw:
I've merged the PR directly after small changes:
- updated the commit message slightly
- added a bunch of newlines
- switch iterator types to `int`

Issue #1593, opened 2022-07-05
nvme verbose log printed with "nvme list" cmd


yizhanglinux:
Hello
I compiled the latest nvme-cli, found some nvme verbose log output even with "nvme list" cmd, is that expected?

```
$ uname -r
5.19.0-rc5
$ nvme --version
nvme version 2.0 (git 2.0-165-gc9b7d00)
libnvme version 1.0 (git 1.0-105-ge757ffe)

$ nvme list
Node                  Generic               SN                   Model                                    Namespace Usage                      Format           FW Rev  
--------------------- --------------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
/dev/nvme0n1          /dev/ng0n1            S48CNC0N400972B      Samsung SSD 983 DCT 960GB                1           0.00   B / 960.20  GB    512   B +  0 B   EDA5302Q

$ dmesg
[ 2408.204381] nvme0: Identify(0x6), Invalid Field in Command (sct 0x0 / sc 0x2) MORE

$ nvme show-regs /dev/nvme0n1
cap     : 3028033fff
version : 10200
cc      : 460001
csts    : 1
nssr    : 0
crto    : 0
intms   : 0
intmc   : 0
aqa     : 1f001f
asq     : 284baf000
acq     : 284bae000
cmbloc  : 0
cmbsz   : 0
bpinfo  : 0
bprsel  : 0
bpmbl   : 0
cmbmsc	: 0
cmbsts	: 0
pmrcap  : 0
pmrctl  : 0
pmrsts  : 0
pmrebs  : 0
pmrswtp : 0
pmrmscl	: 0
pmrmscu	: 0

$ dmesg
[ 2578.505337] nvme0: Identify(0x6), Invalid Field in Command (sct 0x0 / sc 0x2) MORE 
[ 2578.518015] nvme0: Unknown(0x7f), Invalid Command Opcode (sct 0x0 / sc 0x1) 
[ 2578.528673] nvme0: Identify(0x6), Invalid Field in Command (sct 0x0 / sc 0x2) MORE 

```

yizhanglinux:
> The messages are coming from the kernel, and they are expected. We may be able to avoid some of them, though.

Thanks for the update, so is it something we need to improve from the kernel or nvme-cli side?

keithbusch:
I think the kernel is purposefully logging these. Perhaps it's too noisy and we can discuss on that list about suppressing them since they are a bit alarming.

[...]

PR #1721, opened 2022-11-03
nvme: Change to use pci resource bar at first by show-regs command

ikegami-t:
This is to prevent the nvme error log mentioned by the issue #1593.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4543,6 +4543,8 @@ static int nvme_get_properties(int fd, void **pbar)
 			err = 0;
 			value = -1;
 		} else if (err) {
+			fprintf(stderr, "get-property: %s\n",
+				nvme_strerror(errno));
 			free(*pbar);
 			break;
 		}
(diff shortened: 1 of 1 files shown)


igaw:
It is a kernel bug which has been addressed by https://git.infradead.org/nvme.git/commitdiff/ecaec17576195ad45022d41a7092d895acdf1d0c

This fix should be back ported to all affected LTS kernels.

There are a few more places where nvme-cli/libnvme would trigger to verbose logging. So I don't think it helps just to patch up one place. IIRC, these register are also not available when `CONFIG_IO_STRICT_DEVMEM` is set.

ikegami-t:
Seems as basically the unnecessary or unsupported commands should not be submitted to device. But is it really not necessary to fix for the prevention?

igaw:
We can't really know upfront which command will fail. Thus libnvme is probing if a feature is supported before using it. IIRC we do the same thing in the kernel driver.

igaw:
Sure, if we can reliable detect the `CONFIG_IO_STRICT_DEVMEM` config option.

ikegami-t:
Thanks for your comments. I will do consider more for your advices.

igaw:
> Sure, if we can reliable detect the `CONFIG_IO_STRICT_DEVMEM` config option.

I think you could just try to mmap the regs and when this fails use fabrics approach. Though how should we handle the error message. Currently we will print `Did your kernel enable CONFIG_IO_STRICT_DEVMEM`. Should we just turn it into a debug message which will only be visible with the `--verbose` command line option?

igaw:
Ah that is actually what your patch does. Right, so please turn the error message into a debug message and I think you should add the error message when both approaches fail or are we printing anything useful when the fabrics approach fails?

ikegami-t:
Thanks for your understanding. Just fixed for your comments so please review again.

igaw:
Thanks! I've applied the patch directly after making the error message on `CONFIG_IO_STRICT_DEVMEM` conditional on the debug flag.

PR #1722, opened 2022-11-06
zsh: move SPDX identifier

RPigott:
The zsh completion system requires compdef to be set in the first line of the completion definition. The SPDX identifier introduced breaks this requirement, so let's just move it a line down.


igaw:
Oops, thanks for catching this!

PR #1724, opened 2022-11-08
fix an error message in scaleflux plugin

maurizio-lombardi:
"write" is a const string which cannot be null, we should check cfg.write instead


igaw:

PR #1725, opened 2022-11-08
fix unreacheable code in verify_cmd()

maurizio-lombardi:
the "if (err < 0)" condition should be checked before "if (err)", otherwise it would be never caught.


igaw:

PR #1726, opened 2022-11-08
fix freed memory dereference.

maurizio-lombardi:
dev_close() is already called after the jump to "out". Calling it 2 times against the same pointer is unsafe because it will dereference freed memory.


igaw:

PR #1727, opened 2022-11-09
Support telemetry-log over nvme-mi.

jinliangw:
Copy and re-write the [nvme_get_telemetry_log](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/linux.c#L120-L224) function from `libnvme`, so that we can use the existing wrappers below to get telemetry log both in-band and over nvme-mi.

We can stop using the legacy in-band functions:
- nvme_get_ctrl_telemetry
- nvme_get_host_telemetry
- nvme_get_new_host_telemetry

Start to use the new wrappers:
- nvme_cli_get_log_telemetry_ctrl
- nvme_cli_get_log_create_telemetry_host
- nvme_cli_get_log_telemetry_host

Tested in band:
```
# ./nvme_old telemetry-log --host-generate=1 --output-file=./telemetry_host_gen_1.bin.old /dev/nvme1
# ./nvme_old telemetry-log --host-generate=0 --output-file=./telemetry_host_gen_0.bin.old /dev/nvme1
# ./nvme_old telemetry-log --controller-init --output-file=./telemetry_ctrl_gen.bin.old /dev/nvme1


# ./nvme_new telemetry-log --host-generate=1 --output-file=./telemetry_host_gen_1.bin.new /dev/nvme1
# ./nvme_new telemetry-log --host-generate=0 --output-file=./telemetry_host_gen_0.bin.new /dev/nvme1
# ./nvme_new telemetry-log --controller-init --output-file=./telemetry_ctrl_gen.bin.new /dev/nvme1

# ls -l
-rw-r--r-- 1 root root 33554432 Nov  8 17:48 telemetry_ctrl_gen.bin.new
-rw-r--r-- 1 root root 33554432 Nov  8 17:46 telemetry_ctrl_gen.bin.old
-rw-r--r-- 1 root root   172032 Nov  8 17:46 telemetry_host_gen_0.bin.new
-rw-r--r-- 1 root root   172032 Nov  8 17:44 telemetry_host_gen_0.bin.old
-rw-r--r-- 1 root root   172032 Nov  8 17:46 telemetry_host_gen_1.bin.new
-rw-r--r-- 1 root root   172032 Nov  8 17:44 telemetry_host_gen_1.bin.old
# diff ./telemetry_ctrl_gen.bin.new ./telemetry_ctrl_gen.bin.old
#
```

**Not Tested over nvme-mi**
During the verification, we met a SSD FW issue. 
Will test it again once the FW issue get fixed.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -540,6 +540,108 @@ ret:
 	return err;
 }
 
+static int get_telemetry_log_helper(struct nvme_dev *dev, bool create,
+				    bool ctrl, struct nvme_telemetry_log **buf,
+				    enum nvme_telemetry_da da,
+				    size_t *size)
+{
+	static const __u32 xfer = NVME_LOG_TELEM_BLOCK_SIZE;
+	struct nvme_telemetry_log *telem;
+	struct nvme_id_ctrl id_ctrl;
+	void *log, *tmp;
+	int err;
+	*size = 0;
+
+	log = calloc(1, xfer);
+	if (!log) {
+		errno = ENOMEM;
+		return -1;
+	}
+
+	if (ctrl) {
+		/* set rae = true so it won't clear the current telemetry log in controller */
+		err = nvme_cli_get_log_telemetry_ctrl(dev, true, 0, xfer, log);
+	} else {
+		if (create)
+			err = nvme_cli_get_log_create_telemetry_host(dev, log);
+		else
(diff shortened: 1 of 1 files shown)


igaw:
The frontend code doesn't need to to error codes returns in the same way as we do in the library. I've updated the code accordingly and applied it directly. Thanks!

Issue #1559, opened 2022-06-08
nvme-sanitize documentation conflicting statements


tuxrenci:
The documentation for **nvme-sanitize** provides conflicting statements. This problem exists in both the text and HTML versions, and also in the **man** page.

Specifically, the documentation says (highlighting added):
> The <device> parameter is mandatory NVMe character device (ex: **/dev/nvme0**)

However, in the **EXAMPLES** section, the syntax shown references the Block device (highlighting added):
> nvme sanitize **/dev/nvme0n1** --sanact=0x01

This is confusing. Is the first statement accurate, or is the usage example correct? The conflict needs to be resolved.

igaw:
I think the first example is correct. The spec (section `5.24 Sanitize command`) doesn't list any namespace id which could be provided. Also section `8.21 Sanitize Operations` says

> A sanitize operation alters all user data in the NVM subsystem such that recovery of any previous user data
> from any cache, the non-volatile media, or any Controller Memory Buffer is not possible. It is implementation
> specific whether Submission Queues and Completion Queues within a Controller Memory Buffer are altered
> by a sanitize operation; all other data stored in all Controller Memory Buffers is altered by a sanitize
> operation.

tuxrenci:
At the core of this problem is the phrase **_I think_**.

The purpose of documentation is to remove any ambiguity and turn **_I think_** into **_I know_**.

I suggest that specifications don't help; plenty of software deviates (intentionally or not) from underlying specifications (otherwise every TCP/IP stack implementation would perform identically, to the dismay of OS fingerprinters). Moreover, documentation exists not to recite the spec, but to tell the user what the software **actually does**.

igaw:
> The purpose of documentation is to remove any ambiguity and turn I think into I know.

To make it short, I don't know the answer, hence I am just guessing.

ikegami-t:
It is described by the specification as below.
> 3.3.3.1 Submission Queue Entry
> Figure 87: Common Command Format
> Namespace Identifier (NSID): ... If the namespace identifier is not used for the command, then this field shall be cleared to 0h. ...

So it seems thta if the block device is specified instead of the character device it is cleared to zero by the device controller software but actually it is depended on the device controller software behavior. I think that the device driver sets the name space ID value to submit the commnad to device if the block device is specified as same with other commands.
By the way is this really required to describe this by the document? (Sorry I am not sure actual the command behavior since on my local environment if the command executed the device data is erased.)

[...]

PR #1728, opened 2022-11-10
doc: Fix sanitize command examples as character device for issue #1559

ikegami-t:


igaw:
Add a little bit to the commit message and dropped the generated files from the commit.

igaw:
Fixes: #1728

PR #1738, opened 2022-12-02
solidigm, ocp: Added OCP-Clear-FW-Update-History support.

lgdacunh:


igaw:
@arthurshau please have a look at the OCP bits.

plugins/ocp/ocp-utils.c
```diff
@@ -0,0 +1,29 @@
...
+	0x6f, 0xbe, 0x56, 0x8f, 0x99, 0x29, 0x1d, 0xa2, 0x94, 0x47,
+	0x94, 0xe0, 0x5b, 0xd5, 0x94, 0xc1 };
+
+int ocp_get_uuid_index(struct nvme_dev *dev, int *index)
+{
+	struct nvme_id_uuid_list uuid_list;
+	int err = nvme_identify_uuid(dev_fd(dev), &uuid_list);
+
+	*index = 0;
+	if (!err) {
```
igaw, line 20: Return early is the preferred style. So this would be

```c
if (err)
    return err;

for (...) {
}
```

igaw:
@lgdacunh please add documentation for the new command under Documentation. While at it also add something the commit message. It is very terse.

plugins/ocp/ocp-utils.c
lgdacunh: Done!

lgdacunh:
I added the requested documentation, elaborated the commit message a bit more, and modified the the requested code style change.

igaw:
@lgdacunh sorry for the slow response. Next week will be back to normal response times :)

arthurshau:
@lgdacunh @igaw Apologies for the slow response as well. Yes, the bits all look correct to me.

lgdacunh:
I force-pushed some small changes, to  fix some typos and white spaces introduced on my last force-push, and to make the documentation about spec versions a bit more accurate.

igaw:
The implementation looks good. I am just not sure if the new command should live in new file. The other to commands from the ocp plugin are implemented in ocp-nvme.c. @arthurshau any feedback on this?

lgdacunh:
@igaw, just a comment to help giving a better perspective on where the code should live... and to share my assumptions.
The OCP specification asks to expose some common generic feature on each vendor plug-in. I opted to add this generic OCP feature on the the generic OCP plug-in, and redirect the Solidigm plug-in here. I did this to reduce code duplication. This way every vendor can reuse the same OCP feature implementation.
There is a few other generic OCP features that I am planning on doing the same.
I am open to change how the code is organized, if you want to emphasize other aspects of the code, instead of avoiding code duplication. 
If we decide to follow the current structure of this pull request, the code that should be common to multiple OCP features goes on ocp-utils.c, and  each feature that can be shared by multiple plug-ins goes on its own ocp-*.c file.

igaw:
The reason I asked for a review from Arthur, is because he asked me to do so whenever a non trivial patch touches the ocp plugin. I've sent an email out, maybe the github notification do not  reach him. Anyway, let's give him a chance to review.

Personally, I would merge the new code into the existing file but we can do that also afterwards.

arthurshau:
Sorry, just caught up on the thread now.

Hmm... My first reaction was that I would prefer to keep the code all in one file, as it is right now with the other plugins. I do think @lgdacunh's explanation makes sense, but that would also require sign-on by the other vendor plugin developers to do it this way. I am personally not the point of contact between Meta and the other vendors, so let me bring this up with the team.

I'll discuss with the team when I get back from vacation after the holidays. For now, I'll leave it up to you @igaw. I think in this one instance it's probably OK to just accept the pull request. Like you said, we can always merge it back into ocp-nvme.c later if we decide we want to do that.

igaw:
Alright, let's merge it as it is. We can still move bits around later.

Issue #1742, opened 2022-12-15
Negative values when parsing structures with output format JSON


Neney13rx:
Noticed that with v2+ that nvmecli commands with --output-format=json parses large integers as signed resulting in negative values and converts integers to floating point.

```
# nvme version
nvme version 2.2.1 (git 2.2.1)
libnvme version 1.2 (git 1.2)

# nvme list -o json
{
  "Devices":[
    {
      "NameSpace":1,
      ...
      "UsedBytes":-2147483648,
      "MaximumLBA":937684566,
      "PhysicalSize":-2147483648,
      "SectorSize":4096
    },
```

On another drive, we can observe that calling `id-ctrl` with `--output-format=json` returns floating values for the total and unused capacities entries, in addition to a negative value for the sanicap field.
```
nvme id-ctrl /dev/nvme1 --output-format=json
{
  ...
  "tnvmcap":960197124096.0,
  "unvmcap":0.0,
  ...
  "sanicap":-1610612733,
  ...
```

Normal output displays the correct values.
```
$ sudo nvme id-ctrl /dev/nvme1 --output-format=normal
NVME Identify Controller:
...
tnvmcap   : 960197124096
unvmcap   : 0
...
sanicap   : 0xa0000003
```

igaw:
Fixed by #https://github.com/linux-nvme/nvme-cli/pull/1743

PR #1743, opened 2022-12-16
nvme-print: Fix to add uint and uint64 json object values to output

ikegami-t:
This fix is for the issue #1742 opened.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -169,13 +169,13 @@ static void json_nvme_id_ns(struct nvme_id_ns *ns, bool cap_only)
 		}
 
 		json_object_add_value_int(root, "mssrl", le16_to_cpu(ns->mssrl));
-		json_object_add_value_int(root, "mcl", le32_to_cpu(ns->mcl));
+		json_object_add_value_uint(root, "mcl", le32_to_cpu(ns->mcl));
 		json_object_add_value_int(root, "msrc", ns->msrc);
 	}
 	json_object_add_value_int(root, "nulbaf", ns->nulbaf);
 
 	if (!cap_only) {
-		json_object_add_value_int(root, "anagrpid", le32_to_cpu(ns->anagrpid));
+		json_object_add_value_uint(root, "anagrpid", le32_to_cpu(ns->anagrpid));
 		json_object_add_value_int(root, "endgid", le16_to_cpu(ns->endgid));
 
 		memset(eui64, 0, sizeof(eui64_buf));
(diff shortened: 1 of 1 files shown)


nvme-print.c
```diff
@@ -169,13 +169,13 @@ static void json_nvme_id_ns(struct nvme_id_ns *ns, bool cap_only)
 		}
 
 		json_object_add_value_int(root, "mssrl", le16_to_cpu(ns->mssrl));
-		json_object_add_value_int(root, "mcl", le32_to_cpu(ns->mcl));
+		json_object_add_value_uint(root, "mcl", le32_to_cpu(ns->mcl));
```
igaw, line 172: We should also update the normal terminal output for this value. The printf format string `%d` is used currently.

nvme-print.c
```diff
@@ -281,19 +281,19 @@ static void json_nvme_id_ctrl(struct nvme_id_ctrl *ctrl,
 	json_object_add_value_int(root, "hctma", le16_to_cpu(ctrl->hctma));
 	json_object_add_value_int(root, "mntmt", le16_to_cpu(ctrl->mntmt));
 	json_object_add_value_int(root, "mxtmt", le16_to_cpu(ctrl->mxtmt));
-	json_object_add_value_int(root, "sanicap", le32_to_cpu(ctrl->sanicap));
-	json_object_add_value_int(root, "hmminds", le32_to_cpu(ctrl->hmminds));
+	json_object_add_value_uint(root, "sanicap", le32_to_cpu(ctrl->sanicap));
+	json_object_add_value_uint(root, "hmminds", le32_to_cpu(ctrl->hmminds));
```
igaw, line 285: also for `hmminds` we should update the `%d` format specifier.

Could you double check that all the print statements for JSON and terminal output agree on the output format? Thanks!

ikegami-t: Just pushed the additional patch to update the mcl and hmminds normal terminal output print format.
Also just checked the outputs as [test.txt](https://github.com/linux-nvme/nvme-cli/files/10260841/test.txt) but sorry some command outputs are not supported on my environment so not tested actually.

igaw:
FWIW, I've updated all places where we print a uint32_t type to use `%u` or 
`json_object_add_value_uint()`

One thing I of 'beauty' I stumbled upon is

```c
        long long lba = nvme_ns_get_lba_size(n);
        double nsze = nvme_ns_get_lba_count(n) * lba;
        double nuse = nvme_ns_get_lba_util(n) * lba;

        const char *s_suffix = suffix_si_get(&nsze);
        const char *u_suffix = suffix_si_get(&nuse);
        const char *l_suffix = suffix_binary_get(&lba);
```
`nvme_ns_get_lba_size()` returns an `int`, `nvme_ns_get_lba_count()` return a `uint64_t` and `suffix_si_get()` does internally do a float type division. IIRC, this is not going to work correctly on 32 bit platforms.

igaw:
Fixes: #1742

ikegami-t:
> FWIW, I've updated all places where we print a uint32_t type to use `%u` or `json_object_add_value_uint()`
> 
> One thing I of 'beauty' I stumbled upon is
> 
> ```c
>         long long lba = nvme_ns_get_lba_size(n);
>         double nsze = nvme_ns_get_lba_count(n) * lba;
>         double nuse = nvme_ns_get_lba_util(n) * lba;
> 
>         const char *s_suffix = suffix_si_get(&nsze);
>         const char *u_suffix = suffix_si_get(&nuse);
>         const char *l_suffix = suffix_binary_get(&lba);
> ```
> 
> `nvme_ns_get_lba_size()` returns an `int`, `nvme_ns_get_lba_count()` return a `uint64_t` and `suffix_si_get()` does internally do a float type division. IIRC, this is not going to work correctly on 32 bit platforms.

Thanks for your reviewing and commnets.
Just FYI: About the 32 bit platforms issue mentioned by you I have just tested the code on my local 32 bit platform envinronment then seems it works correctly as same with the 64 bit plotform as same output printed so looks no problem.

igaw:
Thanks for testing. I got properly confused with all the type conversion.

Issue #1745, opened 2022-12-26
Display Terabytes written for data_units_read and data_units_written (nvme smart-log -H)


jmhands:
enhancement
When executing `nvme smart-log` with `-H`, display TB (terabytes) alongside the raw output. Vendors specify endurance rating in TBW, and NVMe raw output is a bit confusing without looking at the specification.

`sudo nvme smart-log -H /dev/nvme0n1`
example
```
data_units_read                         : 1,526,808,040
data_units_written                      : 1,277,863,316
```

example with proposed enhancement (similar to smartctl output)

```
Data Units Read:                    1,526,808,040 [781 TB]
Data Units Written:                 1,277,863,316 [654 TB]
```

per Figure 207: SMART / Health Information Log Page in NVM-Express-Base-Specification-2.0c-2022.10.04-Ratified.pdf

**Data Units Read**: Contains the number of 512 byte data units the host has read from the
controller as part of processing a SMART Data Units Read Command; this value does not
include metadata. This value is reported in thousands (i.e., a value of 1 corresponds to 1,000
units of 512 bytes read) and is rounded up (e.g., one indicates that the number of 512 byte
data units read is from 1 to 1,000, three indicates that the number of 512 byte data units read
is from 2,001 to 3,000).
Refer to the specific I/O Command Set specification for the list of SMART Data Units Read
Commands that affect this field.
A value of 0h in this field indicates that the number of SMART Data Units Read is not reported.

**Data Units Written**: Contains the number of 512 byte data units the host has written to the
controller as part of processing a User Data Out Command; this value does not include
metadata. This value is reported in thousands (i.e., a value of 1 corresponds to 1,000 units of
512 bytes written) and is rounded up (e.g., one indicates that the number of 512 byte data
units written is from 1 to 1,000, three indicates that the number of 512 byte data units written
is from 2,001 to 3,000).
Refer to the specific I/O Command Set specification for the list of User Data Out Commands
that affect this field.
A value of 0h in this field indicates that the number of Data Units Written is not reported

jmhands:
excellent! thanks all, looking to rewrite the NVMe-cli guide on nvmexpresss.org at some point soon, and making things more user-friendly (don't always have to look at NVMe spec) will be incredibly helpful

igaw:
Just merged @ikegami-t patch.

igaw:
with `LC_ALL=de_DE.UTF-8`

```
[...]

PR #1749, opened 2022-12-28
nvme-print: Display smart log data units read and written SI value

ikegami-t:
This resolves for the issue #1745.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -6255,10 +6255,14 @@ void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid,
 		smart->percent_used);
 	printf("endurance group critical warning summary: %#x\n",
 		smart->endu_grp_crit_warn_sumry);
-	printf("data_units_read				: %s\n",
-		uint128_t_to_string(le128_to_cpu(smart->data_units_read)));
-	printf("data_units_written			: %s\n",
-		uint128_t_to_string(le128_to_cpu(smart->data_units_written)));
+	printf("Data Units Read				: %s (%s)\n",
+		uint128_t_to_string(le128_to_cpu(smart->data_units_read)),
+		uint128_t_to_si_string(le128_to_cpu(smart->data_units_read),
+				       1000 * 512));
+	printf("Data Units Written			: %s (%s)\n",
+		uint128_t_to_string(le128_to_cpu(smart->data_units_written)),
+		uint128_t_to_si_string(le128_to_cpu(smart->data_units_written),
+				       1000 * 512));
 	printf("host_read_commands			: %s\n",
 		uint128_t_to_string(le128_to_cpu(smart->host_reads)));
 	printf("host_write_commands			: %s\n",
--- a/util/suffix.c
+++ b/util/suffix.c
@@ -38,35 +38,31 @@
 #include <math.h>
 
 static struct si_suffix {
-	double magnitude;
+	long double magnitude;
 	const char *suffix;
 } si_suffixes[] = {
+	{1e30, "Q"},
+	{1e27, "R"},
+	{1e24, "Y"},
+	{1e21, "Z"},
+	{1e18, "E"},
 	{1e15, "P"},
 	{1e12, "T"},
 	{1e9, "G"},
 	{1e6, "M"},
 	{1e3, "k"},
 	{1e0, ""},
-	{1e-3, "m"},
-	{1e-6, "u"},
-	{1e-9, "n"},
-	{1e-12, "p"},
-	{1e-15, "f"},
 	{0}
 };
 
 const char *suffix_si_get(double *value)
 {
-	struct si_suffix *s;
+	long double value_ld = *value;
+	const char *suffix = suffix_si_get_ld(&value_ld);
 
-	for (s = si_suffixes; s->magnitude != 0; s++) {
-		if (*value >= s->magnitude) {
-			*value /= s->magnitude;
-			return s->suffix;
-		}
-	}
+	*value = value_ld;
 
-	return "";
+	return suffix;
 }
 
 uint64_t suffix_si_parse(const char *value, bool *suffixed)
@@ -92,6 +88,20 @@ uint64_t suffix_si_parse(const char *value, bool *suffixed)
 		errno = EINVAL;
 
 	return (uint64_t)ret;
+}
+
+const char *suffix_si_get_ld(long double *value)
+{
+	struct si_suffix *s;
+
+	for (s = si_suffixes; s->magnitude != 0; s++) {
+		if (*value >= s->magnitude) {
+			*value /= s->magnitude;
+			return s->suffix;
+		}
+	}
+
+	return "";
 }
 
 static struct binary_suffix {
--- a/util/types.c
+++ b/util/types.c
@@ -3,10 +3,12 @@
 #include <inttypes.h>
 #include <stdio.h>
 #include <string.h>
+#include <locale.h>
 
 #include <ccan/endian/endian.h>
 
 #include "types.h"
+#include "util/suffix.h"
 
 nvme_uint128_t le128_to_cpu(__u8 *data)
 {
@@ -46,14 +48,22 @@ uint64_t int48_to_long(__u8 *data)
 
 char *uint128_t_to_string(nvme_uint128_t val)
 {
-	static char str[40];
-	int idx = 40;
+	static char str[60];
+	int idx = 60;
 	__u64 div, rem;
+	char *sep = localeconv()->thousands_sep;
+	int len = sep ? strlen(sep) : 0;
+	int i;
 
 	/* terminate at the end, and build up from the ones */
 	str[--idx] = '\0';
 
 	do {
+		if (len && !((sizeof(str) - idx) % (3 + len))) {
+			for (i = 0; i < len; i++)
+				str[--idx] = sep[i];
+		}
+
 		rem = val.words[0];
 
 		div = rem / 10;
@@ -76,6 +86,35 @@ char *uint128_t_to_string(nvme_uint128_t val)
 	} while (val.words[0] || val.words[1] || val.words[2] || val.words[3]);
 
 	return str + idx;
+}
+
+static long double uint128_t_to_double(nvme_uint128_t data)
+{
+	int i;
+	long double result = 0;
+
+	for (i = 0; i < sizeof(data.words) / sizeof(*data.words); i++) {
+		result *= 4294967296;
+		result += data.words[i];
+	}
+
+	return result;
+}
+
+char *uint128_t_to_si_string(nvme_uint128_t val, __u32 bytes_per_unit)
+{
+	static char str[40];
+	long double bytes = uint128_t_to_double(val) * bytes_per_unit;
+	const char *suffix = suffix_si_get_ld(&bytes);
+	int n = snprintf(str, sizeof(str), "%.2Lf %sB", bytes, suffix);
+
+	if (n <= 0)
+		return "";
+
+	if (n >= sizeof(str))
+		str[sizeof(str) - 1] = '\0';
+
+	return str;
 }
 
 const char *util_uuid_to_string(unsigned char uuid[NVME_UUID_LEN])
--- a/unit/meson.build
+++ b/unit/meson.build
@@ -2,7 +2,7 @@
 
 test_uint128 = executable(
     'test-uint128',
-    ['test-uint128.c', '../util/types.c'],
+    ['test-uint128.c', '../util/types.c', '../util/suffix.c'],
     include_directories: [incdir, '..'],
     dependencies: [libnvme_dep],
 )
@@ -17,3 +17,12 @@ test_suffix_si_parse = executable(
 )
 
 test('suffix_si_parse', test_suffix_si_parse)
+
+test_uint128_si = executable(
+    'test-uint128-si',
+    ['test-uint128-si.c', '../util/types.c', '../util/suffix.c'],
+    include_directories: [incdir, '..'],
+    dependencies: [libnvme_dep],
+)
+
+test('uint128-si', test_uint128_si)
--- a/unit/test-uint128-si.c
+++ b/unit/test-uint128-si.c
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "../util/types.h"
+
+#define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
+
+/* create a uint128_t from four uint32_ts. w0 is the most significant value,
+ * w2 the least */
+#define U128(w0, w1, w2, w3) { .words = { w0, w1, w2, w3 } }
+
+static int test_rc;
+
+static void check_str(nvme_uint128_t val, __u32 bytes_per_unit, const char *exp,
+		      const char *res)
+{
+	if (!strcmp(res, exp))
+		return;
+
+	printf("ERROR: printing {%08x.%08x.%08x.%08x} (bytes per unit %u), got '%s', expected '%s'\n",
+	       val.words[3], val.words[2], val.words[1], val.words[0],
+	       bytes_per_unit, res, exp);
+
+	test_rc = 1;
+}
+
+struct tostr_test {
+	nvme_uint128_t val;
+	__u32 bytes_per_unit;
+	const char *exp;
+};
+
+static struct tostr_test tostr_tests[] = {
+	{ U128(0, 0, 0, 0), 1, "0.00 B" },
+	{ U128(0, 0, 0, 1), 1, "1.00 B" },
+	{ U128(0, 0, 0, 10), 1, "10.00 B" },
+	{ U128(4, 3, 2, 1), 1, "316.91 RB" },
+	{ U128(0xffffffff, 0xffffffff, 0xffffffff, 0xffffffff), 1,
+	  "340282366.92 QB" },
+	{ U128(0, 0, 0, 0xae0dc2), 1000 * 512, "5.84 TB" },
+	{ U128(0, 0, 0, 0xf9c546), 1000 * 512, "8.38 TB" },
+	{ U128(0, 0, 0, 0x4c2aa594), 1000 * 512, "654.27 TB" },
+	{ U128(0, 0, 0, 0x5b013de8), 1000 * 512, "781.73 TB" },
+};
+
+void tostr_test(struct tostr_test *test)
+{
+	char *str;
+	str = uint128_t_to_si_string(test->val, test->bytes_per_unit);
+	check_str(test->val, test->bytes_per_unit, test->exp, str);
+}
+
+int main(void)
+{
+	unsigned int i;
+
+	test_rc = 0;
+
+	for (i = 0; i < ARRAY_SIZE(tostr_tests); i++)
+		tostr_test(&tostr_tests[i]);
+
+	return test_rc ? EXIT_FAILURE : EXIT_SUCCESS;
+}
--- a/util/suffix.h
+++ b/util/suffix.h
@@ -37,6 +37,7 @@
 
 const char *suffix_si_get(double *value);
 uint64_t suffix_si_parse(const char *value, bool *suffixed);
+const char *suffix_si_get_ld(long double *value);
 const char *suffix_binary_get(long long *value);
 const char *suffix_dbinary_get(double *value);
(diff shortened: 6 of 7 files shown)


util/types.c
```diff
@@ -45,16 +45,19 @@ uint64_t int48_to_long(__u8 *data)
...
+	static char str[60];
+	int idx = 60;
 	__u64 div, rem;
 
 	/* terminate at the end, and build up from the ones */
 	str[--idx] = '\0';
 
 	do {
+		if (thousands_grouping && !((sizeof(str) - idx) % 4))
+			str[--idx] = ',';
```
igaw, line 59: Just rely on printf localization and use `"'%d"` 

```
The five flag characters above are defined in the C99 standard.  The Single UNIX Specification spec-
       ifies one further flag character.

       '      For  decimal  conversion  (i,  d,  u, f, F, g, G) the output is to be grouped with thousands'
              grouping characters if the locale information indicates any.  (See setlocale(3).)  Note  that
              many  versions  of  gcc(1) cannot parse this option and will issue a warning.  (SUSv2 did not
              include %'F, but SUSv3 added it.)
```

See also https://stackoverflow.com/questions/10602729/digit-grouping-in-cs-printf

nvme-print.c
```diff
@@ -6255,10 +6261,15 @@ void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid,
...
 	printf("endurance group critical warning summary: %#x\n",
 		smart->endu_grp_crit_warn_sumry);
-	printf("data_units_read				: %s\n",
-		uint128_t_to_string(le128_to_cpu(smart->data_units_read)));
-	printf("data_units_written			: %s\n",
-		uint128_t_to_string(le128_to_cpu(smart->data_units_written)));
+	printf("Data Units Read				: %s %s\n",
+		uint128_t_to_string(le128_to_cpu(smart->data_units_read)),
+		human ?
+		int128_to_si_string(smart->data_units_read, 1000 * 512) : "");
```
igaw, line 6267: `data_units_read` is unsigned, so we should have a `uint128_to_si_string` here and not `int128_to_si_string`

util/suffix.c
```diff
@@ -69,6 +69,47 @@ const char *suffix_si_get(double *value)
...
+	{1e-15, "f"},
+	{1e-18, "a"},
+	{1e-21, "z"},
+	{1e-24, "y"},
+	{1e-27, "r"},
+	{1e-30, "q"},
+	{0}
+};
+
+const char *suffix_si_get_ld(long double *value)
```
igaw, line 99: If we process 128 bit values, we should use either uint128_t or int128_t. `long double` is the wrong type IMO

igaw: BTW,  this is also what I don't like on the other `suffix_*` functions.

util/types.c
```diff
@@ -78,6 +79,18 @@ char *uint128_t_to_string(nvme_uint128_t val)
 	return str + idx;
 }
 
+char *int128_to_si_string(__u8 *data, __u32 bytes_per_unit)
+{
+	static char str[40];
+	long double val = int128_to_double(data) * bytes_per_unit;
```
igaw, line 93: Does this work on 32bit machines? 

BTW, I would like to ask you to add unit tests for these functions, like the ones in `tests/test-uint128.c`.

igaw (changes requested):
I think these util function deserve a unit test. At least I would be way more confident to accept such changes when I see that some tests say all is good.

util/types.c
ikegami-t: Yes but the variable type nvme_uint128_t is not used by printf I think. Instead of this the 128 bit value can be converted t o long doulbe then we can use printf to group the value with thousands' grouping characters.

nvme-print.c
ikegami-t: Actually int128_to_si_string passes `data_units_read ` as __u8 pointer to int128_to_double.
```
char *int128_to_si_string(__u8 *data, __u32 bytes_per_unit)
{
	static char str[40];
	long double val = int128_to_double(data) * bytes_per_unit;
...
```
Then also int128_to_double converts the unsigned pointer values to long double so seems no problem.
```
long double int128_to_double(__u8 *data)
{
	int i;
	long double result = 0;

	for (i = 0; i < 16; i++) {
		result *= 256;
		result += data[15 - i];
	}
	return result;
}
```

util/suffix.c
ikegami-t: suffix_si_get converts decimal values as for example from 0.001 to 1m or from 0.000001 to 1u etc. also so double or long double values are necessary but not used uint128_t or int128_t integer types I think.

util/types.c
ikegami-t: Yes I will do test on 32bit environment and add unit tests as mentioned.
Note: My local 32bit enbironment is virtual environment as I used 32bit Ubuntu as guest OS on virtual box but use 64bit PC and Windows host OS.

ikegami-t:
Thanks for your reviewing and comments. I see I will do add a unit test.

ikegami-t:
Just pushed additional commit to add unit tests. Still I will do test on 32 bit environment later.

util/types.c
ikegami-t: Tested on 32 bit environment and confirmed as works correctly.

nvme-print.c
```diff
@@ -6237,12 +6237,18 @@ void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid,
...
+		printf("      Temp. Threshold[1]             : %d\n",
+		       (smart->critical_warning & 0x02) >> 1);
+		printf("      NVM subsystem Reliability[2]   : %d\n",
+		       (smart->critical_warning & 0x04) >> 2);
+		printf("      Read-only[3]                   : %d\n",
+		       (smart->critical_warning & 0x08) >> 3);
+		printf("      Volatile mem. backup failed[4] : %d\n",
+		       (smart->critical_warning & 0x10) >> 4);
+		printf("      Persistent Mem. RO[5]          : %d\n",
+		       (smart->critical_warning & 0x20) >> 5);
```
igaw, line 6251: This is not change is not relevant, please drop it or just move into a new patch. But I think it is not necessary at all.

nvme-print.c
igaw: The data members are either `unsigned int` or `int` (for example `smart->data_units_read`).  They can't be both.  

Isn't  `int128_to_double(data) * bytes_per_unit`  prone to overflow as the multiplication is done as integer type?

tests/meson.build
```diff
@@ -57,7 +57,7 @@ endif
 
 test_uint128 = executable(
     'test-uint128',
-    ['test-uint128.c', '../util/types.c'],
+    ['test-uint128.c', '../util/types.c', '../util/suffix.c'],
```
igaw, line 60: This looks like a leftover. As the unit test is added in patch #3.

util/suffix.c
```diff
@@ -69,6 +69,47 @@ const char *suffix_si_get(double *value)
...
+	{1e-3, "m"},
+	{1e-6, "u"},
+	{1e-9, "n"},
+	{1e-12, "p"},
+	{1e-15, "f"},
+	{1e-18, "a"},
+	{1e-21, "z"},
+	{1e-24, "y"},
+	{1e-27, "r"},
+	{1e-30, "q"},
```
igaw, line 96: You can drop the unit types for `< 0`.  It's not relevant for nvme. And instead duplicating these constants from just above, make these reusable.

util/types.c
```diff
@@ -78,6 +79,18 @@ char *uint128_t_to_string(nvme_uint128_t val)
...
+	const char *suffix = suffix_si_get_ld(&val);
+
+	if (sprintf(str, "(%.2Lf %sB)", val, suffix) > 0)
```
igaw: Maybe `snprintf` would be a better choice or are you sure that we never write more than 40 chars?

util/types.c
igaw: Ah yes, you are right. There is not support from glibc for this type. But we still can't just use a comma as grouping blindly as this is a localization question.  So I I think you should use `localeconv()` ( to retrieve the correct char for grouping and do that on default. This also avoids the bool argument.

util/suffix.c
```diff
@@ -76,6 +76,7 @@ static struct si_suffix_ld {
 	{1e30, "Q"},
 	{1e27, "R"},
 	{1e24, "Y"},
+	{1e21, "Z"},
```
igaw, line 79: this change should be in the patch which introduces this.

nvme-print.c
```diff
@@ -6261,15 +6261,16 @@ void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid,
...
 		smart->endu_grp_crit_warn_sumry);
-	printf("Data Units Read				: %s %s\n",
+	printf("Data Units Read				: %s %s%s%s\n",
 		uint128_t_to_string_tg(le128_to_cpu(smart->data_units_read),
-		human), human ?
-		int128_to_si_string(smart->data_units_read, 1000 * 512) : "");
-	printf("Data Units Written			: %s %s\n",
+		human), human ? "(" : "", human ?
+		int128_to_si_string(smart->data_units_read, 1000 * 512) : "",
+		human ? ")" : "");
```
igaw, line 6268: The same function prints the temperature always in Celsius and Kelvin (in brackets). I think I would just always print both strings.

igaw (changes requested):

nvme-print.c
ikegami-t: 'smart->data_units_read' is defined as __u8 array as below and `le128_to_cpu` converts it to `nvme_uint128_t`.
```
	__u8			data_units_read[16];
```
`int128_to_si_string` uses existing `int128_to_double` and it converts the array to 'long double'. (existing `int48_to_long` also does convert the 48bit size array to uint64_t.)
Now I am thiking to change the name `int128_to_si_string` to `le128_to_si_string` but seems better to change existing `int128_to_double` and `int48_to_long` names also to `le(num)_to_(type)`.

tests/meson.build
ikegami-t: No 'int128_to_si_string' added in patch #1 uses `suffix_si_get_ld` so needed this to add and a build error occurred if not added this.

util/types.c
ikegami-t: Okay I will do use `localeconv()` mentioned. But about the bool argument it is to keep the existing `uint128_t_to_string()` it does not use a comma as grouping and seems that it should be not changed.

ikegami-t:
Pushed an additional patch to fix for the review comments. If needed I will rebase the commits and add commit message decription more later.

util/types.c
igaw: Indeed, if we do localization (which includes thousends grouping) we have to do it everywhere.

nvme-print.c
igaw: You should really use the `le128_to_cpu` getter to extract the value and pass it into the print function as we do it for `uint128_t_to_string`. Having two different API types for this is not good.

util/suffix.c
igaw: My argument is that we have implicit type conversion which can be trouble. If we can avoid those, it's better solution in the long term.

tests/meson.build
igaw: Ah got it. :)

util/types.c
```diff
@@ -50,13 +51,21 @@ char *uint128_t_to_string_tg(nvme_uint128_t val, bool thousands_grouping)
 	static char str[60];
 	int idx = 60;
 	__u64 div, rem;
+	char *sep = localeconv()->thousands_sep;
+	int len = sep ? strlen(sep) : 0;
```
igaw, line 55: If I got this right, the separator is always one byte. So if `sep != NULL` than `len == 1`.

igaw (changes requested):
- drop uint128_t_to_string_tg, always use localization if available.
- add localization also for `int` and `uint` outputs.
- suffix_si_get_ld() askes for a `long double` but `magnitude` is only a `double`. Again, this might be okay but implicit type conversion. Best avoided.
- use le128_to_cpu() getter and use the uint128 type as for `uint128_to_si_string` (instead of `le128_to_si_string`)
- snprintf() usage needs to handle the case null terminating is missing.

util/suffix.c
keithbusch: Regarding the 128bit types, isn't it not guaranteed to be available with many compiler+arch combinations? The long-double usage was initially a compromise to have an easily printable value with acceptable precision.

igaw: We have defined our own uint128_t type. As we don't do any real arithmetic operation on this type we can get away with a few simple hacks. Basically we just need read and print operations. This even works on 32bit platforms.

igaw: https://github.com/linux-nvme/nvme-cli/blob/c045535fb0e0e3d3059ed65f18ff9f541e2ad0ac/util/types.h#L20
https://github.com/linux-nvme/nvme-cli/blob/c045535fb0e0e3d3059ed65f18ff9f541e2ad0ac/util/types.c#L11

util/types.c
ikegami-t: About the struct lconv char * pointer members it is described as below so looks `len == 0` is possible if `sep != NULL` since `""` that is `len == 0` but not `len == 1` is set to the thousands_sep member.
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf
```
7.11.2 Numeric formatting convention inquiry
7.11.2.1 The localecon
...
Description
3 The members of the structure with type char * are pointers to strings, any of which
(except decimal_point) can point to "", to indicate that the value is not available in
the current locale or is of zero length.
```

igaw: Argh, so the spec allows "" as a valid separator, which is for obvious reasons pretty stupid. Thanks for looking this up.

ikegami-t:
Thanks for your review comments. I have just fixed for them and rebased the branch with the latest master.
But only atbout the comment for snprintf() NULL terminating below I have confirmed the followng things so not  fixed yet but seems not necessary.
> * snprintf() usage needs to handle the case null terminating is missing.
1. It is described by the linux man page as below so seems not necessary to handle by the code.
https://linux.die.net/man/3/snprintf
```
snprintf(3) - Linux man page
...
Description
...
The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.
```
2. On my local tests environment actually it is terminated with NULL by the snpintf() as set the size to 9 for debugging and the str is only written as 8 characters.
```
tokunori@tokunori-desktop:~/nvme-cli$ .build/tests/test-uint128-si
ERROR: printing {00000001.00000002.00000003.00000004} (bytes per unit 1), got '316.91 R', expected '316.91 RB'
ERROR: printing {ffffffff.ffffffff.ffffffff.ffffffff} (bytes per unit 1), got '34028236', expected '340282366.92 QB'
ERROR: printing {4c2aa594.00000000.00000000.00000000} (bytes per unit 512000), got '654.27 T', expected '654.27 TB'
ERROR: printing {5b013de8.00000000.00000000.00000000} (bytes per unit 512000), got '781.73 T', expected '781.73 TB'
```
Note: If needed I will merge the commits to one or more so please let me know.

igaw:
from the same man page:
```
       The  functions snprintf() and vsnprintf() do not write more than size bytes (including the terminat-
       ing null byte ('\0')).  If the output was truncated due to this limit, then the return value is  the
       number  of characters (excluding the terminating null byte) which would have been written to the fi-
       nal string if enough space had been available.  Thus, a return value of size or more means that  the
       output was truncated.  (See also below under NOTES.)
```

ikegami-t:
About the snprintf handling still I am not sure if really required it but to make sure just changed to add the checking if NULL terminating needed to add. Also merged the commits to one so please review the changes again.

ikegami-t:
Just confirmed the snprintf implementation as below so seems it only writes `maxlen (size) - 1` charactors to string buffer and terminates the string with NULL correctly.
https://sourceware.org/git/?p=glibc.git;a=blob;f=libio/vsnprintf.c;h=8dae667;hb=c804cd1 (tag: glibc-2.36)
```
static int
_IO_strn_overflow (FILE *fp, int c)
{
  /* When we come to here this means the user supplied buffer is
     filled.  But since we must return the number of characters which
     would have been written in total we must provide a buffer for
     further use.  We can do this by writing on and on in the overflow
     buffer in the _IO_strnfile structure.  */
  _IO_strnfile *snf = (_IO_strnfile *) fp;

  if (fp->_IO_buf_base != snf->overflow_buf)
    {
      /* Terminate the string.  We know that there is room for at
	 least one more character since we initialized the stream with
	 a size to make this possible.  */
      *fp->_IO_write_ptr = '\0';
...
    }
...
  return c;
}
...
int
__vsnprintf_internal (char *string, size_t maxlen, const char *format,
		      va_list args, unsigned int mode_flags)
{
  _IO_strnfile sf;
  int ret;
...
  string[0] = '\0';
  _IO_str_init_static_internal (&sf.f, string, maxlen - 1, string);
  ret = __vfprintf_internal (&sf.f._sbf._f, format, args, mode_flags);

  if (sf.f._sbf._f._IO_buf_base != sf.overflow_buf)
    *sf.f._sbf._f._IO_write_ptr = '\0';
  return ret;
}
```
https://sourceware.org/git/?p=glibc.git;a=blob;f=stdio-common/snprintf.c;h=717e6a1;hb=c804cd1 (tag: glibc-2.36)
```
int
__snprintf (char *s, size_t maxlen, const char *format, ...)
{
  va_list arg;
  int done;

  va_start (arg, format);
  done = __vsnprintf_internal (s, maxlen, format, arg, 0);
  va_end (arg);

  return done;
}
libc_hidden_def (__snprintf)
ldbl_weak_alias (__snprintf, snprintf)
```

igaw:
glibc decided to make snprintf to behave like this. But the POSIX standard is pretty clear on this. The string termination might missing. There are other C library implementations (uclib, musl, ...) which might not do this.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/snprintf.html

ikegami-t:
> glibc decided to make snprintf to behave like this. But the POSIX standard is pretty clear on this. The string termination might missing. There are other C library implementations (uclib, musl, ...) which might not do this.
> 
> https://pubs.opengroup.org/onlinepubs/9699919799/functions/snprintf.html

Thanks for your explanation. I could understood well. Currently the patch implemented to handle the snprintf NULL terminating I think.

ikegami-t:
By the way can we build nvme-cli on FreeBSD environment? Now I am trying to build to confirm the POSIX standard snprintf behavior but the `linux/types.h` not found error caused on it.
```
tokunori@tokunori:~/nvme-cli $ ninja -C .build
ninja: Entering directory `.build'
[1/105] Compiling C object subprojects...rc/libnvme.so.1.2.0.p/nvme_filters.c.o
FAILED: subprojects/libnvme/src/libnvme.so.1.2.0.p/nvme_filters.c.o
cc -Isubprojects/libnvme/src/libnvme.so.1.2.0.p -Isubprojects/libnvme/src -I../s                                                                                                                                                             ubprojects/libnvme/src -Isubprojects/libnvme -I../subprojects/libnvme -Isubproje                                                                                                                                                             cts/libnvme/ccan -I../subprojects/libnvme/ccan -Isubprojects/libnvme/internal -I                                                                                                                                                             ../subprojects/libnvme/internal -Isubprojects/json-c-0.16 -I../subprojects/json-                                                                                                                                                             c-0.16 -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=gnu99                                                                                                                                                              -O3 -fomit-frame-pointer -D_GNU_SOURCE -include internal/config.h -fPIC -pthrea                                                                                                                                                             d -MD -MQ subprojects/libnvme/src/libnvme.so.1.2.0.p/nvme_filters.c.o -MF subpro                                                                                                                                                             jects/libnvme/src/libnvme.so.1.2.0.p/nvme_filters.c.o.d -o subprojects/libnvme/s                                                                                                                                                             rc/libnvme.so.1.2.0.p/nvme_filters.c.o -c ../subprojects/libnvme/src/nvme/filter                                                                                                                                                             s.c
In file included from ../subprojects/libnvme/src/nvme/filters.c:21:
In file included from ../subprojects/libnvme/src/nvme/filters.h:13:
In file included from ../subprojects/libnvme/src/nvme/tree.h:19:
In file included from ../subprojects/libnvme/src/nvme/ioctl.h:15:
../subprojects/libnvme/src/nvme/types.h:16:10: fatal error: 'linux/types.h' file                                                                                                                                                              not found
#include <linux/types.h>
...
```

igaw (approved):

Issue #1747, opened 2022-12-26
nvme create-ns input of --nsze adn --ncap in standard SI units


jmhands:
accept SI units in `--nsze` and `--ncap` for `nvme create-ns`

```
  [  --nsze=<IONUM>, -s <IONUM> ]       --- size of ns (NSZE)
  [  --ncap=<IONUM>, -c <IONUM> ]       --- capacity of ns (NCAP)
```

example current command
`nvme create-ns /dev/nvme0 –nsze 11995709440 –ncap 1199570940 –flbas 0 –dps 0 –nmic 0`

proposed command
`nvme create-ns /dev/nvme0 –nsze 6.14T –ncap 6.14T –flbas 0 –dps 0 –nmic 0`

ikegami-t:
FYI: Just confirmed as currenty the binaly units can be used as below but not SI units then the value is just parsed to 6,597,069,766,656 (= 6 x 1,024 x 1,024 x 1,024 x 1,024) as not divided by the LBA size.
```
nvme create-ns /dev/nvme0 -nsze 6Ti -ncap 1199570940 -flbas 0 -dps
```
Note: `6T` is also can be used insed but the prased value is same with `6Ti`. Also `6.14Ti` decimal value using `.` is not parsed correctly as parsed to 0.

igaw:
The change is merged.

PR #1750, opened 2022-12-30
nvme: Allow standard SI units for ns-create nsze and ncap input

ikegami-t:
This is proposed by the issue #1747.

### DIFF
--- a/Documentation/nvme-create-ns.txt
+++ b/Documentation/nvme-create-ns.txt
@@ -19,6 +19,9 @@ SYNOPSIS
 			[--lbstm=<lbstm> | -l <lbstm>]
 			[--block-size=<block-size> | -b <block-size>]
 			[--timeout=<timeout> | -t <timeout>]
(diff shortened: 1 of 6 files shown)


keithbusch:
I feel like we need entirely new arguments, and ensure they're mutually exclusive with the existing nsze/ncap args. The old ones follow the spec defined units, so I think it'd actually be more confusing to repurpose them to include si units.

ikegami-t:
Okay I will do split as new arguments. Thank you.

igaw (changes requested):
- Merge the 3 patch into 1.
- Provide documentation for these new arguments.
- Please add commit message text not just referencing the github issues.. Commit messages should be self-explanatory. This also avoids strong dependency on infrastructure we don't control.

Thanks!

ikegami-t:
Updated the commit as mentioned.

igaw (approved):

igaw:
Looks good! Thanks!

PR #1751, opened 2022-12-30
Bash completion: move global declare into function

QiangWeiSel:
* To fix completion issue by moving global vairables into function _nvme_subcmds. It is not only suite for bash completion system but also support manually source bash script.
* Fix pluginx_sfx_opts spell error to plugin_sfx_opts.
* Correct variable NO_OPTS used in bash assginment.

Signed-off-by: Qiang Wei <qiang.wei@suse.com>


igaw:
Thanks!

PR #1753, opened 2023-01-01
nvme: sanitize: Allow sanitize action string values

ikegami-t:

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4588,6 +4588,28 @@ ret:
 	return err;
 }
 
+static int parse_sanact(char *str, __u8 *val)
+{
+	int len = strlen(str);
+
+	if (!strncasecmp(str, "exit-failure", len > 1 ? len : 1))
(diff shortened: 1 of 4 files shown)


nvme.c
```diff
@@ -4588,6 +4588,28 @@ static int ns_rescan(int argc, char **argv, struct command *cmd, struct plugin *
 	return err;
 }
 
+static int parse_sanact(char *str, __u8 *val)
+{
+	int len = strlen(str);
+
+	if (!strncasecmp(str, "exit-failure", len > 1 ? len : 1))
```
igaw, line 4595: Couldn't also accept the old accepted values here? E.g. Also the documentation should be updated accordingly

igaw (changes requested):
at least the documentation needs to be updated.

nvme.c
ikegami-t: By the change both the old accepted values and new string values are accepted. Noted I will update the documentation also.

ikegami-t:
Updated the documentation.

PR #1756, opened 2023-01-04
Add 'fdp' plugin for managing devices supporting TP4146 ("Flexible Data Placement")

birkelund:
This adds support for managing and inspecting devices and endurance groups supporting TP4146 ("Flexible Data Placement") through a plugin similar to that for zoned namespaces.

Marking as draft since the top commit will be dropped when https://github.com/linux-nvme/libnvme/pull/552 is merged.


nvme.c
```diff
@@ -1893,6 +1893,210 @@ static int get_supp_cap_config_log(int argc, char **argv, struct command *cmd,
...
+close_dev:
+	dev_close(dev);
+	return err;
+}
+
+static int io_mgmt_recv(int argc, char **argv, struct command *cmd, struct plugin *plugin)
+{
+	const char *desc = "I/O Management Receive";
+	const char *mos = "Management Operation Specific";
+	const char *mo = "Management Operation";
```
igaw, line 2001: @ikegami-t  was busy grouping common strings at the beginning of the file. I suppose we should these also there.

nvme-print.c
```diff
@@ -2292,6 +2292,149 @@ static void json_supported_cap_config_log(
...
+	__u8 rgif = fdpa & 0xf;
+
+	printf("  [7:7] : %#x\tFDP Configuration %sValid\n",
+		valid, valid ? "" : "Not ");
+	if (rsvd)
+		printf("  [6:5] : %#x\tReserved\n", rsvd);
+	printf("  [4:4] : %#x\tFDP Volatile Write Cache %sPresent\n",
+		fdpvwc, fdpvwc ? "" : "Not ");
+	printf("  [3:0] : %#x\tReclaim Group Identifier Format\n", rgif);
+}
```
igaw, line 2309: I can find the corresponding JSON output for this? Do I miss something?

igaw (changes requested):
looks in good shape.

Just one thing: is the JSON output missing?

nvme.c
birkelund: Okidokie.

nvme-print.c
birkelund: Generally I don't think nvme-cli does this for the json output (this is individual bits in a "flag" value).

nvme.c
```diff
@@ -1893,6 +1896,204 @@ static int get_supp_cap_config_log(int argc, char **argv, struct command *cmd,
 	return err;
 }
 
+static int io_mgmt_send(int argc, char **argv, struct command *cmd, struct plugin *plugin)
+{
+	const char *desc = "I/O Management Send";
+	const char *data = "optional file for data (default stdin)";
+	const char *data_len = "length of data in bytes";
```
ikegami-t, line 1903: Seems this is also can be moved at the beginning of the file as common strings. By the way there is also the folloing common string already so it also can be used instead.
```
static const char *buf_len = "buffer len (if) data is returned";
```

ikegami-t:
Seems documentations should be updated and added.

igaw:
Yes, please also update the documentation accordingly. Thanks!

igaw:
I think you need to do something like -Ddbus:werror=false

birkelund:
> I think you need to do something like `-Ddbus:werror=false`, for the static builds

Yeah, and on the fallback one as well.

birkelund:
Looking at the commit history, I am not 100% sure if I should regenerate the documentation, or if that is something you do on release?

birkelund:
> looks in good shape.
> 
> Just one thing: is the JSON output missing?

Added json output.

igaw:
I update the doc on release, see the `release.sh` script. I am trying to document the release steps as code :)

nvme.c
birkelund: Switch to `buf_len`, but updated the description a little bit since it is used for both sends and receives.

igaw (approved):

Issue #1759, opened 2023-01-06
`nvme(1)` should get a switch to stay in the foreground


moviuro:
# Why?

I'm trying to write a `systemd.unit(5)` for `nvme-device-self-test(1)` but because there is no background/wait-until-complete switch, it is impossible to write a unit that can be properly started and stopped.

# Example

```
# nvme --foreground device-self-test --self-test-code=2 /dev/nvme0n1
... useful data can be printed
... test complete!
#
```

When the nvme command is complete it can even, depending on the command, return an exit status (in the case of device-self-test, 0 if no issues were found, 2 if the test was aborted, etc.)

```ini
[Unit]
Description=Run a short nvme device test on %f
Documentation=man:nvme-device-self-test(1)

[Service]
Type=oneshot
ExecStart=/usr/bin/nvme --foreground device-self-test -s 1 %f
ExecStop=/usr/bin/nvme device-self-test -s 0xf %f
```

keithbusch:
I'm unfamiliar with writing such units, but is it possible to do a more complicated script instead? If you can, loop on `nvme self-test-log -e 0 %f | grep "Current Completion"`, and wait for the line to indicate 100%?

I suppose we could also do that within a single command too. The extended self tests can go for several hours, though.

moviuro:
@keithbusch this would cause issues with the "real" PID of the service though: what process is in charge of the test, who do we kill/signal when something happens, etc.? The loop would also add unnecessarily "load" to the system for limited gain (though a while sleep loop is not that much, it could add up).

keithbusch:
Well, *something* is going to have to do a sleep loop on the self test log, so I'm not sure if there's anything we can do about the load concern.

igaw:
I was pondering if the wakeup could be done via udev? Though I have no clue if this makes sense.

keithbusch:
If the self test completion was accompanied by AEN, then udev might be okay, but it's not. We can only poll the log, so something will need to take on that responsibility.

igaw:
So the idea is that when `nvme device-self-test --self-test-code=1 /dev/nvme0n1 --wait` (or whatever is called), we first check if there is ongoing self test. If so we wait until it finishes in a poll loop or start a new self test and then do the poll loop.

moviuro:
Sounds like a good starting point. Further question: what happens when `nvme(1)` gets a signal?

* TERM: stop following/looping or stop the test?
* USR1: print data?
* HUP ?

[...]

PR #1762, opened 2023-01-10
add wait for device self test

keithbusch:
Link: https://github.com/linux-nvme/nvme-cli/issues/1759
Signed-off-by: Keith Busch <kbusch@kernel.org>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -3957,6 +3957,47 @@ ret:
 	return err;
 }
 
+static const char dash[50] = {[0 ... 49] = '='};
+static const char space[50] = {[0 ... 49] = ' '};
+
+static void wait_self_test(struct nvme_dev *dev)
+{
+	static const char spin[] = {'-', '\\', '|', '/' };
+	struct nvme_self_test_log log;
+	int err, i = 0, p = 0;
+
+	printf("Waiting for self test completion...\n");
+	while (true) {
+		if (p == 100) {
+			printf("[%.*s] %2d%%\n", 50, dash, p);
+			break;
+		}
+
+		printf("\r[%.*s%c%.*s] %3d%%", p / 2, dash, spin[i % 4], 49 - p / 2, space, p);
+		fflush(stdout);
+		sleep(1);
+
+		err = nvme_cli_get_log_device_self_test(dev, &log);
+		if (err) {
+			printf("\n");
+			if (err < 0)
+				perror("self test log\n");
+			else
+				nvme_show_status(err);
+			break;
+		}
+
+		if (log.completion < p) {
+			printf("\n");
+			fprintf(stderr, "progress broken\n");
+			break;
+		}
+
+		p = log.completion;
+		i++;
+	}
+}
+
 static int device_self_test(int argc, char **argv, struct command *cmd, struct plugin *plugin)
 {
 	const char *desc  = "Implementing the device self-test feature"\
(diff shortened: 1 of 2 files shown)


igaw:
I think in this case we need something like

```c
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wformat-overflow"

[...]

#pragma GCC diagnostic pop

```

nvme.c
```diff
@@ -3957,6 +3957,47 @@ static int list_secondary_ctrl(int argc, char **argv, struct command *cmd, struc
...
+
+static void wait_self_test(struct nvme_dev *dev)
+{
+	static const char spin[] = {'-', '\\', '|', '/' };
+	struct nvme_self_test_log log;
+	int err, i = 0, p = 0;
+
+	printf("Waiting for self test completion...\n");
+	while (true) {
+		if (p == 100) {
```
ikegami-t, line 3971: The value is described by the specfication as below but can we check the value as 100% correctly? Also not neccesary to check the value as 0% as completed but not `progress broken`?
```
Current Device Self-Test Completion
If bits 3:0 in the Current Device Self-Test Operation field are cleared to 0h (indicating there is no device self-test operation in progress), then this field is ignored.
```

keithbusch:
> I think in this case we need something like
> 
> ```c
> #pragma GCC diagnostic push
> #pragma GCC diagnostic ignored "-Wformat-overflow"
> 
> [...]
> 
> #pragma GCC diagnostic pop
> ```

We can just null terminate the array instead if the compiler isn't able to figure out we'll never overflow it.

nvme.c
ikegami-t: The wait function sleeps every one second so I think whether the value is possible to be changed to zero in the one second period. Anyway I will test it later. By the way about the termination I thought the wait option can be set the timeout threshold also as optionally.

igaw: I'd rather not exposing the wait timeout on the command line. If you want to be fancy the wait time could be calculated from the previous loop. So if you don't do a lot of progress in 1 second increase the timeout. But this is fancy stuff, I would also like to avoid at this point.

ikegami-t: Okay I see. Thanks for your comments. Sorry on my local environment the dst command is not supported as below so if any chance I will retry on other environment.
```
[2023-01-13 20:29:06.603] tokunori@tokunori-desktop:~/nvme-cli$ sudo .build/nvme device-self-test /dev/nvme0 -w
[2023-01-13 20:29:07.603] NVMe status: Invalid Field in Command: A reserved coded value or an unsupported value in a defined field(0x2)
[2023-01-13 20:29:07.618] tokunori@tokunori-desktop:~/nvme-cli$ sudo .build/nvme device-self-test /dev/nvme0 -w
[2023-01-13 20:29:09.509] NVMe status: Invalid Field in Command: A reserved coded value or an unsupported value in a defined field(0x2)
[2023-01-13 20:29:09.509] tokunori@tokunori-desktop:~/nvme-cli$
```

igaw (changes requested):
I couldn't help and done the changes myself: https://github.com/linux-nvme/nvme-cli/pull/1770

nvme.c
ikegami-t: Thanks for your advice. Yes I could do execute the dst command with the '-s' option as you mentioned as below. And also confirmed the waiting failed as 'progress broken'. (The '-w' option is not checked correctly as you mentoned as always waits without the option.)
```
tokunori@tokunori-desktop:~/nvme-cli$ sudo .build/nvme device-self-test /dev/nvme0 -s 1
Short Device self-test started
Waiting for self test completion...
[===============================================/  ]  95%
progress broken
```
By the way it seems that it is better to be set the '-s' option default valeus as 1 but not as 0.

igaw:
I've merged the updated version of this PR (see https://github.com/linux-nvme/nvme-cli/pull/1770)

PR #1769, opened 2023-01-19
Create AppImage

### LINKED ISSUE #1764: make appimage format or binary file
inxi bat or other command , they allow have binary file.

appimage or binary they not use apt-get ,it can run every e:)

igaw:
Fixes #1764


igaw:
After the AppImage build an artifact is generated which should bring all the dependency etc with it:

  https://github.com/linux-nvme/nvme-cli/suites/10471832765/artifacts/518698524

igaw:
The generated AppImage artifact seems to work, though we don't have a nice place to put it yet.

PR #1777, opened 2023-01-26
TP4146 fixes

birkelund:
A set of various fixups for the TP4146 support.

### DIFF
--- a/subprojects/libnvme.wrap
+++ b/subprojects/libnvme.wrap
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = 568c5b80d1387a1491b09a8254fe039f0675d073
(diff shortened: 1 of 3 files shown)


subprojects/libnvme.wrap
```diff
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = 568c5b80d1387a1491b09a8254fe039f0675d073
+revision = 662799b5961ac1f4739825181480a32e1effa669
```
igaw, line 3: Could you split the libnvme.wrap update of this patch. In case someone has to backport this patch it will always collide with the wrap update. Yeah  I know it's a bit awkward. Though all will get better in future with libnvme 2.x :)

birkelund: Alright! :)

igaw:
Thanks!

Issue #1776, opened 2023-01-26
IO commands report error in getting 64-bit storage tag from command line


alexmchou:
Read command reports "Expected long suffixed integer argument for 'storage-tag' but got '0xdeadbeefcafebabe' when using option '-g 0xdeadbeefcafebabe' 
If storage tag is less than 64-bit, for example 48-bit, the input '-g 0xdeadbeefcafe' is OK

PR #1779, opened 2023-01-26
util: Parse suffix binary value as unsinged long long correctly

ikegami-t:


ikegami-t:
Fix for the issue #1776. Also the strage-tag option is not described by the compare, read and write commands dorumentation so added them also but desribed it only by the verify and write-zeroes commands.

igaw:
Thanks!

PR #1780, opened 2023-01-26
json_discovery_log: avoid buffer overrun

calebsander:
`nvme_strip_spaces()` has an off-by-one error and starts stripping spaces at the byte AFTER the end of the buffer. Use `space_strip_len()` instead, which doesn't have this bug and is already used in `print_discovery_log()`.


igaw:
Ah, good catch. BTW, I'll going to update libnvme as well (strchomp) accordingly.

PR #1782, opened 2023-01-27
Device self-test changes

ikegami-t:
Still thinking and trying to add changes below also etc.
1. Add device self-test SIGINT signal handling for wait option to abort
2. Use abort self-test code definition value instead hard-code value
3. Update device-self-test command documentation

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -152,6 +152,8 @@ static const char *timeout = "timeout value, in milliseconds";
 static const char *uuid_index = "UUID index";
 static const char *uuid_index_specify = "specify uuid index";
 static const char *verbose = "Increase output verbosity";
+static const char dash[51] = {[0 ... 49] = '=', '\0'};
+static const char space[51] = {[0 ... 49] = ' ', '\0'};
 
 static void *mmap_registers(nvme_root_t r, struct nvme_dev *dev);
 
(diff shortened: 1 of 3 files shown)


ikegami-t:
All changes pushed and the last patch required the libnvme changes separately the pull requrest submitted also.

nvme.c
```diff
@@ -4308,6 +4346,9 @@ static int device_self_test(int argc, char **argv, struct command *cmd, struct p
 		fprintf(stderr, "Device self-test: %s\n", nvme_strerror(errno));
 
 close_dev:
+	if (err == -EINTR)
+		abort_self_test(&args);
```
igaw, line 4349: I am wondering if this is something what the user expects? Well, I see both arguments, a) we just want to stop the foreground wait but not the background operation and b) we always want to stop everything with ctrl-c. There is properly for both situation an good reason why a certain behavior is preferred.

I think we just go with your proposal and address any feedback (if there is any).

igaw:
Just added the bump for libnvme.wrap

igaw:
Thanks!

nvme.c
ikegami-t: I have just checked the dst command implementation again so as far as I checked the err `-EINTR' is only set by the foreground wait so seems no problem and concern.

igaw: Thanks for testing. As said, I think we just see what people think of it.

Issue #1787, opened 2023-02-01
2.3: test site is failing in `suffix_si_parse` unit


kloczek:
Looks like something is wrong.
```console
+ /usr/bin/meson test -C x86_64-redhat-linux-gnu --num-processes 48 --print-errorlogs
ninja: no work to do.
ninja: Entering directory `/home/tkloczko/rpmbuild/BUILD/nvme-cli-2.3/x86_64-redhat-linux-gnu'
ninja: no work to do.
1/3 uint128         OK              0.02s
2/3 suffix_si_parse FAIL            0.01s   exit status 1
>>> MALLOC_PERTURB_=125 /home/tkloczko/rpmbuild/BUILD/nvme-cli-2.3/x86_64-redhat-linux-gnu/unit/test-suffix-si-parse
――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
ERROR: printing {11995709440} (lbas 512), got '23429120', expected '11995709440'
ERROR: printing {1199570940} (lbas 512), got '2342911', expected '1199570940'
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

3/3 uint128-si      OK              0.01s

Summary of Failures:

2/3 suffix_si_parse FAIL            0.01s   exit status 1

Ok:                 2
Expected Fail:      0
Fail:               1
Unexpected Pass:    0
Skipped:            0
Timeout:            0
```
I'm not sure what kind of details I should provide to diagnose that issue.

igaw:
@ikegami-t

igaw:
So we need something like this:

```diff
@@ -84,9 +84,12 @@ uint64_t suffix_si_parse(const char *value, bool *suffixed)
                }
        }
 
-       if (suffix[0] != '\0')
+       if (suffix[0] != '\0') {
                errno = EINVAL;
+               return 0;
+       }
 
+       *suffixed = false;
        return (uint64_t)ret;
 }
```

But when I look at the test code, I think we have another bug in `suffix_si_parse`:

```c
static struct tonum_test tonum_tests[] = {
	{ "11995709440", 512, 11995709440 },
	{ "1199570940", 512, 1199570940 },
	{ "6.14T", 512, 11992187500 },
	{ "6.14T", 520, 11807692307 },
	{ "6.14T", 4096, 1499023437 },
	{ "6.14", 512, 0 },
	{ "6.14#", 512, 0 },
};
```
[...]

PR #1788, opened 2023-02-01
test: Initialize suffixed flag for suffix-si-parse test

ikegami-t:

### DIFF
--- a/unit/test-suffix-si-parse.c
+++ b/unit/test-suffix-si-parse.c
@@ -34,14 +34,14 @@ static struct tonum_test tonum_tests[] = {
 	{ "6.14T", 512, 11992187500 },
 	{ "6.14T", 520, 11807692307 },
 	{ "6.14T", 4096, 1499023437 },
-	{ "6.14", 512, 0 },
+	{ "6.14", 512, 6 },
 	{ "6.14#", 512, 0 },
 };
 
 void tonum_test(struct tonum_test *test)
 {
 	__u64 num;
-	bool suffixed;
+	bool suffixed = false;
 
 	num = suffix_si_parse(test->val, &suffixed);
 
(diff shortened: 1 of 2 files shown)


ikegami-t:
Sorry this is a fix for the issue #1787.

igaw:
No worries, that's why we have unit tests :)

unit/test-suffix-si-parse.c
```diff
@@ -41,7 +41,7 @@ static struct tonum_test tonum_tests[] = {
 void tonum_test(struct tonum_test *test)
 {
 	__u64 num;
-	bool suffixed;
+	bool suffixed = false;
```
igaw, line 44: Why not fixing the conversion function? I should always set suffixed when it doesn't return an error.

ikegami-t: Fixed the patch and chaged the test also again as the expected value for "6.14" to `6` from `0` since the `suffix_si_parse` and its `strtod` behavior was changed to return `6` from `0`. I am not sure the behavior change reason but seems not caused by the fix changes.

igaw: This could be caused by the `suffixed` variable to false. This looks wrong too. Let me address this.

ikegami-t: Thank you. Noted it.

igaw: I've re implemented the function. So don't invest more time on it :)

ikegami-t: Okay I will do close this pull request. Thank you.

igaw: Thanks anyway. Your patch is a good fix for the v2.3 release :)

PR #1789, opened 2023-02-01
nvme: Check invalid field status code value correctly for get feature…

ikegami-t:
… command

Since previously unnecessary DNR, M and CAD status fields were checked.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4450,6 +4450,12 @@ static int get_feature_id(struct nvme_dev *dev, struct feat_cfg *cfg,
 	return nvme_cli_get_features(dev, &args);
 }
 
+static int filter_out_flags(int status)
+{
+	return status & (NVME_GET(NVME_SCT_MASK, SCT) |
+			 NVME_GET(NVME_SC_MASK, SC));
+}
+
 static void get_feature_id_print(struct feat_cfg cfg, int err, __u32 result,
 				 void *buf)
 {
@@ -4471,7 +4477,8 @@ static void get_feature_id_print(struct feat_cfg cfg, int err, __u32 result,
 			d_raw(buf, cfg.data_len);
 		}
 	} else if (err > 0) {
-		if (!nvme_status_equals(err, NVME_STATUS_TYPE_NVME,
+		if (!nvme_status_equals(filter_out_flags(err),
+					NVME_STATUS_TYPE_NVME,
 					NVME_SC_INVALID_FIELD))
 			nvme_show_status(err);
 	} else {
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -4450,6 +4450,19 @@ static int get_feature_id(struct nvme_dev *dev, struct feat_cfg *cfg,
 	return nvme_cli_get_features(dev, &args);
 }
 
+static bool nvme_status_invalid_field(int status)
+{
+	enum nvme_status_type type = NVME_STATUS_TYPE_NVME;
+	unsigned int value = NVME_SC_INVALID_FIELD;
```
igaw, line 4456: Just use the values directly in `nvme_status_equals()`, no need to define them as local variable.

igaw (changes requested):
1dacde708f44 ("nvme: use helpers for checking status types") updated also `nvme_get_properties()` but dropped the DNR mask filter. So this patch should update this call side as well.

I am not really happy with a helper calling a helper to test flags. It's very convoluted and I am wondering we should have something in libnvme? @jk-ozlabs any ideas?

nvme.c
ikegami-t: Noted and will do fix. (I just thought to shorten the 'nvme_status_equals()` line since the definition value names NVME_S*** length not short.)

ikegami-t:
It seems that it is better to change `nvme_status_equals()` to check the value without DNR, M and CAD status fields values but I am not sure the changes impact for other caller. As you mentioned `nvme_get_properties()` dropped the mask so seems same fix needed as this patch.

igaw:
I think we should not change `nvme_status_equals()` as there might be users which depend on reading the `DNR` bit.

igaw:
What about naming the function `status_equals(int status, enum nvme_status_field value)` which makes the function identical to `nvme_status_equals` except stripping the DNR bits?  This would make it at least a bit more consistent with the existing `nvme_status_equals`

igaw:
Or maybe even better is just to introduce a helper which strips the bits from the status.

```c
static int filter_out_flags(status)
{
         return status & NVME_GET(NVME_SCT_MASK, SCT) | NVME_GET(NVME_SC_MASK, SC));
}

[...]
       if (err && !nvme_status_equals(filter_out_flags(err), 
                                      NVME_STATUS_TYPE_NVME,
                                      NVME_SC_INVALID_FIELD))
[...]```

In fact I like this approach way better, because you don't hide the filtering.

ikegami-t:
Fixed the patch as mentioned. Thank you.

igaw:
Thanks!

jk-ozlabs:
> @jk-ozlabs any ideas?

A bit late to the party here, but: we should probably just mask out DNR in `nvme_status_equals`; it's not like this defines the type of the status...

ikegami-t:
Can we fix and change the libnvme API `nvme_status_equals()` behavior? (By the way in the past the libnvme API `nvme_cli_get_feature_length2()` was added to keep the `nvme_cli_get_feature_length()' behavior.)
Also if we can change it seems better to add the API to check the status flag DNR, M and CAD fields too.

jk-ozlabs:
Yeah, I think that's justifiable as a fix, rather than a change that breaks API. @igaw would that be okay with you?

PR #1790, opened 2023-02-02
util: Reimplement suffix_si_parse

igaw:
suffix_si_parse() has an very awkward interface with using errno and suffixed as return value. Let's just use the return code as status value and use an argument for returning the value and avoid setting the errno.

Furthermoe, this function should not know anything about LBAs, it just supposed to parse a string with a suffix. In order to support the previous use case we also return the endprt which allows to implement the nsze-si parse use case and the caller side (see documentation on create-ns).

Next problem is that suffix_is_parse uses double for calculation which tend to get inaccurate, e.g. 6.14T results in 61399999997. Instead let's do do all in pure integer arithmetic.

Also handle the decimal point according the locale settings.

Fixes: #1788

### DIFF
--- a/unit/test-suffix-si-parse.c
+++ b/unit/test-suffix-si-parse.c
@@ -3,6 +3,8 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <errno.h>
+#include <locale.h>
 
 #include "../util/suffix.h"
 #include "../util/types.h"
@@ -11,44 +13,54 @@
 
 static int test_rc;
 
-static void check_num(const char *val, int lbas, __u64 exp, __u64 num)
+static void check_num(const char *val, __u64 exp, __u64 num)
 {
 	if (exp == num)
 		return;
 
-	printf("ERROR: printing {%s} (lbas %d), got '%llu', expected '%llu'\n",
-	       val, lbas, (unsigned long long)num, (unsigned long long)exp);
+	printf("ERROR: printing {%s},  got '%llu', expected '%llu'\n",
+	       val, (unsigned long long)num, (unsigned long long)exp);
 
 	test_rc = 1;
 }
 
 struct tonum_test {
 	const char *val;
-	int lbas;
-	const __u64 exp;
+	const uint64_t exp;
+	int ret;
 };
 
 static struct tonum_test tonum_tests[] = {
-	{ "11995709440", 512, 11995709440 },
-	{ "1199570940", 512, 1199570940 },
-	{ "6.14T", 512, 11992187500 },
-	{ "6.14T", 520, 11807692307 },
-	{ "6.14T", 4096, 1499023437 },
-	{ "6.14", 512, 0 },
-	{ "6.14#", 512, 0 },
+	{ "11995709440", 11995709440, 0 },
+	{ "1199570940", 1199570940, 0},
+	{ "234.567M", 234567000, 0 },
+	{ "1.2k", 1200, 0 },
+	{ "6.14T", 6140000000000, 0 },
+	{ "123.4567k", 123456, 0 },
+	{ "12345.6789101112M", 12345678910, 0},
+	{ "6.14", 6, 0 },
+	{ "6.14#", 0, -EINVAL },
+	{ "2,33", 0, -EINVAL },
+	{ "3..3", 0, -EINVAL },
+	{ "123.12MM", 0, -EINVAL },
 };
 
 void tonum_test(struct tonum_test *test)
 {
-	__u64 num;
-	bool suffixed;
+	char *endptr;
+	uint64_t num;
+	int ret;
 
-	num = suffix_si_parse(test->val, &suffixed);
+	ret = suffix_si_parse(test->val, &endptr, &num);
+	if (ret != test->ret) {
+		printf("ERROR: converting {%s} failed\n", test->val);
+		test_rc = 1;
+		return;
+	}
+	if (ret)
+		return;
 
-	if (suffixed)
-		num /= test->lbas;
-
-	check_num(test->val, test->lbas, test->exp, num);
+	check_num(test->val, test->exp, num);
 }
 
 int main(void)
@@ -56,6 +68,7 @@ int main(void)
 	unsigned int i;
 
 	test_rc = 0;
+	setlocale(LC_NUMERIC, "C");
 
 	for (i = 0; i < ARRAY_SIZE(tonum_tests); i++)
 		tonum_test(&tonum_tests[i]);
--- a/util/suffix.c
+++ b/util/suffix.c
@@ -167,14 +167,15 @@ static struct binary_suffix {
 	{30, "Gi"},
 	{20, "Mi"},
 	{10, "Ki"},
-	{0, ""}
 };
 
 const char *suffix_binary_get(long long *value)
 {
-	struct binary_suffix *s;
+	int i;
 
-	for (s = binary_suffixes; s->shift != 0; s++) {
+	for (i = 0; i < ARRAY_SIZE(binary_suffixes); i++) {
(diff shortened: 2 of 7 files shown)


util/suffix.c
```diff
@@ -31,28 +31,31 @@
...
 #include <ctype.h>
 #include <errno.h>
 #include <math.h>
+#include <float.h>
+#include <limits.h>
+#include <locale.h>
 
 static struct si_suffix {
 	long double magnitude;
+	unsigned int power;
```
ikegami-t, line 46: Is it really both magnitude and power required? Seems magnitude value can be calcurated from the power without the definition.

ikegami-t:
> suffix_si_parse() has an very awkward interface with using errno and suffixed as return value. Let's just use the return code as status value and use an argument for returning the value and avoid setting the errno.

Yes agree with you as the function should be more simply.
Note: The function is based on `suffix_binary_get()` that also using errno but just changed to use suffixed as mentioned. Yes just confirmed you changed the function 'suffix_binary_parse()' also. Thank you.

> Furthermoe, this function should not know anything about LBAs, it just supposed to parse a string with a suffix. In order to support the previous use case we also return the endprt which allows to implement the nsze-si parse use case and the caller side (see documentation on create-ns).

I think that the function does not care about the LBAs as you mentioned but just use suffixed as above.

> Next problem is that suffix_is_parse uses double for calculation which tend to get inaccurate, e.g. 6.14T results in 61399999997. Instead let's do do all in pure integer arithmetic.

I see.

> Also handle the decimal point according the locale settings.

Yes you are right.

Anyway the patch changes basically look good but I am still reviewing so will comment later.

unit/test-suffix-si-parse.c
```diff
@@ -11,51 +13,61 @@
...
-	check_num(test->val, test->lbas, test->exp, num);
+	check_num(test->val, test->exp, num);
 }
 
 int main(void)
 {
 	unsigned int i;
 
 	test_rc = 0;
+	setlocale(LC_NUMERIC, "C");
```
ikegami-t, line 70: About the locale is it enough to test "C"? It seems so but to make sure let me confirm it.

unit/test-suffix-si-parse.c
```diff
@@ -11,51 +13,61 @@
...
-	{ "11995709440", 512, 11995709440 },
-	{ "1199570940", 512, 1199570940 },
-	{ "6.14T", 512, 11992187500 },
-	{ "6.14T", 520, 11807692307 },
-	{ "6.14T", 4096, 1499023437 },
-	{ "6.14", 512, 0 },
-	{ "6.14#", 512, 0 },
+	{ "11995709440", 11995709440, 0 },
+	{ "1199570940", 1199570940, 0},
+	{ "234.567M", 234567000, 0 },
```
ikegami-t, line 36: I think that the val "234.567MM" is parsed as "234.567M". If so is it expected?

util/suffix.c
```diff
@@ -145,24 +198,30 @@ const char *suffix_dbinary_get(double *value)
...
 		return 0;
+	}
+
+	for (i = 0; i < ARRAY_SIZE(binary_suffixes); i++) {
+		struct binary_suffix *s = &binary_suffixes[i];
 
-	struct binary_suffix *s;
-	for (s = binary_suffixes; s->shift != 0; s++) {
-		if (tolower(suffix[0]) == tolower(s->suffix[0])) {
+		if (tolower(*endptr[0]) == tolower(s->suffix[0])) {
```
ikegami-t, line 219: The character `i` of the suffix "Xi" is not checked but ignored and is it not necessary?

util/suffix.c
igaw: Yes, properly it works, though I am slightly scared by the math function might result in small errors. Though this pow() should return the correct magnitude.

Hmm, but the name `power` is wrong, should be `exponent` or short `exp`.

unit/test-suffix-si-parse.c
igaw: Should be okay, I've checked the documentation. Obviously, we could add a test which tests for ',' decimal point but I guess but there are better things to test :)

unit/test-suffix-si-parse.c
igaw: Good point, this should be handled as well.

util/suffix.c
igaw: Don't know... existing code. But this parser could also get some improvements.

ikegami-t: I see. Thanks for your replies.

util/suffix.c
igaw: I think we should keep the magnitude as it avoid pulling in the math library (`libm`). I think the space tradeoff is better.

util/suffix.c
igaw: I've updated the test so it tests also for the possible second char.

unit/test-suffix-si-parse.c
igaw: Updated the check and also the tests case.

igaw:
Or maybe even better is just to introduce a helper which strips the bits from the status.

```c
static int filter_out_flags(status)
{
         return status & NVME_GET(NVME_SCT_MASK, SCT) | NVME_GET(NVME_SC_MASK, SC));
}

[...]
       ig (err && !nvme_status_equals(filter_out_flags(err), 
                                      NVME_STATUS_TYPE_NVME,
                                      NVME_SC_INVALID_FIELD))
[...]```

In fact I like this approach way better, because you don't hide the filtering.

File: util/suffix.c [hunk 5]
```diff
@@ -65,36 +68,87 @@ const char *suffix_si_get(double *value)
...
+	char *sep, *tmp;
+	int frac_len, len, i;
+
+	num = strtoull(str, endptr, 0);
+	if (str == *endptr ||
+	    ((num == ULLONG_MAX) && errno == ERANGE))
+		return -EINVAL;
+
+	/* simple number, no decimal point not suffix */
+	if (*endptr[0] == '\0') {
```
ikegami-t, line 83: This also should be fixed as same with other casting `endptr` parts. (By the way it seems that it is same with the case without the casting for the pointer of pointer as `(*endptr)[n]` and `*endptr[n]` are same but not sure..)
```diff
- *endptr[0]
+ (*endptr)[0]
```

ikegami-t, line 83: This also should be fixed as same with other casting `endptr` parts. (By the way it seems that it is same with the case without the casting for the pointer of pointer as `(*endptr)[n]` and `*endptr[n]` are same but not sure..)
```diff
- *endptr[0]
+ (*endptr)[0]
```

ikegami-t, line 83: This also should be fixed as same with other casting `endptr` parts. (By the way it seems that it is same with the case without the casting for the pointer of pointer as `(*endptr)[n]` and `*endptr[n]` are same but not sure..)
```diff
- *endptr[0]
+ (*endptr)[0]
```

ikegami-t:
Just commented a minor comment but basically looks good.
Reviewed-by: Tokunori Ikegami <ikegami.t@gmail.com>

util/suffix.c
igaw: Oh I missed those. I'll fix them as well.

The operator precedence says  `[]` wins over `*`. That means `*endptr[1]` will point to invalid memory. It happens 
that `*endptr[0]` and `(*endptr)[0]` point to the same memory address.

ikegami-t: Thanks for your reply and explanations. Yes I know the precedence but now I could underand the diffence between `*endptr[1]` and `(*endptr)[1]` but same for `*endptr[0]` and `(*endptr)[0]`.
P.S. Sorry my same comments were repeated as missed so deleted them. (Sometimes github comments seems not updated to add so repeated to push the comment button.)

PR #1791, opened 2023-02-02
Added NBFT table support

johnmeneghini:
Added support for parsing and printing the contents of the NBFT table (per NVMe-oF boot specification v1.0).

```
 $ .build/nvme nbft --help
 nvme-2.4
 usage: nvme nbft  command [device] [args]
 
 The  <device> may be either an NVMe character device (ex: /dev/nvme0), an 
 nvme block device (ex: /dev/nvme0n1), or a mctp address in the form
 mctp net,eid ctrl-id 

 ACPI NBFT table extensions

 The following are all implemented sub-commands:
   show            Show contents of ACPI NBFT tables
   version         Shows the program version
   help            Display this help

 See nvme nbft help <command> for more information on a specific command

 $ .build/nvme nbft show --help
 Usage: nvme nbft show <device> [OPTIONS]

 Display contents of the ACPI NBFT files.

 Options:
   [  --output-format=<FMT>, -o <FMT> ]  --- Output format: normal|json
   [  --subsystem, -s ]                  --- show NBFT subsystems
   [  --hfi, -H ]                        --- show NBFT HFIs
   [  --discovery, -d ]                  --- show NBFT discovery controllers
   [  --nbft-path=<STR> ]                --- user-defined path for NBFT tables

```

Signed-off-by: Stuart Hayes <stuart_hayes@dell.com>
Signed-off-by: Tomas Bzatek <tbzatek@redhat.com>
Signed-off-by: Martin Wilck <mwilck@suse.com>
Signed-off-by: John Meneghini <jmeneghi@redhat.com>

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -40,6 +40,7 @@
 
 #include "common.h"
 #include "nvme.h"
+#include "nbft.h"
 #include "libnvme.h"
 #include "nvme-print.h"
 #include "nvme-print-json.h"
@@ -710,6 +711,7 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 {
 	char *subsysnqn = NVME_DISC_SUBSYS_NAME;
 	char *hostnqn = NULL, *hostid = NULL, *hostkey = NULL;
+	char *hostnqn_arg, *hostid_arg;
 	char *transport = NULL, *traddr = NULL, *trsvcid = NULL;
 	char *config_file = PATH_NVMF_CONFIG;
 	char *hnqn = NULL, *hid = NULL;
@@ -724,6 +726,8 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 	char *device = NULL;
 	bool force = false;
 	bool json_config = false;
+	bool nbft = false, nonbft = false;
+	char *nbft_path = NBFT_SYSFS_PATH;
 
 	OPT_ARGS(opts) = {
 		OPT_STRING("device",   'd', "DEV", &device, "use existing discovery controller device"),
@@ -736,6 +740,9 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 		OPT_INCR("verbose",      'v', &verbose,       "Increase logging verbosity"),
 		OPT_FLAG("dump-config",  'O', &dump_config,   "Dump configuration file to stdout"),
 		OPT_FLAG("force",          0, &force,         "Force persistent discovery controller creation"),
+		OPT_FLAG("nbft",           0, &nbft,          "Only look at NBFT tables"),
+		OPT_FLAG("no-nbft",        0, &nonbft,        "Do not look at NBFT tables"),
+		OPT_STRING("nbft-path",  0, "STR", &nbft_path, "user-defined path for NBFT tables"),
 		OPT_END()
 	};
 
@@ -768,6 +775,8 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 	if (!nvme_read_config(r, config_file))
 		json_config = true;
 
+	hostnqn_arg = hostnqn;
+	hostid_arg = hostid;
 	if (!hostnqn)
 		hostnqn = hnqn = nvmf_hostnqn_from_file();
 	if (!hostnqn)
@@ -789,6 +798,14 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 		nvme_host_set_dhchap_key(h, hostkey);
 
 	if (!device && !transport && !traddr) {
+		if (!nonbft)
+			discover_from_nbft(r, hostnqn_arg, hostid_arg,
+					   hostnqn, hostid, desc, connect,
+					   &cfg, nbft_path, flags, verbose);
+
+		if (nbft)
+			goto out_free;
+
 		if (json_config)
 			ret = discover_from_json_config_file(r, h, desc,
 							     connect, &cfg,
--- a/fabrics.c
+++ b/fabrics.c
@@ -43,6 +43,7 @@
 #include "libnvme.h"
 #include "nvme-print.h"
 #include "nvme-print-json.h"
+#include "fabrics.h"
 
 #define PATH_NVMF_DISC		SYSCONFDIR "/nvme/discovery.conf"
 #define PATH_NVMF_CONFIG	SYSCONFDIR "/nvme/config.json"
@@ -109,15 +110,6 @@ static const char *nvmf_config_file	= "Use specified JSON configuration file or 
 	OPT_FLAG("hdr-digest",        'g', &c.hdr_digest,         nvmf_hdr_digest),	\
 	OPT_FLAG("data-digest",       'G', &c.data_digest,        nvmf_data_digest), \
 	OPT_FLAG("tls",                 0, &c.tls,                nvmf_tls)	\
-
-struct tr_config {
-	const char *subsysnqn;
-	const char *transport;
-	const char *traddr;
-	const char *host_traddr;
-	const char *host_iface;
-	const char *trsvcid;
-};
 
 /*
  * Compare two C strings and handle NULL pointers gracefully.
@@ -202,7 +194,7 @@ static nvme_ctrl_t lookup_discovery_ctrl(nvme_root_t r, struct tr_config *trcfg)
 	return __lookup_ctrl(r, trcfg, disc_ctrl_config_match);
 }
 
-static nvme_ctrl_t lookup_ctrl(nvme_root_t r, struct tr_config *trcfg)
+nvme_ctrl_t lookup_ctrl(nvme_root_t r, struct tr_config *trcfg)
 {
 	return __lookup_ctrl(r, trcfg, ctrl_config_match);
 }
--- a/nbft.c
+++ b/nbft.c
@@ -1,0 +1,202 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <errno.h>
+#include <stdio.h>
+#include <fnmatch.h>
+
+#include "nvme.h"
+#include "nbft.h"
+#include "libnvme.h"
+#include "fabrics.h"
+
+#define NBFT_SYSFS_FILENAME	"NBFT*"
+
+static void print_connect_msg(nvme_ctrl_t c)
+{
+	printf("device: %s\n", nvme_ctrl_get_name(c));
+}
+
+static void json_connect_msg(nvme_ctrl_t c)
+{
+	struct json_object *root;
+
+	root = json_create_object();
+	json_object_add_value_string(root, "device", nvme_ctrl_get_name(c));
+
+	json_print_object(root, NULL);
+	printf("\n");
+	json_free_object(root);
+}
+
+int nbft_filter(const struct dirent *dent)
+{
+	return !fnmatch(NBFT_SYSFS_FILENAME, dent->d_name, FNM_PATHNAME);
+}
+
+int read_nbft_files(struct list_head *nbft_list, char *path)
+{
+	struct dirent **dent;
+	char filename[PATH_MAX];
+	int i, count, ret;
+	struct nbft_file_entry *entry;
+	struct nbft_info *nbft;
+
+	count = scandir(path, &dent, nbft_filter, NULL);
+	if (count < 0) {
+		fprintf(stderr, "Failed to open %s.\n", path);
+		return -1;
+	}
+
+	for (i = 0; i < count; i++) {
+		snprintf(filename, sizeof(filename), "%s/%s", path, dent[i]->d_name);
+		ret = nvme_nbft_read(&nbft, filename);
+		if (!ret) {
+			entry = calloc(1, sizeof(*entry));
+			entry->nbft = nbft;
+			list_add_tail(nbft_list, &entry->node);
+		}
+		free(dent[i]);
+	}
+	free(dent);
+	return 0;
+}
+
+void free_nbfts(struct list_head *nbft_list)
+{
+	struct nbft_file_entry *entry;
+
+	while ((entry = list_pop(nbft_list, struct nbft_file_entry, node))) {
+		nvme_nbft_free(entry->nbft);
+		free(entry);
+	}
+}
+
+int discover_from_nbft(nvme_root_t r, char *hostnqn_arg, char *hostid_arg,
+		       char *hostnqn_sys, char *hostid_sys,
+		       const char *desc, bool connect,
+		       const struct nvme_fabrics_config *cfg, char *nbft_path,
+		       enum nvme_print_flags flags, bool verbose)
+{
+	char *hostnqn = NULL, *hostid = NULL, *host_traddr = NULL;
+	nvme_host_t h;
+	nvme_ctrl_t c;
+	int ret, i;
+	struct list_head nbft_list;
+	struct nbft_file_entry *entry;
+	struct nbft_info_subsystem_ns **ss;
+	struct nbft_info_hfi *hfi;
+
+	if (!connect)
+		/* to do: print discovery-type info from NBFT tables */
+		return 0;
+
+	list_head_init(&nbft_list);
+	ret = read_nbft_files(&nbft_list, nbft_path);
+	if (ret)
+		goto out_free_2;
+
+	list_for_each(&nbft_list, entry, node)
+		for (ss = entry->nbft->subsystem_ns_list; ss && *ss; ss++)
+			for (i = 0; i < (*ss)->num_hfis; i++) {
+				nvme_ctrl_t cl;
+
+				hfi = (*ss)->hfis[i];
+				if (hostnqn_arg)
+					hostnqn = hostnqn_arg;
+				else {
+					hostnqn = entry->nbft->host.nqn;
+					if (!hostnqn)
+						hostnqn = hostnqn_sys;
+				}
+
+				if (hostid_arg)
+					hostid = hostid_arg;
+				else if (*entry->nbft->host.id) {
+					hostid = (char *)util_uuid_to_string(entry->nbft->host.id);
+					if (!hostid)
+						hostid = hostid_sys;
+				}
+
+				h = nvme_lookup_host(r, hostnqn, hostid);
+				if (!h) {
+					errno = ENOMEM;
+					goto out_free;
+				}
+
+				if (!cfg->host_traddr) {
+					host_traddr = NULL;
+					if (!strncmp((*ss)->transport, "tcp", 3))
+						host_traddr = hfi->tcp_info.ipaddr;
+				}
+
+				struct tr_config trcfg = {
+					.subsysnqn	= (*ss)->subsys_nqn,
+					.transport	= (*ss)->transport,
+					.traddr		= (*ss)->traddr,
+					.host_traddr	= host_traddr,
+					.host_iface	= NULL,
+					.trsvcid	= (*ss)->trsvcid,
+				};
+
+				/* Already connected ? */
+				cl = lookup_ctrl(r, &trcfg);
+				if (cl && nvme_ctrl_get_name(cl))
+					continue;
+
+				c = nvme_create_ctrl(r, (*ss)->subsys_nqn, (*ss)->transport,
+						     (*ss)->traddr, host_traddr, NULL,
+						     (*ss)->trsvcid);
+				if (!c) {
+					errno = ENOMEM;
+					goto out_free;
+				}
+
+				errno = 0;
+				ret = nvmf_add_ctrl(h, c, cfg);
+
+				/*
+				 * With TCP/DHCP, it can happen that the OS
+				 * obtains a different local IP address than the
+				 * firmware had. Retry without host_traddr.
+				 */
+				if (ret == -1 && errno == ENVME_CONNECT_WRITE &&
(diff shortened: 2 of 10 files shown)


igaw:
Just two quick remarks as we have to agree on this first (the implementation can be changed update/changed without breaking any users):

- I think this should be a plugin and not a top level command (e.g. like `zns`).
-  The next thing is the name, `nbft` stand for `NVMe Boot Firmware Table`. Is there any plans to extend this in future? Would it make sense to name the plugin `firmware` and have a subcommand called nbft?

mwilck:
> Would it make sense to name the plugin `firmware` and have a subcommand called nbft?

This would result in 3-level subcommands like `nvme firmware nbft show` . I don't think we need this complexity. Currently the "firmware" command would support NBFT only. If any time in the future we get a different FW API, firmware supporting that API will most likely *not* support NBFT, and the kind of API used by the firmware could be an implementation detail hidden by the  `nvme firmware` command.

Let's summarize the functionality we need, independent of the FW implementation:

 1. print configuration from firmware in machine-readable format (JSON), today `nvme show-nbft -o json`;
 2. print configuration from firmware in human-readable format, currently `nvme show-nbft -o normal`;
 3. connect to subsystems configured by the firmware, currently `nvme connect-nbft`;
 4. connect to everything, including firmware-configured subsystems;
 5. connect to everything, excluding firmware-configured subsystems (for troubleshooting).

No. 4) and 5) would need integration of NBFT support in the standard subcommands `connect` and/or `connect-all`. 3) could be implemented by `nvme firmware connect`,  or e.g. using `nvme connect --firmware`.

igaw:
- a) I was wondering if `nbtf` is a good name (== plugin name), e.g. `nvme firmware show` and not `nvme firmware nbtf show`.

- b) We currently have two main connect user interfaces, `nvme connect`  and `nvme connect-all`. Introducing yet another connect call is confusing from the user point of view, e.g how does it interact with the existing `connect` and `connect-all` command and why does `connect-all` not do what it says in the name? 

I think a) is not really controversial just the question of the name. I don't mind going with `nbft`.

The bigger issue for me is b). This needs a strong argument why it should exists in the first place. This includes **documentation** how to use it correctly (if we agree it should exist).

igaw:
Just to make it clear: Why do we need another 'connect' command? This needs to be explained.

johnmeneghini:
>
 >     a) I was wondering if |nbtf| is a good name (== plugin name), e.g. |nvme firmware show| and not |nvme firmware nbtf show|.
 >

`nvme firmware` sounds like a utility used to download/upgrade the firmware on a pcie device.

I don't think we should use that for reading what is basically connection information for an nvme-of device.

Oh wait... what's this?

 > man nvme | grep -C 1 firmware
        1mnvme-micron-selective-download22m(1)
            Performs selective firmware download

--
        1mnvme-wdc-clear-fw-activate-history22m(1)
            Clears the firmware activate history table

--
        1mnvme-wdc-vs-internal-log22m(1)
            Retrieve WDC device’s internal firmware log and save to file

See what I mean?

> 
>     b) We currently have two main connect user interfaces, |nvme connect| and |nvme connect-all|. Introducing yet another
>     connect call is confusing from the user point of view, e.g how does it interact with the existing |connect| and
>     |connect-all| command and why does |connect-all| not do what it says in the name?

How about: `nvme-connect --nbft` ?

Basically, the nbft, if present, replaces:

   [  --transport=<LIST>, -t <LIST> ]    --- transport type
   [  --nqn=<LIST>, -n <LIST> ]          --- nqn name
   [  --traddr=<LIST>, -a <LIST> ]       --- transport address
   [  --trsvcid=<LIST>, -s <LIST> ]      --- transport service id (e.g. IP
                                             port)
   [  --host-traddr=<LIST>, -w <LIST> ]  --- host traddr (e.g. FC WWN's or IP
                                             source address)
   [  --host-iface=<LIST>, -f <LIST> ]   --- host transport interface (e.g. IP
                                             eth1, enp2s0)
   [  --hostnqn=<LIST>, -q <LIST> ]      --- user-defined hostnqn
   [  --hostid=<LIST>, -I <LIST> ]       --- user-defined hostid (if default


> I think a) is not really controversial just the question of the name. I don't mind going with |nbft|.
> 
> The bigger issue for me is b). This needs a strong argument why it should exists in the first place. This includes 
> *documentation* how to use it correctly (if we agree it should exist).


Yes, agreed. We talked about the fact that we don't have a man page yet.  I agree that documentation is needed.

/John

mwilck:
I've pondered this some more. The main difference between `connect` and `connect-all` is that the former takes an NQN argument while the latter does not. `connect-all` does take other options that limit the selection of discovery controllers to retrieve discovery information from, though. The NBFT could be seen as a source of "discovery" information.  Thus it would make sense to use `nvme connect-all --nbft` rather than `nvme connect --nbft`. `connect-all`, without arguments, would include the NBFT. For troubleshooting, we might also need `nvme connect-all --no-nbft`. Thus we would end up with the following new commands:

```
nvme nbft show 
nvme connect-all --nbft
nvme connect-all --no-nbft
```

Idea: we could also add `nvme discover --nbft`, which would print the subsystem information from the NBFT in `nvme discover` format. This would allow us to keep the analogy between `discover` and `connect-all`. We could add this in a later PR if we agree on it.

igaw:
- About the name of the plugin: I agree `firmware` is not a good name as it is more associated with actually firmware on a device. As I said `nbft` is okay just a bit hard to type and remember what it is.
- `nvme connect` and `nvme connect-all`  are only taking options, no command line arguments. In fact both accept the same options.
 Given the argument, if there is need for `--nbtf` and `--no-nbtf`, in this case I think both commands need to handle them. I am not sure, because no exact description what they are doing and how the interact with any of the existing options is explained. Maybe it's just me understanding things wrong. But I rather have these thing explicitly discussed and documented now because changing it later is more pain.

igaw:
So the argument is that we only need either `nvme connect --nbtf` or `nvme connect-all --nbtf` and your preference is that `nvme connect-all --nbtf` makes more sense because it connects all connection listed in the table. Alright, then do this and let's how this goes. I don't care too much anyway, except I don't want another `connect` command and that is what I tried to say from the beginning on.

hreinecke:
Actually, I would do it exactly the other way round.
'nvme connect-all' really should be doing what it advertised, namely connecting to all sources. And all sources should include nbft one.
'nvme connect', OTOH, requires the user to specify _where_ to connect to, so having an '--nbft' argument here is far more sensible.

nbft.c
```diff
@@ -0,0 +1,691 @@
...
+					errno = 0;
+					ret = nvmf_add_ctrl(h, c, &cfg);
+					if (ret == 0 && verbose >= 1)
+						fprintf(stderr,
+							"connect with host_traddr=\"%s\" failed, success after omitting host_traddr\n",
+							host_traddr);
+				}
+
+				if (ret)
+					fprintf(stderr, "no controller found\n");
```
mwilck, line 670: We should handle the case `ENVME_CONNECT_ALREADY` here. Dracut may need to do multiple connection attempts. "no controller found" is not a helpful error message if the subsystem at hand is already connected.

mwilck:
> Actually, I would do it exactly the other way round. 'nvme connect-all' really should be doing what it advertised, namely connecting to all sources.

This applies only to `connect-all` without arguments / options. `connect-all` supports plenty of options that would restrict the "all" to a subset of connections only offered by a certain discovery controller.

> And all sources should include nbft one. 'nvme connect', OTOH, requires the user to specify _where_ to connect to, so having an '--nbft' argument here is far more sensible.

Hm. AFAICS, the distinctive difference between `connect -t $transport -a $traddr -s $trsvcid` and `connect-all -t $transport -a $traddr -s $trsvcid` is that the former *requires* a `-n $subsysnqn` option, whereas the latter doesn't. Connecting to the NBFT would not require a subsysnqn option. Therefore I think it's more like `connect-all` (with options).

I don't want to turn this into a religious matter. If everyone else agrees `connect` is the way to go, fine with me. I just wouldn't want to have to implement both.

nbft.c
```diff
@@ -0,0 +1,691 @@
...
+}
+
+static struct json_object *nbft_to_json(struct nbft_info *nbft, bool show_subsys, bool show_hfi, bool show_discovery)
+{
+	struct json_object *nbft_json;
+
+	nbft_json = json_create_object();
+	if (!nbft_json)
+		return NULL;
+
```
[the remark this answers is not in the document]
igaw: In this case the schema and parser in libnvme needs to be updated accordingly.

nbft.c
igaw: There is `const char *nvme_status_to_string(int status, bool fabrics);` which translates the error codes into something more useful.

mwilck: We might actually consider not returning an error if we find that (all?) connections specified in the NBFT are already set up.

nbft.c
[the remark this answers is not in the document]
mwilck: I was referring to the format of the JSON only here, i.e. the API presented to consumers of the JSON by nvme-cli. Currently the code for converting the NBFT to JSON is only in nvme-cli, not in libnvme. So I'm unsure what changes in schema and parser you're referring to.

You are right that we should reflect possible changes in the NBFT format itself, too. So maybe we need to version fields, or a 2-digit version, one part indicating the version of the NBFT, and the other the version of the JSON output. When the NBFT format changes, the JSON will almost certainly change as well, but not vice versa.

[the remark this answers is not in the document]
igaw: I was referring the JSON config file. But this seems to be a different one. Still a schema for it wouldn't hurt, I suppose.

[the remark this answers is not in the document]
mwilck: > Still a schema for it wouldn't hurt, I suppose.

ack.

nbft.c
igaw: FWIW `nvme connect-all` ignores already connected targets. It only logs the warning.

stuarthayes:
I don't have any objections to "connect-all --nbft".

igaw:
As I understood @mwilck, he wants to keep the effort limited hence the wish only to do one connect command. As far I can tell the newly nbtf connect code will life in a separate function and `connect` or `connect-all` call it. Given that, it would be simple to add/remove the calls. This feels a bit like bike shedding :)

mwilck:
> This feels a bit like bike shedding 

I just want clarification how the command line API should look like. And I believe this decision should be made now.

To me, the important thing is that the final command line API provides commands to execute any of the following:
1. connect everything, including NBFT,
2. connect everything, except NBFT,
3. connect NBFT targets only.

mwilck:
I realize that the code in this PR doesn't build because it doesn't contain the changes to `libnvme.wrap` that would be necessary to make it build against the Timberland-enhanced libnvnme (basically, cc8f7443). I suggest that we add these changes back. We can (and will) still remove them once the Timberland libnvme changes have been merged upstream.

igaw:
FWIW, we can also progress quickly on the libnvme part and then look/update this PR.

johnmeneghini:
I have squashed and rebased both https://github.com/linux-nvme/libnvme/pull/572 and https://github.com/linux-nvme/nvme-cli/pull/1791 onto the HEAD of the master branch. After some work on the libnvme.spec file in my rh_linux_poc [here](https://github.com/timberland-sig/rh-linux-poc/pull/7/commits/f86161912acc126208d55197576c9f91110cc442) I was able to get my fedora RPMs to compile and install. However, testing these new rebased bits in my QEMU POC fails. After rebasing everything to the HEAD of the master branch in libnvme and nvme-cli, these changes no longer work with https://github.com/dracutdevs/dracut/pull/2184 and with https://github.com/timberland-sig/edk2/commit/9e63dc0cdfa052c3185d9451f2a4fb67886b959e. I am no longer able to boot my QEMU host with NVMe/TCP using these bits.

johnmeneghini:
Pushed a major update which should address all comments.

Please review.

johnmeneghini:
> I have squashed and rebased both [linux-nvme/libnvme#572](https://github.com/linux-nvme/libnvme/pull/572) and #1791 onto the HEAD of the master branch. After some work on the libnvme.spec file in my rh_linux_poc [here](https://github.com/timberland-sig/rh-linux-poc/pull/7/commits/f86161912acc126208d55197576c9f91110cc442) I was able to get my fedora RPMs to compile and install. However, testing these new rebased bits in my QEMU POC fails. After rebasing everything to the HEAD of the master branch in libnvme and nvme-cli, these changes no longer work with [dracutdevs/dracut#2184](https://github.com/dracutdevs/dracut/pull/2184) and with [timberland-sig/edk2@9e63dc0](https://github.com/timberland-sig/edk2/commit/9e63dc0cdfa052c3185d9451f2a4fb67886b959e). I am no longer able to boot my QEMU host with NVMe/TCP using these bits.

All issues have been resolved.  The newest version of the rh_linux_poc which includes all the needed prebuilt rpms and artifacts used to test this latest nvme-cli  changes can be seen [here](https://github.com/timberland-sig/rh-linux-poc/commit/63bbfd9a220acd0faf534edb67470435a93eee29).  See the dracut patches uses with this version of nvme-cli can be seen [here]( https://github.com/timberland-sig/dracut/pull/6).  The prebuilt Fedora RPMs are all publicly available  in my [COPR repository] (https://copr.fedorainfracloud.org/coprs/johnmeneghini/timberland-v14.2/)

fabrics.c
```diff
@@ -110,15 +112,6 @@ static const char *nvmf_config_file	= "Use specified JSON configuration file or
...
 
-struct tr_config {
-	const char *subsysnqn;
-	const char *transport;
-	const char *traddr;
-	const char *host_traddr;
-	const char *host_iface;
-	const char *trsvcid;
-};
-
```
igaw, line 121: The refactoring part should be in a preparation patch.

fabrics.c
```diff
@@ -744,6 +740,9 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 		OPT_INCR("verbose",      'v', &verbose,       "Increase logging verbosity"),
 		OPT_FLAG("dump-config",  'O', &dump_config,   "Dump configuration file to stdout"),
 		OPT_FLAG("force",          0, &force,         "Force persistent discovery controller creation"),
+		OPT_FLAG("nbft",           0, &nbft,          "Only look at NBFT tables"),
+		OPT_FLAG("no-nbft",        0, &nonbft,        "Do not look at NBFT tables"),
+		OPT_STRING("nbft-path",  0, "STR", &nbft_path, "user-defined path for NBFT tables"),
```
igaw, line 745: Looks about right. Personally I would have gone for `nbft-only`, `nbft-ignore` and `nbft-path`. But I can live with this. Anyway, it would be cool if you could also update the documentation (just the `*.txt` file, the rest will be automatically regenerated on release anyway)

@hreinecke are you also fine with this?

nbft.c
```diff
@@ -0,0 +1,208 @@
...
+					.trsvcid	= (*ss)->trsvcid,
+				};
+
+				/* Already connected ? */
+				cl = lookup_ctrl(r, &trcfg);
+				if (cl && nvme_ctrl_get_name(cl))
+					continue;
+
+				//if (hostkey)
+				//	nvme_host_set_dhchap_key(h, hostkey);
```
igaw, line 147: I suppose you have disable this part, because it didn't work during testing and will be enabled later?

nbft.c
```diff
@@ -0,0 +1,208 @@
...
+				c = nvme_create_ctrl(r, (*ss)->subsys_nqn, (*ss)->transport,
+						     (*ss)->traddr, host_traddr, NULL,
+						     (*ss)->trsvcid);
+				if (!c) {
+					errno = ENOMEM;
+					goto out_free;
+				}
+
+				//if (ctrlkey)
+				//	nvme_ctrl_set_dhchap_key(c, ctrlkey);
```
igaw, line 158: same as above?

nvme-builtin.h
```diff
@@ -105,7 +105,7 @@ COMMAND_LIST(
 	ENTRY("virt-mgmt", "Manage Flexible Resources between Primary and Secondary Controller ", virtual_mgmt)
 	ENTRY("rpmb", "Replay Protection Memory Block commands", rpmb_cmd)
 	ENTRY("lockdown", "Submit a Lockdown command,return result", lockdown_cmd)
-	ENTRY("dim", "Send Discovery Information Management command to a Discovery Controller", dim_cmd) \
+	ENTRY("dim", "Send Discovery Information Management command to a Discovery Controller", dim_cmd)
```
igaw, line 108: unrelated change

plugins/meson.build
```diff
@@ -12,6 +12,7 @@ if json_c_dep.found()
     'plugins/intel/intel-nvme.c',
     'plugins/memblaze/memblaze-nvme.c',
     'plugins/micron/micron-nvme.c',
+    'plugins/nbft/nbft-plugin.c',
```
igaw, line 15: Could you split the plugin change into a separate commit?

igaw (changes requested):
Only minor change requests/question
- just splitting the patch slightly
- questions on some comments in the connect algorithm code.
- updating of the man pages.

Looks in great shape

nbft.c
johnmeneghini: There is no support for NVMe-oF authentication or security of any kind with NBFT.  It will take a new TP to update the spec before it's supported.  I'm going to remove this commented out code.

nvme-builtin.h
johnmeneghini: checkpatch.pl did this....

nbft.c
johnmeneghini: Removing this dead code.

nbft.c
johnmeneghini: @mwilck are we good here?  Do you still want to make this change?

nvme-builtin.h
johnmeneghini: I just ran checkpatch again and it removed the second continueation as well.

plugins/meson.build
johnmeneghini: Done

fabrics.c
johnmeneghini: I've broken this change out into a separate commit.

fabrics.c
johnmeneghini: Resolving this thread.

nbft.c
johnmeneghini: resolving this thread

nbft.c
johnmeneghini: Resolving this thread

nvme-builtin.h
johnmeneghini: checkpatch.pl was complaining about these extra `\` so I removed them.

johnmeneghini:
I've addressed all of @igaw s comments.  There are now three commits. Documentation has been added, and `checkpatch.pl` has been run over everything.  Note that I opened https://github.com/linux-nvme/nvme-cli/issues/1905 to demonstrait that I have introduced no regression in `fabric.c`.  (The file needs some attention).  Everything here runs and passes the nvme/tcp boot sniff test with the [rh_linux_poc](https://github.com/timberland-sig/rh-linux-poc). The prebuilt Fedora RPMs are all publicly available in my COPR repository [here](https://copr.fedorainfracloud.org/coprs/johnmeneghini/timberland-v14.2/).

```
[root@host-vm ~]# nvme connect-all --help
Usage: nvme connect-all <device> [OPTIONS]

Discover NVMeoF subsystems and connect to them

Options:
.
.
.
  [  --force ]                          --- Force persistent discovery
                                            controller creation
  [  --nbft ]                           --- Only look at NBFT tables
  [  --no-nbft ]                        --- Do not look at NBFT tables
  [  --nbft-path=<STR> ]                --- user-defined path for NBFT tables

[root@host-vm ~]# nvme nbft --help
nvme-2.4
usage: nvme nbft <command> [<device>] [<args>]

The '<device>' may be either an NVMe character device (ex: /dev/nvme0), an
nvme block device (ex: /dev/nvme0n1), or a mctp address in the form
mctp:<net>,<eid>[:ctrl-id]

ACPI NBFT table extensions

The following are all implemented sub-commands:
  show            Show contents of ACPI NBFT tables
  version         Shows the program version
  help            Display this help

See 'nvme nbft help <command>' for more information on a specific command

[root@host-vm ~]# nvme nbft show
/sys/firmware/acpi/tables/NBFT:

NBFT Subsystems:

Idx|NQN                                                                 |Trsp|Address       |SvcId|HFIs
---+--------------------------------------------------------------------+----+--------------+-----+----
1  |nqn.2014-08.org.nvmexpress:uuid:0c468c4d-a385-47e0-8299-6e95051277db|tcp |192.168.101.20|4420 |1   

NBFT HFIs:

Idx|Trsp|PCI Addr  |MAC Addr         |DHCP|IP Addr       |Mask|Gateway |DNS     
---+----+----------+-----------------+----+--------------+----+--------+--------
1  |tcp |0:0:4.0   |ea:eb:d3:58:89:58|no  |192.168.101.30|24  |0.0.0.0 |0.0.0.0 
```

nbft.c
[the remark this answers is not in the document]
johnmeneghini: I think this is the last remaining open issues.

igaw:
Thanks a lot! 

We have some time to stabilize or change things without any concerns of breaking stuff (inside this project, obviously only), since we still have time until the next release. But I suspect you really want to keep the command line options now as it is :)

PR #1792, opened 2023-02-03
ocp: Bug Fix. OCP UUID value initialization changed to use network by…

lgdacunh:
…te order.


igaw:
Thanks! Directly applied after updating the commit message slightly.

PR #1793, opened 2023-02-03
Solidigm: Added automated uuid_index resolution when retrieving laten…

lgdacunh:
…cy-tracking log, garbage collection log, and SMART log.


igaw:
Thanks. I've updated the commit message slightly (the commit subject should be a bit more terse (under 80 chars)) and applied it directly.

PR #1796, opened 2023-02-04
fix a mistake about json_nvme_zns_report_zones

sunxing0210:


igaw:
Thanks. I've applied the patch directly on master after tweaking the commit message slightly.

PR #1805, opened 2023-02-09
docs: Fixed typos and add missing option

martin-belanger:
For `nvme-discover` the options `--hdr-digest` and `--data-digest` were documented with an underscore instead of a hyphen (i.e. `--hdr_digest` and `--data_digest`).

For `nvme-connect-all` the option `--device` was missing.

I also made a few other changes for consistency between `nvme-discover` and `nvme-connect-all`.


igaw:
Thanks!

PR #1811, opened 2023-02-10
build: Add muon build target

igaw:
Build nvme-cli using muon.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -244,3 +244,47 @@ jobs:
         with:
           name: cross s390x build log
           path: build/meson-logs/meson-log.txt
+
+  build-minimal:
+    name: muon minimal
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: build muon and samurai build-tool
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/muon.yml
```diff
@@ -0,0 +1,43 @@
...
+
+          build/muon setup -Dlibcurl=disabled -Dlibarchive=disabled -Ddocs=disabled -Dsamurai=disabled build
+          build/samu -C build
+          build/muon -C build test
+      - name: fetch libnvme
+        run: |
+          cd subprojects
+          git clone https://github.com/linux-nvme/libnvme.git
+          # libnvme_ref=$(sed -n "s/revision = \([0-9a-z]\+\)/\1/p" libnvme.wrap)
+          # git -C libnvme checkout $libnvme_ref
```
igaw, line 36: Need to figure out why this part doesn't work. The version check for libnvme fails in the main meson.build file.

igaw: FTR, the problem was that muon somehow had troubles with a ref which also got a tag pointing to. It works fine between the releases. Ah, something to figure out later :)

Issue #1806, opened 2023-02-09
udev rules could be simplified


martin-belanger:
I submit this issue for your review. Please let me know if you agree with the following analysis and I will submit a pull request if you do.

The udev rules specified in [70-nvmf-autoconnect.rules.in](https://github.com/linux-nvme/nvme-cli/blob/master/nvmf-autoconnect/udev-rules/70-nvmf-autoconnect.rules.in) are used to start a oneshot service defined by [nvmf-connect%40.service.in](https://github.com/linux-nvme/nvme-cli/blob/master/nvmf-autoconnect/systemd/nvmf-connect%40.service.in). This oneshot service simply invokes the command `nvme connect-all`.

The command `nvme connect-all` supports a `--device` option. When that option is specified, it is not necessary to provide all the other options like `--transport`, `--traddr`, etc. In fact, when `--device` is specified the other options are ignored.

The udev rules are currently passing not only `--device` but also all the other  options to `nvme connect-all`, which is redundant.  

I propose to only use the `--device` option in the udev rules. Let me know what you think.

igaw:
One thing which needs to be double checked is if  `connect` and `connect-all` actually do the same thing. If I got this right then `connect-all` will try to use the arguments from the command line and does a full discovery. Anything it finds it will try to connect. Where as  `connect` only does only connect to one controller specified from the command line.  Obviously, if the arguments handed in just a controller than it should be equivalent.

martin-belanger:
You are correct, `connect-all` connects to a Discovery Controller (DC), performs a discovery command to retrieve the Discovery Log Pages (DLP), and then connects to every I/O controller listed in the DLP. The `connect-all` command can be invoked by specifying an existing persistent DC connection, in which case one need only provide the `--device` option. And in this case, the options `--transport`, `--traddr`, `--trsvcid` are redundant. 

I need to take a closer look at the code because I think that `--host-traddr` and `--host-iface` are needed because they get applied to the I/O controller connections. However, these two options could be retrieved from the existing DC connection, which means that when invoking `connect-all` for an existing persistent DC connection all we would need to provide is the `--device` option, which would simplify usage quite a bit.

martin-belanger:
For the `connect-all` command, I can confirm that when option `--device` is specified all other options are ignored ([fabrics.c#L888](https://github.com/linux-nvme/nvme-cli/blob/master/fabrics.c#L888)) and we simply retrieve the DLP from the existing DC connection.

[...]

PR #1812, opened 2023-02-10
fabrics: Ensure host-traddr/iface consistency for "connect-all"

martin-belanger:
Fix for: https://github.com/linux-nvme/nvme-cli/issues/1806

When using `nvme connect-all --device` without specifying the options `--host-traddr` and `--host-iface`, the children connections may not have the same `host-traddr` / `host-iface` as the connection to the discovery controller (DC).

This is specific to the `--device` option where one reuses a persistent connection to a discovery controller to update the discovery log pages (DLP) and connect to new controllers listed in the new DLP. If `--device` is used w/o also specifying `--host-traddr` and `--host-iface`, then the new connections may not use the same attributes that were used for the DC connection.

This patch checks if `--host-traddr` and `--host-iface` have been provided, and if not, it will use the DC's `--host-traddr` and `--host-iface` for the new connections made as a result of changes to the DLP.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -908,6 +908,20 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 				 * on exit.
 				 */
 				persistent = true;
+				/*
+				 * When --host-traddr/--host-iface are not specified on the
+				 * command line, use the discovery controller's (c) host-
+				 * traddr/host-iface for the connections to controllers
+				 * returned in the Discovery Log Pages. This is essential
+				 * when invoking "connect-all" with --device to reuse an
+				 * existing persistent discovery controller (as is done
+				 * for the udev rules). This ensures that host-traddr/
+				 * host-iface are consistent with the discovery controller (c).
+				 */
+				if (!cfg.host_traddr)
+					cfg.host_traddr = (char *)nvme_ctrl_get_host_traddr(c);
+				if (!cfg.host_iface)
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -937,6 +937,20 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
...
+	 * the connections to controllers returned in the Discovery Log Pages.
+	 * This is essential when invoking "connect-all" with --device to
+	 * reuse a persistent discovery controller (as is done for the udev
+	 * rules). In this case, we want host-traddr/host-iface to be
+	 * consistent with the discovery controller (c).
+	 */
+	if (!cfg.host_traddr)
+		cfg.host_traddr = (char *)nvme_ctrl_get_host_traddr(c);
+	if (!cfg.host_iface)
+		cfg.host_iface = (char *)nvme_ctrl_get_host_iface(c);
```
igaw, line 952: I think this should be inside the `if (device & !force)` block. The the `force` option should be able to overwrite anything.

igaw (changes requested):

fabrics.c
martin-belanger: Good point. Let me rework it.

martin-belanger: @igaw - Requested changes applied.

igaw (approved):

PR #1817, opened 2023-02-15
completion: fix typo, add missing args

sc108-lee:
write-zeros -> write-zeroes
--nmic short name is -m
add missing space
add missing args

Reported-by: Youngjin Jung <yj4369.jung@samsung.com>


igaw:
Thanks a lot! I keep forgetting to update/check the 'completions'.

PR #1820, opened 2023-02-15
nvme-print-json: Add Generic device name to list ouptut

igaw:
The Generic device name is missing in the 'list' output.

see https://lore.kernel.org/linux-nvme/7f958c6d-d6e6-2346-a680-875c7fafe7e5@grimberg.me/


igaw:
I could just have picked up this one here: https://lore.kernel.org/linux-nvme/20230215090020.620994-1-sagi@grimberg.me/

PR #1824, opened 2023-02-16
OCP/Solidigm: Reorg OCP smart extended log and integrate into solidigm plugin

kdedow:
**Summary:**
In support of meeting OCP 2.0 spec. vendor plugin requirements, I did the following:

1. Moved OCP SMART extended log into its own  source file, so that it could be referenced by both OCP and Solidigm plugins
2. Added UUID index logic to SMART extended log. Code now attempts to obtain non-zero UUID index associated with OCP GUID. If this fails, logic still defaults to zero and proceeds with the log pull.
3. Referenced the OCP SMART extended log from OCP plugin in Solidigm plugin (in order to share implementation)

I will have a few more reviews implementing more OCP plugin features after this pull request has been reviewed, updated (based on review comments), and merged. Thanks!

**Note:** I am on Leo da Cunha's team at Solidigm and have been working with him


igaw (changes requested):
Thanks for cleaning up the code base. This is highly appreciated!

There are several different sort of changes in big commit. Please split them up into several  commits and add commit message which explains why/what a commit does (commits should be self contained and not rely on external services)

For example you fixup many whitespace damages, e.g. spaces to tabs conversion. This should be one change. The commit message would say, just whitespace changes. And I can verify by just comparing the object files after compile step. If this is true, no changes in the binary should happen. Easy review :)

Separating these steps will make your life way simpler in the long run and makes reviewing actually possible. Currently, I can't really see anything.

igaw:
@arthurshau Please have a look at the OCP changes. Thanks!

kdedow:
@igaw - Makes sense. Would this make sense as how to split this review?

1. Whitespace (and other cosmetic changes based on code linter)
2. Minor bug fixes with SMART extended log output
3. Splitting SMART extended log into it's own header/source
4. Solidigm plugin SMART extended integration

I will proceed unless I hear any other suggestion. Thanks!

igaw:
That sounds like an excellent plan to me! Go for it! :+1:

kdedow:
I am going to close this pull request in favor the suggestion provided above. Should see a sequence of requests over the coming days/weeks. Thanks all.

Issue #1803, opened 2023-02-09
unit test for argparser


igaw:
I think it would good idea to also add a unit test for the argparser, e.g. #1801 should be caught by it.

Issue #1801, opened 2023-02-09
invalid argument parse from #1790


sc108-lee:
[root@localhost .build]# ./nvme read /dev/nvme0n1 --data-size=4096
Expected long suffixed integer argument for 'data-size' but got '4096'!

igaw:
`suffix_binary_suffix()` function is used to parse the `--data-size` command line argument. The commit 1e5abd545622 ("util: Update suffix_binary_parse API") introduce this regression. Just a sec.

igaw:
Thanks for the report!

PR #1828, opened 2023-02-16
unit: Add test-argconfig-parse

ikegami-t:
Fixes: #1803
Sorry for the issue #1801 as reviewed #1790 but missed it to find.


igaw:
Nice! Thank a lot!

PR #1829, opened 2023-02-17
OCP: Fix whitespace issues based on code linter results in OCP plugin

kdedow:
As per request, first code review that should only have changed whitespace.

I compared the generated ocp object file via 'cmp' between master and this branch and the object files looked to be the same.


igaw:
No, this is not how to do it. You can't just run a tool and not verify the result. The formatting is really ugly. I've manually updated you patch and I am going to a PR because I can't modify your patch here.

PR #1830, opened 2023-02-17
plugins/ocp: Fix whitespace and style issues

igaw:
As reported by a code linter results in OCP plugin

[dwagner:
 - lot's of manual cleanup
 - fixed non white space changes reported by checktool
 - verified result with 'git diff --word-diff' ]

### DIFF
--- a/plugins/ocp/ocp-nvme.h
+++ b/plugins/ocp/ocp-nvme.h
@@ -3,7 +3,7 @@
  *
  * Authors: Arthur Shau <arthurshau@fb.com>,
  *          Wei Zhang <wzhang@fb.com>,
- *   	    Venkat Ramesh <venkatraghavan@fb.com>
+ *          Venkat Ramesh <venkatraghavan@fb.com>
  */
 #undef CMD_INC_FILE
 #define CMD_INC_FILE plugins/ocp/ocp-nvme
@@ -16,10 +16,8 @@
 PLUGIN(NAME("ocp", "OCP cloud SSD extensions", NVME_VERSION),
 	COMMAND_LIST(
 		ENTRY("smart-add-log", "Retrieve extended SMART Information", ocp_smart_add_log)
-		ENTRY("latency-monitor-log", "Get Latency Monitor Log Page",
-		      ocp_latency_monitor_log)
-		ENTRY("clear-fw-activate-history", "Clear firmware update history log",
-		      clear_fw_update_history)
+		ENTRY("latency-monitor-log", "Get Latency Monitor Log Page", ocp_latency_monitor_log)
+		ENTRY("clear-fw-activate-history", "Clear firmware update history log", clear_fw_update_history)
 	)
 );
 
--- a/plugins/ocp/ocp-clear-fw-update-history.h
+++ b/plugins/ocp/ocp-clear-fw-update-history.h
@@ -6,4 +6,5 @@
  *          leonardo.da.cunha@solidigm.com
  */
 
-int ocp_clear_fw_update_history(int argc, char **argv, struct command *cmd, struct plugin *plugin);
+int ocp_clear_fw_update_history(int argc, char **argv,
+                                struct command *cmd, struct plugin *plugin);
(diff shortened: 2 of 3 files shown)


igaw:
@arthurshau there are some non whitespace change in this patch. I verified it by looking at the chagnes with 'git diff --word-diff' which will show only those non whitespace changes. I am pretty sure I didn't break anything. Let me know if this is okay for you.

kdedow:
@igaw - Couple questions based on other review comment.

- What editor/IDE would you recommend for viewing files visually? I did neglect to look at output via github, but it looked fine in visual studio code.
- What did you mean by "verify the result"? I'm assuming you just mean the source code changes, because the feature (ocp smart-add-log) output is the same.

Thanks for jumping on this. Just hoping for some clarification so I can avoid in the future.

kdedow:
My default tab size in VS code was 4, as opposed to 8. Which  is what I think the problem was.

plugins/ocp/ocp-nvme.h
```diff
@@ -17,9 +17,9 @@ PLUGIN(NAME("ocp", "OCP cloud SSD extensions", NVME_VERSION),
 	COMMAND_LIST(
 		ENTRY("smart-add-log", "Retrieve extended SMART Information", ocp_smart_add_log)
 		ENTRY("latency-monitor-log", "Get Latency Monitor Log Page",
-		      ocp_latency_monitor_log)
```
ikegami-t, line 20: Seems the original code indentation correct as described by Linux kernel coding style as below.
https://docs.kernel.org/process/coding-style.html#breaking-long-lines-and-strings
```
2) Breaking long lines and strings
...
Descendants are always substantially shorter than the parent and are placed substantially to the right. A very commonly used style is to align descendants to a function open parenthesis.
```

plugins/ocp/ocp-nvme.h
```diff
@@ -17,9 +17,9 @@ PLUGIN(NAME("ocp", "OCP cloud SSD extensions", NVME_VERSION),
...
+			ocp_latency_monitor_log)
 		ENTRY("clear-fw-activate-history", "Clear firmware update history log",
-		      clear_fw_update_history)
```
ikegami-t, line 22: This also same.

kdedow: @ikegami-t - The reason this is occurring is because tabs are exclusively being used. So function signatures with new lines are tabbing an extra level. Mixing tabs with whitespace seemed odd, but I'm not familiar with Linux code-style.

@igaw - Do you want me to change this, where it occurs? I'm happy to do so.

arthurshau:
Looks good to me. Will defer to @igaw about using tabs + spaces in function signature alignment.

igaw:
Sorry for the delay. Lot of high prio interrupts last few days.

>     * What editor/IDE would you recommend for viewing files visually? I did neglect to look at output via github, but it looked fine in visual studio code.

That is a difficult question. I think any editor works as long you have the defaults (tab == 8 spaces) and the
(horrible) tab/char indention style. Personally I am using emacs for a long time and recently started to use doomemacs which is awesome. It gets all the alignment etc correct without having to tinker with ~/.emacs. But have to learn mode-evil...

>     * What did you mean by "verify the result"? I'm assuming you just mean the source code changes, because the feature (ocp smart-add-log) output is the same.

I meant look at the result. Obviously, if your editor has tab == 4 spaces it looks correct.

> Thanks for jumping on this. Just hoping for some clarification so I can avoid in the future.

No worries.  I was just slightly annoyed by the submission hence my rampage. But with some days distance it's not that bad.

To answer @arthurshau question: We follow the kernel style. I hear you, though I think we should stay consistent. It helps in the long run even though we don't agree with Linus' taste.

plugins/ocp/ocp-nvme.h
igaw: Let's got with the same style as we have in `nvme-builtin.h`, so one command per line even though the lines get a bit longer.

plugins/ocp/ocp-nvme.h
igaw: I am pushing my changes, then you can do a followup if there are still obvious whitespaces damages

kdedow: Sounds good. I will make a note to align to @ikegami-t's comment on subsequent pull requests

Issue #1827, opened 2023-02-16
tnvmcap and unvmcap in id-ctrl wrong data type and inconsistent formatting?


darren-c:
I noticed that in nvme-cli v2.3, the tnvmcap & unvmcap value in the id-ctrl query now includes commas.

nvme-cli 2.2.1
> tnvmcap   : 3200631791616

nvme-cli 2.3
> tnvmcap   : 3,200,631,791,616

I also noticed that if you use output-format=json, these values are set as strings (shouldn't they be numbers?)

>   "tnvmcap":"3,200,631,791,616",
     "unvmcap":"0",

I'm wondering if this is the intended behavior.

Of course it's mostly visual, however I had a Python script that was using these values and casting int() on them. However in nvme-cli v2.3, this causes an exception due to the commas. It's not a big deal to workaround it but thought it would make sense for nvme-cli to treat them as numbers. 

ie. in JSON:
>   {"tnvmcap": 3200631791616,
      "unvmcap": 0}

would make more sense to me

Thanks.

ikegami-t:
Yes the commas was added by the commit: 2a8bc94 ("nvme-print: Display smart log data units read and written SI value
"). About the json output changes I am not sure if we can fix so let me consult with @igaw . Probably it was changed for the review comment but seems to keep the previous behavior for JSON or change the localization as not available if needed to use. Do you have any comment or advice?
https://github.com/linux-nvme/nvme-cli/pull/1749#pullrequestreview-1234770082
> - drop uint128_t_to_string_tg, always use localization if available.

igaw:
I think adding the localization formatting to the JSON output was wrong (sorry.  We should revert it to plain numbers.

igaw:
This made me thinking we might want to write a schema. Though I don't know how we could easily test it. Without a regular test (CI) the schema is pretty worthless , IMO.

PR #1831, opened 2023-02-17
util: Revert JSON output to plain numbers

ikegami-t:
Add uint128_t_to_string_with_sep() to unset the thousands separator. Fixes: 2a8bc94 ("nvme-print: Display smart log data units read and written SI value") Fixes: #1827

### DIFF
--- a/util/json.c
+++ b/util/json.c
@@ -35,10 +35,24 @@ struct json_object *util_json_object_new_uint64(uint64_t i)
 
 }
 
+static int util_json_object_string_to_number(struct json_object *jso,
+					     struct printbuf *pb, int level,
+					     int flags)
+{
+	ssize_t len = json_object_get_string_len(jso);
+
+	printbuf_memappend(pb, json_object_get_string(jso), len);
+
+	return 0;
+}
+
 struct json_object *util_json_object_new_uint128(nvme_uint128_t  val)
 {
 	struct json_object *obj;
+
 	obj = json_object_new_string(uint128_t_to_string(val));
+	json_object_set_serializer(obj, util_json_object_string_to_number, NULL, NULL);
+
 	return obj;
 }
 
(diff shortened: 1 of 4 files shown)


util/json.c
```diff
@@ -38,7 +38,7 @@ struct json_object *util_json_object_new_uint64(uint64_t i)
 struct json_object *util_json_object_new_uint128(nvme_uint128_t  val)
 {
 	struct json_object *obj;
-	obj = json_object_new_string(uint128_t_to_string(val));
+	obj = json_object_new_string(uint128_t_to_string_with_sep(val, NULL));
```
igaw, line 41: I think we need to introduce a new json object for the type uint128 and write our own serializer (see  `json_object_set_serializer()`, https://github.com/json-c/json-c/blob/master/tests/test_set_serializer.c) , because this will still generates a string and not a number.

igaw (changes requested):

util/json.c
igaw: `util_json_object_new_double()` and `util_json_object_new_uint64() ` are fallbacks implementation which cheat a bit and generate just a string. But it supposed to generate actual numbers. 

```c
					json_object_add_value_uint64(jns, "UsedBytes", nuse);
					json_object_add_value_uint64(jns, "MaximumLBA", nvme_ns_get_lba_count(n));
					json_object_add_value_uint64(jns, "PhysicalSize", nsze);
```

gives

```json
      "UsedBytes":512110190592,
      "MaximumLBA":1000215216,
      "PhysicalSize":512110190592,
```

with a current version of json-c. The type is supposed to be a number and not string (in JSON spec terms). The reason why it works for the bug reported in #1827 is that he uses Python which is able to switch between types very easy. It is a bug since introduction of the uint128 type.

ikegami-t: Thanks for your explanations. I could understand and just pushed the patch fixed.

igaw (approved):

igaw:
I've made the localization explicit. This makes it clear in the code what happens when you call the stringify function.  Obviously, had to rename all the calls in nvme-print.c. So it's an opt-in for the plugins too.

igaw:
Thanks!

ikegami-t:
Thank you so much for your reviewing and making!
Let me comment as checked the changes but just a trivial comment: The commit message `l10` should be `l10n`.
> [dwagner: introduce an explicit l10 function]

PR #1832, opened 2023-02-19
log: high latency timestamp issue

djukyo:
1. The old nvme-cli set-feature uses 32bit to represent timestamp, but the spec definition is 48bit.
2. memblaze made a slight correction to allow for overflow.
3. Now this correction is no longer needed.


djukyo:
@igaw could you help review this patch?

igaw:
I will review it, no worry, I am just bogged down with other stuff.

PR #1833, opened 2023-02-22
Tp4146

sc108-lee:


sc108-lee:
it looks muon does not recognize commit on PR
https://github.com/linux-nvme/libnvme/pull/582/commits/c32d360954bd655f2de19c8a08297b41ac56ce3c

sc108-lee:
This commit will be modified after libnvme TP4146 merged

igaw:
> it looks muon does not recognize commit on PR [linux-nvme/libnvme@c32d360](https://github.com/linux-nvme/libnvme/commit/c32d360954bd655f2de19c8a08297b41ac56ce3c)

The muon build is trying to fetch it from the orig repo and I think anything which is just a PR wont be fetched correctly. The CI instruction are a bit too simple. On the other hand, it will prevent to merge something which isn't in libnvme. So it's a feature :)

igaw:
I've merged PR #1894. It looks like this one doesn't have new changes ontop of #1894.

PR #1834, opened 2023-02-24
nvme: Continue get feature on invalid NS error to get multiple featur…

ikegami-t:
…e ids

Some features returned invalid NS if NSID all 0xffffffff default specified.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4549,6 +4549,9 @@ static int filter_out_flags(int status)
 static void get_feature_id_print(struct feat_cfg cfg, int err, __u32 result,
 				 void *buf)
 {
+	int status = filter_out_flags(err);
+	enum nvme_status_type type = NVME_STATUS_TYPE_NVME;
+
 	if (!err) {
 		if (!cfg.raw_binary || !buf) {
 			printf("get-feature:%#0*x (%s), %s value:%#0*x\n",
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -4519,10 +4521,10 @@ static int get_feature_id(struct nvme_dev *dev, struct feat_cfg *cfg,
 		cfg->data_len = 0;
 
 	if (cfg->data_len) {
-		if (posix_memalign(buf, getpagesize(), cfg->data_len)) {
+		size = ((cfg->data_len - 1) / getpagesize() + 1) * getpagesize();
```
igaw, line 4524: I think this line deserves a comment. Something like 'rounding up size to page size'.

igaw:
I think the first two commits could be merged, because I think the second commit removes the code from the first commit.

ikegami-t:
Fixed the commits as mentioned. Thank you.

nvme.c
```diff
@@ -4626,15 +4630,21 @@ static int get_feature_ids(struct nvme_dev *dev, struct feat_cfg cfg)
 	for (i = cfg.feature_id; i < feat_max; i++, feat_num++) {
 		cfg.feature_id = i;
 		err = get_feature_id_changed(dev, cfg, changed);
-		if (err && !nvme_status_equals(filter_out_flags(err),
-					       NVME_STATUS_TYPE_NVME,
-					       NVME_SC_INVALID_FIELD))
+		if (!err)
+			continue;
```
igaw, line 4637: `get_features_id_changed()` calls `get_feature_id()` which returns the the status from `nvme_get_features()`. This would be 0/1 and errno. 

(need to look a bit closer tomorrow)

ikegami-t: The status is checked by `get_feature_id_print()` but seems to be `0`: success, `> 0`:  error status and `-errno`. Also it is checked as same by other commands.

PR #1835, opened 2023-02-26
Changed 'Kelvin' to 'kelvin', as recommended by the BIPM.

tfpf:
The BIPM recommends that SI units (even those named after scientists) start with a lowercase letter when spelled out.

**References**
* https://www.bipm.org/en/si-base-units/kelvin
* https://www.nist.gov/pml/owm/writing-si-metric-system-units
* https://english.stackexchange.com/questions/173906

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -10576,7 +10576,7 @@ static int wdc_vs_temperature_stats(int argc, char **argv,
 	if (ret != 0)
 		goto out;
 
-   	/* convert from kelvin to degrees Celsius */
+	/* convert from kelvins to degrees Celsius */
 	temperature = ((smart_log.temperature[1] << 8) | smart_log.temperature[0]) - 273;
 
 	/* retrieve HCTM Thermal Management Temperatures */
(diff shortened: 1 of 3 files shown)


plugins/wdc/wdc-nvme.c
```diff
@@ -10576,7 +10576,7 @@ static int wdc_vs_temperature_stats(int argc, char **argv,
 	if (ret != 0)
 		goto out;
 
-   	/* convert from Kelvin to degrees Celsius */
+   	/* convert from kelvin to degrees Celsius */
```
igaw, line 10579: What about celsius? Shouldn't this also be lower cased?

tfpf: This is an exception.

> … the correct spelling of the name of the unit with the symbol °C is “degree Celsius” (the unit degree begins with a lower-case d and the modifier Celsius begins with an upper-case C because it is a proper name).
>
> —Section 5.3, _The International System of Units_, 9th Edition (2019)

ikegami-t: By the way NVMe specification revision 1.4c described as "in degrees Kelvin" for example in MNTMT. But it was changed as "in Kelvins" by revision 2.0c. Is this related to this issue? (Seems the specification also should be fixed to be lower case.)

ikegami-t: Thanks for the confirmation. Yes I can agree with the change to use 'K' instead of 'Kelvin'. Sorry a space is not used before '°C' so seems to need to fix it also.

igaw:
Thanks!

PR #1837, opened 2023-02-28
plugins/ocp: SMART cloud log bug fixes.

kdedow:
- Fix incorrect SMART cloud log description
- Fix SMART cloud log end-to-end correction offsets. See offsets from 2.0 spec below:

https://www.opencompute.org/documents/datacenter-nvme-ssd-specification-v2-0r21-pdf

image


kdedow:
In case anyone is wondering, the force pushes were only because my local git config was not mapping/linking nicely with my github account. It should be now.

igaw:
Looks good to me.

@arthurshau

arthurshau:
Looks good to me.

I am also aware of incorrect offsets in the latency monitor log command that hasn't been updated to match the 2.0 spec. I have plans to update them later this month but feel free to change them yourself as well if you'd like.

PR #1838, opened 2023-03-01
nvme: Change feature default NSID to 0 from all namespaces ID FFFFFFFFh

ikegami-t:
Since some features requires the NSID as 0 but not able set previously.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -408,7 +408,7 @@ static int get_dev(struct nvme_dev **dev, int argc, char **argv, int flags)
 
 int parse_and_open(struct nvme_dev **dev, int argc, char **argv,
 		   const char *desc,
-		   const struct argconfig_commandline_options *opts)
+		   struct argconfig_commandline_options *opts)
 {
 	int ret;
 
@@ -4700,7 +4700,7 @@ static int get_feature(int argc, char **argv, struct command *cmd,
 	if (err)
 		goto ret;
 
-	if (!cfg.namespace_id) {
+	if (!opts[1].seen) {
 		err = nvme_get_nsid(dev_fd(dev), &cfg.namespace_id);
 		if (err < 0) {
 			if (errno != ENOTTY) {
@@ -5866,14 +5866,13 @@ static int set_feature(int argc, char **argv, struct command *cmd, struct plugin
 	if (err)
 		goto ret;
 
-	if (!cfg.namespace_id) {
+	if (!opts[0].seen) {
 		err = nvme_get_nsid(dev_fd(dev), &cfg.namespace_id);
 		if (err < 0) {
 			if (errno != ENOTTY) {
 				fprintf(stderr, "get-namespace-id: %s\n", nvme_strerror(errno));
 				goto close_dev;
 			}
-
 			cfg.namespace_id = NVME_NSID_ALL;
 		}
 	}
--- a/nvme.h
+++ b/nvme.h
@@ -96,7 +96,7 @@ void register_extension(struct plugin *plugin);
  * parse_and_open - parses arguments and opens the NVMe device, populating @dev
  */
 int parse_and_open(struct nvme_dev **dev, int argc, char **argv, const char *desc,
-	const struct argconfig_commandline_options *clo);
+	struct argconfig_commandline_options *clo);
 
 void dev_close(struct nvme_dev *dev);
 
--- a/plugins/intel/intel-nvme.c
+++ b/plugins/intel/intel-nvme.c
@@ -1558,7 +1558,7 @@ static int enable_lat_stats_tracking(int argc, char **argv,
 		.disable = false,
 	};
 
-	const struct argconfig_commandline_options command_line_options[] = {
+	struct argconfig_commandline_options command_line_options[] = {
 		{"enable", 'e', "", CFG_FLAG, &cfg.enable, no_argument, enable_desc},
 		{"disable", 'd', "", CFG_FLAG, &cfg.disable, no_argument, disable_desc},
 		{NULL}
--- a/plugins/memblaze/memblaze-nvme.c
+++ b/plugins/memblaze/memblaze-nvme.c
@@ -1140,7 +1140,7 @@ static int mb_set_lat_stats(int argc, char **argv,
 		.disable = false,
 	};
 
-	const struct argconfig_commandline_options command_line_options[] = {
+	struct argconfig_commandline_options command_line_options[] = {
 		{"enable", 'e', "", CFG_FLAG, &cfg.enable, no_argument, enable_desc},
 		{"disable", 'd', "", CFG_FLAG, &cfg.disable, no_argument, disable_desc},
 		{NULL}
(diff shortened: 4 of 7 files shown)


igaw:
> Since some features requires the NSID as 0 but not able set previously.

What does this mean? So some features the `NSID` to be `0xfffffff` and others `0`?

igaw:
What's the point exchanging one hard coded value with another if both a valid in for certain features. Why not allowing to user to set it from the command line?

ikegami-t:
Since the NSID parameter is initialized as zero then it is set to 0xffffffff if the character parameter is the nvme character device /dev/nvme0. Also user can set the NSID as zero but reset to 0xffffffff as same.

igaw:
Sorry, I don't understand what you are saying. 

If the user is able to overwrite the nsid, why do we need to change the existing defaults **IF** some of the features expect 0 and others 0xfffffff. If this is the case, we can't set any defaults which always work. My point is that changing the current settings is not fixing anything.

ikegami-t:
> Sorry, I don't understand what you are saying. 
> 
> 
> 
> If the user is able to overwrite the nsid, why do we need to change the existing defaults **IF** some of the features expect 0 and others 0xfffffff. If this is the case, we can't set any defaults which always work. My point is that changing the current settings is not fixing anything.

No the user is not able to set NSID to zero actually but it can be only input the command parameter as zero then it is changed to all F by the current implementation.

ikegami-t:
The flow is below and the fix is changing the number 4 reset value to `0` from `0xffffffff`.
1. The local variable: `cfg.namespace_id` is initialized as `0`.
2. If the command parameter: `namespace-id` is input for example `0` then the local variable: `cfg.namespace_id` is reset to the input in this case as `0` as example.
3. The local variable: `cfg.namespace_id` is checked if it is 0 then checked the block device 'NSID' also but it is not exisited the device `NSID: 0` as `/dev/nvme*n0` I think.
4. If the character device: `/dev/nvme*` is input then the local variable: `cfg.namespace_id` is reset to `0xffffffff` so the NSID: 0 is not able to set.
5. The local variable: `cfg.namespace_id` is passed into the vnme drvicer.

ikegami-t:
To make sure let me comment again. I think the problem is not able to set the NSID `0` currently. Also basically the NSID can be set by the command paramter except `0`. The fix is only for the default NSID behavior only to changge `0xffffffff` to `0` so I think not incorrect completely.

keithbusch:
So then I guess we need to remove the whole check for "!cfg.namespace_id" and rely on the user to set it properly.

ikegami-t:
> So then I guess we need to remove the whole check for "!cfg.namespace_id" and rely on the user to set it properly.

But currently if !cfg.namespace_id is 0 the block device NSID is set to the variable. So if removed this the block device NSID not used to set. Is this expected?

keithbusch:
> > So then I guess we need to remove the whole check for "!cfg.namespace_id" and rely on the user to set it properly.
> 
> But currently if !cfg.namespace_id is 0 the block device NSID is set to the variable. So if removed this the block device NSID not used to set. Is this expected?

What other choice have we got? If 0 is a valid value, then we have to let users select it even when going through a namespace handle.

ikegami-t:
How about this? (This is same with the commit changes.)
1. If the NSID paramter except 0 is input then it is used.
2. If the NSID paramter is not intut or intput 0 and the block devie is input then the block device NSID is used.
3. Other case the NSID 0 as default is used.

| Input parameter NSID | Input block device NSID | Pass NSID to driver |
| --- | --- | --- |
| 1 ... 0xfffffff | * (Any) | 1 ... 0xfffffff |
| 0 or - (Not set) | * (Any): /dev/nvmeXn* | * (Block device NSID): /dev/nvmeXn* |
| 0 or - (Not set) | - (Not set) | 0 (Changed 0xffffffff to 0 by the fix) |

keithbusch:
That makes it features that require nsid 0 unreachable through the block handle. Are you okay with that?

ikegami-t:
I think it can be okay to input the NSID paramter but not the block device NSID. Also I think that the block device does not have NSID is this not correct? Since the NS management command requires NSID 0 to create and the avairable NSID is selected by the drive and usually I think the NSID 0 is not selected by the drive.

keithbusch:
Correct, 0 is not a valid NSID for any namespace. Its usage is for commands that are not targeted at namespaces.

keithbusch:
Let's say I have just one namespace with NSID 1.

If I run: `nvme get-feature /dev/nvme0n1 -f X --namespace-id=0`, the tool will replace the user's provided nsid with '1'.

The tool would have allowed the user input if they said `--namespace-id=12345` instead.

You are proposing to allow `--namespace-id=0` only if the user provides the ctrl handle, like `/dev/nvme0`. This makes nsid 0 features unreachable through block handles. I'm just making sure that's what you want.

ikegami-t:
Thanks for your explanation. I see. Yes the behavior seems okay to me. (Also I do not have any other idea to resolve this.)

keithbusch:
That's why I suggested to just get rid of all the namespace id parameter checks and just trust the user.

ikegami-t:
Anyway updated the patch to delete the NSID checking as suggested.

keithbusch:
Maybe something like this? (note, I'm not including any of the fallout from the 'const' modifier in the below diff)

```
diff --git a/nvme.c b/nvme.c
index e10e3823..b3058025 100644
--- a/nvme.c
+++ b/nvme.c
@@ -4602,7 +4602,7 @@ static int get_feature(int argc, char **argv, struct command *cmd,
        if (err)
                goto ret;

-       if (!cfg.namespace_id) {
+       if (!opts[1].seen) {
                err = nvme_get_nsid(dev_fd(dev), &cfg.namespace_id);
                if (err < 0) {
                        if (errno != ENOTTY) {
diff --git a/util/argconfig.c b/util/argconfig.c
index 972b8d6f..5e7b2857 100644
--- a/util/argconfig.c
+++ b/util/argconfig.c
@@ -232,6 +232,7 @@ int argconfig_parse(int argc, char *argv[], const char *program_desc,
                }

                s = &options[option_index];
+               s->seen = true;
                value_addr = (void *)(char *)s->default_value;
                if (s->config_type == CFG_STRING) {
                        *((char **)value_addr) = optarg;
diff --git a/util/argconfig.h b/util/argconfig.h
index 6ef3b6a1..41ea8326 100644
--- a/util/argconfig.h
+++ b/util/argconfig.h
@@ -37,6 +37,7 @@
 #include <string.h>
 #include <getopt.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <stdio.h>

 enum argconfig_types {
@@ -109,6 +110,7 @@ struct argconfig_commandline_options {
        void *default_value;
        int argument_type;
        const char *help;
+       bool seen;
 };

 #define CFG_MAX_SUBOPTS 500
```

ikegami-t:
Thanks for your advices. Just fixed the patch as suggested. Note: The option args are changed as not `const` variables to set the seen flag parameter.

PR #1839, opened 2023-03-02
plugins/ocp: OCP 2.0 field updates for SMART cloud log

kdedow:
- Add power state change field. This is an OCP 2.0 field.
- Parse full two bytes for point version and minor version fields. Although this should have no functional effect (since versions fields should be zero for 2.0) this should be correct (and if version ever increments past 255 we won't have a bug!)

Screenshots from OCP 2.0:
https://www.opencompute.org/documents/datacenter-nvme-ssd-specification-v2-0r21-pdf

image

image


arthurshau:
Looks good to me

lgdacunh:
Looks good to me too!

igaw:
Thanks everyone!

PR #1840, opened 2023-03-03
nvme.c: Check Firmware Update Granularity in fwdl

sc108-lee:
FWUG indicates the granularity and alignment requirement of the firmware image being updated by the Firmware Image Download command

FWUG is reported in 4 KiB units
0h indicates that no information on granularity is provided FFh indicates there is no restriction
so when this is 0 value, use 4 KiB split as a default

Reported-by: Minsik Jeon <hmi.jeon@samsung.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4842,6 +4842,7 @@ static int fw_download(int argc, char **argv, struct command *cmd, struct plugin
 	struct stat sb;
 	void *fw_buf;
 	bool huge;
+	struct nvme_id_ctrl ctrl;
 
 	struct config {
 		char	*fw;
@@ -4894,7 +4895,18 @@ static int fw_download(int argc, char **argv, struct command *cmd, struct plugin
 		goto close_fw_fd;
 	}
 
-	if (cfg.xfer == 0 || cfg.xfer % 4096)
+	if (cfg.xfer == 0) {
+		err = nvme_cli_identify_ctrl(dev, &ctrl);
+		if (err) {
+			fprintf(stderr, "identify-ctrl: %s\n", nvme_strerror(errno));
+			goto close_fw_fd;
+		}
+		if (ctrl.fwug == 0 || ctrl.fwug == 0xff)
+			cfg.xfer = 4096;
+		else
+			cfg.xfer = ctrl.fwug * 4096;
+	}
+	else if (cfg.xfer % 4096)
 		cfg.xfer = 4096;
 
 	if (cfg.xfer < HUGE_MIN)


igaw:
Looks good to me. @jk-ozlabs any objection?

keithbusch:
I see what you're saying, but I always expected the user to supply the correct value rather than automate it. Let's say current firmware has a bug that reports fwug incorrectly: a value too high that exceeds mdts. Now you have to install new firmware that fixes it, but this patch makes it impossible to send the new firmware.

sc108-lee:
@keithbusch I understood your concern. 
At the first I ignored that case since both cases should be failed.
fwug > mdts => fwug send => Invalid Field in Command
fwug > mdts => 4k send =>  Invalid Field in Command

But after your review i got second thought, FWUG could be ignored when ssd vendor allowed in some way.
so I updated commit to get FWUG when xfer is not given.

Thanks for the review

keithbusch:
Yes, this is good. Thanks for changing, despite the obscure usecase.

keithbusch (approved):

jk-ozlabs:
This is going to cause issues if the transport maximum is lower than 4k, do we care about that case?

keithbusch:
The minimum possible maximum the spec allows is 8k.

jk-ozlabs:
the maximum possible size that the MI spec allows is 4k :)

jk-ozlabs:
(unless we use the Management Endpoint Buffer, but I'm not sure whether it's common for devices to support the MEB...)

igaw:
I don't know if we should care or not. It sounds like the spec is inconsistent on this topic. If this the case we might want to let the spec folks know about it :)

jk-ozlabs:
Yep, agreed.

But to get some input from the implementation side: @sc108-lee : what is your expectation for this when using the MI channel to update firmware?

sc108-lee:
@jk-ozlabs Good point to think about it.
I just looked MI SPEC (I didn't see the whole spec.)
As far as I understand,
Out-of-band(The maximum sized NVMe-MI Message is 4,224 bytes (i.e., 4 KiB + 128 bytes).)  -> split into MCTP
in-band -> not split into MCTP (same as MDTS)

If I understand right, your concern is about out-of-band case
For that case I think if we want use that, It should be split into MCTP, not split of fwdl command.
If split of fwdl command, it cause Invalid Field in Command by fwug.

jk-ozlabs:
> it should be split into MCTP, not split of fwdl command.

I'm not clear what you man by "split into MCTP", but: When using the MCTP transport, the maximum data length of any Admin command is 4kB; we can't do any further splitting of the command here.

[We do split the MCTP message into MCTP packets, but that's a different concept, and has no relation to the 4kB maximum of NVMe-MI messages]

sc108-lee:
oh I misunderstood.
1 MI message could be multiple of MCTP message.!
Over 4k of FWUG, it need to use MEB to send the data. (it may require some changes)
you were right, it's optional but if not support MEB then it should be failed anyway. (no need to care, it's common or not I think)

**In spec**
The amount of supported Request or Response Data is Command Message specific due to the presence of additional command specific fields. In some cases, it is desirable to service Command Messages that contain more
Request Data or Response Data than allowed to be transferred in an NVMe-MI Message. For example, one may wish to issue an NVM Express Admin Command Set Get Log Page command to transfer a log page that is greater in size than that allowed in the Response Data.
A Management Endpoint may support an optional Management Endpoint Buffer that facilitates Request Data and Response Data transfers that exceed that maximum size allowed by an NVMe-MI Message

jk-ozlabs:
Yes, we will need to use the MEB if we want to transfer > 4kB in one NVMe command.

However:
 * we don't yet have support for the MEB in libnvme (I'm happy to add it though!)
 * some devices may not support the MEB

We may need a little API for the command implementation to query the max message size, and use that to clamp the negotiated fwug size...

PR #1845, opened 2023-03-08
nvme: Fix invalid memory access in show_registers

mjoerg:
In case nvme_get_property fails, nvme_get_properties frees the allocated memory.
The memory should not be accessed and freed again in show_registers.


igaw (changes requested):
Could update the allocation path to use a local variable in `nvme_get_properties()`.

And please add the prefix 'nvme: ' the subject line. Thanks!

igaw:
Thanks!

PR #1848, opened 2023-03-09
nvme-print: sanitize supported-log-pages output

martin-gpy:
The current nvme supported-log-pages output is a little cryptic as seen below:

Support Log Pages Details for nvme1n1:
LID 0x0 (Supported Log Pages), supports 0x3

LID 0x1 (Error Information), supports 0x3
...

So sanitize this output to make it more meaningful with better formatting. These changes apply to the verbose and json outputs too. With these changes, the sanitized normal output shows up as follows:

Support Log Pages Details for nvme1n1:
LID 0x0 - Supported Log Pages
LID 0x1 - Error Information
...

And the sanitized verbose output would show up as:

Support Log Pages Details for nvme1n1:
LID 0x0 - Supported Log Pages
  LSUPP is supported
  IOS is supported
LID 0x1 - Error Information
  LSUPP is supported
  IOS is supported
...

And finally the sanitized json output would look as:

{
  "Supported_Logs":[
    {
      "LID_0x0":"Supported Log Pages"
    },
    {
      "LID_0x1":"Error Information"
    },
...

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -2447,13 +2447,13 @@ void json_support_log(struct nvme_supported_log_pages *support_log)
 		support = le32_to_cpu(support_log->lid_support[lid]);
 		if (support & 0x1) {
 			valid_attrs = json_create_object();
-			sprintf(key, "lid_0x%x ", lid);
-			json_object_add_value_uint(valid_attrs, key, support);
+			sprintf(key, "LID_0x%x", lid);
+			json_object_add_value_string(valid_attrs, key, nvme_log_to_string(lid));
 			json_array_add_value_object(valid, valid_attrs);
 		}
 	}
 
-	json_object_add_value_object(root, "supported_logs", valid);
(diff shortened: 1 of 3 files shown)


igaw:
I am fine with updating the `stdout`, though I am a bit reluctant with the JSON part. 

The existing JSON naming scheme pattern seems to be mostly to use the variable
name, e.g. 
```c
		json_object_add_value_uint128(entry, "unalloc_dom_cap", unalloc_dom_cap);
```

but when some getters() are involved it's camel casing
```c
		json_object_add_value_string(hss, "HostNQN", nvme_host_get_hostnqn(h));
```

or strings with spaces
```c
		json_object_add_value_int(valid_attrs, "Self test code", self_test->result[i].dsts >> 4);
```

Normally, I wouldn't mind too much but as it breaks the API, I rather be sure we want to change this.

martin-gpy:
OK, fair enough. Will update the patch avoiding modifying the JSON output then. Thanks.

PR #1849, opened 2023-03-09
Fix printing of u32 values

calebsander:
Fields `HMPRE`, `HMMIN`, and `NN` in the Identify Controller data structure are 32-bit unsigned values but are being printed as signed integers. Print them using `%u` so large values are not displayed as negative ones. Also make JSON `uint` objects use `uint64` instead of `int64` internally so `u32` values are stored accurately.

Before (on a device which reports `NN` = 0xFFFFEFFF):
```
$ ./.build/nvme id-ctrl /dev/nvme0 | grep nn
nn        : -4097
$ ./.build/nvme id-ctrl /dev/nvme0 -o json | grep nn
  "nn":-4097,
```
After:
```
$ ./.build/nvme id-ctrl /dev/nvme0 | grep nn
nn        : 4294963199
$ ./.build/nvme id-ctrl /dev/nvme0 -o json | grep nn
  "nn":4294963199,
```


igaw:
Thanks!

PR #1854, opened 2023-03-13
fdp: bug fix for data size of fdp status

sc108-lee:


igaw:
Please provide a better commit message. 'Bug fix' is a bit terse.

@birkelund Any objections?

birkelund:
> @birkelund Any objections?

LGTM!

sc108-lee:
@igaw rebased and modified.
I thought data size looks obviously wrong.
But if someone doesn't know the spec, might confuse
So added more details on message as you recommended

fdp: Fix wrong Reclaim Unit Handle Status size

size should include reserved, NRUHSD bytes
that exist before Reclaim Unit Handle Status Descriptor List

igaw:
Thanks. Well it might be obvious when reading all the code around. Just by looking at the commit I can't decide.

PR #1856, opened 2023-03-14
Check that cfg.len is dw-aligned before the get_log_page function to prevent it from mismatching numd in the get_log_page function

duanhaoqiang:
Check the cfg.len in the get_log function to prevent the buffer of malloc from being less than the get log length

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2257,7 +2257,7 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 		cfg.log_id = (cfg.aen >> 16) & 0xff;
 	}
 
-	if (!cfg.log_len || cfg.log_len &0x3) {
+	if (!cfg.log_len || cfg.log_len & 0x3) {
 		fprintf(stderr, "non-zero or non-dw alignment log-len is required param\n");
 		err = -EINVAL;
 		goto close_dev;
--- a/nvme.c
+++ b/nvme.c
@@ -2257,7 +2257,7 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 		cfg.log_id = (cfg.aen >> 16) & 0xff;
 	}
 
-	if (!cfg.log_len || cfg.log_len & 0x3) {
+	if (!cfg.log_len || cfg.log_len &0x3) {
 		fprintf(stderr, "non-zero or non-dw alignment log-len is required param\n");
 		err = -EINVAL;
 		goto close_dev;
--- a/nvme.c
+++ b/nvme.c
@@ -2257,8 +2257,8 @@ static int get_log(int argc, char **argv, struct command *cmd, struct plugin *pl
 		cfg.log_id = (cfg.aen >> 16) & 0xff;
 	}
 
-	if (!cfg.log_len) {
-		fprintf(stderr, "non-zero log-len is required param\n");
+	if (!cfg.log_len || cfg.log_len & 0x3) {
+		fprintf(stderr, "non-zero or non-dw alignment log-len is required param\n");
 		err = -EINVAL;
 		goto close_dev;
 	}


igaw:
The commit message doesn't really match with what the change is doing. The patch enforces dword alignment. 

I found in the spec  `Figure 117: SGL Data Block descriptor`:
```
Length: The Length field specifies the length in bytes of the data block. A Length field cleared to
0h specifies that no data is transferred. An SGL Data Block descriptor specifying that no data is
transferred is a valid SGL Data Block descriptor. If the controller requires dword alignment and
granularity as specified in the SGL Support (SGLS) field the of Identify Controller data structure,
then the two least significant bits shall be cleared to 00b.
```

So it depends on the controller, no?

duanhaoqiang:
> The commit message doesn't really match with what the change is doing. The patch enforces dword alignment.
> 
> I found in the spec `Figure 117: SGL Data Block descriptor`:
> 
> ```
> Length: The Length field specifies the length in bytes of the data block. A Length field cleared to
> 0h specifies that no data is transferred. An SGL Data Block descriptor specifying that no data is
> transferred is a valid SGL Data Block descriptor. If the controller requires dword alignment and
> granularity as specified in the SGL Support (SGLS) field the of Identify Controller data structure,
> then the two least significant bits shall be cleared to 00b.
> ```
> 
> So it depends on the controller, no?
I think the memory length of calling malloc should be the same length as the NUMDL and NUMDU of the get log page to avoid memory overflow or waste. In the original code, for example, cfg.len is 1 byte and malloc has 1 byte, then "__u32 numd = (args->len >> 2) - 1" in the nvme_get_log function is 0xffffffff. The length of the transmitted data has exceeded malloc's memory, causing unpredictable problems. Thank you

igaw:
Ah I see. I forgot that we cut the size down for `numd`. Could you add the info on `numd` also to the commit message. So it's just a bit of too much we allocate but then don't use it. 

Again, information like this should be part of the commit message.

duanhaoqiang:
> Ah I see. I forgot that we cut the size down for `numd`. Could you add the info on `numd` also to the commit message. So it's just a bit of too much we allocate but then don't use it.
> 
> Again, information like this should be part of the commit message.

duanhaoqiang:
> Ah I see. I forgot that we cut the size down for `numd`. Could you add the info on `numd` also to the commit message. So it's just a bit of too much we allocate but then don't use it.
> 
> Again, information like this should be part of the commit message.

igaw:
Please update the commit message (I am not asking to update github issue title).

igaw:
We use the Linux kernel style of writing change logs (https://github.com/linux-nvme/nvme-cli#how-to-contribute). Especially the section https://docs.kernel.org/process/submitting-patches.html#describe-your-changes is of importance. The problem with your commit message is that I had to ask you what's the problem. The initial description was not enough for me to figure this out. So adding the original explanation as commit message is not really helping here.

"Fix the issue in the get_log_page" is not a good commit title. It should really tell what it is changing because 'fix a bug' could be anything. So something like 'nvme: Enforce length to a multiple of 4 in get_log_page' and the commit message includes the reference to the `numdl` part?

duanhaoqiang:
> We use the Linux kernel style of writing change logs (https://github.com/linux-nvme/nvme-cli#how-to-contribute). Especially the section https://docs.kernel.org/process/submitting-patches.html#describe-your-changes is of importance. The problem with your commit message is that I had to ask you what's the problem. The initial description was not enough for me to figure this out. So adding the original explanation as commit message is not really helping here.
> 
> "Fix the issue in the get_log_page" is not a good commit title. It should really tell what it is changing because 'fix a bug' could be anything. So something like 'nvme: Enforce length to a multiple of 4 in get_log_page' and the commit message includes the reference to the `numdl` part?
> 
> And please squash all changes into one commit. Thanks.

Thanks for your correction, I submitted a new request

PR #1859, opened 2023-03-15
nvme: Fix parameter limit range

sc108-lee:
3 Bits - Max value 7
SEL : Bits [10:08] of Get Features Command Dword 10

4 Bits - Max value 15
OWPASS : Bits [07:04] of Sanitize – Command Dword 10

7 Bits - Max value 127
LSP : Bits [14:08] of Get Log Page Command Dword 10 UUID Index : Bits [06:00] of Get Log Page, Get Features, Set Features Command Dword 14

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -1815,7 +1815,7 @@ static int get_boot_part_log(int argc, char **argv, struct command *cmd, struct 
 		goto close_dev;
 	}
 
-	if (cfg.lsp > 128) {
+	if (cfg.lsp > 127) {
 		fprintf(stderr, "invalid lsp param: %u\n", cfg.lsp);
 		err = -1;
 		goto close_dev;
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -5244,8 +5244,8 @@ static int sanitize(int argc, char **argv, struct command *cmd, struct plugin *p
 	}
 
 	if (sanact == NVME_SANITIZE_SANACT_START_OVERWRITE) {
-		if (cfg.owpass > 16) {
-			fprintf(stderr, "OWPASS out of range [0-16]\n");
+		if (cfg.owpass > 15) {
```
igaw, line 5247: If `owpass` is 3 bits, how can it hold a value of 15?

sc108-lee: As I mentioned on commit message owpass is 4bit value

Figure 303: Sanitize – Command Dword 10

Bits [07:04]
Overwrite Pass Count (OWPASS): This field specifies the number of overwrite passes (i.e., how many times the media is to be overwritten) using the data from the Overwrite Pattern field of this command. A value of 0h specifies 16 overwrite passes. This field shall be ignored unless the Sanitize Action field is set to 011b (i.e., Overwrite).

PR #1860, opened 2023-03-15
Fix the mismatch between the log memory of malloc and the numd length in the command in the get_log function

duanhaoqiang:
Check that cfg.len is dw-aligned before the get_log_page function to prevent it from mismatching numd in the get_log_page function.


igaw:
If you could update the commit message too, that would be really helpful.

duanhaoqiang:
> If you could update the commit message too, that would be really helpful.

done

PR #1861, opened 2023-03-15
plugins/ocp: Add support for clear PCIe correctable error counters fe…

ikegami-t:
…ature

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -861,25 +861,37 @@ static int eol_plp_failure_mode_get(struct nvme_dev *dev, const __u32 nsid,
 }
 
 static int eol_plp_failure_mode_set(struct nvme_dev *dev, const __u32 nsid,
-				    const __u8 fid, __u8 mode, bool save)
+				    const __u8 fid, __u8 mode, bool save,
+				    bool uuid)
 {
 	__u32 result;
 	int err;
+	int uuid_index = 0;
+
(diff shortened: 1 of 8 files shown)


plugins/ocp/ocp-nvme.c
```diff
@@ -942,3 +945,65 @@ static int eol_plp_failure_mode(int argc, char **argv, struct command *cmd,
...
+	OPT_ARGS(opts) = {
+		OPT_FLAG("no-uuid", 'n', &no_uuid,
+			 "Skip UUID index search (UUID index not required for OCP 1.0)"),
+		OPT_END()
+	};
+
+	err = parse_and_open(&dev, argc, argv, desc, opts);
+	if (err)
+		return err;
+	if (no_uuid == false) {
```
igaw, line 970: Personally I get easily confused by inverted logic. Wouldn't it make sense have a `bool uuid = true` as default?

plugins/ocp/ocp-nvme.c
```diff
@@ -942,3 +945,65 @@ static int eol_plp_failure_mode(int argc, char **argv, struct command *cmd,
...
+		// OCP 2.0 requires UUID index support
```
igaw, line 971: `/* comment */`

igaw (commented):
@arthurshau

ikegami-t:
Thanks for your reviewing and comments. Just fixed the changes as mentioned.

PR #1863, opened 2023-03-20
print: print csi and opcode in error log (TP4113)

minwooim:
Print newly added fields in the error information log page to show which command is errored by opcode. This commit is based on TP4113.

---
This needs (https://github.com/linux-nvme/libnvme/pull/598) to be accepted :)


minwooim:
Pushed with bumping up `libnvme`

igaw:
Thanks!

PR #1864, opened 2023-03-21
nvme: fix block count and data size logic

martin-gpy:
The required block count and data size is not automatically set for a namespace device in submit_io(). Also it is the user specified block count and not the actual required block count that is passed to the ioctl data structure, often leading to nvme passthrough io errors. For e.g. an nvme read on a 512b block size namespace device with data size set to 4K ends up with the below error:

nvme read /dev/nvme0n1 -s 0 -z 4096
NVMe status: Data SGL Length Invalid: The length of a Data SGL is too short or too long and the controller does not support SGL transfers longer than the amount of data to be transferred(0x400f)

This read is successful only when the appropriate block count is passed from the command line:

nvme read /dev/nvme0n1 -s 0 -z 4096 -c 7
read: Success

So fix this by deriving the required block count and data size based on the logical block size and passing them appropriately to the respective ioctl data structure. And while we are at it, remove a couple of superfluous braces in this part of the code.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -7213,7 +7213,7 @@ static int submit_io(int opcode, char *command, const char *desc,
 	int dfd, mfd;
 	int flags = opcode & 1 ? O_RDONLY : O_WRONLY | O_CREAT;
 	int mode = S_IRUSR | S_IWUSR |S_IRGRP | S_IWGRP| S_IROTH;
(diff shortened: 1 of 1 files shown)


igaw:
Thanks. Funny my local device seems to handle the wrong size whereas the Linux soft target reports an error.

PR #1868, opened 2023-03-22
nvme-gen-tls-key: add options to derive a TLS key

hreinecke:
Add options to derive a TLS key and store it into the given keyring.

### DIFF
--- a/Documentation/nvme-gen-tls-key.txt
+++ b/Documentation/nvme-gen-tls-key.txt
@@ -8,7 +8,10 @@ nvme-gen-tls-key - Generate a NVMe TLS PSK
 SYNOPSIS
 --------
 [verse]
-'nvme gen-tls-key' [--hmac=<hmac-id> | -h <hmac-id>]
+'nvme gen-tls-key' [--keyring=<keyring-name> | -k <keyring-name>]
+		      [--hostnqn=<nqn> | -n <nqn>]
+		      [--subsysnqn=<nqn> | -c <nqn>]
+		      [--hmac=<hmac-id> | -h <hmac-id>]
 		      [--secret=<secret> | -s <secret> ]
 
 DESCRIPTION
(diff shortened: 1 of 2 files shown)


nvme.c
```diff
@@ -8858,16 +8871,32 @@ static int gen_tls_key(int argc, char **argv, struct command *command, struct pl
...
+		raw_secret[key_len++] = (crc >> 16) & 0xff;
+		raw_secret[key_len++] = (crc >> 24) & 0xff;
 
-	memset(encoded_key, 0, sizeof(encoded_key));
-	base64_encode(raw_secret, key_len, encoded_key);
+		memset(encoded_key, 0, sizeof(encoded_key));
+		base64_encode(raw_secret, key_len, encoded_key);
+		printf("NVMeTLSkey-1:%02x:%s:\n",
+		       cfg.hmac, encoded_key);
+		return 0;
```
sagi-lb, line 8885: I don't understand the semantics of this function. if host and subsys nqns not given you just print the key and exit, and if they were given then it is stored in a keyring?

hreinecke: Correct. The 'NVMeTLSkey' thingie is the so-called 'PSK Interchange Format' (cf section 3.6.1.5 of the NVMe TCP spec), and is meant to be used if you want to generate a PSK to be installed on (other) machines.
This format has _no_ indication to which NQN is should match.
From this the actual TLS key should be derived (cf section 3.6.1.3 of the NVMe TCP spec), for which you have to specify the host NQN and subsysten NQN as they are factored in when deriving the TLS key.

Documentation/nvme-gen-tls-key.txt
```diff
@@ -20,6 +23,19 @@ and prints it to stdout.
 
 OPTIONS
 -------
+-k <keyring-name>::
```
igaw, line 26: I think the new feature deserves some explanation in the description sections.

nvme.c
```diff
@@ -8858,16 +8871,32 @@ static int gen_tls_key(int argc, char **argv, struct command *command, struct pl
...
+	}
+	if ((!cfg.hostnqn && cfg.subsysnqn) ||
+	    (cfg.hostnqn && !cfg.subsysnqn)) {
```
igaw, line 8888: What about `!(a && b)`?

igaw (changes requested):
I think with a bit documentation this good to go.

hreinecke:
Superseded by PR https://github.com/linux-nvme/nvme-cli/pull/1875 .

PR #1869, opened 2023-03-22
MAINT: util/nvme: Maintenance changes

ikegami-t:
1. MAINT: util: Use argconfig commandline options pointer directly
Also change print help function options argument as not const variable.
Fix ";;" also incorrectly used to ";".
2. MAINT: nvme: Fix format command function indentation errors


nvme.c
```diff
@@ -5749,11 +5751,14 @@ static int format(int argc, char **argv, struct command *cmd, struct plugin *plu
 	}
 
 	if (!cfg.force) {
-		fprintf(stderr, "You are about to format %s, namespace %#x%s.\n",
+		fprintf(stderr,
+			"You are about to format %s, namespace %#x%s.\n",
 			dev->name, cfg.namespace_id,
-			cfg.namespace_id == NVME_NSID_ALL ? "(ALL namespaces)" : "");
+			cfg.namespace_id == NVME_NSID_ALL ? "(ALL namespaces)" :
+			"");
```
igaw, line 5758: I would leave the last line on the previous line, looks more readable to me.

We can tolerate a few more chars than 80 these days (IIRC, some parts of the kernel is 100 chars okay) especially if it makes it a bit more readable.

util/argconfig.c
igaw, line 1: Nice cleanup.

igaw (changes requested):
Nice cleanup. I'd say leave some of the overlong lines (> 80) intact because of readability. 

Could you also drop the `MAIN:` prefix on the commits. It doesn't really add additional information. It's okay for the PR though.

Thanks!

igaw (approved):

igaw:
Thanks!

PR #1871, opened 2023-03-22
plugins/ocp: Move SMART cloud log methods to separate file for reusab…

kdedow:
Move SMART cloud log methods to separate file for reusability. Specifically, I plan to reference the `ocp_smart_add_log` method in solidigm plugin.

No methods/code was modified except for the static ocp_smart_add_log method in ocp-nvme.c/.h (the name of the method was updated to smart_add_log, and the method logic now simply references ocp_smart_add_log in ocp-smart-extended-log.c/.h. Change was focused on moving OCP SMART cloud logic into its own file.

**Before Results (SMART Cloud):**
SMART Cloud Before Changes

SMART Cloud Before Changes - JSON

**After Results (SMART Cloud)**:
SMART Cloud After Changes

SMART Cloud After Changes - JSON

### DIFF
--- a/plugins/ocp/meson.build
+++ b/plugins/ocp/meson.build
@@ -2,5 +2,6 @@ sources += [
   'plugins/ocp/ocp-utils.c',
   'plugins/ocp/ocp-nvme.c',
   'plugins/ocp/ocp-clear-fw-update-history.c',
+  'plugins/ocp/ocp-smart-extended-log.c',
 ]
 
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -22,26 +22,17 @@
 #include "linux/types.h"
 #include "util/types.h"
 #include "nvme-print.h"
+
+#include "ocp-smart-extended-log.h"
 #include "ocp-clear-fw-update-history.h"
 
 #define CREATE_CMD
 #include "ocp-nvme.h"
 #include "ocp-utils.h"
 
-/* C0 SCAO Log Page */
-#define C0_SMART_CLOUD_ATTR_LEN			0x200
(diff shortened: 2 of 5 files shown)


kdedow:
@igaw @arthurshau @lgdacunh

lgdacunh:
@kdedow, as you wrote on the first comment " I plan to reference the ocp_smart_add_log method in solidigm plugin."
Maybe you could add to this pull request, the actual  soldigm plug-in code that actually reference  ocp_smart_add_log() to give some more context of what you are going for. 
Because feature is a vector that keeps growing on this plug-ins,  I personally like the code to be organized with a "feature file" for each feature unique logic and data structures. But I already read comments from @igaw and @arthurshau exposing their preference for one file per plug-in, so they may not see a clear purpose for the proposed change.

igaw:
I am not a big fan of single function per file approach, compile overhead and navigating the code base is a bit more difficult if you don't use a IDE.

Though also too big files are not good to handle either. In the end I try to keep it pragmatic, so no hard rules.

In the end if all OCP involved parties want to do it this way, I wont stop you guys. I see @arthurshau as the one who makes the call in this part of the code base.

kdedow:
@igaw - Happy to follow the recommendation on files.

The one thing that is required for me to use the OCP plugin implementation in the Solidigm plugin is to split the smart cloud method into a static method (to follow the command macro design) and non-static method (so I can link).

kdedow:
I will follow what @arthurshau recommends

arthurshau:
I agree with @igaw's reasoning. As long we keep it pragmatic, I think it's ok for now. LGTM

igaw:
Okay, so let's go with this one.

PR #1872, opened 2023-03-23
nvme: validate storage tag size correctly

Laiwenhu:
when pif =2, sts=48, ref tag value not zero, write cmd succees

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -6358,7 +6358,7 @@ static int invalid_tags(__u64 storage_tag, __u64 ref_tag, __u8 sts, __u8 pif)
 			result = 1;
 		break;
 	case 1:
-		if (sts > 16 && ref_tag >= (1LL << (80 - sts)))
+		if (sts >= 16 && ref_tag >= (1LL << (80 - sts)))
 			result = 1;
 		break;
 	case 2:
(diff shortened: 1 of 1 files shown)


igaw (changes requested):
Thanks for your contribution, could update your patch and include following changes:

- use a more descriptive commit title, e.g. "nvme: validate storage tag size correctly"
- update the STS=1 case as well (min 16,  max  value 64)
- and please add the reference to the spec in the commit message body:  Figure 101: Extended LBA Format Data Structure, NVM Command Set Specific

nvme.c
```diff
@@ -6358,7 +6358,7 @@ static int invalid_tags(__u64 storage_tag, __u64 ref_tag, __u8 sts, __u8 pif)
 			result = 1;
 		break;
 	case 1:
-		if (sts > 16 && ref_tag >= (1LL << (80 - sts)))
+		if (sts >= 16 && ref_tag >= (1LL << (80 - sts)))
```
igaw, line 6361: After looking again on it, I am sure it was correct. Sorry about that. I'll fixup the commit.

igaw:
I've pushed the change directly. Thanks!

PR #1873, opened 2023-03-23
RFC: xnvme backend support

vikash-k5:
Extended “struct nvme_dev” with the following device-handle types:

* struct dev_handle
* enum nvme_dev_type

The  ‘dev_handle' contains ‘struct xnvme_dev*’, file-descripter and ‘nvme_dev_type’ to manage library-side dispatch.

Adding  ‘open_dev_xnvme(struct nvme_dev **devp, char *devname)’ function to open xNVMe library device handle and same being passed to ‘libnvme’ through ‘struct dev_handle’.

This xnvme plugin can work on different platform such as linux, freebsd and even windows. As well as this can work with different driver on linux(kernel and userspace driver such as SPDK)

Note: this will require libnvme changes, which can be found here
https://github.com/linux-nvme/libnvme/pull/600

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -49,6 +49,7 @@ libnvme_dep = dependency('libnvme', version: '>=1.3', required: true,
                          fallback : ['libnvme', 'libnvme_dep'])
 libnvme_mi_dep = dependency('libnvme-mi', required: true,
                          fallback : ['libnvme', 'libnvme_mi_dep'])
+libxnvme_dep = dependency('xnvme', version: '>=0.2.0', required: true)
 
 # Check for libjson-c availability
 if get_option('json-c').disabled()
@@ -268,7 +269,7 @@ subdir('Documentation')
 executable(
   'nvme',
   sources,
-  dependencies: [ libnvme_dep, libnvme_mi_dep, json_c_dep,
+  dependencies: [ libxnvme_dep, libnvme_dep, libnvme_mi_dep, json_c_dep,
                   libhugetlbfs_dep ],
   link_args: '-ldl',
   include_directories: incdir,
--- a/nvme-wrap.c
+++ b/nvme-wrap.c
@@ -20,8 +20,8 @@
  */
 #define do_admin_op(op, d, ...) ({					\
 	int __rc;							\
-	if (d->type == NVME_DEV_DIRECT)					\
-		__rc = nvme_ ## op(d->direct.fd, __VA_ARGS__);		\
+	if (d->type == NVME_DEV_DIRECT || d->type == NVME_DEV_XNVME)	\
+		__rc = nvme_ ## op(&d->direct.hnd, __VA_ARGS__);		\
 	else if (d->type == NVME_DEV_MI)				\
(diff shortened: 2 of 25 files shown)


meson.build
```diff
@@ -49,6 +49,7 @@ libnvme_dep = dependency('libnvme', version: '>=1.3', required: true,
                          fallback : ['libnvme', 'libnvme_dep'])
 libnvme_mi_dep = dependency('libnvme-mi', required: true,
                          fallback : ['libnvme', 'libnvme_mi_dep'])
+libxnvme_dep = dependency('xnvme', version: '>=0.2.0', required: true)
```
igaw, line 52: This should be guarded by meson option setup option to disable it completely, see `json-c` as example. Some folks really don't like if we add hard dependencies.

nvme-wrap.c
```diff
@@ -20,8 +20,8 @@
  */
 #define do_admin_op(op, d, ...) ({					\
 	int __rc;							\
-	if (d->type == NVME_DEV_DIRECT)					\
-		__rc = nvme_ ## op(d->direct.fd, __VA_ARGS__);		\
+	if (d->type == NVME_DEV_DIRECT || d->type == NVME_DEV_XNVME)	\
```
igaw, line 23: The complete `nvme-wrap.c` file shouldn't be needed after introducing the transport abstraction.

nvme.c
```diff
@@ -387,6 +387,42 @@ static int check_arg_dev(int argc, char **argv)
...
+	*devp = dev;
+	return 0;
+
+perror:
+        perror(devname);
+        return err;
+}
+
+
+
```
igaw, line 425: Please avoid doing whitespace changes or adding too many newlines here.

nvme.c
```diff
@@ -387,6 +387,42 @@ static int check_arg_dev(int argc, char **argv)
...
 static int get_dev(struct nvme_dev **dev, int argc, char **argv, int flags)
```
igaw, line 426: I thought we are going to use the newly introduce transport abstraction from the library through out nvme-cli? Ideally we have this open device in libnvme, no?

nvme.c
```diff
@@ -5535,8 +5575,8 @@ static int set_property(int argc, char **argv, struct command *cmd, struct plugi
 	}
 
 	struct nvme_set_property_args args = {
+		.hnd		= dev_fd(dev),
```
igaw, line 5578: `dev_fd` should be renamed of course.

nvme.h
```diff
@@ -69,15 +66,15 @@ struct nvme_dev {
 
 #define dev_fd(d) __dev_fd(d, __func__, __LINE__)
 
-static inline int __dev_fd(struct nvme_dev *dev, const char *func, int line)
+static inline struct dev_handle* __dev_fd(struct nvme_dev *dev, const char *func, int line)
```
igaw, line 69: as previously said, this function should be renamed.

igaw (changes requested):
It looks more of the transport abstraction should be in libnvme.

vikash-k5:
Created new PR addressing review comments

Issue #1867, opened 2023-03-22
fw-commit: Print the 'MUD status' if supported


igaw:
The tool will always print the 'MUD status' line because it doesn't test first if it's supported or not. So it's a bit confusing but harmless. Obviously, the `fw-commit` command could be improved to only print the MUD status if it is actually supported.

_Originally posted by @igaw in https://github.com/linux-nvme/nvme-cli/issues/1866#issuecomment-1479713594_

igaw:
Thanks @ikegami-t for providing the new feature!

PR #1874, opened 2023-03-23
nvme: Check fw-commit command support MUD

ikegami-t:
Changes for the issue #1867.


igaw:
Just updated the commit title of the second patch.

igaw:
Thanks!

Issue #1797, opened 2023-02-05
How to retrieve errno code of IO commands in shell?


alexmchou:
Hello,
I am trying to test NVMe cli IO commands' error status code in bash script. If I got following error message for a write command: "NVMe status: End-to-End Storage Tag Check Error: The command was aborted due to an end-to-end storage tag check failure(0x288)", I tried to use "echo $?",  but got 1 instead of 136 (0x88).  From the nvme.c code, err (not errno)  is returned by IO command handler. This is different than what I got from history issue #862. Any suggestions?
Thanks

ikegami-t:
As below currently only returned 0 on success and 1 on error that was changed by the commit 13f74d2.
```
int main(int argc, char **argv)
{
	int err;

	nvme.extensions->parent = &nvme;
	if (argc < 2) {
		general_help(&builtin);
		return 0;
	}
	setlocale(LC_ALL, "");

	err = handle_plugin(argc - 1, &argv[1], nvme.extensions);
	if (err == -ENOTTY)
		general_help(&builtin);

	return err ? 1 : 0;
}
```

igaw:
I think for the normal shell output keeping I think we should leave it as it is (see reasons in https://github.com/linux-nvme/nvme-cli/commit/13f74d20e1d9e571368b1d06fe9889a2fdd29849). This should be fine as it is thought to be used by humans which are able to read/parse the output.

Though I think we could make it for scripts and tooling around nvme a bit more useful which is supposed to use the JSON output. hat is when a command fails we return 1 but also a JSON formatted output with the error code and error message. What do you think of this idea?

ikegami-t:
Yes I can agree with your idea.

ikegami-t:
By the PR #1876 merged into maser a new `json` parameter has been added as below and then the errno code is able to be retrieved by the `value` output as below for example. Can we close the issue now?
```
[2023-03-25 12:16:14.221] tokunori@tokunori-desktop:~/nvme-cli$ sudo .build/nvme get-feature /dev/nvme0 -f 0xc0 --json
[2023-03-25 12:16:17.768] {
[2023-03-25 12:16:17.768]   "error":"Invalid Field in Command: A reserved coded value or an unsupported value in a defined field",
[2023-03-25 12:16:17.768]   "type":"nvme",
[2023-03-25 12:16:17.768]   "value":8194
[2023-03-25 12:16:17.768] }
```
[...]

PR #1876, opened 2023-03-24
util: Cleanup argconfig parse function

ikegami-t:
1. Split to parse config type function
2. Change if else statement to switch case statement
3. Use calloc instead of malloc
4. Delete unused while 0 statement
5. Add argconfig error common function

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <assert.h>
+#include <errno.h>
 
 #include "nvme-print-json.h"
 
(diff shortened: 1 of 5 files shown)


ikegami-t:
The second patch changes resolves the issue #1797.

igaw:
Nice cleanups.

FYI, my plan was to introduce a plugin interface for the different output formats instead of forwarding the calls in `nvme-print.c` to `nvme-print-json.c`.

So we would introduce an ops struct

```c
struct nvme_serialize_ops {
	void (*id_iocs)(struct nvme_id_iocs *iocs);
	void (*id_ctrl)(struct nvme_id_ctrl *ctrl, unsigned int mode,
			void (*vendor_show)(__u8 *vs, struct json_object *root));
	void (*id_ns)(struct nvme_id_ns *ns, unsigned int nsid,
		      unsigned int lba_index, bool cap_only, enum nvme_print_flags flags);
	[...]
};

```
and then hookup` nvme-print.c` (which we could then rename to `nvme-stdout.c`) and `nvme-print-json.c` (rename to `nvme-json.c`?) and the `nvme.c` just calls the right ops type.  Let me know if you want to work on this. I can also create an issue for this.

ikegami-t:
Thanks for your reviewing and comments. Sure yes I would like to work on it if you need. (I can also continue it as for the issue #1797 instead of creating an issue but I can start it from this weekend and I need to understand the plan detail more.)

PR #1877, opened 2023-03-26
nvme-print: Unify number base format for nsid

minwooim:
`nvme list` is one of the most frequently used subcommand to list up nvme devices in the system.  `nvme list -v` prints much more detail in it.  But, those two commands have different number base format which are decimal and hexadecimal without any prefix (e.g., 0x) or something.

nvme list -v shows hexadecimal nsid without `0x` which means it might be confused between decimal and hexadecimal.  To unify this, this patch expands NSID field width from 8 to 10 to cover `0x` two chracters as a prefix.  Chainging it to decimal format might cause issues for the existing systems looking at this value, so just put the prefix on it.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -5156,7 +5156,7 @@ void nvme_show_list_item(nvme_ns_t n)
 	nvme_dev_full_path(n, devname, sizeof(devname));
 	nvme_generic_full_path(n, genname, sizeof(genname));
 
-	printf("%-21s %-21s %-20s %-40s %-9d %-26s %-16s %-8s\n",
+	printf("%-21s %-21s %-20s %-40s %#-10x %-26s %-16s %-8s\n",
 		devname, genname, nvme_ns_get_serial(n),
 		nvme_ns_get_model(n), nvme_ns_get_nsid(n), usage, format,
 		nvme_ns_get_firmware(n));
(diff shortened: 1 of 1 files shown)


igaw:
I like the change, but there are a few more places which we print the nsid and we switch between decimal and hexadecimal. I think we should agree on one format and use it consistently. I would go with the decimal version.

minwooim:
I have actually been thinking about unifying `nsid` in decimal as well. Can we print out decimal `nsid` in --verbose case only?
I am worried if there are any places already parsing the result as hexadecimal, there are parsing formats such as JSON, though.

igaw:
I think those tools which are screen scraping they are more likely just do the `nvme list` command. So I don't think there is a big issue when we update the verbose version of it. And as pointed out we have JSON for a reason ;)

minwooim:
Sure, then please let me update it soon :)

igaw:
One thing which might be worth to note, there are two nsid types. The one which Linux assigns and the one on the protocol layer. Maybe it would be good to update only the nsid for the Linux to decimal and leave the rest to hex (which I think was the case already)

minwooim:
If I correctly understood what you meant about `protocol layer` (as device `nsid`),

In case of `nvme list` and `nvme list -v`, I guess all the `nsid` fields are representing the device `nsid`, not the `head_instance` from the kernel.  As you said, it's already printed in hexadecimal without any hexa-prefix.

Then, do you think it's good to go with `#` prefix for the `nvme list` as well by changing the decimal to hexadecimal ?

igaw:
Yes I think so. It doesn't really make sense to have different bases for the same variable.

minwooim:
Updated the commit, Thanks!

PR #1881, opened 2023-03-29
Not issue IDENTIFY command to get a LBA size

minwooim:
Hello,

This patchset is to prevent `submit_io()` from issuing an admin command (IDENTIFY) to figure out a LBA size of the target namespace. `nvme-cli` is a command line utility that composed of many sub-commands which is pretty based on command-by-command unit.  Someone might not want for any admin commands to proceed before `nvme read` or `nvme write`.  Kernel already has initialized namespaces and store the `logical_block_size` in the sysfs and we can easily read from the attribute value rather than issuing a command.

Please review :)

Thanks,

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -274,6 +274,21 @@ static bool is_blkdev(struct nvme_dev *dev)
 	return S_ISBLK(dev->direct.stat.st_mode);
 }
 
+static bool is_gendev(struct nvme_dev *dev)
+{
+	int instance, head_instance;
+	int ret;
+
+	if (!is_chardev(dev))
+		return false;
+
+	ret = sscanf(dev->name, "ng%dn%d", &instance, &head_instance);
+	if (ret != 2)
+		return false;
+
+	return true;
+}
+
 static int open_dev_direct(struct nvme_dev **devp, char *devstr, int flags)
 {
 	struct nvme_dev *dev;
@@ -7364,6 +7379,14 @@ static int submit_io(int opcode, char *command, const char *desc,
 			}
 			goto ret;
 		}
+	}
+
+	if (!(is_blkdev(dev) || is_gendev(dev))) {
+		fprintf(stderr, "Invalid device type, should be either "
+				"block device or generic device: (%s)\n",
+				dev->name);
+		err = -EINVAL;
+		goto close_dev;
 	}
 
 	if (!cfg.namespace_id) {
--- a/nvme.c
+++ b/nvme.c
@@ -7446,8 +7446,8 @@ static int submit_io(int opcode, char *command, const char *desc,
 		goto close_mfd;
 	}
 
-	if (nvme_get_logical_block_size(dev_fd(dev), cfg.namespace_id,
-					&logical_block_size) < 0)
+	if (nvme_get_logical_block_size_attr(dev->name,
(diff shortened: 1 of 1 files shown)


minwooim:
This PR should be done once https://github.com/linux-nvme/libnvme/pull/608 is accepted.

igaw:
> Someone might not want to issue any commands before a I/O command that they try to issue.

But there is also the other person who wants to read via a read command. If you want to read the attributes parsed by the kernel why not just use this info from the `sysfs`? Could you elaborate on your use case?

Not sure about this feature.

minwooim:
> But there is also the other person who wants to read via a read command. 

I guess in that case we would rather issue a `nvme id-ns` command before the I/O command.

> If you want to read the attributes parsed by the kernel why not just use this info from the sysfs? Could you elaborate on your use case?

It's retrieving the information from the sysfs in `libnvme` side which I've posted as a PR.  Am I missing something here ;) ?

igaw:
> It's retrieving the information from the sysfs in libnvme side which I've posted as a PR.

That is not explaining why it should read it from sysfs? What are you trying to solve? 

If you look at the current libnvme API we do not have any sysfs dependency except for the
fabric implementation and Linux specific helpers. This makes this part of the API usable
also for non-Linux operating systems.

minwooim:
> That is not explaining why it should read it from sysfs? What are you trying to solve?

In testing NVMe, some testcases are expecting that `nvme read` and `nvme write` is a single
command by retrieving that value by `BLKGETSZ` ioctl() for the blkdev.  I just tried to remove
the redundant IDENTIFY for the logical_block_size which already have been parsed by kernel
driver in the reset_work.

> If you look at the current libnvme API we do not have any sysfs dependency except for the
> fabric implementation and Linux specific helpers. This makes this part of the API usable
> also for non-Linux operating systems.

Now It makes sense to me. I didn't even consider the platform dependency of `libnvme`.

igaw:
> In testing NVMe, some testcases are expecting that nvme read and nvme write is a single command by retrieving that value by BLKGETSZ ioctl() for the blkdev.

Are these test case some official compliance test framework, or just some internal tests suites?

minwooim:
On March 31, 2023 10:48:28 PM GMT+09:00, Daniel Wagner ***@***.***> wrote:
>> In testing NVMe, some testcases are expecting that nvme read and nvme write is a single command by retrieving that value by BLKGETSZ ioctl() for the blkdev.
>
>Are these test case some official compliance test framework, or just some internal tests suites?

I meant some of intermal suites ;)

minwooim:
> > In testing NVMe, some testcases are expecting that nvme read and nvme write is a single command by retrieving that value by BLKGETSZ ioctl() for the blkdev.
> 
> Are these test case some official compliance test framework, or just some internal tests suites?

I mean some of the internal suites ;)

I don't want to mess up things just because it has a different behavior from the way it used to be, and if it's not for official tests, even more so. If some use cases are coming up about not issuing an IDENTIFY command in I/O commands, I would like to participate in the discussion, but before that, I would rather close this issue :)

Thanks for the kind review!

igaw:
No worries, I am just trying to understand what the problems are. So nothing really broken. That is nice :)

PR #1886, opened 2023-03-31
fabrics: fix fc config JSON file handling

martin-gpy:
Unlike other nvme transports, nvme/fc connection requires a valid host_traddr in addition to traddr and transport type. Current fc config JSON handling is broken due to the host_traddr not getting updated even if explicitly listed in the JSON file, as shown below:

nvme connect-all -J /usr/local/etc/nvme/config.json Failed to write to /dev/nvme-fabrics: Invalid argument

And the below error is logged in the messages file for the same:

nvme_fabrics: missing parameter 'host_traddr=%s'

Fix this by ensuring the relevant host_traddr string is appropriately passed to the respective nvme controller structure.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -630,7 +630,7 @@ static int discover_from_json_config_file(nvme_root_t r, nvme_host_t h,
 					  enum nvme_print_flags flags,
 					  bool force)
 {
-	const char *transport, *traddr, *trsvcid, *subsysnqn;
+	const char *transport, *traddr, *host_traddr, *trsvcid, *subsysnqn;
 	nvme_subsystem_t s;
 	nvme_ctrl_t c, cn;
 	struct nvme_fabrics_config cfg;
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -668,7 +675,7 @@ static int discover_from_json_config_file(nvme_root_t r, nvme_host_t h,
 				.subsysnqn	= subsysnqn,
 				.transport	= transport,
 				.traddr		= traddr,
-				.host_traddr	= cfg.host_traddr,
+				.host_traddr	= host_traddr,
```
igaw, line 678: What about host_iface? Shouldn't we take care of that one too?

martin-gpy: > What about host_iface? Shouldn't we take care of that one too?

host_iface is permitted for tcp alone, and not fc or even rdma for that matter. That's the reason why I didn't include it here. But this should be doable. How about we include this fix (for addressing fc for now), and I will send an incremental patch later to take care of host_iface too?

martin-gpy:
Closing this. Will submit a new git pull request for the new changes.

PR #1887, opened 2023-04-02
nvme: Add support for NVMe-MI receive command

ikegami-t:
This PR includes a patch to fix passthru command function indentation errors also.


igaw:
The only minor complain is the first patch should be two, one which fixes the stderr thingy and the second one the whitespace fixes. Just for next time :)

PR #1892, opened 2023-04-04
fabrics: fix fc config JSON file handling

martin-gpy:
Unlike other nvme transports, nvme/fc connection requires a valid host_traddr in addition to traddr and transport type. Current fc config JSON handling is broken due to the host_traddr not getting updated even if explicitly listed in the JSON file, as shown below:

nvme connect-all -J /usr/local/etc/nvme/config.json Failed to write to /dev/nvme-fabrics: Invalid argument

And the below error is logged in the messages file for the same:

nvme_fabrics: missing parameter 'host_traddr=%s'

Fix this by ensuring the relevant host_traddr string is appropriately passed to the respective nvme controller structure. And while we are at it, ensure the host_iface string is updated too but noting that it is only applicable to tcp alone, and not rdma or fc.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -630,7 +630,7 @@ static int discover_from_json_config_file(nvme_root_t r, nvme_host_t h,
 					  enum nvme_print_flags flags,
 					  bool force)
 {
-	const char *transport, *traddr, *trsvcid, *subsysnqn;
+	const char *transport, *traddr, *host_traddr, *host_iface, *trsvcid, *subsysnqn;
 	nvme_subsystem_t s;
 	nvme_ctrl_t c, cn;
 	struct nvme_fabrics_config cfg;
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -650,6 +652,18 @@ static int discover_from_json_config_file(nvme_root_t r, nvme_host_t h,
 			    strcmp(transport, "fc"))
 				continue;
 
+			/* ignore if no host_traddr for fc */
+			if (!strcmp(transport, "fc")) {
+				if (!host_traddr)
+					continue;
```
martin-belanger, line 658: Should we print an error message to `stderr` indicating that there is an error in the JSON file?

fabrics.c
```diff
@@ -650,6 +652,18 @@ static int discover_from_json_config_file(nvme_root_t r, nvme_host_t h,
...
+			}
+
+			/* ignore if host_iface set for any transport other than tcp */
+			if (!strcmp(transport, "rdma") || !strcmp(transport, "fc")) {
+				if (host_iface)
+					continue;
```
martin-belanger, line 664: Same as above.

martin-belanger (approved):

igaw:
Loosk good. Squashed the commits.

PR #1895, opened 2023-04-05
nvme-print: Use enum nvme_print_flags for nvme_show_id_ctrl

igaw:
gcc is complaining that the function declaration and definition do not match. Update the declaration accordingly.


minwooim:
Looks good to me.

PR #1900, opened 2023-04-12
nvme: Fix format command to skip to reread NSID all block device

ikegami-t:
Since if NSID all specified NS data is not identified so the reread failed.


igaw:
Great catch!

PR #1901, opened 2023-04-13
Change default fallback behavior

igaw:
meson's default setting for wrap mode is to attempt to download missing
dependencies. Disable this feature as the community is unhappy with
this default behavior.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,7 @@ project(
       'prefix=/usr/local',
       'warning_level=1',
       'sysconfdir=etc',
+      'wrap_mode=nofallback',
     ]
 )
 
--- a/meson.build
+++ b/meson.build
@@ -55,7 +55,7 @@ libnvme_mi_dep = dependency('libnvme-mi', required: true,
 if get_option('json-c').disabled()
     json_c_dep = dependency('', required: false)
 else
-    json_c_dep = dependency('json-c', required: true, version: '>=0.13',
+    json_c_dep = dependency('json-c', required: get_option('json-c'), version: '>=0.13',
                             fallback : ['json-c', 'json_c_dep'])
     if json_c_dep.version().version_compare('>=0.14')
         conf.set('CONFIG_JSONC_14', true, description: 'Is json-c at least 0.14?')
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -54,7 +55,7 @@ libnvme_mi_dep = dependency('libnvme-mi', required: true,
 if get_option('json-c').disabled()
     json_c_dep = dependency('', required: false)
 else
-    json_c_dep = dependency('json-c', required: true, version: '>=0.13',
+    json_c_dep = dependency('json-c', required: get_option('json-c'), version: '>=0.13',
                             fallback : ['json-c', 'json_c_dep'])
     if json_c_dep.version().version_compare('>=0.14')
```
eli-schwartz, line 60: If json-c is no longer mandatory then you should be checking that before inserting `requires = 'Requires: json-c >= 0.13'` in the case where it is not found.

eli-schwartz: Right, but what happens if it is the value "auto" and the library isn't present?

igaw: That idea is that should be treated like disabled.

Background some members of the community have expressed that they dislike all the newly introduced dependencies, that's why I am trying to cut them down. Obviously, the feature set is also cut down, not much I can do there.

eli-schwartz: > That idea is that should be treated like disabled.

This is not how meson treats anything at all. Instead, meson treats disabled like "not found".

And programmatically what you get is that `json_c_dep.found() == false` and `json_c_dep.version() == 'unknown'`. Your above check for `get_option('json-c').disabled()` doesn't match because it is possible to get "not found" without "disabled".

...

Personally I'd tend to think it makes more sense to not manually check this. It's too complicated and meson can do it builtin.

```meson
json_c_dep = dependency('json-c', required: get_option('json-c'), version: '>=0.13',
                        fallback : ['json-c', 'json_c_dep'])
if json_c_dep.found()
    if json_c_dep.version().version_compare('>=0.14')
    .........
```

Why do you need to open-code your own disabled handling?

igaw:
The problem I try to solve is that it's possible to forcefully disable a dependency even though it is available on the system. I know from reading various github's issue on this, that this is something you are not really happy about it. 

There are a couple of reasons why I went for this approach here. The most prominent problem is that meson picks up the host Python libraries in a cross build or it wants or it links statically against the Python libraries when  `LDFLAGS=-static`.

eli-schwartz:
It's not obvious to me why you think I am unhappy about the ability to forcefully disable a dependency. The only thing I'm concerned about is the fact that currently your build setup is adding "Requires: json -c" in the case where json-c has been disabled due to lacking availability. And that is only because I'm pretty sure it's not what you intended to do.

Issue #1902, opened 2023-04-13
nvme-cli version 1.9-1 on Ubuntu release focal returns invalid json format 


bunkar-akamai:
nvme-cli version 1.9-1 with Ubuntu focal release returns empty string as output when asked for JSON and no NVMEs are present on the system.

**Actual Scenario:**

command: nvme list -o json
output: ""  (Empty string)

**Expected Scenario:**
command: nvme list -o json
{

}

bunkar-akamai:
The possible fix for the issue could be as below:

cat patches/nvme-cli-fix.patch 
--- pkgsrc.orig/nvme.c	2023-03-16 10:46:31.157298636 +0100
+++ pkgsrc/nvme.c	2023-03-16 12:20:12.908420032 +0100
@@ -1998,7 +1998,7 @@
 		}
 	}
 
-	if (list_cnt) {
+	if (list_cnt >= 0) {
 		if (fmt == JSON)
 			json_print_list_items(list_items, list_cnt);
 		else

ikegami-t:
The PR #1906 exmaple output is below.
```sh
tokunori@tokunori:~/nvme-cli$ env LANGUAGE=en_US.UTF-8 sudo .build/nvme list --output-format=json
{
  "error":"Failed to scan topology: No such file or directory"
}
```

igaw:
Thanks for the report. @ikegami-t fixed this in upstream (thanks!) .

That said, we are not maintaining the downstream version of nvme-cli in Ubuntu, thus we can't fix it. You need to open a bug report against Ubuntu and the downstream maintainer has to apply a fix.

Anyway, if you want to test it, we build an `AppImage` from the latest version: https://monom.org/linux-nvme/upload/AppImage/nvme-cli-latest-x86_64.AppImage

bunkar-akamai:
> Thanks for the report. @ikegami-t fixed this in upstream (thanks!) .
> 
> That said, we are not maintaining the downstream version of nvme-cli in Ubuntu, thus we can't fix it. You need to open a bug report against Ubuntu and the downstream maintainer has to apply a fix.
> 
> Anyway, if you want to test it, we build an `AppImage` from the latest version: https://monom.org/linux-nvme/upload/AppImage/nvme-cli-latest-x86_64.AppImage

Thanks for the update @igaw

PR #1906, opened 2023-04-14
Json output error

ikegami-t:
Changes for the issue #1902.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -9,6 +9,8 @@
 #include "nvme.h"
 #include "common.h"
 
+#define ERROR_MSG_LEN 100
+
 static const uint8_t zero_uuid[16] = { 0 };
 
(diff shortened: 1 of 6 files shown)


nvme-print-json.c
```diff
@@ -2854,3 +2856,27 @@ void json_output_status(int status)
...
+	if (!error)
+		return;
+
+	len = vsnprintf(error, len, msg, ap) + 1;
+	if (len > ERROR_MSG_LEN) {
+		error = (char *)realloc(error, len);
+		if (!error)
+			return;
+		vsnprintf(error, len, msg, ap);
+	}
```
igaw, line 2875: We are already using `asprintf` in the code base. So this could just use it as well. Avoids the open coded realloc etc.

ikegami-t:
Thanks for your reviewing and comments. Just created the PR #1911 to use `vasprintf` so review it again.

Issue #1905, opened 2023-04-14
fabrics.c total: 6 errors, 19 warnings, 1468 lines checked


johnmeneghini:
```
jmeneghi:nvme-cli(master) > ~/data/repos/CENTOS9/centos-stream-9/scripts/checkpatch.pl --no-tree fabrics.c
WARNING: Improper SPDX comment style for 'fabrics.c', please use '//' instead
#1: FILE: fabrics.c:1:
+/* SPDX-License-Identifier: GPL-2.0-only */

WARNING: Missing or malformed SPDX-License-Identifier tag in line 1
#1: FILE: fabrics.c:1:
+/* SPDX-License-Identifier: GPL-2.0-only */

ERROR: Macros with complex values should be enclosed in parentheses
#87: FILE: fabrics.c:87:
+#define NVMF_OPTS(c)                                                                   \
+       OPT_STRING("transport",       't', "STR", &transport,   nvmf_tport), \
+       OPT_STRING("nqn",             'n', "STR", &subsysnqn,   nvmf_nqn), \
+       OPT_STRING("traddr",          'a', "STR", &traddr,      nvmf_traddr), \
+       OPT_STRING("trsvcid",         's', "STR", &trsvcid,     nvmf_trsvcid), \
+       OPT_STRING("host-traddr",     'w', "STR", &c.host_traddr,       nvmf_htraddr), \
+       OPT_STRING("host-iface",      'f', "STR", &c.host_iface,        nvmf_hiface), \
+       OPT_STRING("hostnqn",         'q', "STR", &hostnqn,     nvmf_hostnqn), \
+       OPT_STRING("hostid",          'I', "STR", &hostid,      nvmf_hostid), \
+       OPT_STRING("dhchap-secret",   'S', "STR", &hostkey,     nvmf_hostkey), \
+       OPT_INT("nr-io-queues",       'i', &c.nr_io_queues,       nvmf_nr_io_queues),   \
+       OPT_INT("nr-write-queues",    'W', &c.nr_write_queues,    nvmf_nr_write_queues),\
+       OPT_INT("nr-poll-queues",     'P', &c.nr_poll_queues,     nvmf_nr_poll_queues), \
+       OPT_INT("queue-size",         'Q', &c.queue_size,         nvmf_queue_size),     \
+       OPT_INT("keep-alive-tmo",     'k', &c.keep_alive_tmo,     nvmf_keep_alive_tmo), \
[...]

PR #1907, opened 2023-04-15
fabrics: Fix linux kernel check patch errors and warnings

ikegami-t:
Note: Still remaining some errors warnings.


ikegami-t:
Changes for the issue #1905 but still remaining some check patch error and warinings.

PR #1908, opened 2023-04-17
nvme: fix io_mgmt args parse error

sc108-lee:
Since mos defined __u16, it should be parsed with OPT_SHRT if mos given right after mo, it cause parse error
mo always set as 0

Reported-by: Youngjin Jung <yj4369.jung@samsung.com>

### DIFF
--- a/subprojects/libnvme.wrap
+++ b/subprojects/libnvme.wrap
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = c90eeff5f5bfab597a6152d979aa7049524dd79f
+revision = ef27cf88b1aeed3ba778740f5f71dbd98a79073d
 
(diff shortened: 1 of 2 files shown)


subprojects/libnvme.wrap
```diff
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = c90eeff5f5bfab597a6152d979aa7049524dd79f
+revision = b8ea8bee138af7a336874b47c160fed91c2b1969
```
igaw, line 3: I've applied the MOS/MO fixes, the reference has changed. Could you please update this accordingly? Thanks!

sc108-lee: @igaw  Thanks for the fast response, rebased and modified with merged hash :)

PR #1913, opened 2023-04-18
fabrics: Refactor NVMF_OPTS macro approach into functions approach

ikegami-t:


util/argconfig.c
```diff
@@ -619,3 +619,49 @@ bool argconfig_parse_seen(struct argconfig_commandline_options *s,
...
+		return NULL;
+	}
+
+	memcpy(opts, opts1, sizeof(*opts) * num1);
+	memcpy(&opts[num1], opts2, sizeof(*opts) * num2);
+
+	return opts;
+}
+
+int argconfig_parse_merge(int argc, char *argv[], const char *program_desc,
```
igaw, line 649: I would keep the name `argconfig_parse()` and rename the function above to something like `option_merge()` or so. This avoids that all the callsides get a new function name. The option merge aspect is just an internal thing.

igaw: Ah I see we still have `argconfig_parse()` users left. Maybe in this case just introduce a `argconfig_nvmf_parse()` function which adds the default options automatically?

igaw: And now I see why you left the define :) It's because it depends on local variables of the calling context. IMO, this is a bit ugly and I think it would be better to use a `struct nvmf_args` and hand it into the parser.

igaw (changes requested):

ikegami-t:
Thanks for your reviewing and comments. Just refixed the patch as suggested so review it again.

fabrics.c
```diff
@@ -86,31 +86,58 @@ static const char *nvmf_data_digest	= "enable transport protocol data digest (TC
...
+	char **transport;
+	char **subsysnqn;
+	char **traddr;
+	char **trsvcid;
+	char **hostnqn;
+	char **hostid;
+	char **hostkey;
+	struct nvme_fabrics_config *c;
+	struct argconfig_commandline_options *opts;
+};
```
igaw, line 127: My idea was to have this structure embedded into each command.

Something like

```c
struct nvmf_args {
	char *transport;
	char *subsysnqn;
	char *traddr;
	char *trsvcid;
	char *hostnqn;
	char *hostid;
	char *hostkey;
	struct nvme_fabrics_config *c;
	struct argconfig_commandline_options *opts;
};
```

fabrics.c
```diff
@@ -86,31 +86,58 @@ static const char *nvmf_data_digest	= "enable transport protocol data digest (TC
...
+		.transport = &transport, \
+		.subsysnqn = &subsysnqn, \
+		.traddr = &traddr, \
+		.trsvcid = &trsvcid, \
+		.hostnqn = &hostnqn, \
+		.hostid = &hostid, \
+		.hostkey = &hostkey, \
+		.c = &cfg, \
+		.opts = opts, \
+	}
```
igaw, line 140: ... and this wouldn't be necessary. So there wouldn't be any local open coded variables. Only the `args` struct. Obviously all commands need to be updated where we currently do 

` nvme_foo(transport, subsysysnqn) -> nvme_foo(args.transport, args.subsysnqn)`

igaw: Hmm, but I don't think this makes actually sense. As we always have some commends with different arguments.

So your patch seems reasonable. Maybe we could also introduce a macro which defines all these common variables like hostnqn etc which are expected by NVMF_ARGS.

igaw: But given that the goal here is just to get rid of the checkpatch warning, I find it a bit too much code churn. Would the first part of this patch be enough? I mean the change to

`#define NVME_OPTS(args)` ->  `#define NVMF_OPTS(args) {`

ikegami-t: No only the first part below causes a build error.
```c
#define NVMF_OPTS(args) {                                                                       \
	OPT_STRING("transport",       't', "STR", args->transport,       nvmf_tport),           \
...
	OPT_FLAG("tls",                 0, &args->c->tls,                nvmf_tls),             \
}
...
static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
				   const char *desc, bool connect,
				   const struct nvme_fabrics_config *defcfg)
{
...
	OPT_ARGS(opts) = {
		NVMF_OPTS(cfg),
		OPT_FMT("output-format", 'o', &format,        output_format),
...
		OPT_FLAG("force",          0, &force,      "Force persistent discovery controller creation"),
		OPT_END()
	};
```
Since the macro is expanded as below but the `{` and `}` expanded are not able to be used in side of the OPT_ARGS `{` and `}`.
```c
static int discover_from_conf_file(nvme_root_t r, nvme_host_t h,
				   const char *desc, bool connect,
				   const struct nvme_fabrics_config *defcfg)
{
...
	OPT_ARGS(opts) = {
		{
		OPT_STRING("transport",       't', "STR", args->transport,       nvmf_tport),
...
		OPT_FLAG("tls",                 0, &args->c->tls,                nvmf_tls),
		}
		OPT_FMT("output-format", 'o', &format,        output_format),
...
		OPT_FLAG("force",          0, &force,      "Force persistent discovery controller creation"),
		OPT_END()
	};

...

ikegami-t:
Attached the linux kernel checkpatch.pl tool result of nvme.c [nvme-linux-kernel-check-patch.txt](https://github.com/linux-nvme/nvme-cli/files/11304518/nvme-linux-kernel-check-patch.txt).

ikegami-t:
Thanks for your review comments. I have removed the first patch and reverted the overlong line strings changes. By the way seems the command description should be shorten basically since there is the commnd documentaton also so it can be described the detail by it.

igaw (approved):

igaw:
Thanks!

PR #1924, opened 2023-05-02
util: Delete argconfig value length to calcurate instead

ikegami-t:


ikegami-t:
Thanks for merging. Sorry I missed to delete the macro `OPT_VAL` and `CFG_VAL` as now unnecessary definitions so will do fix later it.

PR #1932, opened 2023-05-06
Zsh command completions

ikegami-t:


igaw:
Nice! This was on my TODO list for a while. Thanks!

PR #1936, opened 2023-05-13
util: Add argconfig debug option to show command output

ikegami-t:


ikegami-t:
The changes depended on the PR linux-nvme/libnvme#646.

igaw:
There seems to be a declaration or import issue: `../nvme-wrap.c:429:24: error: implicit declaration of function ‘nvme_debug’; did you mean ‘mc_debug’? `

ikegami-t:
The build error can be resolved by the libnvme PR https://github.com/linux-nvme/libnvme/pull/646 changes. Also updated the changes since the libnvme PR fixed.

ikegami-t:
Rebased and split the patch changes. Also changed for some parts of refactoring.

igaw:
> The build error can be resolved by the libnvme PR [linux-nvme/libnvme#646](https://github.com/linux-nvme/libnvme/pull/646) changes. Also updated the changes since the libnvme PR fixed.

D'oh. Right, so let's get the libnvme part added first.

ikegami-t:
Just rebased.

igaw:
Sorry for the long wait for a review. The first two patches look good.

I am not convinced that the third patch is the right approach though. We already have the `--verbose` flag for some of the commands. I think it would be better to add the `--verbose` option to all commands which are missing it and make the `--verbose` flag also enable the debug flag of libnvme.

ikegami-t:
Thanks for your review comments. Changed to add `--verbose` option as mentioned but still changing argconfig for all commands.

util/argconfig.c
```diff
@@ -389,11 +402,15 @@ int argconfig_parse(int argc, char *argv[], const char *program_desc,
 	long_opts[option_index++].val = 'h';
 
 	long_opts[option_index].name = "json";
-	long_opts[option_index].val = 'j';
+	long_opts[option_index++].val = 'j';
```
igaw, line 405: This looks odd. The output format is handled via the `-o` `--output` options, which not all command currently support. So having this here doens't look right. We should remove it.

util/argconfig.c
```diff
@@ -389,11 +402,15 @@ int argconfig_parse(int argc, char *argv[], const char *program_desc,
...
+
+	long_opts[option_index].name = "verbose";
+	long_opts[option_index].val = 'v';
```
igaw, line 408: The commands need to support this. Just adding to the parser is not enough. Every single command needs to be updated so that also the help/usage output matches. Maybe there is way to have some default options.

ikegami-t: Fixed this as the default option mentioned.

fabrics.c
```diff
@@ -88,7 +88,7 @@ static const char *nvmf_config_file	= "Use specified JSON configuration file or
 static const char *nvmf_context		= "execution context identification string";
 
 #define NVMF_ARGS(n, c, ...)                                                                     \
-	struct argconfig_commandline_options opts[] = {                                          \
+	struct argconfig_commandline_options n[] = {                                             \
```
igaw, line 91: Ooops :)

util/argconfig.c
igaw: Cool! In the long run I would like to see getting the `-o/--output-format` option supported by all commands as well. So this is going into the right direction. Thanks!

igaw:
Thanks!

PR #1940, opened 2023-05-17
nvme: Fix io-passthru command read metadata output

ikegami-t:


soojung56:
I checked the changed contents. Thanks you.

PR #1943, opened 2023-05-19
fabrics: Change NVMF_OPTS as NVMF_ARGS to fix linux kernel check patc…

ikegami-t:
…h error


ikegami-t:
Changes for the issue #1905 to retry to fix the remaing error.

ikegami-t:
Thank you!

PR #1946, opened 2023-05-21
Nvme show error

ikegami-t:


nvme-print.c
```diff
@@ -5462,8 +5462,10 @@ void nvme_show_error(const char *msg, ...)
 	va_list ap;
 	va_start(ap, msg);
 
-	if (argconfig_output_format_json(false))
-		return json_output_error(msg, ap);
+	if (argconfig_output_format_json(false)) {
+		json_output_error(msg, ap);
+		return va_end(ap);
```
igaw, line 5470: `return va_end()` for a void function doesn't really make sense.

ikegami-t: Thansks for your review comments. To make sure let confirm below.
The function to return is defined as the void function as below so seems it correct but still should it be separated the function call and retrun as mentioned?
```c
void va_end(va_list ap);
```

igaw: Yes, the idea is to use plain `return;` if the function is defined as 'void', so don't return void. 

Obviously, compilers are able to handle this but my eyes are trained that I always check if the return types match. A plain return says, we do not return anything. It's a readability thing in short.

ikegami-t: Thanks for your explanation. I see so just fixed the patch as mentioned.

PR #1949, opened 2023-05-23
Revert "nvme: Masks SSTAT in sanize-log output"

maurizio-lombardi:
The NVME_SANITIZE_SSTAT_STATUS_MASK macro is used to mask the 3 least significant bits (representing the status of the most recent sanitize operation) of the SSTAT field.

The SSTAT field is 16 bits wide and contains other information too, such as the Global Data Erased bit and the number of completed passes.

Revert this commit so nvme-cli will print the entire SSTAT field

See NVME Base Specification, page 236
https://nvmexpress.org/wp-content/uploads/NVM-Express-Base-Specification-2_0-2021.06.02-Ratified-5.pdf

This reverts commit 7092ff55c5d9017231a68fa4dbb9a37b42d61d61.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -4423,7 +4423,7 @@ void nvme_show_sanitize_log(struct nvme_sanitize_log_page *sanitize,
 		printf("\n");
 
 	printf("Sanitize Status                        (SSTAT) :  %#x\n",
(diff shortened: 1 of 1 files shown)


igaw:
D'oh. Can't remember what I thought. Anyway, thanks!

PR #1953, opened 2023-05-24
fabrics: fix conditions in nvmf-autoconnect.service

mwilck:
The unit should be started if either config.json or discovery.conf exists. The current syntax would start it if either one does not exist. Fix it.

Fixes: eec7634 ("fabrics: Trigger auto connect if config.json exists")


igaw:
Fixes https://github.com/linux-nvme/nvme-cli/pull/1953

PR #1959, opened 2023-05-28
plugins/ocp: Fix ocp-nvme.c linux kernel check patch errors

ikegami-t:


common.h
```diff
@@ -11,6 +11,11 @@
 #define min(x, y) ((x) > (y) ? (y) : (x))
 #define max(x, y) ((x) > (y) ? (x) : (y))
 
+#ifdef __packed
+#else /* __packed */
+#define __packed __attribute__((__packed__))
+#endif /* __packed */
```
igaw, line 17: Good idea! I like this!

plugin.c
```diff
@@ -14,13 +14,13 @@ static int version(struct plugin *plugin)
 	struct program *prog = plugin->parent;
 
 	if (plugin->name) {
-		printf("%s %s version %s (git %s)\n",
```
igaw, line 17: Hmm, I think this is ment to be the literal meaning 'version' not the function called version. So I don't think changing this part makes sense.

plugin.c
```diff
@@ -60,18 +60,19 @@ void usage(struct plugin *plugin)
 	struct program *prog = plugin->parent;
 
 	if (plugin->name)
-		printf("usage: %s %s %s\n", prog->name, plugin->name, prog->usage);
```
igaw, line 63: Same here, it is really the literal string `usage` not the function.

plugins/ocp/ocp-nvme.c
```diff
@@ -983,7 +982,7 @@ static int extract_dump_get_log(struct nvme_dev *dev, char *featurename, char *f
 		}
 
 		if (i != total_loop_cnt - 1) {
-			if (i == 0) {
+			if (!i) {
```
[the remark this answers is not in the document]
igaw: On the other hand we often do `if (!ret) {...}`. Well, let's put it this way, if the whole file is using the same pattern, then go for this. Just try to keep it consistent :)

plugin.c
igaw: let me fix this myself.

plugin.c
igaw: let me fix this myself.

ikegami-t: Thanks for your comments for the tool warnings below let us change the function namings as `version`/`usage` to `version_cmd`/`usage_cmd` by the PR #1966.
```sh
tokunori@tokunori-desktop:~/nvme-cli$ git reset --hard linux-nvme/master
HEAD is now at 9077d042 fabrics: prevent 'nvme connect' to well-known discovery NQN
tokunori@tokunori-desktop:~/nvme-cli$ ~/centos-stream-9/scripts/checkpatch.pl --no-tree plugin.c
WARNING: Prefer using '"%s...", __func__' to using 'version', this function's name, in a string
#17: FILE: plugin.c:17:
+               printf("%s %s version %s (git %s)\n",

WARNING: Prefer using '"%s...", __func__' to using 'version', this function's name, in a string
#20: FILE: plugin.c:20:
+               printf("%s version %s (git %s)\n",

WARNING: Prefer using '"%s...", __func__' to using 'version', this function's name, in a string
#23: FILE: plugin.c:23:
+       printf("libnvme version %s (git %s)\n",

WARNING: Prefer using '"%s...", __func__' to using 'usage', this function's name, in a string
#63: FILE: plugin.c:63:
+               printf("usage: %s %s %s\n", prog->name, plugin->name, prog->usage);

WARNING: Prefer using '"%s...", __func__' to using 'usage', this function's name, in a string
#65: FILE: plugin.c:65:
+               printf("usage: %s %s\n", prog->name, prog->usage);

total: 0 errors, 5 warnings, 218 lines checked

NOTE: For some of the reported defects, checkpatch may be able to
      mechanically convert to the typical style using --fix or --fix-inplace.

plugin.c has style problems, please review.

NOTE: If any of the errors are false positives, please report
      them to the maintainer, see CHECKPATCH in MAINTAINERS.
```

igaw: Alright, don't mind renaming these functions.

PR #1961, opened 2023-05-30
plugins/innogrit vendor command get-eventlog and get-cdump

innogritkenny:
cdw2 add signature
nsid modify to 0xffffffff
for support fw checking rule.

### DIFF
--- a/plugins/innogrit/innogrit-nvme.c
+++ b/plugins/innogrit/innogrit-nvme.c
@@ -146,11 +146,13 @@ static int nvme_vucmd(int fd, unsigned char opcode, unsigned int cdw12,
 
 	memset(&cmd, 0, sizeof(cmd));
 	cmd.opcode = opcode;
+	cmd.cdw2  = 0x69677673;
+	cmd.cdw10 = data_len / 4;
 	cmd.cdw12 = cdw12;
 	cmd.cdw13 = cdw13;
 	cmd.cdw14 = cdw14;
 	cmd.cdw15 = cdw15;
-	cmd.nsid = 0;
+	cmd.nsid = 0xffffffff;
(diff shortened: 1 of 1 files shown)


plugins/innogrit/innogrit-nvme.c
```diff
@@ -146,11 +146,13 @@ static int nvme_vucmd(int fd, unsigned char opcode, unsigned int cdw12,
 
 	memset(&cmd, 0, sizeof(cmd));
 	cmd.opcode = opcode;
+	cmd.cdw2  = 0x69677673;
```
igaw, line 149: Usually, it's a good practice to use a define for a magic number.

innogritkenny: I have modify to add a define for magic number

igaw:
I've fixed up this PR again and applied it directly.

@innogritkenny just to let you know, I will not do cleanups or refactoring for new PRs from you in future.

I'm fine with doing reviewing and fixing up small things once in a while. That's not the issue. Though I expect that a PR is in reasonable shape. And here is the problem. So far many PRs from you were not ready to be merged. There is always something completely wrong. Commit messages, merges in the PR, coding style ignored (e.g. spaces vs tabs), on top fixes instead in patch fixes.

Every time I have to spend extra time. I wouldn't mind doing this once or twice if the following PRs improve. But here it isn't the case.

There is plenty of documentation on how to do it and if that doesn't help there are also Opens Source consultants which can help you to get up to speed. And if our documentation needs improvement I am willing to spend the necessary time to update it so that I don't have to explain it every single time what the expectation are on code submissions.

It's just not fair to everyone else if I always spend extra time on your PRs.

PR #1962, opened 2023-05-31
Refactor print code

igaw:
Continuing with the code split for the different printing backends. This will allow to make use of the upcoming auto generated printf code from the  2.1 spec which will ship a JSON file which describes all fields.

This is a first rough version which shows my ideas how to abstract the printing code.

One thing I do not like in this version is the passing of the `enum nvme_print_flags` to all functions. This makes the API overloaded and a bit difficult to read. One solution to this problem could that I pass in the flags and any additional parameters to the `nvme_get_[stdout|json]_print_ops()` functions and store it locally, like a static member variable in C++. I don't want to introduce `this` pointer for this.

Thoughs?


igaw:
Did some more cleanups
 - minimizing unnecessary white space changes 
 - dropped flags argument passing
 - moved changes into the right patch

nvme-print.c
```diff
@@ -267,6 +267,18 @@ static void nvme_show_persistent_event_entry_ehai(__u8 ehai)
 		"Event not associated with any port and PELPID does not apply");
 }
 
+static void add_bitmap(int i, __u8 seb)
+{
+        for (int bit = 0; bit < 8; bit++) {
+                if (nvme_pel_event_to_string(bit + i * 8)) {
+			if (nvme_pel_event_to_string(bit + i * 8))
```
ikegami-t: Duplicated the `if` condition checking line.

nvme-print.c
```diff
@@ -125,6 +132,16 @@ void nvme_show_predictable_latency_per_nvmset(
 	__u16 nvmset_id, const char *devname,
 	enum nvme_print_flags flags)
 {
+	struct print_ops *ops = nvme_print_ops(flags);
+
+	if (ops) {
```
ikegami-t, line 139: Seems the statement implementations are almost same so can be defined as a macro as below for example.
```c
#define nvme_print(name, flags, ...) \
    do { \
        struct print_ops *ops = nvme_print_ops(flags); \
        if (ops) { \
            if (ops->name) \
                ops->name(__VA_ARGS__); \
            return; \
    } while (0);
...
void nvme_show_predictable_latency_per_nvmset(
...
{
    nvme_print(predictable_latency_per_nvmset, flags, plpns_log, nvmset_id, devname);
```

nvme-print.c
```diff
@@ -35,6 +32,8 @@ static struct print_ops *nvme_print_ops(enum nvme_print_flags flags)
 
 	if (flags & JSON)
 		ops = nvme_get_json_print_ops(flags);
+	if (!(flags & (JSON | BINARY)))
```
ikegami-t, line 35: By changing the `if` to `else if` the `JSON` flag condition checking seems to be delete as below.
```c
    else if (!(flags & BINARY))
```

nvme-print.c
igaw: I see, though this is not code I am adding, just moving around. Let's fix this up on top of this series. I'd like to avoid changing too many things in one go.

nvme-print.c
igaw: Sounds like a good idea.

igaw:
Obviously, the binary output code should also be updated.

igaw:
changes:
- introduce `nvme_print` macro to avoid boiler plate code repetition
- dropped `nvme-print-[json|stdout].h and merged the content into `nvme-print.h`
- refactored binary code also to use the newly print_ops infrastructure

igaw:
There are a lot more cleanups possible on top of this series, e.g.
- drop unnecessary arguments
- cleanup all calling places of `nvme_show_` (maybe replace them directly with the new macro?)
- figure out if we can improve the situation the special cases where we do a `nvme_get_print_ops(0)`
- add missing implementations
- maybe more cleanup in the `nvme-print-stdout.c` file
- ...

Though I think we can go with this first step and do the rest of the cleanup later. As usual if someone wants to help out, highly appreciated :)

ikegami-t:
Looks good. Thank you.

igaw:
Thanks @ikegami-t for reviewing. 

So let's get some more cleanups done on this code. I think the most important one is to review the arguments we pass into the functions. Ideally we should avoid passing any non nvme type arguments , so only those defined in libnvme's types.h. 

Though I haven't seen the auto generated print code yet based on the nvme 2.1 spec but I expect it will only know 'nvme types' and not any made up nvme-cli types. Just as background info.

PR #1966, opened 2023-06-02
plugin: Change version and usage function names for linux kernel chec…

ikegami-t:
…k patch warnings

Also change usage function as a static function since not used externally.


igaw:
Thanks!

PR #1968, opened 2023-06-02
plugins/wdc: Refactor wdc-nvme.c for linux kernel check patch errors …

ikegami-t:
…and warnings


ikegami-t:
Now all plugins source files are checked by the linux kernel check patch tool and fixed the errors and warinings. Some refactoring changes for the fix included but it does not change the command behavior I think as long as I checked.

common.h
```diff
@@ -16,6 +16,12 @@
 #define __packed __attribute__((__packed__))
 #endif /* __packed */
 
+#if __has_attribute(__fallthrough__)
```
martin-belanger, line 19: We may want to do the same thing as `libnvme` and have `fallthrough` configured be meson and defined in `config.h`.

Ref: https://github.com/linux-nvme/libnvme/blob/master/meson.build#L234

ikegami-t: Thanks for your reviewing and comments. Just fixed the patch as mentioned and rebased the changes with the latest master branch.

igaw:
I've just splitted the first patch, so that the `fallthtrough` feature check is a separate patch.

igaw:
Thanks a lot. This was on my TODO list and never found time to do it.

PR #1969, opened 2023-06-05
nit: ocp: rename FW activation history command

arthurshau:
Get rid of the "vs-" prefix in the fw-activate-history command to bring it in line with the other command names


igaw:
Could you also update the bash/zsh tab completion too? 

There doesn't seem to be a man page on this command, so maybe also add one? Thanks!

igaw:
updated tab completion

Issue #1967, opened 2023-06-02
When using 'nvme read', output form is different from data and metadata.


soojung56:
When using 'nvme read', output form is different from data and metadata.
why data output size is 'cfg.data_size'?

cfg.block_count is considered when setting the buffer_size.
but when it is output as the current operation, 'cfg.block_count'  for data is not reflected.

In addition metadata setting output size is mbuffer_size.

- data
`write(dfd, (void *)buffer, cfg.data_size)`
- metadata
` write(mfd, (void *)mbuffer, mbuffer_size)`

**data/metadata setting code in submit_io, nvme.c +7547**
```
buffer_size = ((long long)cfg.block_count + 1) * logical_block_size;
if (cfg.data_size < buffer_size)
    nvme_show_error("Rounding data size to fit block count (%lld bytes)", buffer_size);
else
    buffer_size = cfg.data_size;

/* Get the required block count. Note this is a zeroes based value. */
nblocks = ((buffer_size + (logical_block_size - 1)) / logical_block_size) - 1;

/* Update the data size based on the required block count */
buffer_size = (nblocks + 1) * logical_block_size;

buffer = nvme_alloc(buffer_size, &huge);

...

mbuffer_size = ((unsigned long long)cfg.block_count + 1) * ms;
if (ms && cfg.metadata_size < mbuffer_size)
    nvme_show_error("Rounding metadata size to fit block count (%lld bytes)",
            mbuffer_size);
else
    mbuffer_size = cfg.metadata_size;

mbuffer = malloc(mbuffer_size);
```

**data/metadata output code in submit_io, nvme.c +7668**
```
if (err < 0) {
	nvme_show_error("submit-io: %s", nvme_strerror(errno));
} else if (err) {
	nvme_show_status(err);
} else {
	if (!(opcode & 1) && write(dfd, (void *)buffer, cfg.data_size) < 0) {
		nvme_show_error("write: %s: failed to write buffer to output file",
			strerror(errno));
		err = -EINVAL;
	} else if (!(opcode & 1) && cfg.metadata_size &&
		   write(mfd, (void *)mbuffer, mbuffer_size) < 0) {
		nvme_show_error(
		    "write: %s: failed to write meta-data buffer to output file",
		    strerror(errno));
		err = -EINVAL;
	} else {
		printf("%s: Success\n", command);
	}
}
```

igaw:
[...]

PR #1973, opened 2023-06-06
nvme: Use correct data buffer size

igaw:
The submit_io command is using the buffer size provided by the user. But this size argument might be updated and the correct length for the buffer is buffer_size. Thus use this value and not the initial length.

Fixes: #1967


ikegami-t:
Looks good. Thank you.

PR #1978, opened 2023-06-14
Add support for execution context filtering

igaw:
fabrics: Add support for volatile configuration

    The fabric resources are global resources and as such should only
    modified by one component. So far nvme-cli was the only component
    managing connections. With the introduction of nvme-stas we need to
    coordinate the access to these resources.

    Introduce access coordination feature which allows the external softare
    to drop volatie configurations json files into the /var/nvme directory.
    When nvme-cli is execututed (e.g. via the udev rules), the execution
    context is evaluated and only resources which belong the application are
    considered.

    TODO: implement context checking for disconnect(-all)


igaw:
@martin-belanger this might of interest for you :)

igaw:
I am testing this with blktests and using this configuration:

```json
[
  {
    "hostnqn": "nqn.2014-08.org.nvmexpress:uuid:242d4a24-2484-4a80-8234-d0169409c5e8",
    "hostid": "242d4a24-2484-4a80-8234-d0169409c5e8",
    "subsystems": [
      {
        "application": "blktests",
        "nqn": "blktests-subsystem-1",
        "ports": [
          {
            "transport": "fc",
            "traddr": "nn-0x10001100aa000001:pn-0x20001100aa000001",
            "host_traddr": "nn-0x10001100aa000002:pn-0x20001100aa000002"
          }
        ]
      }
    ]
  }
]

```

fabrics.c
```diff
@@ -47,6 +47,7 @@
 
 #define PATH_NVMF_DISC		SYSCONFDIR "/nvme/discovery.conf"
 #define PATH_NVMF_CONFIG	SYSCONFDIR "/nvme/config.json"
+#define PATH_NVMF_RUNDIR	RUNDIR "/nvme"
```
martin-belanger, line 50: Should `PATH_NVMF_RUNDIR` be set to `/run/nvme`. I believe that since FHS 3.0 the use of `/run` is encouraged for volatile run time data (but I may be wrong...).

Ref: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

igaw: `RUNDIR` is set to `run` as default, thus it will looking under `${PREFIX}/run/nvme`

martin-belanger: > BTW, I using `/etc/tmpfiles.d/nvme-cli.conf ` to create the directory:
> 
> `d /run/nvme 0755 - - -`

Good idea!

martin-belanger: I assume that each application will be in charge of writing their configuration to this directory using a unique file name (e.g. `/run/nvme/nvme-stas.json`). This ensures that there won't be collisions between two apps trying to write to the same file at the same time.

igaw:
BTW, just discussed this PR offline with @hreinecke and we decided we concentrate on the 'nvme connect' use case and leave the 'nvme disconnect(-all)' for later. The current udev rules are only doing the connect part. Reason being that the disconnect patch is completely different to the connect call. 

We need to do some more cleanups in the disconnect path first, e.g. do not issue any additional commands except the disconnect. This avoids addressed the issue when the controller is not working correctly anymore.

Documentation/nvme-connect-all.txt
```diff
@@ -38,6 +38,7 @@ SYNOPSIS
                 [--nbft]
                 [--no-nbft]
                 [--nbft-path=<STR>]
+                [--application=<id>]
```
martin-belanger, line 41: I'm wondering if we should define this as `--application=`**`<STR>`** to make it clear that a string is expected.

igaw: yes, that make sense.

I am wondering if we should call it `--context` or something similar. `--application` seems a bit wierd. e.g.

```systemd
[Unit]
Description=Connect NVMe-oF subsystems automatically during boot
ConditionPathExists=|@SYSCONFDIR@/nvme/config.json
ConditionPathExists=|@SYSCONFDIR@/nvme/discovery.conf
After=network-online.target
Before=remote-fs-pre.target

[Service]
Type=oneshot
ExecStartPre=/sbin/modprobe nvme-fabrics
ExecStart=@SBINDIR@/nvme connect-all --application=autoboot

[Install]
WantedBy=default.target
```

martin-belanger: >
> I am wondering if we should call it `--context` or something similar. `--application` seems a bit wierd. e.g.

Yes. I think `context` makes more sense from a user's point of view.

igaw:
I've tested with blktests and it works as designed, yay!

See also https://github.com/osandov/blktests/pull/121

PR #1979, opened 2023-06-14
nvme-print: Print PCI physical slot number with -v switch

usaleem-ix:
https://github.com/linux-nvme/libnvme/pull/657 adds PCI physical slot information to the controller.

This PR prints the Slot number in detailed/verbose output on stdout. JSON object is also updated to include the Slot number in detailed list.


usaleem-ix:
> Add as first a patch which updates subproject/libnvme.wrap to the latest libnvme ref to fix the build. Thanks!

Updated.

PR #1980, opened 2023-06-14
plugins/solidigm: Added Identify controller VU extensions parsing.

lgdacunh:


plugins/solidigm/solidigm-id-ctrl.c
```diff
@@ -0,0 +1,75 @@
...
+	uint8_t	rsvd1[3];
+	uint8_t ss;
+	char health[20];
+	uint8_t cls;
+	uint8_t nlw;
+	uint8_t scap;
+	uint8_t sstat;
+	char bl[8];
+	uint8_t rsvd2[38];
+	uint64_t ww; /* little endian */
```
igaw, line 23: Couldn't you use `__le64` instead here?

plugins/solidigm/solidigm-id-ctrl.c
```diff
@@ -0,0 +1,75 @@
...
+ * Copyright (c) 2023 Solidigm.
+ *
+ * Author: leonardo.da.cunha@solidigm.com
+ */
+
+#include <inttypes.h>
+#include "common.h"
+#include "solidigm-id-ctrl.h"
+
+#pragma pack(push, vu_id_ctrl, 1)
```
igaw, line 12: Why not using the attribute `__packed` ?

plugins/solidigm/solidigm-id-ctrl.c
lgdacunh: Done!

plugins/solidigm/solidigm-id-ctrl.c
lgdacunh: Done!

PR #1987, opened 2023-06-24
nbft: Refactor use of nvme_msg in nbft.c

ikegami-t:


ikegami-t:
Changes for the linux-nvme/libnvme#619 and also depended on the changes linux-nvme/libnvme#669.

ikegami-t:
The changes linux-nvme/libnvme#669 fixed so not necessary the nvme-cli changes.

PR #1990, opened 2023-06-28
id-ctrl JSON output power descriptor bitfield update

bao-tran-wdc:
Split max power flags into max power scale and non-operational state bits for clarity


igaw:
Sorry for delay. I've updated the commit message slightly and applied the patch directly. Thanks!

PR #2006, opened 2023-07-19
nvme-print: fix counter while looping through uuid_list

martin-gpy:
While iterating through the uuid_list in json_nvme_id_uuid_list() as part of the 'id-uuid' json output, it always misses out displaying the first uuid entry here due to wrongly initialized counter value. Fix it.

And while we are at it, remove the respective comment in stdout_id_uuid_list() as well where this counter is already properly initiatlized.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -2378,8 +2378,8 @@ static void json_nvme_id_uuid_list(const struct nvme_id_uuid_list *uuid_list)
 
 	root = json_create_object();
 	entries = json_create_array();
-	/* The 0th entry is reserved */
-	for (i = 1; i < NVME_ID_UUID_LIST_MAX; i++) {
+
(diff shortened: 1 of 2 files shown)


igaw:
Okay, did some digging to understand why it was added initially.

The reserved was introduced in 4ef5a032d50c ("nvme: UUIDs for Vendor-Specific Information"). It's not necessary
anymore because `struct nvme_id_uuid_list` takes care of it.

```c
/**
 * struct nvme_id_uuid_list - UUID list
 * @rsvd0:	reserved
 * @entry:	UUID list entry
 */
struct nvme_id_uuid_list {
	__u8	rsvd0[32];
	struct nvme_id_uuid_list_entry entry[NVME_ID_UUID_LIST_MAX];
};
```

PR #2008, opened 2023-07-21
nvme-print-stdout: List only ctrls which are part of ns

igaw:
We should only list the controllers which are associated with the current namespace. Previously we added all controllers of the subsystem to the list.

Fixes: a660c506906d ("nvme-print-stdout: Support subsys reachable from several hosts")

### DIFF
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -4674,6 +4674,7 @@ static bool stdout_detailed_ns(const char *name, void *arg)
 	struct htable_ns_iter it;
 	struct strset ctrls;
 	nvme_ctrl_t c;
+	nvme_path_t p;
 	nvme_ns_t n;
 	bool first;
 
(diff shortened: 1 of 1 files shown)


nvme-print-stdout.c
```diff
@@ -4693,8 +4694,10 @@ static bool stdout_detailed_ns(const char *name, void *arg)
 			return true;
 		}
 
-		nvme_subsystem_for_each_ctrl(nvme_ns_get_subsystem(n), c)
+		nvme_namespace_for_each_path(n, p) {
```
ikegami-t, line 4697: Looks good but to make sure let me confirm if is this worked for not multipath case since the function `nvme_namespace_for_each_path()` is used to check multipath as below.
```c
static bool nvme_is_multipath(nvme_subsystem_t s)
{
...
	nvme_subsystem_for_each_ns(s, n)
		nvme_namespace_for_each_path(n, p)
			return true;
...
	return false;
}
```

igaw: In the case of !multipath, `nvme_ns_get_ctrl(n))` will return a valid pointer right above  (ca. line 4691)

igaw: And what's worth, I've tested this for both scenarios

ikegami-t: Thanks for your explanations and confirmation. I could understand them.

PR #2012, opened 2023-07-25
nvme-print-stdout: fix storage tag size label

vincentkfu:
The NVM Command Set Extended LBA Format Data Structure has Protection Information Format (pif) and Storage Tag Size (sts) fields. The sts field is incorrectly labeled in the non-verbose normal output.  Fix the label.


igaw:
Thanks!

Issue #1963, opened 2023-06-01
Add I/O scheduler configuration to the topology command


igaw:
Since we are heading towards supporting different I/O scheduler in the kernel I was wondering if we should also add the currently used I/O scheduler setting in the topology command?

Thoughts?

@hreinecke @johnmeneghini @martin-gpy

keithbusch:
This would have been useful for a recent issue I was looking into. Which command are you thinking of adding this to? The verbose list?

igaw:
I would try to get it added to `nvme show-topology` because it shows all the ANA states. As  I/O schedulers are likely going to make use of this information. Thought not really sure if this makes sense. 

Or would it make sense to introduce a scheduler commands on itself which shows more detailed information? Hmm, I think I would only do this when @hreinecke scheduler patches have landed in the kernel.

So something like this:

```shell
# /usr/sbin/nvme show-topology
nvme-subsys23 - NQN=nqn.1992-08.com.netapp:sn.44ede5aaa52511e8abb200a09890f5bf:subsystem.nvme-svm-dolin-qla-subsystem iopolicy=round-robin
\
 +- ns 2
 \
  +- nvme23 fc traddr=nn-0x200200a09890f5bf:pn-0x203900a09890f5bf,host_traddr=nn-0x20000024ff7fa448:pn-0x21000024ff7fa448 live non-optimized
  +- nvme29 fc traddr=nn-0x200200a09890f5bf:pn-0x203600a09890f5bf,host_traddr=nn-0x20000024ff7fa448:pn-0x21000024ff7fa448 live optimized
  +- nvme31 fc traddr=nn-0x200200a09890f5bf:pn-0x203800a09890f5bf,host_traddr=nn-0x20000024ff7fa448:pn-0x21000024ff7fa448 live optimized
  +- nvme4 fc traddr=nn-0x200200a09890f5bf:pn-0x203700a09890f5bf,host_traddr=nn-0x20000024ff7fa448:pn-0x21000024ff7fa448 live non-optimized
 +- ns 1
 \
  +- nvme23 fc traddr=nn-0x200200a09890f5bf:pn-0x203900a09890f5bf,host_traddr=nn-0x20000024ff7fa448:pn-0x21000024ff7fa448 live non-optimized
[...]

PR #2015, opened 2023-07-26
Add I/O scheduler configuration to the topology command

igaw:
- A few small fixes upfront.
- Add the iopolicy scheduler to the topology output 

Fixes: https://github.com/linux-nvme/nvme-cli/issues/1963


igaw:
I forgot to add the `iopolicy' field to `list-subsys`...

PR #2019, opened 2023-08-02
Fix multi hostnqn support for discovery controllers

martin-belanger:
When setting up a persistent connection to a Discovery Controller (DC), `nvme discover` would reuse any existing connection matching the transport parameters even when the Host NQN is different between two connections.

For example, let's say one configures a persistent connection to a DC as follows:
```
nvme discover -t tcp -a 100.94.69.55 -s 8009 -q nqn.2014-08.org.nvmexpress:uuid:11111111-1111-1111-1111-111111111111 -p
```

Now, let's say one configures another persistent connection to the same DC, but using a different Host NQN.
```
nvme discover -t tcp -a 100.94.69.55 -s 8009 -q nqn.2014-08.org.nvmexpress:uuid:22222222-2222-2222-2222-222222222222 -p
```

This second command should result in a separate connection being created because it uses a different Host NQN and therefore a separate host object is required. However, nvme-cli would reuse the first connection instead of creating a new one.

This pull request also changes the way IP Addresses (TRADDR) are compared for TCP and RDMA. A simple `strcmp()` or `strcmpcase()` is not sufficient when comparing IP addresses (especially IPv6). This change follows a recent change that was made to `libnvme` (Ref: https://github.com/linux-nvme/libnvme/pull/656).

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -117,28 +117,32 @@ static const char *nvmf_context		= "execution context identification string";
 		OPT_END()                                                                        \
 	}
 
-/*
+/**
  * Compare two C strings and handle NULL pointers gracefully.
- * If either of the two strings is NULL, return 0
- * to let caller ignore the compare.
+ * Return true if both pointers are equal (including both set to NULL).
+ * Return false if one and only one of the two pointers is NULL.
+ * Perform string comparisong only if both pointers are not NULL and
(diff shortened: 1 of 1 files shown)


igaw:
Sorry, this fell through the cracks. Let me quickly go over it again, I can remember I looked at it but wanted first to get the libnvme changes in.

PR #2027, opened 2023-08-08
Change handles from file-descriptors to 'struct dev_handle'

vikash-k5:
Change handles from file-descriptors to 'struct dev_handle'

This replaces the use of file-descriptors as handles with the use of a struct dev_handle. This is done in preparation to separate the nvme-cli "frontend" code from the "backend" code provided by libraries such as as libnvme, xNVMe, and potentially other backends.


igaw:
Let's close this one. It is nvme-cli v3 material anyway.

PR #2039, opened 2023-08-23
nvme-copy: support cross-namespace copy

jgu-pure:
Add support for NVMe TP4130 ("Cross-Namespace Copy")
- Add Copy Descriptor Formats 2h and 3h
- Add "--snsids" option to specify source namespaces to copy from
- Add "--sopts" option to specify source options (Fast Copy Only)
- Print new Host Behavior Support bits, new ONCS bits, and new Optional Copy Formats Supported bits
- Extend unit test case to test cross-namespace copy formats

---

There are TBD constants because TP4130 hasn't been ratified yet, but I hope I can get some feedback on this change before then!

A recent version of TP4130 can be found here: [TP4130 Cross-Namespace Copy 2023.07.06 - Integration](https://workspace.nvmexpress.org/wg/Technical/document/5015)


jgu-pure:
Could this pull request be removed? https://support.github.com/request/remove-data

PR #2047, opened 2023-09-13
nvme: Change to use NVME_NS_FLBAS_META_EXT() macro to check flbas value

ikegami-t:


ikegami-t:
Change for the linux-nvme/libnvme#662 changes.

ikegami-t:
Just rebased and fixed a conflict.

PR #2051, opened 2023-09-14
 nvme: allocate aligned payloads for all nvme commands

igaw:
The kernel supports since v5.2 direct mapped DMA buffers to userspace.
Up to this point a bounce buffer was involved. Because the buffers are
now directly accessed by the device, the rules of alignment also apply
for the payloads.

Furthermore, ensure that the buffer is a multiple of 4k, because there
are devices on the market which will always transfer a multiple of 4k,
even if we ask for less, e.g 512 bytes. This avoid stack smashes.

Fixes: #https://github.com/linux-nvme/libnvme/issues/684

### DIFF
--- a/nvme-print.h
+++ b/nvme-print.h
@@ -8,8 +8,8 @@
 #include <ccan/list/list.h>
 
 typedef struct nvme_effects_log_node {
+	struct nvme_cmd_effects_log effects; /* needs to be first member because of alignment requirement. */
 	enum nvme_csi csi;
-	struct nvme_cmd_effects_log effects;
 	struct list_node node;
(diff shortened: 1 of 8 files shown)


igaw:
I've added a cleanup patch on top for the `nvme_dev` resource. This makes the code slightly more cleaner and easier to follow. Obviously, there is more things we can tidy up with the auto cleanup hooks. I think we could also close all file handles this way in `submit_io` and co which could make the whole error handling way simpler. Currently, this function is really really hard to read and I wouldn't surprised if there are still bugs lurking in these code paths.

PR #2054, opened 2023-09-20
nvme: Add support for the OAQD ID controller field

bpaupore-wdc:
Depends on the changes in commit `6de7efe` for libnvme


igaw:
just rebasing, testing if the build breaks due to the missing dependency update.

igaw:
and now rebased with dependency updated for libnvme.

PR #2056, opened 2023-09-20
Add Support for Phy rx eom log

bpaupore-wdc:
This primarily introduces a new command `phy-rx-eom-log` for getting that log page and allowing the user to specify the lsi/lsp values for it. These changes depend on libnvme commit `e923016`.

One question I have is how to handle failing allocations in `nvme-print-json.c`. Currently these changes just silently fail to include the printable eye data in the json root, but if there's some way this error should be reported I can make that change.

The second commit addresses a very minor bug obscuring non-zero reserved fields for fdpa. I could split this out into its own PR if that would be preferred, otherwise it seemed simple enough to include here as a separate commit.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -69,6 +69,7 @@
 #include "fabrics.h"
 #define CREATE_CMD
 #include "nvme-builtin.h"
+#include "malloc.h"
 
 struct feat_cfg {
 	enum nvme_features_id feature_id;
@@ -198,6 +199,20 @@ static void *nvme_alloc(size_t len)
 	return p;
 }
 
+void *nvme_realloc(void *p, size_t len)
+{
+	size_t old_len = malloc_usable_size(p);
+
+	void *result = nvme_alloc(len);
+
+	if (p) {
+		memcpy(result, p, min(old_len, len));
+		free(p);
+	}
+
+	return result;
+}
+
 static void *__nvme_alloc_huge(size_t len, bool *huge)
 {
 	void *p;
@@ -248,6 +263,21 @@ void *nvme_alloc_huge(size_t len, bool *huge)
 	return __nvme_alloc_huge(len, huge);
 }
 #endif
+
+void *nvme_realloc_huge(void *p, size_t len, bool *huge)
+{
+	size_t old_len = malloc_usable_size(p);
+	bool was_huge = *huge;
+
+	void *result = nvme_alloc_huge(len, huge);
+
+	if (p) {
+		memcpy(result, p, min(old_len, len));
+		nvme_free_huge(p, was_huge);
+	}
+
+	return result;
+}
 
 const char *nvme_strerror(int errnum)
 {
--- a/nvme.c
+++ b/nvme.c
@@ -1965,6 +1965,104 @@ static int get_boot_part_log(int argc, char **argv, struct command *cmd, struct 
 	return err;
 }
 
+static int get_phy_rx_eom_log(int argc, char **argv, struct command *cmd,
+		struct plugin *plugin)
+{
+	const char *desc = "Retrieve Physical Interface Receiver Eye Opening\n"
+		"Measurement log for the given device in decoded format\n"
+		"(default), json or binary.";
+	const char *controller = "Target Controller ID.";
+	_cleanup_free_ struct nvme_phy_rx_eom_log *phy_rx_eom_log = NULL;
+	size_t phy_rx_eom_log_len;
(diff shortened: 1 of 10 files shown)


igaw:
rebased on current master to get latest libnvme changes.

igaw:
> One question I have is how to handle failing allocations in nvme-print-json.c. Currently these changes just silently fail to include the printable eye data in the json root, but if there's some way this error should be reported I can make that change.

The printing backend rework is still WIP. The goal is to have all commands producing output for all formats (at least stdout and json). Currently there are a lot of missing hooks. So it would be great to have also the json output part added. I am unsure if we want have a binary output as well. I suspect for some commands it would make sense.

igaw:
> The second commit addresses a very minor bug obscuring non-zero reserved fields for fdpa. I could split this out into its own PR if that would be preferred, otherwise it seemed simple enough to include here as a separate commit.

I don't mind it that it is in this PR.

nvme.c
```diff
@@ -1927,6 +1927,115 @@ static int get_boot_part_log(int argc, char **argv, struct command *cmd, struct
...
+	};
+
+	err = parse_and_open(&dev, argc, argv, desc, opts);
+	if (err)
+		goto ret;
+
+	err = flags = validate_output_format(cfg.output_format);
+	if (err < 0) {
+		nvme_show_error("Invalid output format");
+		goto close_fd;
```
igaw, line 1970: With the auto cleanup this will be just

```
if (err < 0) {
    nvme_show_error(...)
    return err;
}
```

nvme.c
```diff
@@ -1927,6 +1927,115 @@ static int get_boot_part_log(int argc, char **argv, struct command *cmd, struct
...
+	} else if ((cfg.lsp & 12) == 12) {
+		nvme_show_error("invalid action: %u", cfg.lsp & 12);
+		err = -1;
+		goto close_fd;
+	}
+
+	/* Fetching header to calculate total log length */
+	phy_rx_eom_log_len = sizeof(struct nvme_phy_rx_eom_log);
+	phy_rx_eom_log = malloc(phy_rx_eom_log_len);
```
igaw, line 1989: Use the `nvme_alloc` helper to ensure that the alignment of the payload is correct and we have a buffer size of multiple of 4k. There are some buggy firmware out there.

```
phy_rx_oem_log = nvme_alloc(phy_rx_log_len);
if (!phy_rx_oem_log)
    return -ENOMEM;
```

nvme.c
```diff
@@ -2015,7 +2015,7 @@ static int get_phy_rx_eom_log(int argc, char **argv, struct command *cmd,
 		phy_rx_eom_log_len = le16_to_cpu(phy_rx_eom_log->hsize);
 	}
 
-	phy_rx_eom_log = realloc(&phy_rx_eom_log, phy_rx_eom_log_len);
+	phy_rx_eom_log = realloc(phy_rx_eom_log, phy_rx_eom_log_len);
```
igaw, line 2018: This should be part of the first patch. Though we don't have a `nvme_realloc` yet. Just introduce a simple `nvme_realloc` function which does the alloc new block if necessary things. Doesn't need to be fancy, correctness is more important than premature optimization.

igaw (changes requested):
Looks good overall. Thanks for the different print implementations. That is very cool!

I've just added the auto cleanup hooks to get rid of the many gotos which makes the code a bit more readable (there are more places to cleanup WIP).

nvme.c
```diff
@@ -1927,6 +1927,115 @@ static int get_boot_part_log(int argc, char **argv, struct command *cmd, struct
...
 }
 
+static int get_phy_rx_eom_log(int argc, char **argv, struct command *cmd,
+		struct plugin *plugin)
+{
+	const char *desc = "Retrieve Physical Interface Receiver Eye Opening\n"
+		"Measurement log for the given device in decoded format\n"
+		"(default), json or binary.";
+	const char *controller = "Target Controller ID.";
+	struct nvme_phy_rx_eom_log *phy_rx_eom_log;
```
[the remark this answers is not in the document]
bpaupore-wdc: The phy_rx_**oem**_log tripped me up a bit lol, but it seems like this is sorted out now.

nvme.c
bpaupore-wdc: Pulled this change into the other commit and added a `nvme_realloc` function.

igaw (approved):

PR #2057, opened 2023-09-21
udev-rule: apply round-robin iopolicy on i/o subsystems alone

martin-gpy:
The current 71-nvmf-iopolicy-netapp.rules.in sets the round-robin iopolicy not just on nvme i/o subsystems, but on discovery subsystems too. Update this rule to have this applied to i/o subsystems alone.

Cc: Clayton Skaggs <claytons@netapp.com>


igaw:
Thanks!

Issue #2013, opened 2023-07-25
Invalid FW image error (0x2107) with split files


tkenbo:
When we try downloading split firmware file with offset, it reported an error 2107 from fw_commit command with version 2.4.
Tried it with v1.16 and it did not report an error.

here is the command I executed.
```
nvme fw-download /dev/nvme2 -f file_0.bin -x 262144 -o 0
nvme fw-download /dev/nvme2 -f file_1.bin -x 262144 -o 50176
nvme fw-download /dev/nvme2 -f file_2.bin -x 262144 -o 161792
nvme fw-download /dev/nvme2 -f file_3.bin -x 262144 -o 256000
nvme fw-download /dev/nvme2 -f file_4.bin -x 262144 -o 385024
nvme fw-download /dev/nvme2 -f file_5.bin -x 262144 -o 481280
nvme fw-download /dev/nvme2 -f file_6.bin -x 262144 -o 560128
nvme fw-commit /dev/nvme2 -s 1 -a 1
NVMe status: Invalid Firmware Image: The firmware image specified for activation is invalid and not loaded by the controller(0x2107)
```
fw_commit completes successfully with v1.16.
```
nvme fw-commit /dev/nvme2 -s 1 -a 1
Success committing firmware action:1 slot:1
Multiple Update Detected (MUD) Value: 0
```
Looking at the trace in v2.4, we only see a few download commands sent to the drive.
```
            nvme-249925  [027] ..... 424627.900917: nvme_setup_cmd: nvme2: qid=0, cmdid=49176, nsid=0, flags=0x0, meta=0x0, cmd=(nvme_admin_download_fw cdw10=ff c3 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00)
          <idle>-0       [047] d.h1. 424627.901595: nvme_sq: nvme2: qid=0, head=22, tail=22
          <idle>-0       [027] .Ns1. 424627.901608: nvme_complete_rq: nvme2: qid=0, cmdid=49176, res=0x0, retries=0, flags=0x2, status=0x0
            nvme-249949  [024] ..... 424747.717830: nvme_setup_cmd: nvme2: qid=0, cmdid=20493, nsid=0, flags=0x0, meta=0x0, cmd=(nvme_admin_download_fw cdw10=ff ff 00 00 00 c4 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00)
[...]

PR #2060, opened 2023-09-23
nvme: fw-download offset only describes FW offset, not file offset

jk-ozlabs:
In e7ca3bcbdba, we (unintentionally) changed the semantics of the fw-download's --offset argument to also apply an offset to the source file.

According to https://github.com/linux-nvme/nvme-cli/issues/2013, the old behaviour is useful when the source firmware image was provided in separate chunks.

This change restored to the old hehaviour, where --offset only applies to the destination firmware buffer on the device. This means that the start of the source file will end up at the destination offset.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/2013


jk-ozlabs:
@tkenbo, can you test this against your use-case?

tkenbo:
Hi @jk-ozlabs 

Thanks for the update.  I tried to install it but having json-types.h missing error on compile.  I have Rocky 8.7 and json-c is v1.3 installed.  will try figuring out.

```[root@sm-110p-140 nvme-cli-pr-fw-offset]# meson compile -C .build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/local/bin/ninja -C /root/Downloads/nvme-cli-pr-fw-offset/.build
ninja: Entering directory `/root/Downloads/nvme-cli-pr-fw-offset/.build'
[3/160] Compiling C object subprojects/libnvme/src/libnvme.so.1.5.0.p/nvme_json.c.o
FAILED: subprojects/libnvme/src/libnvme.so.1.5.0.p/nvme_json.c.o
cc -Isubprojects/libnvme/src/libnvme.so.1.5.0.p -Isubprojects/libnvme/src -I../subprojects/libnvme/src -Isubprojects/libnvme -I../subprojects/libnvme -Isubprojects/libnvme/ccan -I../subprojects/libnvme/ccan -Isubprojects/libnvme/internal -I../subprojects/libnvme/internal -I/usr/include/json-c -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=gnu99 -O0 -g -fomit-frame-pointer -D_GNU_SOURCE -include subprojects/libnvme/internal/config.h -DCCAN_LIST_DEBUG=1 -DCCAN_STR_DEBUG=1 -fPIC -MD -MQ subprojects/libnvme/src/libnvme.so.1.5.0.p/nvme_json.c.o -MF subprojects/libnvme/src/libnvme.so.1.5.0.p/nvme_json.c.o.d -o subprojects/libnvme/src/libnvme.so.1.5.0.p/nvme_json.c.o -c ../subprojects/libnvme/src/nvme/json.c
In file included from /usr/include/json-c/json.h:27,
                 from ../subprojects/libnvme/src/nvme/json.c:15:
/usr/include/json-c/json_object.h:27:10: fatal error: json_types.h: No such file or directory
 #include "json_types.h"
          ^~~~~~~~~~~~~~
compilation terminated.
[56/160] Compiling C object subprojects/libnvme/src/libnvme.so.1.5.0.p/nvme_tree.c.o
ninja: build stopped: subcommand failed.
```

jk-ozlabs:
I haven't used Rocky here, but looking at the `json-c-devel` RPM:

```
$ rpm2cpio json-c-devel-0.13.1-3.el8.x86_64.rpm | cpio -t  | grep include
221 blocks
./usr/include/json-c
./usr/include/json-c/arraylist.h
./usr/include/json-c/bits.h
./usr/include/json-c/debug.h
./usr/include/json-c/json.h
./usr/include/json-c/json_c_version.h
./usr/include/json-c/json_config.h
./usr/include/json-c/json_inttypes.h
./usr/include/json-c/json_object.h
./usr/include/json-c/json_object_iterator.h
./usr/include/json-c/json_pointer.h
./usr/include/json-c/json_tokener.h
./usr/include/json-c/json_util.h
./usr/include/json-c/json_visit.h
./usr/include/json-c/linkhash.h
./usr/include/json-c/printbuf.h
```

so there is indeed no `json_types.h` file there. However:

```
$ rpm2cpio json-c-devel-0.13.1-3.el8.x86_64.rpm | cpio -i --to-stdout ./usr/include/json-c/json_object.h | grep include
#include <stddef.h>
#include "json_inttypes.h"
#include "printbuf.h"
```

it doesn't look like the packaged `json_object.h` references a `json_types.h` file anyway? Are you using the vanilla packages there? or is there a chance your json_object.h has been overwritten?

jk-ozlabs:
I haven't used Rocky here, but looking at the `json-c-devel` RPM:

```
$ rpm2cpio json-c-devel-0.13.1-3.el8.x86_64.rpm | cpio -t  | grep include
221 blocks
./usr/include/json-c
./usr/include/json-c/arraylist.h
./usr/include/json-c/bits.h
./usr/include/json-c/debug.h
./usr/include/json-c/json.h
./usr/include/json-c/json_c_version.h
./usr/include/json-c/json_config.h
./usr/include/json-c/json_inttypes.h
./usr/include/json-c/json_object.h
./usr/include/json-c/json_object_iterator.h
./usr/include/json-c/json_pointer.h
./usr/include/json-c/json_tokener.h
./usr/include/json-c/json_util.h
./usr/include/json-c/json_visit.h
./usr/include/json-c/linkhash.h
./usr/include/json-c/printbuf.h
```

so there is indeed no `json_types.h` file there. However:

```
$ rpm2cpio json-c-devel-0.13.1-3.el8.x86_64.rpm | cpio -i --to-stdout ./usr/include/json-c/json_object.h | grep include
#include <stddef.h>
#include "json_inttypes.h"
#include "printbuf.h"
```

it doesn't look like the packaged `json_object.h` references a `json_types.h` file anyway? Are you using the vanilla packages there? or is there a chance your json_object.h has been overwritten?

igaw:
Looks good. Should I wait with the merge until we have a positive test result?

igaw:
I think you don't need libjson-c for this tests, so you could just disable the build for json-c.

tkenbo:
cool.  Thanks @igaw!

The error I was getting was random order of segment download.
I downloaded segment 2 -> 5 -> 1 -> 4 -> 0 -> 6 -> 3, then commit returned the error.
```
nvme fw-download /dev/nvme3 -f file_2.bin -x 262144 -o 146432
nvme fw-download /dev/nvme3 -f file_5.bin -x 262144 -o 479232
nvme fw-download /dev/nvme3 -f file_1.bin -x 262144 -o 41984
nvme fw-download /dev/nvme3 -f file_4.bin -x 262144 -o 357376
nvme fw-download /dev/nvme3 -f file_0.bin -x 262144 -o 0
nvme fw-download /dev/nvme3 -f file_6.bin -x 262144 -o 605184
nvme fw-download /dev/nvme3 -f file_3.bin -x 262144 -o 260096
nvme fw-commit /dev/nvme3 -s 1 -a 1
CRITICAL NvmeCli Error : NVMe status: Invalid Firmware Image: The firmware image specified for activation is invalid and not loaded by the controller(0x2107)
```
I looked at the change and noticed one issue at line 5072.
`cfg.xfer = min(cfg.xfer, fw_size);`

in here I believe fw_size should be the actual remaining file size.
For example, 
 fw_size = 3500
 cfg.xfer =1000
then, cfg.xfer for the last loop should be 500 instead of 1000.

jk-ozlabs:
Ah! yes, this should clamp to the min of the remaining size. New patch pushed.

jk-ozlabs:
@tkenbo purely out of curiosity: why are you transferring this as separate files / chunks?

(I'm glad we're covering this use-case, just want to understand more about your requirements here)

tkenbo:
@jk-ozlabs we are testing NVMe devices for the spec compliance in various system configurations in the market with various OS platforms as a part of our system integration testing.  We test the device for a unique customer environment and also in the off-the-shelf environment.  I believe the original change you made was to improve a user experience, but we ran into this behavior change, which made little harder to test along the spec definition using nvmecli. so I though I would ask.  Hope this answers.
By the way, I tested the new change and it passed the test!  Thanks again!

jk-ozlabs:
>  which made little harder to test along the spec definition using nvmecli

ah, that makes sense  - this restored behaviour allows you to transfer the offsets out-of-order.

(We can do the same thing with an offset and fixed length argument, which means you don't need to split the file up into chunks - you would just use the same file for all transfers. However, I'm fine with the behaviour restored by this change if you prefer)

> By the way, I tested the new change and it passed the test! Thanks again!

Excellent, thanks for that!

in which case: @igaw , I'm ok to merge if you are.

PR #2064, opened 2023-09-26
nvme: Add output-format default option instead of argconfig json option

ikegami-t:


igaw:
Nice!  Many thanks!

Cool, so the next big task is to add the missing callbacks in json (and maybe binary) print files. Just in case you want to work on this.

ikegami-t:
I see. Yes will do it. Thanks for your suggestion.

PR #2065, opened 2023-09-26
Feature/nvme cli log page directory add log page names

lgdacunh:


nvme-print.c
```diff
@@ -664,12 +664,22 @@ const char *nvme_log_to_string(__u8 lid)
 	case NVME_LOG_LID_ENDURANCE_GROUP:		return "Endurance Group Information";
 	case NVME_LOG_LID_PREDICTABLE_LAT_NVMSET:	return "Predictable Latency Per NVM Set";
 	case NVME_LOG_LID_PREDICTABLE_LAT_AGG:		return "Predictable Latency Event Aggregate";
+	case NVME_LOG_LID_MEDIA_UNIT_STATUS:
+		return "Media Unit Status";
+	case NVME_LOG_LID_SUPPORTED_CAP_CONFIG_LIST:
+		return "Supported Capacity Configuration List";
```
igaw, line 670: Although the lines get a bit long with the new entries, I'd rather stay consistent and keep every entry  one one line.

igaw (changes requested):
A formatting nitpicking and I've noticed that shankara is listed as author, so the `Signed-off-by`  from shankara would be necessary.

Rest looks good.

nvme-print.c
lgdacunh: We are working to get Shankar the proper GitHub user and keys. Soon we will have a signed commit with the requested changes.

lgdacunh:
Shankar amended commits with his signature, and performed the requested code format changes.

igaw (approved):

igaw:
Thanks!

PR #2070, opened 2023-09-28
fix for create-ns command

vikash-k5:
"data" field is being used without memory allocation
this commit add fix for that


igaw:
Updated commit message

PR #2073, opened 2023-10-03
udev-rules: set ctrl_loss_tmo to -1 for ONTAP NVMe/TCP

martin-gpy:
Setting ctrl_loss_tmo to -1 for ONTAP NVMe/TCP controllers would enable indefinite reconnect attempts during a path loss and help avoid purging the path on the host, which otherwise may lead to mounted fs read-only behavior. So add a rule towards enabling the same.


igaw:
Thanks!

PR #2080, opened 2023-10-08
Print json directive

ikegami-t:


igaw:
Looks good so far. Though there is a bit more work ahead, e.g. 

`./nvme get-feature /dev/nvme0n1  -o json`

will output a mixture between stdout and json. I understand this is WIP as we just added the gobal `--output` option. So no problem with this for the moment.

Do you push more changes on this branch or do you expect me to merge it?

ikegami-t:
Thanks for your reviewing. Yes please do merge the PR changes if you okay since both okay for me. As you mentioned the get-feature command mixed output issue should be fixed later and also still remaining json print function will be implemented so I will rebase the implementation if merged.

PR #2083, opened 2023-10-11
Standalone

sc108-lee:
* Support stand alone binary build on Makefile
* Add libhugetlbfs option to not included in stand alone (not supported)


igaw:
It might be worth to add a CI build target for this? What do you think?

sc108-lee:
wrapmode modified as forcefallback (it might be better to support stand alone)
@igaw I think thats great idea.
it would be helpful to catch problems like libhugetlbfs (which is not support)

igaw:
I've updated the build container to include make. If you add something like:

```yaml
  build-make-standalone:
    name: make standalone
    runs-on: ubuntu-latest
    container:
      image: ghcr.io/igaw/linux-nvme/debian:latest
    steps:
      - uses: actions/checkout@v3
      - name: build
        run: |
          make standalone
```

to `.github/workflows/build.yml` it should work.

igaw:
> could you install libkeyutils-dev to build keyutils in libnvme?

The problem is that libkeyutils-dev is installed but only the shared version not the static one. Obviously, it would be nice if meson could catch this, maybe we are using it wrong? No idea.

Anyway, either we add libkeytutils a wrapper so or disable it for the time being.

BTW, I think we should call the build target 'make static' instead of standalone.

sc108-lee:
> The problem is that libkeyutils-dev is installed but only the shared version not the static one. Obviously, it would be nice if meson could catch this, maybe we are using it wrong? No idea.
As far as I know, That would be .pc(pkgconfig) file problem.
if -lname is given pkgconfig find libname.so then libname.a,
so usually to avoid that confusion, usually release split file of pc (  name.pc, name-static.pc )
And if meson using that as dependency('name-static').

so I just disabled libkeytutils as you mentioned

standalone -> static addressed :)

igaw:
Excellent! Thanks!

PR #2086, opened 2023-10-12
nvme: Fixed segmentation fault when getting host initiated telemetry

lgdacunh:


igaw:
Oops :)

Thanks for the fix!

PR #2088, opened 2023-10-12
nvme-print-json: Add remaining feature fields print functions

ikegami-t:


igaw:
Could you split the change into two bits? One which updates the formatting/codying style and one which just adds the new code? It makes the review a bit simpler. Thanks!

ikegami-t:
Thanks for your comment. Split the patch and added additional a patch to fix the get feature command mixed stdout and json outputs issue.
Note: Still the get feature command outputs d() and d_raw() to stdout and error to stderr so it will be fixed separately.

ikegami-t:
Removed the patches below to the PRs #2090 and #2091 since not related to the json output changes.
1. 5cc7bac nvme: Change short option -o and -v duplicated to upper case
2. 10e9c65 doc: Change short option -o and -v duplicated to upper case
3. c2510ad completions: Change short option -o and -v duplicated to upper case
4. e98bc40 nvme: Change phy-rx-eom-log command to use NVME_ARGS instead

nvme-print-json.c
```diff
@@ -47,17 +47,51 @@
...
+static const char *supported_str = "Supported";
+static const char *temperature_str = "Temperature";
+static const char *threshold_str = "Threshold";
+static const char *transport_str = "Transport";
+static const char *true_str = "True";
+static const char *type_str = "Type";
+static const char *unknow_str = "Unknown";
+static const char *value_str = "Value";
+static const char *version_str = "Version";
+static const char *yes_str = "Yes";
```
igaw, line 106: I am not really convinced we should start to move these strings out. These days, the compiler and linker are very smart and are able to avoid data duplication. And these are only a few of the keys and we end up with things like:

```c
					obj_add_str(jns, namespace_str, nvme_ns_get_name(n));
					obj_add_str(jns, generic_str, nvme_ns_get_generic_name(n));
					obj_add_int(jns, nsid_str, nvme_ns_get_nsid(n));
					obj_add_uint64(jns, "UsedBytes", nuse);
					obj_add_uint64(jns, "MaximumLBA", nvme_ns_get_lba_count(n));
					obj_add_uint64(jns, "PhysicalSize", nsze);
```

two styles as of key. Which I really do not like because it's not clear when to use what. I assume if a key used more than once use an `const char *` definition.

Sure the argument that allows to update the string in one place is nice but I don't think we have a huge problem here.

In other words I think we should not do it at all. Could you drop this patch and also remove the already existing `const char *` variables? Thanks!

nvme-print-json.c
```diff
@@ -11,14 +11,68 @@
...
+#define STR_LEN 100
 #define BUF_LEN 320
 #define VAL_LEN 4096
 #define BYTE_TO_BIT(byte) ((byte) * 8)
+#define POWER_OF_TWO(exponent) (1 << (exponent))
+#define MS500_TO_SEC(time) ((time) / 2)
+
+#define array_add_obj(o, k) json_array_add_value_object(o, k);
+#define obj_add_str(o, k, v) json_object_add_value_string(o, k, v)
+#define root_add_array(k, v) json_object_add_value_array(root, k, v)
```
igaw: I was wondering if wouldn't make sense to just use `r` instead of `root` as variable name in the functions and the use `obj_add_str(r, ..., ...)` instead introducing a wrapper macro just for this. What do you think?

As this series is pretty big and I don't want to make you suffer too much by updating every patch again, you can  just add a cleanup patch add the end of the series, that is fine with me.

igaw:
The feature output is getting closer to be valid JSON:

`./nvme get-feature  /dev/nvme0 -o json`
```json
{
  "Feature":"0x01",
  "Name":"Arbitration",
  "Current":"0x03030302"
}
{
  "Feature":"0x02",
  "Name":"Power Management",
  "Current":"0x00000003"
}
{
  "Error":"get-feature:0x03 (LBA Range Type): "
}
{
  "Error":"Invalid Namespace or Format: The namespace or the format of that namespace is invalid",
  "Type":"nvme",
  "Value":16395
}
{
  "Feature":"0x04",
  "Name":"Temperature Threshold",
  "Current":"0x00000162"
}
{
  "Error":"get-feature:0x05 (Error Recovery): "
}
{
  "Error":"Invalid Namespace or Format: The namespace or the format of that namespace is invalid",
  "Type":"nvme",
  "Value":16395
}
{
  "Feature":"0x06",
  "Name":"Volatile Write Cache",
  "Current":"0x00000001"
}
```

We need to figure out how handle the error case above and all features should be reported in an array. There problem is the `show_feature` API is not really good for the JSON case. Maybe we need to split into `show_feature_init`, `show_feature` and `show_feature_finish` calls. The `show_feature_init` callback would allow the JSON code to allocate the array and then the pointer is handed into the `show_feature` function for every single invocation. `show_feature_finish` allows the JSON code then print the result in one go.

nvme-print-json.c
ikegami-t: Thanks for your comments. Just fixed by the additional commit [7a7b142](https://github.com/linux-nvme/nvme-cli/pull/2088/commits/7a7b142d7704f4bf430160bb7de0c7d375b9779c) but not deleted the original commit since seems some other commits are depended on the original commit. If needed will do delete the original commit.

ikegami-t:
Rebased with the master and fixed some conflicts.

ikegami-t:
Noted about the show_feature API problem. Yes I thought also to use the global variable root json object in nvme-print-json to resolved the issue. Anyway later will do consider the suggestion also. Thank you.

nvme-print.h
```diff
@@ -83,6 +83,8 @@ struct print_ops {
 	void (*lba_range)(struct nvme_lba_range_type *lbrt, int nr_ranges);
 	void (*lba_status_info)(__u32 result);
 	void (*d)(unsigned char *buf, int len, int width, int group);
+	void (*show_init)(void);
```
igaw, line 86: Let this function return an opaque pointer

nvme-print.h
```diff
@@ -83,6 +83,8 @@ struct print_ops {
...
+	void (*show_finish)(void);
```
igaw, line 87: and pass it into the function again...

but this might be  unnecessary complex thing to do or do we expect multithreading? so just have a init/finish call would be enough. let me ponder on this.

nvme-print.h
ikegami-t: I see.

nvme-print.h
ikegami-t: Just I thought no opaque pointer returning and passing seem simple for the caller to use the APIs. Sure if needed I can change them as mentioned.

igaw: Basically this means this API is not thread safe, which is fine for nvme-cli. I think we just go with your version, but maybe just call it the function `init` and `finish`.

igaw:
Thanks a lot.

PR #2089, opened 2023-10-12
plugins/solidigm: internal-logs Telemetry auto detect last data area.

lgdacunh:
Faster telemetry retrieval using large buffers.


igaw:
Thanks!

PR #2090, opened 2023-10-14
Nvme short option

ikegami-t:
Cherry picked the PR #2088 patches below to split since not related to the json output changes.
1. 5cc7bac4 nvme: Change short option -o and -v duplicated to upper case
2. 10e9c650 doc: Change short option -o and -v duplicated to upper case
3. c2510add completions: Change short option -o and -v duplicated to upper case


igaw:
Rebase to current master to trigger new clean build.

igaw:
Thanks!

PR #2094, opened 2023-10-17
plugins/solidigm: Added support for temperature statistics log page

lgdacunh:


plugins/solidigm/solidigm-temp-stats.c
```diff
@@ -0,0 +1,106 @@
...
+	__u8 uuid_idx;
+	int err;
+
+	const char *desc = "Get/show Temperature Statistics log.";
+	const char *raw = "dump output in binary format";
+	struct config {
+		bool  raw_binary;
+	};
+
+	struct config cfg = {
```
ikegami-t, line 53: Is the variable raw_binary is initialized to false correctly by this implementation or not necessary to initialize?
Usually it is zero cleared as below I think.
```c
	struct config cfg = { 0 };
```

plugins/solidigm/solidigm-temp-stats.c
```diff
@@ -0,0 +1,106 @@
...
+		if (guid[1] == 0xC7BB98B7D0324863 && guid[0] == 0xBB2C23990E9C722F) {
+			fprintf(stderr, "Error: Log page has 'OCP unsupported Requirements' GUID\n");
+			err = -EBADMSG;
+			goto closefd;
+		}
+		if (!cfg.raw_binary)
+			show_temp_stats((struct temp_stats *) buffer);
+		else
+			d_raw(buffer, sizeof(struct temp_stats));
+	} else if (err > 0)
```
ikegami-t, line 98: Better to use `{` and `}` for the `else if` case statement also as below since the `if` statement already used.
```diff
  	if (xxx) {
  		aaa;
  		bbb;
- 	} else if
+ 	} else if {
  		ccc;
+ 	}
```

plugins/solidigm/solidigm-temp-stats.c
lgdacunh: Good catch! Now raw_binary is explicitly initialized.

plugins/solidigm/solidigm-temp-stats.c
lgdacunh: Done!

ikegami-t:
Looks good.

igaw:
Thanks!

PR #2095, opened 2023-10-18
create-ns: align the namespaces to 1Mib boundaries when using SI suffixes

maurizio-lombardi:
Some controllers refuse to create namespaces with a size not aligned to a 1 megabyte boundary, this happens when using the -S and -C options.

$ nvme create-ns /dev/nvme0 -S 80M -C 80M -f 0
NVMe status: Invalid Field in Command: A reserved coded value or an unsupported value in a defined field(0x2)

Fix this problem by modifying parse_lba_num_si() to align the values to 1 Megabyte boundaries.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2833,12 +2833,13 @@ static int detach_ns(int argc, char **argv, struct command *cmd, struct plugin *
 }
 
 static int parse_lba_num_si(struct nvme_dev *dev, const char *opt,
-			    const char *val, __u8 flbas, __u64 *num)
+			    const char *val, __u8 flbas, __u64 *num, __u32 align)
 {
 	_cleanup_free_ struct nvme_ns_list *ns_list = NULL;
 	_cleanup_free_ struct nvme_id_ctrl *ctrl = NULL;
 	_cleanup_free_ struct nvme_id_ns *ns = NULL;
 	__u32 nsid = 1;
+	unsigned int remainder;
 	char *endptr;
 	int err = -EINVAL;
 	int i;
@@ -2916,6 +2917,12 @@ static int parse_lba_num_si(struct nvme_dev *dev, const char *opt,
 		return -errno;
 	}
 
+	if (endptr[0]) {
+		remainder = *num % align;
+		if (remainder)
+			*num += align - remainder;
+	}
+
 	if (endptr[0] != '\0')
 		*num /= lbas;
 
@@ -2958,6 +2965,10 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 	__u32 nsid;
 	uint16_t num_phandle;
 	uint16_t phndl[128] = { 0, };
+	_cleanup_free_ struct nvme_id_ctrl *id = NULL;
+	_cleanup_free_ struct nvme_id_ns_granularity_list *gr_list = NULL;
+	__u32 align_nsze = 1 << 20; /* Default 1 MiB */
+	__u32 align_ncap = align_nsze;
 
 	struct config {
 		__u64	nsze;
@@ -3075,11 +3086,56 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 		return -EINVAL;
 	}
 
-	err = parse_lba_num_si(dev, "nsze", cfg.nsze_si, cfg.flbas, &cfg.nsze);
+	id = nvme_alloc(sizeof(*id));
+	if (!id)
+		return -ENOMEM;
+
+	err = nvme_identify_ctrl(dev_fd(dev), id);
+	if (err) {
+		if (err < 0) {
+			nvme_show_error("identify-controller: %s", nvme_strerror(errno));
+		} else {
+			fprintf(stderr, "identify controller failed\n");
+			nvme_show_status(err);
+		}
+		return err;
+	}
+
+	if (id->ctratt & NVME_CTRL_CTRATT_NAMESPACE_GRANULARITY) {
+		gr_list = nvme_alloc(sizeof(*gr_list));
+		if (!gr_list)
+			return -ENOMEM;
+
+		if (!nvme_identify_ns_granularity(dev_fd(dev), gr_list)) {
+			struct nvme_id_ns_granularity_desc *desc;
+			int index = cfg.flbas;
+
+			/* FIXME: add a proper bitmask to libnvme */
+			if (!(gr_list->attributes & 1)) {
+				/* Only the first descriptor is valid */
+				index = 0;
+			} else if (index > gr_list->num_descriptors) {
+				/*
+				 * The descriptor will contain only zeroes
+				 * so we don't need to read it.
+				 */
+				goto parse_lba;
+			}
+			desc = &gr_list->entry[index];
+
+			if (desc->nszegran && desc->nszegran < align_nsze)
+				align_nsze = desc->nszegran;
+			if (desc->ncapgran && desc->ncapgran < align_ncap)
+				align_ncap = desc->ncapgran;
+		}
+	}
+
+parse_lba:
+	err = parse_lba_num_si(dev, "nsze", cfg.nsze_si, cfg.flbas, &cfg.nsze, align_nsze);
 	if (err)
 		return err;
 
-	err = parse_lba_num_si(dev, "ncap", cfg.ncap_si, cfg.flbas, &cfg.ncap);
+	err = parse_lba_num_si(dev, "ncap", cfg.ncap_si, cfg.flbas, &cfg.ncap, align_ncap);
 	if (err)
 		return err;
 
--- a/Documentation/nvme-create-ns.txt
+++ b/Documentation/nvme-create-ns.txt
@@ -98,13 +98,15 @@ OPTIONS
 
 -S::
 --nsze-si::
-	The namespace size (NSZE) in standard SI units.
+	The namespace size (NSZE) in standard SI units (aligned on 1Mib boundaries,
+	unless the controller recommends a smaller value).
 	The value SI suffixed is divided by the namespace LBA size to set as NSZE.
 	If the value not suffixed it is set as same with the nsze option.
 
 -C::
 --ncap-si::
-	The namespace capacity (NCAP) in standard SI units.
+	The namespace capacity (NCAP) in standard SI units (aligned on 1Mib boundaries,
+	unless the controller recommends a smaller value).
 	The value SI suffixed is divided by the namespace LBA size to set as NCAP.
 	If the value not suffixed it is set as same with the ncap option.
 


ikegami-t:
The `-S` and `-C` options were added for the issue #1747 proposed to accept SI units. Is this PR changes still able to keep and follow the issue requirement? By the way the `-s` and `-c` options can be used to set the binaly units instead as below I think. But those binary units values are not divided by the LBA size as different from the `-S` and `-C` options. Is it acceptable to use the binary units options instead of the PR changes?
```sh
$ nvme create-ns /dev/nvme0 -s 80Mi -c 80Mi -f 0
```

igaw:
I think we should have one pair of options which take value verbatim without any modifications and one set which do `the right thing`. The second set is for the casual user which just wants to create a new NS and doesn't want to figure all the low level things, e.g. I want a 8GB namespace.

IIRC, this is what we had in mind when we added the `--nsze-si/-S` and `--ncap-si/-C` option. This patch just brings it a bit closer to `just do the right thing`.
                                            SI units

ikegami-t:
By the way for example is there any user to use the -S and -C options to align the values to 512 Kilobyte boundaries but not 1 Megabyte boundaries. If so in the case the options are not able to use for the case with the PR changes.

HaroPanosyan:
Not sure if namespace granularity (CNS 16h) is being accounted for in current implementation, but if controller supports it then granularity can be accounted for when creating namespaces (with same matching indices 0-15 as for LBA formats).

igaw:
> namespace granularity (CNS 16h)

Not yet. For anyone looking for the documentation it is in `NVM Command Set Specification (4.1.5.6)`. Sounds reasonable to use it if it's available.

HaroPanosyan:
> > namespace granularity (CNS 16h)
> 
> Not yet. For anyone looking for the documentation it is in `NVM Command Set Specification (4.1.5.6 Namespace Granularity List (CNS 16h))`. Sounds reasonable to use it if it's available.

Yup, CNS 01h will report if granularity is supported at bit 7 for "Controller Attributes (CTRATT)"

ikegami-t:
If the granularity is not support should we keep the current command options behavir? Or the controllers needed the size aligned to a 1 megabyte boundary are supported the granularity?

igaw:
> If the granularity is not support should we keep the current command options behavir? Or the controllers needed the size aligned to a 1 megabyte boundary are supported the granularity?

I think it makes sense to align it to 1 MB boundary in absence of the granularity information. The worst case scenario is that we waste 1MB disk space. If you look at  partitioning tools, it is not uncommon that some space is unused to alignment constrains. 

As I said, I see these options here as the user friendly command line options to create namespaces. You should not need to know the spec at all to create a new namespace. As long we have the low level command line arguments around, there is always the possibility to configure a namespace exactly as you want it do be.

ikegami-t:
Agreed. Thanks for your explanation.

igaw:
@maurizio-lombardi do you plan to add the Granularity discovery bit?

maurizio-lombardi:
> @maurizio-lombardi do you plan to add the Granularity discovery bit?

I am going to look at how to do that.

igaw:
@maurizio-lombardi any updates?

maurizio-lombardi:
7. 12. 2023 v 14:27, Daniel Wagner ***@***.***>:﻿
@maurizio-lombardi any updates?Hello, sorry I was busy with other things, I will come up with something as soon as possible

maurizio-lombardi:
I updated the patch but I didn't have the opportunity to test it yet so it may not work as intended. I plan to test it this week

nvme.c
```diff
@@ -3075,11 +3084,43 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 		return -EINVAL;
 	}
 
-	err = parse_lba_num_si(dev, "nsze", cfg.nsze_si, cfg.flbas, &cfg.nsze);
+	id = nvme_alloc(sizeof(*id));
```
ikegami-t, line 3087: The allocated `id` memory is not freed after the usage. By the way is it necessary to allocate really since seems just use as the `id` variable but not a pointer?

nvme.c
```diff
@@ -3075,11 +3084,43 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
+		} else {
+			fprintf(stderr, "identify controller failed\n");
+			nvme_show_status(err);
+		}
+		return err;
+	}
+
+	align_nsze = align_ncap = 1 << 20; /* Default 1 Mb */
+	if (id->ctratt & NVME_CTRL_CTRATT_NAMESPACE_GRANULARITY) {
+		gr_list = nvme_alloc(sizeof(*gr_list));
```
ikegami-t, line 3104: Is this also needed to allocate the gr_list memory insted of a variable but not pointer?

nvme.c
maurizio-lombardi: The pointer is not freed but the nvme process terminates right after the command execution so any memory leak is short-lived, but I can add a free if people prefer, not a problem for me. About the calls to malloc(), those two structures (id and gr_list) are quite big, in the order of a few kilobytes, so it comes natural to me to allocate them on the heap. But, again, it's probably not really necessary and I have not a strong reason to defend it.

igaw: Just do this here:

```c
_cleanup_free_ struct nvme_id_ctrl *id = NULL;
_cleanup_free_ struct nvme_id_ns_granularity_list *gr_list = NULL;
```

and we should be fine. The reason why we really take care of freeing memory is to be able to use valgrind to figure out if we are leaking memory somewhere where it matters. Keeping the house clean is the motto :)

igaw: And yes these structures should be allocated via `nvme_alloc` because of the alignment constraints of certain devices.

ikegami-t: I see. Thanks for your responses.

igaw:
I've lost the overview here. Is the patch ready to be merged (tested)? It looks it is in shape.

maurizio-lombardi:
> I've lost the overview here. Is the patch ready to be merged (tested)? It looks it is in shape.

It is not, on one of my machines I get quite high values from the granularity list, for example 4 Gb of preferred granularity. (desc->nszegran = 4294967296 desc->ncapgran = 4294967296). I didn't expect the values to be so high and this makes some u32 variables to overflow, leading to divisions by zero and crashes.
Plus it makes it impossible to create, for example, a namespace with size 100Mb because it rounds up to a much larger value.

maurizio-lombardi:
> It is not, on one of my machines I get quite high values from the granularity list.

By the way, the specification says that the list gives us a *preferred* granularity, but doesn't say that we are obliged to respect it

igaw:
Uff, this is a big alignment requirement! Does this mean we stick your initial idea to align to 1 MB ? Not really sure what to do here.

maurizio-lombardi:
> Uff, this is a big alignment requirement! Does this mean we stick your initial idea to align to 1 MB ? Not really sure what to do here.

Me neither, I am tempted to align to 1 Mb unless the controller suggests that a smaller value is preferred.

igaw:
Maybe we should just do this and see if someone complains :)

maurizio-lombardi:
> Maybe we should just do this and see if someone complains :)

Will modify the patch and go ahead then, unless @ikegami-t  disagrees

nvme.c
```diff
@@ -3075,11 +3084,58 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
+		if (!nvme_identify_ns_granularity(dev_fd(dev), gr_list)) {
+			struct nvme_id_ns_granularity_desc *desc;
+			int index = cfg.flbas;
+
+			/* FIXME: add a proper bitmask to libnvme */
+			if (!(gr_list->attributes & 1)) {
+				/* Only the first descriptor is valid */
+				index = 0;
+			} else if (index > gr_list->num_descriptors) {
+				/* The descriptor will contain only zeroes
```
ikegami-t, line 3118: Should be split the line to start multiple lines comment as belwo.
```diff
- 				/* The descriptor will contain only zeroes
+ 				/*
+ 				 * The descriptor will contain only zeroes
```

nvme.c
```diff
@@ -3075,11 +3084,58 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
+
+			if (desc->nszegran && desc->nszegran < align_nsze)
+				align_nsze = desc->nszegran;
+			if (desc->ncapgran && desc->ncapgran < align_ncap)
+				align_ncap = desc->ncapgran;
+		}
+	}
+
+parse_lba:
+
```
ikegami-t, line 3133: No need this blank line as below.
```c
  parse_lba:
  	err = parse_lba_num_si(dev, "nsze", cfg.nsze_si, cfg.flbas, &cfg.nsze, align_nsze);
```

nvme.c
```diff
@@ -3075,11 +3084,58 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
+		if (err < 0) {
+			nvme_show_error("identify-controller: %s", nvme_strerror(errno));
+		} else {
+			fprintf(stderr, "identify controller failed\n");
+			nvme_show_status(err);
+		}
+		return err;
+	}
+
+	align_nsze = align_ncap = 1 << 20; /* Default 1 Mb */
```
ikegami-t, line 3102: The value can be initialized by the variable definitions as for example. By the way `1 Mb` should be `1 MiB`. (The lowercase character `b` also should be uppercase charadter `B`. This is also same for the commit summary `1Mib` and description `1Mb`.)
```c
	__u32 nsid:
	__u32 align_nsze = 1 << 20; /* Default 1 MiB */
	__u32 align_ncap = align_nsze;
```

nvme.c
```diff
@@ -3075,11 +3084,58 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
 		return -EINVAL;
 	}
 
-	err = parse_lba_num_si(dev, "nsze", cfg.nsze_si, cfg.flbas, &cfg.nsze);
+	id = nvme_alloc(sizeof(*id));
+	if (!id)
+		return -ENOMEM;
```
ikegami-t, line 3089: Just FYI: It may needed to output any error message but not just the error return. This is not only for the changes but exsiting other codes using `nvme_alloc()` also same. So seems better the error message output by the function itself.

nvme.c
```diff
@@ -3075,11 +3084,58 @@ static int create_ns(int argc, char **argv, struct command *cmd, struct plugin *
...
+		}
+		return err;
+	}
+
+	align_nsze = align_ncap = 1 << 20; /* Default 1 Mb */
+
+	if (id->ctratt & NVME_CTRL_CTRATT_NAMESPACE_GRANULARITY) {
+		gr_list = nvme_alloc(sizeof(*gr_list));
+		if (!gr_list)
+			return -ENOMEM;
```
ikegami-t, line 3107: This only error retrun also same.

ikegami-t (commented):
Basically looks good but only some minor comments let me mention.

nvme.c
maurizio-lombardi: Ok

nvme.c
igaw: Testing small allocation fails is pretty futile in userspace, because you will almost certainty never see it. And thus most of the error paths will not work because no one really tests this.

Libraries like GLib just call `exit` on small allocation fails. For bigger allocations it might makes sense to test if it fails and handle it. 

Even worse is that if the small allocation fails printing any error message is also going to fail because the printing path is likely to use memory allocation as well. 

If we really want to improve the situation I'd say we make `nvme_alloc` fail hard on memory allocation failures and remove all the untested `-ENOMEM` code paths.

igaw:
Thanks!

PR #2098, opened 2023-10-20
nvme: allow set-features to take input from stdin

jgu-pure:
Fixes: fed9e5ee04cd ("nvme: auto free nvme_dev resource")

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5875,16 +5875,18 @@ static int set_feature(int argc, char **argv, struct command *cmd, struct plugin
 		if (cfg.feature_id == NVME_FEAT_FID_TIMESTAMP && cfg.value) {
 			memcpy(buf, &cfg.value, NVME_FEAT_TIMESTAMP_DATA_SIZE);
 		} else {
-			if (strlen(cfg.file)) {
+			if (strlen(cfg.file))
 				ffd = open(cfg.file, O_RDONLY);
-				if (ffd <= 0) {
-					nvme_show_error("Failed to open file %s: %s",
-							cfg.file, strerror(errno));
-					return -EINVAL;
-				}
+			else
+				ffd = dup(STDIN_FILENO);
+
+			if (ffd < 0) {
+				nvme_show_error("Failed to open file %s: %s",
+						cfg.file, strerror(errno));
+				return -EINVAL;
 			}
 
-			err = read(ffd, (void *)buf, cfg.data_len);
+			err = read(ffd, buf, cfg.data_len);
 			if (err < 0) {
 				nvme_show_error("failed to read data buffer from input file: %s",
 						strerror(errno));


nvme.c
```diff
@@ -5888,6 +5888,8 @@ static int set_feature(int argc, char **argv, struct command *cmd, struct plugin
 							cfg.file, strerror(errno));
 					return -EINVAL;
 				}
+			} else {
+				ffd = STDIN_FILENO;
```
ikegami-t, line 5892: Seems the stdin stream will be closed by the close_file() for cleanup. Is it expected?

jgu-pure: Ah, no, sorry. Will fix.

nvme.c
```diff
@@ -5888,9 +5888,10 @@ static int set_feature(int argc, char **argv, struct command *cmd, struct plugin
 							cfg.file, strerror(errno));
 					return -EINVAL;
 				}
+				err = read(ffd, (void *)buf, cfg.data_len);
+			} else {
+				err = read(STDIN_FILENO, (void *)buf, cfg.data_len);
```
calebsander, line 5893: Pre-existing, but doesn't seem like these `void *` casts should be necessary, as that's already the type of `buf`

calebsander (approved):

nvme.c
igaw: Why not just dup the `stdin` so we just can use the common close path.

igaw: Ignore my comment. I though it's something different. This looks good.

There is still some more cleanup stuff to do with the auto close stuff, e.g. `submit_io` could need some cleanup and I thought we were discussing this here.

igaw: Hmm, but I still prefer this here:

```diff
-                       if (strlen(cfg.file)) {
+                       if (strlen(cfg.file))
                                ffd = open(cfg.file, O_RDONLY);
-                               if (ffd <= 0) {
-                                       nvme_show_error("Failed to open file %s: %s",
-                                                       cfg.file, strerror(errno));
-                                       return -EINVAL;
-                               }
-                               err = read(ffd, buf, cfg.data_len);
-                       } else {
-                               err = read(STDIN_FILENO, buf, cfg.data_len);
+                       else
+                               ffd = dup(STDIN_FILENO);
+
+                       if (ffd < 0) {
+                               nvme_show_error("Failed to open file %s: %s",
+                                               cfg.file, strerror(errno));
+                               return -EINVAL;
                        }
+
+                       err = read(ffd, buf, cfg.data_len);
                        if (err < 0) {
                                nvme_show_error("failed to read data buffer from input file: %s",
                                                strerror(errno));
```

Could you update the patch this way? Thanks!

jgu-pure: Done!

igaw:
Thanks!

PR #2099, opened 2023-10-20
nvme-copy: support cross-namespace copy

jgu-pure:
Add support for NVMe TP4130 ("Cross-Namespace Copy")
- Add Copy Descriptor Formats 2h and 3h
- Add "--snsids" option to specify source namespaces to copy from
- Add "--sopts" option to specify source options (Fast Copy Only)
- Print new Host Behavior Support bits, new ONCS bits, and new Optional Copy Formats Supported bits
- Extend unit test case to test cross-namespace copy formats

### DIFF
--- a/libnvme-wrap.c
+++ b/libnvme-wrap.c
@@ -42,6 +42,18 @@ VOID_FN(nvme_init_copy_range_f1,
 	      __u32 *elbats, __u16 nr),
 	ARGS(copy, nlbs, slbas, eilbrts, elbatms, elbats, nr))
 
+VOID_FN(nvme_init_copy_range_f2,
+	PROTO(struct nvme_copy_range_f2 *copy, __u32 *snsids,
+	      __u16 *nlbs, __u64 *slbas, __u16 *sopts, __u32 *eilbrts,
+	      __u32 *elbatms, __u32 *elbats, __u16 nr),
+	ARGS(copy, snsids, nlbs, slbas, sopts, eilbrts, elbatms, elbats, nr))
+
+VOID_FN(nvme_init_copy_range_f3,
+	PROTO(struct nvme_copy_range_f3 *copy, __u32 *snsids,
+	      __u16 *nlbs, __u64 *slbas, __u16 *sopts, __u64 *eilbrts,
+	      __u32 *elbatms, __u32 *elbats, __u16 nr),
+	ARGS(copy, snsids, nlbs, slbas, sopts, eilbrts, elbatms, elbats, nr))
+
 FN(nvme_get_feature_length2,
 	int,
 	PROTO(int fid, __u32 cdw11, enum nvme_data_tfr dir,
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -2167,19 +2167,25 @@ static void stdout_id_ctrl_cqes(__u8 cqes)
 static void stdout_id_ctrl_oncs(__le16 ctrl_oncs)
 {
 	__u16 oncs = le16_to_cpu(ctrl_oncs);
-	__u16 rsvd = (oncs & 0xFE00) >> 9;
-	__u16 copy = (oncs & 0x100) >> 8;
-	__u16 vrfy = (oncs & 0x80) >> 7;
-	__u16 tmst = (oncs & 0x40) >> 6;
-	__u16 resv = (oncs & 0x20) >> 5;
-	__u16 save = (oncs & 0x10) >> 4;
-	__u16 wzro = (oncs & 0x8) >> 3;
-	__u16 dsms = (oncs & 0x4) >> 2;
-	__u16 wunc = (oncs & 0x2) >> 1;
-	__u16 cmp = oncs & 0x1;
+	__u16 rsvd = oncs >> 11;
+	bool afc  = !!(oncs & NVME_CTRL_ONCS_ALL_FAST_COPY);
+	bool csa  = !!(oncs & NVME_CTRL_ONCS_COPY_SINGLE_ATOMICITY);
+	bool copy = !!(oncs & NVME_CTRL_ONCS_COPY);
+	bool vrfy = !!(oncs & NVME_CTRL_ONCS_VERIFY);
+	bool tmst = !!(oncs & NVME_CTRL_ONCS_TIMESTAMP);
+	bool resv = !!(oncs & NVME_CTRL_ONCS_RESERVATIONS);
+	bool save = !!(oncs & NVME_CTRL_ONCS_SAVE_FEATURES);
+	bool wzro = !!(oncs & NVME_CTRL_ONCS_WRITE_ZEROES);
+	bool dsms = !!(oncs & NVME_CTRL_ONCS_DSM);
+	bool wunc = !!(oncs & NVME_CTRL_ONCS_WRITE_UNCORRECTABLE);
+	bool cmp  = !!(oncs & NVME_CTRL_ONCS_COMPARE);
 
 	if (rsvd)
-		printf(" [15:9] : %#x\tReserved\n", rsvd);
+		printf(" [15:11] : %#x\tReserved\n", rsvd);
+	printf("  [10:10] : %#x\tAll Fast Copy %sSupported\n",
+		afc, afc ? "" : "Not ");
+	printf("  [9:9] : %#x\tCopy Single Atomicity %sSupported\n",
+		csa, csa ? "" : "Not ");
 	printf("  [8:8] : %#x\tCopy %sSupported\n",
 		copy, copy ? "" : "Not ");
 	printf("  [7:7] : %#x\tVerify %sSupported\n",
@@ -2288,15 +2294,16 @@ static void stdout_id_ctrl_nwpc(__u8 nwpc)
 static void stdout_id_ctrl_ocfs(__le16 ctrl_ocfs)
 {
 	__u16 ocfs = le16_to_cpu(ctrl_ocfs);
-	__u16 rsvd = (ocfs & 0xfffc) >> 2;
-	__u8 copy_fmt_1 = (ocfs >> 1) & 0x1;
-	__u8 copy_fmt_0 = ocfs & 0x1;
+	__u16 rsvd = ocfs >> 4;
+	__u8 copy_fmt_supported;
+	int copy_fmt;
 	if (rsvd)
-		printf("  [15:2] : %#x\tReserved\n", rsvd);
-	printf("  [1:1] : %#x\tController Copy Format 1h %sSupported\n",
-		copy_fmt_1, copy_fmt_1 ? "" : "Not ");
-	printf("  [0:0] : %#x\tController Copy Format 0h %sSupported\n",
-		copy_fmt_0, copy_fmt_0 ? "" : "Not ");
+		printf("  [15:4] : %#x\tReserved\n", rsvd);
+	for (copy_fmt = 3; copy_fmt >= 0; copy_fmt--) {
+		copy_fmt_supported = ocfs >> copy_fmt & 1;
+		printf("  [%d:%d] : %#x\tController Copy Format %xh %sSupported\n", copy_fmt, copy_fmt,
+			copy_fmt_supported, copy_fmt, copy_fmt_supported ? "" : "Not ");
+	}
 	printf("\n");
 }
 
@@ -4483,6 +4490,10 @@ static void stdout_feature_show_fields(enum nvme_features_id fid,
 			       host_behavior->etdas ? "True" : "False");
 			printf("\tLBA Format Extension Enable (LBAFEE): %s\n",
 			       host_behavior->lbafee ? "True" : "False");
+			printf("\tCopy Descriptor Format 2h Enabled (CDFE): %s\n",
+			       host_behavior->cdfe & (1 << 2) ? "True" : "False");
+			printf("\tCopy Descriptor Format 3h Enabled (CDFE): %s\n",
+			       host_behavior->cdfe & (1 << 3) ? "True" : "False");
 		}
 		break;
 	case NVME_FEAT_FID_SANITIZE:
--- a/nvme.c
+++ b/nvme.c
@@ -6552,6 +6552,8 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 	const char *d_sdlba = "64-bit addr of first destination logical block";
 	const char *d_slbas = "64-bit addr of first block per range (comma-separated list)";
 	const char *d_nlbs = "number of blocks per range (comma-separated list, zeroes-based values)";
+	const char *d_snsids = "source namespace identifier per range (comma-separated list)";
+	const char *d_sopts = "source options per range (comma-separated list)";
 	const char *d_lr = "limited retry";
 	const char *d_fua = "force unit access";
 	const char *d_prinfor = "protection information and check field (read part)";
@@ -6567,22 +6569,26 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 	const char *d_format = "source range entry format";
 
 	_cleanup_nvme_dev_ struct nvme_dev *dev = NULL;
-	uint16_t nr, nb, ns, nrts, natms, nats;
-	__u16 nlbs[128] = { 0 };
-	__u64 slbas[128] = { 0 };
+	__u16 nr, nb, ns, nrts, natms, nats, nids;
+	__u16 nlbs[256] = { 0 };
+	__u64 slbas[256] = { 0 };
+	__u32 snsids[256] = { 0 };
+	__u16 sopts[256] = { 0 };
 	int err;
 
 	union {
-		__u32 f0[128];
-		__u64 f1[101];
+		__u32 short_pi[256];
+		__u64 long_pi[256];
 	} eilbrts;
 
-	__u32 elbatms[128] = { 0 };
-	__u32 elbats[128] = { 0 };
+	__u32 elbatms[256] = { 0 };
+	__u32 elbats[256] = { 0 };
 
 	union {
-		struct nvme_copy_range f0[128];
-		struct nvme_copy_range_f1 f1[101];
+		struct nvme_copy_range f0[256];
+		struct nvme_copy_range_f1 f1[256];
+		struct nvme_copy_range_f2 f2[256];
+		struct nvme_copy_range_f3 f3[256];
 	} *copy;
 
 	struct config {
@@ -6590,6 +6596,8 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 		__u64	sdlba;
 		char	*slbas;
 		char	*nlbs;
+		char	*snsids;
+		char	*sopts;
 		bool	lr;
 		bool	fua;
 		__u8	prinfow;
@@ -6610,6 +6618,8 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 		.sdlba		= 0,
 		.slbas		= "",
 		.nlbs		= "",
+		.snsids		= "",
+		.sopts		= "",
 		.lr		= false,
 		.fua		= false,
 		.prinfow	= 0,
@@ -6630,6 +6640,8 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 		  OPT_SUFFIX("sdlba",                'd', &cfg.sdlba,		d_sdlba),
 		  OPT_LIST("slbs",                   's', &cfg.slbas,		d_slbas),
 		  OPT_LIST("blocks",                 'b', &cfg.nlbs,		d_nlbs),
+		  OPT_LIST("snsids",                 'N', &cfg.snsids,		d_snsids),
+		  OPT_LIST("sopts",                  'O', &cfg.sopts,		d_sopts),
 		  OPT_FLAG("limited-retry",          'l', &cfg.lr,		d_lr),
 		  OPT_FLAG("force-unit-access",      'f', &cfg.fua,		d_fua),
 		  OPT_BYTE("prinfow",                'p', &cfg.prinfow,		d_prinfow),
@@ -6650,13 +6662,15 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 
 	nb = argconfig_parse_comma_sep_array_u16(cfg.nlbs, nlbs, ARRAY_SIZE(nlbs));
 	ns = argconfig_parse_comma_sep_array_u64(cfg.slbas, slbas, ARRAY_SIZE(slbas));
+	nids = argconfig_parse_comma_sep_array_u32(cfg.snsids, snsids, ARRAY_SIZE(snsids));
+	argconfig_parse_comma_sep_array_u16(cfg.sopts, sopts, ARRAY_SIZE(sopts));
 
-	if (cfg.format == 0) {
-		nrts = argconfig_parse_comma_sep_array_u32(cfg.eilbrts, eilbrts.f0,
-							   ARRAY_SIZE(eilbrts.f0));
-	} else if (cfg.format == 1) {
-		nrts = argconfig_parse_comma_sep_array_u64(cfg.eilbrts, eilbrts.f1,
-							   ARRAY_SIZE(eilbrts.f1));
+	if (cfg.format == 0 || cfg.format == 2) {
+		nrts = argconfig_parse_comma_sep_array_u32(cfg.eilbrts, eilbrts.short_pi,
+							   ARRAY_SIZE(eilbrts.short_pi));
+	} else if (cfg.format == 1 || cfg.format == 3) {
+		nrts = argconfig_parse_comma_sep_array_u64(cfg.eilbrts, eilbrts.long_pi,
+							   ARRAY_SIZE(eilbrts.long_pi));
 	} else {
 		nvme_show_error("invalid format");
 		return -EINVAL;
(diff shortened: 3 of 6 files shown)


ikegami-t:
Documentation/nvme-copy.1 and Documentation/nvme-copy.html will be regenerated from Documentation/nvme-copy.txt by the docbook, etc. so not necessary to update but only neeed to update Documentation/nvme-copy.txt.

nvme-print-stdout.c
```diff
@@ -2167,7 +2167,9 @@ static void stdout_id_ctrl_cqes(__u8 cqes)
 static void stdout_id_ctrl_oncs(__le16 ctrl_oncs)
 {
 	__u16 oncs = le16_to_cpu(ctrl_oncs);
-	__u16 rsvd = (oncs & 0xFE00) >> 9;
+	__u16 rsvd = (oncs & 0xF800) >> 11;
+	__u16 afc  = (oncs & 0x400) >> 10;
+	__u16 csa  = (oncs & 0x200) >> 9;
```
calebsander, line 2172: I see these are following the existing style, but seems like this would be simpler?
```c
__u16 rsvd = oncs >> 11;
bool afc = oncs >> 10 & 1;
bool csa = oncs >> 9 & 1;
```
But probably would be even better to use the constants you defined in `libnvme`, e.g. `NVME_CTRL_ONCS_COPY_SINGLE_ATOMICITY`.

nvme-print-stdout.c
```diff
@@ -2289,10 +2295,16 @@ static void stdout_id_ctrl_ocfs(__le16 ctrl_ocfs)
 {
 	__u16 ocfs = le16_to_cpu(ctrl_ocfs);
 	__u16 rsvd = (ocfs & 0xfffc) >> 2;
```
calebsander, line 2297: This shift needs updating?

nvme-print-stdout.c
```diff
@@ -2289,10 +2295,16 @@ static void stdout_id_ctrl_ocfs(__le16 ctrl_ocfs)
...
 	if (rsvd)
-		printf("  [15:2] : %#x\tReserved\n", rsvd);
+		printf("  [15:4] : %#x\tReserved\n", rsvd);
+	printf("  [3:3] : %#x\tController Copy Format 3h %sSupported\n",
+		copy_fmt_3, copy_fmt_3 ? "" : "Not ");
+	printf("  [2:2] : %#x\tController Copy Format 2h %sSupported\n",
+		copy_fmt_2, copy_fmt_2 ? "" : "Not ");
 	printf("  [1:1] : %#x\tController Copy Format 1h %sSupported\n",
 		copy_fmt_1, copy_fmt_1 ? "" : "Not ");
 	printf("  [0:0] : %#x\tController Copy Format 0h %sSupported\n",
```
calebsander, line 2310: Since all these bits are treated similarly, would it make more sense to just have a loop over the 4 formats/bits?

nvme-print-stdout.c
```diff
@@ -4474,8 +4486,13 @@ static void stdout_feature_show_fields(enum nvme_features_id fid,
 		stdout_lba_status_info(result);
 		break;
 	case NVME_FEAT_FID_HOST_BEHAVIOR:
-		if (buf)
+		if (buf) {
 			printf("\tHost Behavior Support: %s\n", (buf[0] & 0x1) ? "True" : "False");
+			printf("\tCopy Descriptor Format 2h Enabled: %s\n",
+					(buf[4] & 0x4) ? "True" : "False");
+			printf("\tCopy Descriptor Format 3h Enabled: %s\n",
+					(buf[4] & 0x8) ? "True" : "False");
```
calebsander, line 4494: Use `1 << 2` and `1 << 3` for clarity?

nvme-print-stdout.c
```diff
@@ -4474,8 +4486,13 @@ static void stdout_feature_show_fields(enum nvme_features_id fid,
 		stdout_lba_status_info(result);
 		break;
 	case NVME_FEAT_FID_HOST_BEHAVIOR:
-		if (buf)
+		if (buf) {
 			printf("\tHost Behavior Support: %s\n", (buf[0] & 0x1) ? "True" : "False");
```
calebsander, line 4490: Pre-existing, but this looks strange. Byte/bit 0 is "Advanced Command Retry Enable (ACRE)" according to the spec?

nvme.c
```diff
@@ -6573,14 +6575,18 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
...
 	__u64 slbas[128] = { 0 };
+	__u32 snsids[128] = { 0 };
+	__u16 sopts[128] = { 0 };
 	int err;
 
 	union {
 		__u32 f0[128];
 		__u64 f1[101];
+		__u32 f2[128];
+		__u64 f3[101];
```
calebsander, line 6589: Can `f2` and `f3` reuse `f0` and `f1`? That would avoid some code duplication below.

nvme.c
```diff
@@ -6573,14 +6575,18 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
...
+	uint16_t nr, nb, ns, nrts, natms, nats, nid;
 	__u16 nlbs[128] = { 0 };
 	__u64 slbas[128] = { 0 };
+	__u32 snsids[128] = { 0 };
+	__u16 sopts[128] = { 0 };
 	int err;
 
 	union {
 		__u32 f0[128];
 		__u64 f1[101];
```
calebsander, line 6587: Pre-existing, but where does the 101 range limit come from?

nvme.c
```diff
@@ -6573,14 +6575,18 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 	const char *d_format = "source range entry format";
 
 	_cleanup_nvme_dev_ struct nvme_dev *dev = NULL;
-	uint16_t nr, nb, ns, nrts, natms, nats;
+	uint16_t nr, nb, ns, nrts, natms, nats, nid;
```
calebsander, line 6578: naming: pluralize, i.e. "nids"?

Could also use `__u16`; that seems to be the preferred integer style.

nvme.c
```diff
@@ -6672,6 +6694,15 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
...
 
 	nr = max(nb, max(ns, max(nrts, max(natms, nats))));
+	if (cfg.format == 2 || cfg.format == 3) {
+		if (nr != nid) {
+			nvme_show_error("formats 2 and 3 require source namespace ids for each source range");
+			err = -EINVAL;
+		}
+	} else if (nid > 0) {
+		nvme_show_error("formats 0 and 1 do not support cross-namespace copy");
+		err = -EINVAL;
```
calebsander, line 6704: Should both these error cases `return`?

nvme.c
```diff
@@ -6672,6 +6694,15 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
...
+	}
 	if (!nr || nr > 128 || (cfg.format == 1 && nr > 101)) {
```
calebsander, line 6706: I don't understand the purpose of `cfg.format == 1 && nr > 101`, but looks like it should include format 3 as well?

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+        print("\nSetting up test...")
+        self.ocfs = self.get_ocfs()
+        cross_namespace_copy = self.ocfs & 0xc
+        if cross_namespace_copy:
+            # get host behavior support data
+            get_features_cmd = "nvme get-feature " + self.ctrl + \
+                    " --feature-id=0x16 -l 512 -b"
+            print("Running command:", get_features_cmd)
+            proc = subprocess.Popen(get_features_cmd,
+                                    shell=True,
```
calebsander, line 46: Shouldn't need `shell=True`, just break the arguments up into a list

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+                                    stdout=subprocess.PIPE)
+            self.host_behavior_data = proc.stdout.read()
```
calebsander, line 48: If just reading the output of the command, could use `subprocess.check_output()`?

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+            # enable cross-namespace copy formats
+            data = list(self.host_behavior_data)
```
calebsander, line 50: Why is the `list` conversion necessary? Should be possible to index a `bytes` object directly?

nvme.c
```diff
@@ -6672,6 +6694,15 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 	nats = argconfig_parse_comma_sep_array_u32(cfg.elbats, elbats, ARRAY_SIZE(elbats));
 
 	nr = max(nb, max(ns, max(nrts, max(natms, nats))));
+	if (cfg.format == 2 || cfg.format == 3) {
+		if (nr != nid) {
+			nvme_show_error("formats 2 and 3 require source namespace ids for each source range");
+			err = -EINVAL;
+		}
+	} else if (nid > 0) {
```
calebsander, line 6702: nit: don't need `> 0`

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+            if data[4] & cross_namespace_copy == cross_namespace_copy:
```
calebsander, line 51: Shorter: `if data[4] & cross_namespace_copy`

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+                # skip if already enabled
+                print("Cross-namespace copy already enabled, skipping set-features")
+                self.host_behavior_data = None
+            else:
+                data[4] = cross_namespace_copy
+                data = bytes(data)
```
calebsander, line 57: Ah if you want to update a `bytes` object, you can just concatenate the old and new pieces:
```py
data = data[:4] + cross_namespace_copy.to_bytes(2, 'little') + data[6:]
```

tests/nvme_test.py
```diff
@@ -228,6 +231,29 @@ def get_ncap(self):
...
+        max_ns_cmd = "nvme id-ctrl " + self.ctrl
+        proc = subprocess.Popen(max_ns_cmd,
+                                shell=True,
+                                stdout=subprocess.PIPE,
+                                encoding='utf-8')
+        err = proc.wait()
+        self.assertEqual(err, 0, "ERROR : reading optional copy formats supported failed")
+
+        for line in proc.stdout:
+            if pattern.match(line):
```
calebsander, line 252: Validate that a match was found?
```py
ocfs_line = next(line for line in proc.stdout if pattern.match(line))
```

tests/nvme_test.py
```diff
@@ -228,6 +231,29 @@ def get_ncap(self):
...
+    def get_ocfs(self):
+        """ Wrapper for extracting optional copy formats supported
+            - Args:
+                - None
+            - Returns:
+                - Optional Copy Formats Supported
+        """
+        pattern = re.compile("^ocfs[ ]+: 0x[0-9a-fA-F]+", re.IGNORECASE)
+        ocfs = 0
+        max_ns_cmd = "nvme id-ctrl " + self.ctrl
```
calebsander, line 243: Name doesn't really make sense

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
-                   " --blocks=" + str(self.range) + \
-                   " --slbs=" + str(self.range)
-        return self.exec_cmd(copy_cmd)
+        # skip if descriptor format not supported (default format is 0)
+        desc_format = kwargs.get("descriptor_format", 0)
+        if not self.ocfs & (1 << desc_format):
+            print(f"Skip copy because descriptor format {desc_format} is not supported")
+            return
+        # build copy command
+        copy_cmd = "nvme copy " + self.ns1 + \
```
calebsander, line 105: Can we assume Python 3? If so, `f`-strings would be much nicer for this.

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+                   " --sdlba=" + str(sdlba) + \
+                   " --blocks=" + str(blocks) + \
+                   " --slbs=" + str(slbs)
+        if kwargs.get("descriptor_format") is not None:
```
calebsander, line 109: `"descriptor_format" in kwargs`?

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+            copy_cmd += " --format=" + str(kwargs.get("descriptor_format"))
```
calebsander, line 110: `kwargs["descriptor_format"]`?

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,124 @@
...
+        if kwargs.get("snsids") is not None:
+            copy_cmd += " --snsids=" + str(kwargs.get("snsids"))
+        if kwargs.get("sopts") is not None:
+            copy_cmd += " --sopts=" + str(kwargs.get("sopts"))
+        # run and assert success
+        print("Running command:", copy_cmd)
+        self.assertEqual(self.exec_cmd(copy_cmd), 0)
```
calebsander, line 117: Huh, so this isn't really verifying anything other than that `nvme-cli` didn't reject the input or crash?

nvme.c
jgu-pure: Not sure. I think it should be ok to be 128 for all of them.

jgu-pure: Nevermind, I think it's in the spec that 1 and 3 should have 102 max? In that case though, it should be 102, not 101.

nvme.c
jgu-pure: Yes! Embarrassing that I didn't catch these on the rebase.

tests/nvme_copy_test.py
jgu-pure:
```
TypeError: 'bytes' object does not support item assignment
```

tests/nvme_test.py
jgu-pure: I think this is based off of another similar function.

tests/nvme_copy_test.py
jgu-pure: Yeah, as with the original test. I'm not sure whether we're able to do otherwise.

tests/nvme_copy_test.py
martin-belanger: You could also retrieve `descriptor_format` from the dict before testing and then just reuse its value if not `None` as follows.
```
descriptor_format = kwargs.get("descriptor_format")
if descriptor_format is not None:
    copy_cmd += f" --format={descriptor_format}"
```

tests/nvme_copy_test.py
martin-belanger: You could also use a list as follows:
```
copy_cmd = [
    "nvme_copy",
    "self.ns1",
    f"--sdlba={sdlba}",
    f"--blocks={blocks}",
    f"--slbs={slbs}",
] 
```
And later join all the elements of the list:
```
cmd = " ".join(copy_cmd)
```

tests/nvme_test.py
jgu-pure: Oh, I see. Yes, it shouldn't be `max_ns_cmd`.

tests/nvme_copy_test.py
calebsander: Yes, that is definitely more performant

nvme.c
calebsander: I don't see anything in the spec saying that there can only be 128 or 102 entries. It looks like that's how many there are in the examples, but up to 256 ranges should be allowed given that NR is an 8-bit 0-based value?

tests/nvme_copy_test.py
calebsander: Yes, so that was my suggestion below; you can reassemble the new bytes with the updated value:
```py
data = data[:4] + cross_namespace_copy.to_bytes(2, 'little') + data[6:]
```
It would be preferable to avoid iterating byte-by-byte twice.

tests/nvme_copy_test.py
calebsander: That's fine, just wanted to understand

nvme-print-stdout.c
jgu-pure: Another change?

nvme.c
jgu-pure: I think this issue goes away if we have 256 as max nr.

nvme-print-stdout.c
calebsander: I was just wondering if you understood why the existing code is doing this. It actually looks like a regression from `nvme-cli` 1.0, which displayed the actual fields:
```
$ nvme version
nvme version 1.16
$ nvme get-feature /dev/nvme0 --feature-id 0x16 -H
get-feature:0x16 (Host Behavior), Current value:00000000
	Host Behavior Support Advanced Command Retry Enable (ACRE): False
	Host Behavior Support Extended Telemetry Data Area 4 Supported (ETDAS): False
	Host Behavior Support LBA Format Extension Enable (LBAFEE): False
```
vs.
```
$ nvme version
nvme version 2.2.1 (git 2.2.1)
libnvme version 1.2 (git 1.2)
$ nvme get-feature /dev/nvme0 --feature-id 0x16 -H
get-feature:0x16 (Host Behavior), Current value:00000000
	Host Behavior Support: False
```
I can submit a separate fix for this.

nvme.c
calebsander: Agreed

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,117 @@
...
-        self.start_block = 0
-        self.range = 1
-        self.slbs = 1
-        self.namespace = 1
+        print("\nSetting up test...")
+        self.ocfs = self.get_ocfs()
+        cross_namespace_copy = self.ocfs & 0xc
+        if cross_namespace_copy:
+            # get host behavior support data
+            get_features_cmd = ["nvme", "get-feature", self.ctrl, "--feature-id=0x16", "-l 512", "-b"]
```
calebsander, line 42: Does `-l 512` not need to be 2 arguments?

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,117 @@
...
+                self.host_behavior_data = None
+            else:
+                data = self.host_behavior_data[:4] + cross_namespace_copy.to_bytes(2, 'little') + self.host_behavior_data[6:]
+                set_features_cmd = ["nvme", "set-feature", self.ctrl, "--feature-id=0x16", "-l 512"]
+                print("Running command:", " ".join(set_features_cmd))
+                proc = subprocess.Popen(set_features_cmd,
+                                        stdout=subprocess.PIPE,
+                                        stdin=subprocess.PIPE)
+                proc.communicate(input=data)
+        get_ns_id_cmd = ["nvme", "get-ns-id", f"{self.ns1}"]
```
calebsander, line 58: `ns1` is already a string, so don't see a need to format it

tests/nvme_copy_test.py
```diff
@@ -4,56 +4,117 @@
...
+                f"--slbs={slbs}",
+        ]
+        if "snsids" in kwargs:
+            copy_cmd.append(f"--snsids={kwargs['snsids']}")
+        if "sopts" in kwargs:
+            copy_cmd.append(f"--sopts={kwargs['sopts']}")
+        # run and assert success
+        copy_cmd = " ".join(copy_cmd)
+        print("Running command:", copy_cmd)
+        self.assertEqual(self.exec_cmd(copy_cmd), 0)
```
calebsander, line 110: Looks like `exec_cmd()` wants a string anyways, so may as well just build `copy_cmd` as a string

nvme-print-stdout.c
calebsander: Submitted https://github.com/linux-nvme/nvme-cli/pull/2102
You may also want to update the definition of the Host Behavior Support feature data structure to include CDFE (see https://github.com/linux-nvme/libnvme/pull/736)

jgu-pure: It looks like the original lines (with ACRE/ETDAS/etc) were added in df4368321500291676984ce7d31dde14154ccd87? Not sure where it was removed though.

tests/nvme_test.py
```diff
@@ -228,6 +231,19 @@ def get_ncap(self):
...
         return int(ncap)
 
+    def get_ocfs(self):
+        """ Wrapper for extracting optional copy formats supported
+            - Args:
+                - None
+            - Returns:
+                - Optional Copy Formats Supported
+        """
+        pattern = re.compile("^ocfs[ ]+: 0x[0-9a-fA-F]+", re.IGNORECASE)
```
calebsander, line 241: The one-element character class `[ ]` is kinda strange. Could you also add a `$` at the end to ensure the whole line matches? And I don't think `re.IGNORECASE` should be necessary.

calebsander (approved):

tests/nvme_test.py
jgu-pure:
```
-        pattern = re.compile("^ocfs[ ]+: 0x[0-9a-fA-F]+", re.IGNORECASE)
+        pattern = re.compile(r'^ocfs\s+: 0x[0-9a-fA-F]+$')
```

nvme-print-stdout.c
calebsander: It looks like the `nvme-cli` version history was rebased for 2.0. So looks like that change was just lost.

igaw: Indeed, there was a long development phase where the two branches drifted away and some things got lost in the attempt to sync those branches up again. Sorry about that.

tests/nvme_copy_test.py
igaw: It's safe to assume Python v3 by now.

igaw:
Could you rebase and resolve the conflicts? Thanks!

nvme-print-stdout.c
```diff
@@ -4483,6 +4490,10 @@ static void stdout_feature_show_fields(enum nvme_features_id fid,
 			       host_behavior->etdas ? "True" : "False");
 			printf("\tLBA Format Extension Enable (LBAFEE): %s\n",
 			       host_behavior->lbafee ? "True" : "False");
+			printf("\tCopy Descriptor Format 2h Enabled: %s\n",
+			       (buf[4] & (1 << 2)) ? "True" : "False");
```
calebsander, line 4494: Can you use the field definition you added in libnvme instead of accessing the raw byte index?

nvme-print-stdout.c
```diff
@@ -2167,19 +2167,25 @@ static void stdout_id_ctrl_cqes(__u8 cqes)
...
+	__u16 csa  = oncs >> 9 & 1;
+	__u16 copy = oncs >> 8 & 1;
+	__u16 vrfy = oncs >> 7 & 1;
+	__u16 tmst = oncs >> 6 & 1;
+	__u16 resv = oncs >> 5 & 1;
+	__u16 save = oncs >> 4 & 1;
+	__u16 wzro = oncs >> 3 & 1;
+	__u16 dsms = oncs >> 2 & 1;
+	__u16 wunc = oncs >> 1 & 1;
+	__u16 cmp  = oncs & 1;
```
calebsander, line 2181: Are there libnvme definitions for these?

jgu-pure: Yeah, I think so. While we're here, we might as well change this to use those for all of them.

nvme-print-stdout.c
jgu-pure:
```
$ nvme get-feature --feature-id 0x16 /dev/nvme0 --human-readable
get-feature:0x16 (Host Behavior), Current value:00000000
        Advanced Command Retry Enable (ACRE): False
        Extended Telemetry Data Area 4 Supported (ETDAS): False
        LBA Format Extension Enable (LBAFEE): False
        Copy Descriptor Format 2h Enabled (CDFE): False
        Copy Descriptor Format 3h Enabled (CDFE): False
```

nvme-print-stdout.c
jgu-pure:
```
nvme id-ctrl /dev/nvme0 -H | grep -A 11 "All Fast Copy"
  [10:10] : 0   All Fast Copy Not Supported
  [9:9] : 0     Copy Single Atomicity Not Supported
  [8:8] : 0     Copy Not Supported
  [7:7] : 0     Verify Not Supported
  [6:6] : 0     Timestamp Not Supported
  [5:5] : 0x1   Reservations Supported
  [4:4] : 0     Save and Select Not Supported
  [3:3] : 0x1   Write Zeroes Supported
  [2:2] : 0x1   Data Set Management Supported
  [1:1] : 0     Write Uncorrectable Not Supported
  [0:0] : 0x1   Compare Supported

```

nvme-print-stdout.c
```diff
@@ -2167,19 +2167,25 @@ static void stdout_id_ctrl_cqes(__u8 cqes)
...
-	__u16 vrfy = (oncs & 0x80) >> 7;
-	__u16 tmst = (oncs & 0x40) >> 6;
-	__u16 resv = (oncs & 0x20) >> 5;
-	__u16 save = (oncs & 0x10) >> 4;
-	__u16 wzro = (oncs & 0x8) >> 3;
-	__u16 dsms = (oncs & 0x4) >> 2;
-	__u16 wunc = (oncs & 0x2) >> 1;
-	__u16 cmp = oncs & 0x1;
+	__u16 rsvd = oncs >> 11;
+	__u16 afc  = !!(oncs & NVME_CTRL_ONCS_ALL_FAST_COPY);
```
calebsander, line 2171: These could be `bool` now

nvme-print-stdout.c
```diff
@@ -2167,19 +2167,25 @@ static void stdout_id_ctrl_cqes(__u8 cqes)
...
-	__u16 vrfy = (oncs & 0x80) >> 7;
-	__u16 tmst = (oncs & 0x40) >> 6;
-	__u16 resv = (oncs & 0x20) >> 5;
-	__u16 save = (oncs & 0x10) >> 4;
-	__u16 wzro = (oncs & 0x8) >> 3;
-	__u16 dsms = (oncs & 0x4) >> 2;
-	__u16 wunc = (oncs & 0x2) >> 1;
-	__u16 cmp = oncs & 0x1;
+	__u16 rsvd = oncs >> 11;
+	bool afc  = oncs & NVME_CTRL_ONCS_ALL_FAST_COPY;
```
calebsander, line 2171: The `!!` is still needed; otherwise the result doesn't fit in a `bool` and may wrap to `false`

jgu-pure: Thanks! Fixed.

calebsander (approved):

libnvme-wrap.c
```diff
@@ -42,6 +42,18 @@ VOID_FN(nvme_init_copy_range_f1,
 	      __u32 *elbats, __u16 nr),
 	ARGS(copy, nlbs, slbas, eilbrts, elbatms, elbats, nr))
 
+VOID_FN(nvme_init_copy_range_f2,
```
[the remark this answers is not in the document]
calebsander: Filed https://github.com/linux-nvme/nvme-cli/pull/2114

igaw:
Thanks a lot! Also for the review. Highly appreciated!

PR #2101, opened 2023-10-25
Add support for codeql sweeps

AM-K-19:
Add codeql.yml configuration file to allow routine static and vulnerability checks. By the way, codeql has detected one issue on the current head of master. I can provide additional details if required.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,70 @@
+# CodeQL build configuration for nvme-cli
+# Mostly based on auto-configuration with additions and tweaks for:
+# * meson install
+# * language detection
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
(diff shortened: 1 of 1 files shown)


igaw:
Thanks! Let's see how this goes.

igaw:
I see a warning reported:

```
Could not process some files due to syntax errors
[View file](https://github.com/linux-nvme/nvme-cli/blob/master//home/runner/work/nvme-cli/nvme-cli/subprojects/libnvme/libnvme/__init__.py#L1-L1)

A parse error occurred while processing /home/runner/work/nvme-cli/nvme-cli/subprojects/libnvme/libnvme/\_\_init\_\_.py, and as a result this file could not be analyzed. Check the syntax of the file using the python -m py\_compile command and correct any invalid syntax.
```

igaw:
We should probably exclude the subprojects from the scan.

AM-K-19:
> We should probably exclude the subprojects from the scan.

Yes, that sounds sensible. Leave it with me. I'll try to figure out how to exclude such directories from the scan.

AM-K-19:
I've created this new pull request:
https://github.com/linux-nvme/nvme-cli/pull/2115

PR #2109, opened 2023-11-02
Fix common misspellings from codespell project

yizhanglinux:
From project https://github.com/codespell-project/codespell


igaw:
- rebased and resolved merge conflict
- dropped auto generated documentation
- dropped ccan changed

igaw:
The ccan fixes should be routed to the upstream project: https://github.com/rustyrussell/ccan

Thanks!

igaw:
Thanks!

Issue #2111, opened 2023-11-02
Multiplication result may overflow 'int' before it is converted to 'unsigned long long'. 


igaw:
Multiplication result may overflow 'int' before it is converted to 'unsigned long long'. 

nvme.c:7322
```c
		/* Update the data size based on the required block count */
		buffer_size = (nblocks + 1) * logical_block_size;
```

Tracking issue for:
- [x] https://github.com/linux-nvme/nvme-cli/security/code-scanning/1

PR #2112, opened 2023-11-03
Reduce cmd call

sc108-lee:


igaw:
Added commit message

Fixes: #2111

PR #2115, opened 2023-11-06
Updates to codeql config

AM-K-19:
* Split c-cpp and python init sections
* Add paths-ignore config for pythin init
* Add json-c wrap during build - increases c codeql coverage

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -47,21 +47,22 @@ jobs:
           sudo apt-get install meson
 
     # Initializes the CodeQL tools for scanning.
-    - name: Initialize CodeQL
+    - if: matrix.language == 'c-cpp'
+      name: Initialize CodeQL C
       uses: github/codeql-action/init@v2
       with:
-        languages: ${{ matrix.language }}
-        # If you wish to specify custom queries, you can do so here or in a config file.
-        # By default, queries listed here will override any specified in a config file.
-        # Prefix the list here with "+" to use these queries and those in the config file.
+        languages: 'c-cpp'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -47,21 +47,22 @@ jobs:
...
+      name: Initialize CodeQL Python
+      uses: github/codeql-action/init@v2
+      with:
+        languages: 'python'
+        config-file: ./.github/codeql/codeql-config.yml
+  
     - name: meson build
       run: |
-         meson setup --force-fallback-for=libnvme .build
+         meson setup --force-fallback-for=libnvme,json-c .build
```
igaw, line 65: I think we should also enable the OpenSSL library here (not sure if's the right place though) because I think it's properly the most interesting thing to cover for security holes.

igaw: Or instead the fallbacks we could try to use the pre-build container. Just as an idea.

AM-K-19: I'll try adding openssl at this point and see what happens after a run through codeql. 

I don't know enough (currently) about the pre-build container approach to comment convincingly, but it sounds interesting :-).

igaw: Sorry for the confusion. I was not interested to check OpenSSL just the code inside nvme-cli which uses OpenSSL. But thinking again on it, the interesting bits are in libnvme anyway. So this is not really needed after all...  Sorray abou that too.

AM-K-19: No problem at all - understood. I've backed out the change to include openssl as a fallback.

igaw:
Thanks!

Issue #2108, opened 2023-11-02
nvme/plugins: nvme-cli 1.16x version issue


jinhuahuang2021:
Does nvme-cli release new version after v1.16?
I found some Linux system (CentOS7, CentOS8...) still use nvme-cli v.1.16.
There is little hard to build nvme-cli 2.x on CentOS 7.

igaw:
There is not active development on the 1.x branch anymore. But if there are build problems etc, I don't see any problems to release a new version. But don't expect features. What is the problem?

(And what is the problem to build 2.x? See scripts/build.sh for different build configuration including samu/muon which avoids the meson python dependencies)

jinhuahuang2021:
Got, thank you.

ikegami-t:
Tried to install on CentOS 7 by muon but failed as below. Also to resolve the failure tried to install latest gcc but failed to configure since `isl` not installed and could not find to install it also.
[build-muon.log](https://github.com/linux-nvme/nvme-cli/files/13256699/build-muon.log)
```sh
[root@localhost nvme-cli]# ./scripts/build.sh -m muon 
which: no samu in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin)
Cloning into '/root/nvme-cli/.build-ci/build-tools/samurai'...
remote: Enumerating objects: 40, done.
...
~/nvme-cli/.build-ci/build-tools/samurai ~/nvme-cli
gcc  -std=c99 -Wall -Wextra -Wshadow -Wmissing-prototypes -Wpedantic -Wno-unused-parameter -c -o build.o build.c
...
gcc  -std=c99 -Wall -Wextra -Wshadow -Wmissing-prototypes -Wpedantic -Wno-unused-parameter -c -o util.o util.c
gcc  -o samu build.o deps.o env.o graph.o htab.o log.o parse.o samu.o scan.o tool.o tree.o util.o
~/nvme-cli
which: no muon in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin)
Cloning into '/root/nvme-cli/.build-ci/build-tools/muon'...
remote: Enumerating objects: 2270, done.
...
Resolving deltas: 100% (158/158), done.
~/nvme-cli/.build-ci/build-tools/muon ~/nvme-cli
+ dir=stage1
+ mkdir -p stage1
+ pkgconf_cmd=
+ command -v pkgconf
+ command -v pkg-config
[...]

PR #2121, opened 2023-11-08
build: Add -std=c99 to CFLAGS for muon on CentOS 7

ikegami-t:


ikegami-t:
Fix for the issue #2108.

PR #2123, opened 2023-11-11
build: Delete libcurl, libarchive and libpkgconf from tools_build_muon

ikegami-t:
Since build failed by dependency on CentOS 7, CentOS 8 and Ubuntu 23.10. Also build completed by the deletion and nvme binary created and works.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -157,9 +157,6 @@ tools_build_muon() {
 
     CC="${CC}" ninja="${SAMU}" stage1/muon setup        \
         -Dprefix="${BUILDDIR}/build-tools"              \
-        -Dlibcurl=enabled                               \
-        -Dlibarchive=enabled                            \
-        -Dlibpkgconf=enabled                            \
         -Ddocs=disabled                                 \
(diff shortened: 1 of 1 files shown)


ikegami-t:
Fix for the muon build mentioned by the issue #2108.

scripts/build.sh
```diff
@@ -157,9 +157,6 @@ tools_build_muon() {
 
     CC="${CC}" ninja="${SAMU}" stage1/muon setup        \
         -Dprefix="${BUILDDIR}/build-tools"              \
-        -Dlibcurl=enabled                               \
-        -Dlibarchive=enabled                            \
-        -Dlibpkgconf=enabled                            \
```
igaw, line 162: This will not work. The only reason why it is passing the CI build is because the `muon` used is prebuild which has support for fetching libnvme.

igaw: That is surprising. It builds without these package. `muon` uses `git` directly, that's why it still works...

igaw:
I've merged this via my PR.

PR #2126, opened 2023-11-16
update build.sh script for muon builds

igaw:
A few improvements for the build.sh script for muon

- do not rebuild samurai/muon every time.
- drop dependency on `libcurl`, `libarchive` and `libpkgconfig`. The `wrap` feature of muon is using `git` directly to fetch `libnvme`.


igaw:
supersede  #/2123

Issue #2105, opened 2023-10-30
admin-passthru data_len: Invalid Argument


alexeyponomarev:
We're heavily utilizing the admin-passthru commands in our test framework including the "data_len" (-l) parameter calculating it as (NUMDU+NUMDL)*4 for our getLogPage requests via admin-passthru.

We previously used this only on the drives with MDTS=5 (128KB max transfer size) and we noticed that the data_len can be only <=MDTS, otherwise we'd get "Invalid Argument" from the tool:

```
// drive with MDTS=128K
nvme id-ctrl /dev/nvme0 | grep mdts
mdts      : 5
nvme show-regs /dev/nvme0 | grep "^cap"
cap     : 1000303c030fff

// getLogPage to LID 2 with 0x8000 NUMD, set data_len to 131072 (128KB) - passing
nvme admin-passthru --opcode 0x02 --cdw10=0x80000002  -l 131072 /dev/nvme0 -b -i /tmp/out.bin
NVMe status: Invalid Field in Command: A reserved coded value or an unsupported value in a defined field(0x4002)

// getLogPage to LID 2 with 0x8000 NUMD, set data_len to 132096 (129KB) - fail (expected, data_len > MDTS)
nvme admin-passthru --opcode 0x02 --cdw10=0x80000002  -l 132096 /dev/nvme0 -b -i /tmp/out.bin
passthru: Invalid argument
```

However with newer drives with MDTS=1MB we get Invalid Argument from the tool even on commands with data_len less than MDTS.

Issue 1 - the tool seems to _require_ libhugetlbfs when data_len >=512K, is this correct? There's no warning for the parameter limits in the man/help, but without libhugetlbfs any requests we make that are >=512K in data_len are failing with Invalid Argument
```
// drive with MDTS=1MB
nvme id-ctrl /dev/nvme1 | grep mdts
mdts      : 8
nvme show-regs /dev/nvme1 | grep "^cap"
cap     : f0003014023fff

// getLogPage to LID 2 with 0x20000 NUMD, set data_len to 524288 (512KB) - fail
nvme admin-passthru --opcode 0x02 --cdw10=0x2 --cdw11=0x2  -l 524288 /dev/nvme1 -b -i /tmp/out.bin
passthru: Invalid argument

mmap(NULL, 532480, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fd3a8264000
[...]

PR #2128, opened 2023-11-17
nvme: replace libhugetlbfs with mmap and madvise

igaw:
Instead depending on libhugetlbfs for large memory allocation mapping
just use mmap and madvise for trying to allocate continuous memory.

Only lightly tested. The small allocation works fine, though I am not so sure 
about the huge allocation part. @keithbusch is that what you had in mind?

Fixes: #2105


nvme.c
```diff
@@ -221,67 +217,53 @@ void *nvme_realloc(void *p, size_t len)
...
-		return p;
-	}
-	return NULL;
-}
-
 #define HUGE_MIN 0x80000
 
-#ifdef CONFIG_LIBHUGETLBFS
-void nvme_free_huge(void *p, bool huge)
+void nvme_free_huge(void *p, size_t len)
```
ikegami-t, line 222: Seems better to remove `huge` from the function name.

nvme.c
```diff
@@ -221,67 +217,53 @@ void *nvme_realloc(void *p, size_t len)
...
+	if (_len < HUGE_MIN) {
 		free(p);
+		return;
 	}
+
+	munmap(p, _len);
 }
 
-void *nvme_alloc_huge(size_t len, bool *huge)
+void *nvme_alloc_huge(size_t len)
```
ikegami-t, line 234: This function name is also same as for removeing `huge` from the name.

nvme.c
```diff
@@ -221,67 +217,53 @@ void *nvme_realloc(void *p, size_t len)
...
+			return p;
+		}
+	}
 
-	p = get_hugepage_region(len, GHR_DEFAULT);
-	if (!p)
-		return __nvme_alloc_huge(len, huge);
+	p = mmap(NULL, _len, PROT_READ | PROT_WRITE,
+		 MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+	if (p < 0)
```
calebsander, line 248: Don't think it is possible for a pointer to be negative? Should this be `if (p == MAP_FAILED)`?

nvme.c
```diff
@@ -221,67 +217,53 @@ void *nvme_realloc(void *p, size_t len)
...
 
-	*huge = true;
-	return p;
-}
-#else
-void nvme_free_huge(void *p, bool huge)
-{
-	free(p);
-}
+	if (madvise(p, _len, MADV_HUGEPAGE) < 0) {
```
calebsander, line 251: Is it possible to just include the `MAP_HUGETLB` flag in the `mmap()` call instead of issuing another syscall?

nvme.c
```diff
@@ -221,67 +217,53 @@ void *nvme_realloc(void *p, size_t len)
...
 
-void *nvme_alloc_huge(size_t len, bool *huge)
-{
-	return __nvme_alloc_huge(len, huge);
+	return p;
 }
-#endif
 
-void *nvme_realloc_huge(void *p, size_t len, bool *huge)
+void *nvme_realloc_huge(void *p, size_t len)
```
calebsander, line 259: Has no callers. Just remove?

nvme.c
```diff
@@ -221,67 +217,53 @@ void *nvme_realloc(void *p, size_t len)
...
 
-#ifdef CONFIG_LIBHUGETLBFS
-void nvme_free_huge(void *p, bool huge)
+void nvme_free_huge(void *p, size_t len)
 {
-	if (huge) {
-		if (p)
-			free_hugepage_region(p);
-	} else {
+	size_t _len = ROUND_UP(len, 0x1000);
```
calebsander, line 224: Just update `len` instead of introducing a new variable?

nvme.c
keithbusch: Need to also include '`MAP_HUGETLB`' right?

keithbusch:
Looks like a great improvement, but I didn't see the mmap with the MAP_HUGETLB. If that fails, we can do the same mmap without the MAP_HUGETLB flag and then run `'madvise()`' on the result with '`MADV_HUGEPAGE`' as a fallback in hopes we can get a transparent one.

nvme.c
igaw: - `MAP_FAILED` is according the man pages `-1` but yes looks better with the define.
- `MAP_HUGETLB`: was not sure if we the flag and `madvise` at the same time. From the man pages I'd get the impression either the mmap flag or the `madvise` call should do the same thing. But I am really just guessing.

nvme.c
igaw: Yes, that doesn't make really sense.

nvme.c
```diff
@@ -221,70 +217,82 @@ void *nvme_realloc(void *p, size_t len)
...
-			free_hugepage_region(p);
-	} else {
-		free(p);
+	memset(mh, 0, sizeof(*mh));
+
+	mh->len = ROUND_UP(len, 0x1000);
+
+	/*
+	 * For smaller allocation we just use malloc and hope the kernel is able
+	 * to convert to continious memory region.
```
keithbusch, line 242: contiguous or continuous?

nvme.c
```diff
@@ -221,70 +217,82 @@ void *nvme_realloc(void *p, size_t len)
...
+		return mh;
 	}
-}
 
-void *nvme_alloc_huge(size_t len, bool *huge)
-{
-	void *p;
+	/*
+	 * Larger allocation will almost certainly fail with the malloc
+	 * approach. Instead try pre-allocating memory from the TBH pool.
```
keithbusch, line 252: HugeTLB pool? Not sure what TBH means.

nvme.c
```diff
@@ -221,70 +217,82 @@ void *nvme_realloc(void *p, size_t len)
...
 
-	if (len < HUGE_MIN)
-		return __nvme_alloc_huge(len, huge);
+	/*
+	 * And if mmap fails because the pool has no pages available, try to use
+	 * malloc/madvise as fallback in order to fullfil the request. This
+	 * gives the kernel a chance to try to claim some huge pages. This might
+	 * still fail though.
+	 */
+	mh->p = nvme_alloc(mh->len);
```
keithbusch, line 266: This is going to request 4k, but if I'm reading the man pages correctly, I think we have a higher chance of transparent huge page success if we align to 2MB.

nvme.c
keithbusch: > And I found following example: https://github.com/torvalds/linux/blob/master/tools/testing/selftests/mm/hugepage-mmap.c
> 
> So I'd say it's okay just to follow the kernel example and don't implement any of the fs-based fallback strategy.

I agree, fs-based fallbacks need not be considered here.

The kernel test example you pointed to using memfd_create()'s flags appear to have the same effects as doing an anonymous mmap. I considered that way since I'm using memfd's extensively in another project, but I don't see the point if we're don't share it with other processes. Unless I'm missing something, I think memfd's are just an unnecessary extra step for nvme-cli's usage.

nvme.c
```diff
@@ -221,70 +217,82 @@ void *nvme_realloc(void *p, size_t len)
...
-		return p;
-	}
-	return NULL;
-}
+struct nvme_mem_huge *nvme_alloc_huge(size_t len)
+{
+	struct nvme_mem_huge *mh;
 
-#define HUGE_MIN 0x80000
+	mh = malloc(sizeof(*mh));
```
calebsander, line 232: Would be nice to avoid another heap allocation and pointer indirection. Could we just return `nvme_mem_huge` by value?
If you drop the `bool nvme_malloc` field, then it's 16 bytes and can be returned in registers.

nvme.c
```diff
@@ -221,70 +217,82 @@ void *nvme_realloc(void *p, size_t len)
 	return result;
 }
 
-static void *__nvme_alloc_huge(size_t len, bool *huge)
-{
+#define HUGE_MIN 0x80000
+
+struct nvme_mem_huge {
+	size_t len;
+	bool nvme_malloc; /* p has been allocated using nvme_malloc */
```
calebsander, line 224: You could probably drop this field and just recompute it based on `len`?

nvme.c
```diff
@@ -221,70 +217,82 @@ void *nvme_realloc(void *p, size_t len)
...
 
-#ifdef CONFIG_LIBHUGETLBFS
-void nvme_free_huge(void *p, bool huge)
-{
-	if (huge) {
-		if (p)
-			free_hugepage_region(p);
-	} else {
-		free(p);
+	memset(mh, 0, sizeof(*mh));
```
calebsander, line 236: Use `calloc()` instead?

nvme.c
```diff
@@ -1609,7 +1619,7 @@ static int get_persistent_event_log(int argc, char **argv,
 	}
 
 free:
-	nvme_free_huge(pevent_log_info, huge);
+	nvme_free_huge(mh);
```
calebsander, line 1622: Consider adding a `__cleanup__` implementation for `nvme_free_huge()`?

nvme.c
igaw: As far I can tell, `len` is not enough to decide if the allocation has used `mmap` or `malloc`. Or do I miss something?

nvme.c
igaw: If I understood you correctly, you would like to see both allocation function combined into one? 

Not really thought about it at so far. Surely, it's possible but I would like to defer this discussion for now.  Let's get the library dependency removed first.

nvme.c
igaw: typping is hard: contiguous

nvme.c
igaw: As pointed out below, I haven't found a way to figure how to differentiate between `mmap` allocation and the `posix_memalign` fallback when `mmap` doesn't work.

nvme.c
igaw: I am using `posix_memalign` here directly now to allocate at a 2MB boundary. So the memset is still necessary.

nvme.c
ikegami-t: I see. Yes your understainding correct and right as not a fix for the issue.

nvme.c
calebsander: Yes, you are correct. I forgot you added the malloc() fallback. There are probably ways you could fit this extra bit in the bits of the other fields that are always 0, but I don't think it's worth the effort.

nvme.c
calebsander: Regardless of its size, it would be more efficient to pass the `nvme_mem_huge` on the stack rather than heap-allocating it. It would also probably simplify the code a bit as you don't need to remember to free it in a bunch of places.

nvme.c
calebsander: I meant for the `nvme_mem_huge`, which is what's being zeroed here, not the data memory

nvme.c
igaw: Non of the callers are in any hotpath and I would be surprised if it's really possible to measure out any performance difference between the variants. Sure in a tight loop comparing heap-allocation vs stack-allocation will result a measurable difference. But I expect that noise around loading an application and stopping in a tight loop will just hide any differences between two allocation types.

I'd rather have the focus on API design and easy/correct use of it. Thus my question here is how can you avoid the call to `nvme_free_huge`? We still need to free the allocated memory somehow or do you mean by using the cleanup hooks?

calebsander: I don't mean avoiding freeing the data memory; I mean freeing the `nvme_mem_huge` allocation itself (there are 4 `free(mh)`s in the current code).
I would suggest one of the following APIs, allowing the `nvme_mem_huge` to be placed on the stack:
```c
struct nvme_mem_huge nvme_alloc_huge(size_t len);
bool nvme_alloc_success(const struct nvme_mem_huge *);
```
or
```c
bool nvme_alloc_huge(size_t len, struct nvme_mem_huge *allocation);
```

keithbusch: Yeah, it's more a convenience thing for the new API than a performance thing. Returning the descriptor by value on the stack just makes it easier to cleanup on error later. Unless you have a need for the allocation descriptor to outlive the scope that allocated it? I didn't see any use for that here, at least.

calebsander: Yeah, and the caller could always move the `nvme_mem_huge` to the heap if they needed that

nvme.c
calebsander: I think a simple solution would be to use `len == 0` to indicate that it needs to be `free()`d, because `free()` doesn't require the length. But don't feel too strongly.

util/cleanup.h
```diff
@@ -2,9 +2,12 @@
 #ifndef __CLEANUP_H
 #define __CLEANUP_H
 
+#include "nvme.h"
```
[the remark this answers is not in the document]
igaw: My editor is eager to add includes. I missed to remove this one afterwards.

nvme.c
```diff
@@ -1588,16 +1492,16 @@ static int get_persistent_event_log(int argc, char **argv,
...
 			nvme_show_status(err);
-			goto free;
+			return err;
 		}
 		pevent_collected = pevent_log_info;
 		if (pevent_collected->gen_number != pevent->gen_number) {
 			printf("Collected Persistent Event Log may be invalid,\n"
 			       "Re-read the log is required\n");
-			goto free;
+			return -EINVAL;
```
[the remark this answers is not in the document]
igaw: Sure, I'll move it into a separate patch.

nvme.c
```diff
@@ -7424,7 +7325,7 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 		printf("sts             : %02x\n", sts);
 	}
 	if (cfg.dry_run)
-		goto free_mbuffer;
```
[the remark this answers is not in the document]
igaw: `submit_io` could profit from using the cleanup hooks a bit more. I avoided to touch this in the previous refactor rampage. Let me add a patch for this part as well.

plugins/scaleflux/sfx-nvme.c
```diff
@@ -1348,13 +1349,13 @@ static int nvme_dump_evtlog(struct nvme_dev *dev, __u32 namespace_id, __u32 stor
 			    char *file, bool parse, char *output)
 {
 	struct nvme_persistent_event_log *pevent;
-	void *pevent_log_info;
+	_cleanup_free_huge_ struct nvme_mem_huge *mh = NULL;
+	struct nvme_mem_huge *pevent_log_info;
```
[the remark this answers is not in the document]
igaw: Ah sorry, changed this code a bunch of times and it got a bit messy. Cleaning it up.

nvme.c
igaw: I decided to follow the existing API and return pointer directly in `nvme_alloc_huge`. Hope this makes everyone happy :)

nvme.c
```diff
@@ -8357,8 +8242,6 @@ static int passthru(int argc, char **argv, bool admin,
 	}
 free_metadata:
 	free(mdata);
-free_data:
-	nvme_free_huge(data, huge);
 close_dfd:
```
ikegami-t, line 8245: Sorry this is not related to the changes but does not work correctly as the `close_dfd:` is not executed after the `close_mfd:` but `goto close mfd;` is executed after the `dfd` opened so needed to execute the `close_dfd:` also.

```c
	if (strlen(cfg.input_file)) {
		dfd = open(cfg.input_file, flags, mode);
		if (dfd < 0) {
			nvme_show_perror(cfg.input_file);
			return -EINVAL;
		}
	}

	if (cfg.metadata && strlen(cfg.metadata)) {
		mfd = open(cfg.metadata, flags, mode);
		if (mfd < 0) {
			nvme_show_perror(cfg.metadata);
			err = -EINVAL;
			goto close_dfd;
		}
	}

	if (cfg.metadata_len) {
		mdata = malloc(cfg.metadata_len);
		if (!mdata) {
			err = -ENOMEM;
			goto close_mfd;
		}
...
free_metadata:
	free(mdata);
close_dfd:
	if (strlen(cfg.input_file))
		close(dfd);
close_mfd:
	if (cfg.metadata && strlen(cfg.metadata))
		close(mfd);
```
By the way I think only one goto label is needed for the function as combined the three goto labels as below for example. Anyway probably this may be needed to be fixed by other PR changes.
```c
	void *mdata = NULL;
	int dfd = -1;
	int mfd = -1;
...
	if (!***) {
		err = -E***;
		goto ret;
	}
...
ret:
	free(mdata);
	if (dfd != -1 && dfd != STDIN_FILENO && dfd != STDOUT_FILENO)
		close(dfd);
	if (mfd != -1 && mfd != STDIN_FILENO && mfd != STDOUT_FILENO)
		close(mfd);
...

Comment (igaw) [reply nvme.c:8245]: Good catch! I missed the `passthru` function to use the cleanup helpers as well.

Comment (igaw) [reply nvme.c]: Fixed in the new version.

File: util/cleanup.h [hunk 19]
```diff
@@ -25,6 +27,18 @@ static inline void freep(void *p)
 }
 #define _cleanup_free_ __cleanup__(freep)
 
+static inline void free_hugep(struct nvme_mem_huge **mh)
+{
+	nvme_free_huge(*mh);
+}
+#define _cleanup_free_huge_ __cleanup__(free_hugep)
```
calebsander, line 34: I don't see any users of this. Are there plans to add some in the future?

util/mem.c
```diff
@@ -0,0 +1,123 @@
...
+	bool posix_memalign; /* p has been allocated using posix_memalign */
+	void *p;
+};
+
+void *nvme_alloc(size_t len)
+{
+	void *p;
+
+	len = ROUND_UP(len, 0x1000);
+	if (posix_memalign((void *)&p, getpagesize(), len))
```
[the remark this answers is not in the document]
calebsander: Looks like you changed the usage in `nvme_alloc_huge()`, but not this one

nvme.c
```diff
@@ -1577,9 +1479,10 @@ static int get_persistent_event_log(int argc, char **argv,
 	if (cfg.action == NVME_PEVENT_LOG_EST_CTX_AND_READ)
 		cfg.action = NVME_PEVENT_LOG_READ;
 
-	pevent_log_info = nvme_alloc_huge(cfg.log_len, &huge);
-	if (!pevent_log_info)
+	pevent_log_info = nvme_alloc_huge(cfg.log_len, &mh);
+	if (pevent_log_info)
```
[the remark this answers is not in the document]
igaw: Argh, thanks for catching this.

nvme.c
```diff
@@ -7289,33 +7195,30 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
...
-		goto close_mfd;
-	}
+	if (!ns)
+		return -ENOMEM;
 
 	err = nvme_cli_identify_ns(dev, cfg.namespace_id, ns);
 	if (err > 0) {
 		nvme_show_status(err);
-		goto close_mfd;
+		return -err;
```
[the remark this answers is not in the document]
igaw: We should properly just return `-EINVAL` because all error codes inside nvme are negative values. That was also the reason why I switched it to `-err`.

nvme.c
```diff
@@ -7473,16 +7368,6 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
...
 
-free_mbuffer:
-	free(mbuffer);
-free_buffer:
-	nvme_free_huge(buffer, huge);
-close_mfd:
-	if (strlen(cfg.metadata))
-		close(mfd);
-close_dfd:
-	close(dfd);
```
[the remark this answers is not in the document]
igaw: I was properly confused by the existing flow and this was also one of the main motivation to introduce the `cleanup` helpers. Looks indeed a bit bogus.

nvme.c
```diff
@@ -7259,7 +7158,13 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 		}
 	}
 
-	dfd = mfd = opcode & 1 ? STDIN_FILENO : STDOUT_FILENO;
+	if (opcode == nvme_cmd_write) {
```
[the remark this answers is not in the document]
igaw: I prefer to use constants because it makes it a bit more readable. But by doing so I introduced a bug.

nvme.c
```diff
@@ -7281,6 +7186,7 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	}
 
 	if (strlen(cfg.data)) {
+		close(dfd);
```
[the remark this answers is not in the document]
igaw: In the last year there were many changes in this code path. It seems it is very hard to get the various conditions right so that there is no leak etc. My approach might look a bit ugly but given the fact that we had some many bugs and ugly conditions to figure out if a file descriptor needs to be close or not I'd prefer my stupid way but easy to follow approach.

util/cleanup.h
igaw: I was not sure if I should leave it around even there is no user right now. It's fairly simple to add if needed anyway.

nvme.c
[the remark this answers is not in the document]
igaw: After looking at the rest, we have pretty much half of  error values positive and negative in nvme.c. So let's keep it. Something for later.

plugins/scaleflux/sfx-nvme.c
```diff
@@ -1453,20 +1454,21 @@ static int nvme_dump_evtlog(struct nvme_dev *dev, __u32 namespace_id, __u32 stor
 	printf("\nDump-evtlog: Success\n");
 
 	if (parse) {
-		nvme_free_huge(pevent_log_info, huge);
-		pevent_log_info = nvme_alloc_huge(log_len, &huge);
+		nvme_free_huge(&mh);
+		single_len = log_len;
```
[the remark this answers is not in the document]
igaw: Indeed, not needed. The `log_len` is not changed thus the allocated buffer has already the right size. Dropped it.

nvme.c
```diff
@@ -9039,6 +8915,16 @@ static int nvme_mi(int argc, char **argv, __u8 admin_opcode, const char *desc)
...
+		flags = O_RDONLY;
+		send = true;
+	} else {
+		fd = dup(STDOUT_FILENO);
+		flags = O_WRONLY | O_CREAT;
+		send = false;
+	}
+
 	if (strlen(cfg.input_file)) {
 		fd = open(cfg.input_file, flags, mode);
```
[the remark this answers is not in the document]
igaw: Fixed the leak.

stdin/stdout is depending on read/write direction. The second conditions is if a file is involved.

nvme.c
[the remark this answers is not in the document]
igaw: You were right :) I'd changed the way you suggested and it's better.

util/cleanup.h
```diff
@@ -25,6 +27,12 @@ static inline void freep(void *p)
...
 
+static inline void free_huge(struct nvme_mem_huge *mh)
+{
+	nvme_free_huge(mh);
+}
+#define _cleanup_huge_ __cleanup__(free_huge)
+
 static inline void close_file(int *f)
 {
 	if (*f >= 0)
```
ikegami-t, line 38: Is this not necessary to check if `*f` is not STDIN_FILENO and not STDOUT_FILENO also?

igaw: Added the filter.

plugins/scaleflux/sfx-nvme.c
```diff
@@ -1453,20 +1454,12 @@ static int nvme_dump_evtlog(struct nvme_dev *dev, __u32 namespace_id, __u32 stor
 	printf("\nDump-evtlog: Success\n");
 
 	if (parse) {
-		nvme_free_huge(pevent_log_info, huge);
-		pevent_log_info = nvme_alloc_huge(log_len, &huge);
```
calebsander, line 1457: Wait how come we can remove this new allocation? The original allocation was with `single_len`, and it seems like we now need a buffer of length `log_len`.

util/cleanup.h
```diff
@@ -25,9 +27,15 @@ static inline void freep(void *p)
 }
 #define _cleanup_free_ __cleanup__(freep)
 
+static inline void free_huge(struct nvme_mem_huge *mh)
+{
+	nvme_free_huge(mh);
+}
```
calebsander, line 33: Could just use `nvme_free_huge()` instead of defining a new function

plugins/scaleflux/sfx-nvme.c
igaw: Right, was confused. Fixed.

util/cleanup.h
igaw: Sure, fixed :)

igaw:
Thanks for the all the feedback.

util/cleanup.h
ikegami-t: Thank you.

PR #2130, opened 2023-11-20
Implement support for version 1 TLS identifiers as specified in TP8018

hreinecke:
TP8018 specifies a new TLS identifier, where the original TLS identifer is suffixed with a PSK hash. This allows to differentiate between keys with different key material and with that enables key rotation.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -8701,9 +8701,18 @@ static int gen_tls_key(int argc, char **argv, struct command *command, struct pl
 		nvme_show_error("Invalid TLS identity version %u", cfg.version);
 		return -EINVAL;
 	}
-	if (cfg.insert && !cfg.subsysnqn) {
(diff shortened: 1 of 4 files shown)


nvme.c
```diff
@@ -8684,6 +8687,7 @@ static int gen_tls_key(int argc, char **argv, struct command *command, struct pl
 		  OPT_STR("subsysnqn",	'c', &cfg.subsysnqn,	subsysnqn),
 		  OPT_STR("secret",	's', &cfg.secret,	secret),
 		  OPT_UINT("hmac",	'm', &cfg.hmac,		hmac),
+		  OPT_UINT("version",	'V', &cfg.version,	version),
```
igaw, line 8690: Are you sure that the `version` option is working here? The main argument parser is handling `version` already.

PR #2133, opened 2023-11-21
plugins/solidigm: Added re_sku_count smart atrribute

ssingonx-sd:


igaw:
@lgdacunh I suppose this is okay? Just asking because it's the first contribution from @ssingonx-sd

PR #2135, opened 2023-11-23
doc: Fix nvme-connect manpage --application option string

tbzatek:
(`--application` -> `--context`)

### DIFF
--- a/Documentation/nvme-connect.1
+++ b/Documentation/nvme-connect.1
@@ -61,7 +61,7 @@ nvme-connect \- Connect to a Fabrics controller\&.
                 [\-\-tls                        ]
                 [\-\-dump\-config            | \-O]
                 [\-\-output\-format=<fmt>    | \-o <fmt>]
-                [\-\-application=<id>                 ]
+                [\-\-context=<id>               ]
 .fi
 .SH "DESCRIPTION"
 .sp


tbzatek:
(spotted this while scouting what the application strings are about and how can we use it in our stack...)

igaw:
Thanks!

BTW, currently the blktests code is using this feature. Before `nvme connect` is called, the framework places following json config to `/run/nvme`:

```shell
        mkdir -p /run/nvme
        cat >> /run/nvme/blktests.json <<-EOF
        [
          {
            "hostnqn": "${def_hostnqn}",
            "hostid": "${def_hostid}",
            "subsystems": [
              {
                "application": "blktests",
                "nqn": "blktests-subsystem-1",
                "ports": [
                  {
                    "transport": "fc",
                    "traddr": "nn-${remote_wwnn}:pn-${remote_wwpn}",
                    "host_traddr": "nn-${local_wwnn}:pn-${local_wwpn}"
                  }
                ]
              }
            ]
          }
        ]
        EOF
```

then it uses `nvme connect --context=blktests`. When the `udev` rules trigger the auto connect, libnvme sees that the context is `blktests` and not `udev` and ignores it. So only one entity is in charge of this particular resource. Hope that helps. If not let me know.

PR #2139, opened 2023-11-30
Implement support for version 1 TLS identifiers as specified in TP8018 (version 2)

hreinecke:
TP8018 specifies a new TLS identifier, where the original TLS identifer is suffixed with a PSK hash. This allows to differentiate between keys with different key material and with that enables key rotation.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -8780,10 +8780,12 @@ static int check_tls_key(int argc, char **argv, struct command *command, struct 
 {
 	const char *desc = "Check a TLS key for NVMe PSK Interchange format.\n";
 	const char *keydata = "TLS key (in PSK Interchange format) to be validated.";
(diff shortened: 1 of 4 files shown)


nvme.c
```diff
@@ -8780,10 +8780,12 @@ static int check_tls_key(int argc, char **argv, struct command *command, struct
 {
 	const char *desc = "Check a TLS key for NVMe PSK Interchange format.\n";
 	const char *keydata = "TLS key (in PSK Interchange format) to be validated.";
+	const char *identity = "TLS identity version to use (0 = NVMe TCP 1.0c, 1 = NVMe TCP 2.0";
```
igaw, line 8778: Closing ')' missing. I'll fix this quickly before merging.

igaw (commented):
Looks good. Just the nitpick in the comment. I'll fix this.

igaw:
Thanks!

PR #2141, opened 2023-12-01
show-regs: Fix to return normal value

hmi-jeon:
If the output format is not normal, the value of err will be non-zero.
If bar is returned the mmap_registers function, it does not return as 0.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5384,6 +5384,8 @@ static int show_registers(int argc, char **argv, struct command *cmd, struct plu
 	if (err < 0) {
 		nvme_show_error("Invalid output format");
 		goto free_tree;
+	} else {
+		err = 0 ;
 	}
 
 	if (cfg.human_readable)


igaw:
I see. The `validate_output_format` is a bit awkward to use. Maybe we should just do something like

```c
int validate_output_format(const char *format, enum nvme_print_flags *flags)
{
    if (...)
        return -EINVAL;

   *flags = FLAG;

    return 0;
}

err = validate_output_format(format, &flags);
if (err)
     return err;
```

hmi-jeon:
@igaw 
Hi
Since validate_output_format function is used in all commands, it is difficult to change it, so I added err = 0.
I think it would be good if it changed according to the method you suggested.
Thank you for checking.

igaw:
>  it is difficult to change

No, it's not. It's just work.

igaw:
Could you test if my PR is fixing your problem? Thanks!

hmi-jeon:
I said something wrong. It's not that it's difficult, I just added err=0 for a simple fix. Sorry.

Could you test if my PR is fixing your problem?
-> Okay. Since I can't do it right now, can I test it next Monday?

igaw:
No worries. Next time you know what to do :)

Sure, Monday is fine. There is no hurry. Have a good weekend.

nvme.c
```diff
@@ -5384,6 +5384,8 @@ static int show_registers(int argc, char **argv, struct command *cmd, struct plu
 	if (err < 0) {
 		nvme_show_error("Invalid output format");
 		goto free_tree;
+	} else {
+		err = 0 ;
```
calebsander, line 5388: How about this instead (moving `err = flags` into the `if` statement)?
```c
flags = validate_output_format(output_format_val);
if (flags < 0) {
	nvme_show_error("Invalid output format");
	err = flags;
	goto free_tree;
}
```

hmi-jeon: That's a good idea. 
However, @igaw changed the validate_output_format function #2142 
thank you

hmi-jeon:
@igaw 
I tested #2142 and confirmed it was working properly.
Thank you.

nvme.c
igaw: @calebsander initially `flags` was just defined as `int` and I changed it to an enum but didn't update the parse function at this point. I don't really like to mix the different types, thus it was about time to also address this by changing the parse function so that return is just an error code and the `flags` variable is passed in as argument.

igaw:
Thanks for the report and testing!

PR #2143, opened 2023-12-01
nvme-print-json: include vs as string for identify namespace

cjp256:
While the spec may allow for arbitrary use of vendor-specific data, some implementations are simply strings.  JSON will encode the string, allowing for safe use of control characters.

The only requirement is that the there be a null-byte in the vendor-specific field terminating the string.  Data beyond the null-byte will be ignored.


nvme-print-json.c
```diff
@@ -267,6 +267,12 @@ static void json_nvme_id_ns(struct nvme_id_ns *ns, unsigned int nsid,
 		array_add_obj(lbafs, lbaf);
 	}
 
+	// If vs is null-terminated, include it. JSON will escape the string.
+	// Data beyond the null byte will be ignored.
+	if (strnlen((const char *)ns->vs, sizeof(ns->vs)) < sizeof(ns->vs)) {
```
ikegami-t, line 272: Please fix the checkpatch.pl warning below.
```sh
tokunori@tokunori-desktop:~/nvme-cli$ ../linux/scripts/checkpatch.pl -g HEAD
WARNING: braces {} are not necessary for single statement blocks
#29: FILE: nvme-print-json.c:272:
+       if (strnlen((const char *)ns->vs, sizeof(ns->vs)) < sizeof(ns->vs)) {
+               obj_add_str(r, "vs", (const char *)ns->vs);
+       }

total: 0 errors, 1 warnings, 12 lines checked

NOTE: For some of the reported defects, checkpatch may be able to
      mechanically convert to the typical style using --fix or --fix-inplace.

Commit eed374d65601 ("nvme-print-json: include vs as string for identify namespace") has style problems, please review.

NOTE: If any of the errors are false positives, please report
      them to the maintainer, see CHECKPATCH in MAINTAINERS.
```

nvme-print-json.c
```diff
@@ -267,6 +267,12 @@ static void json_nvme_id_ns(struct nvme_id_ns *ns, unsigned int nsid,
 		array_add_obj(lbafs, lbaf);
 	}
 
+	// If vs is null-terminated, include it. JSON will escape the string.
```
ikegami-t, line 270: Preferred to use `/*` and `*/` instaed of `//` for the comment as below as described it by https://www.kernel.org/doc/html/v6.7-rc3/process/coding-style.html#commenting.
```diff
- 	// If vs is null-terminated, include it. JSON will escape the string.
- 	// Data beyond the null byte will be ignored.
+ 	/*
+ 	 * If vs is null-terminated, include it. JSON will escape the string.
+ 	 * Data beyond the null byte will be ignored.
+ 	 */
```

nvme-print-json.c
ikegami-t: By the way is there any reason to use `strnlen` since seems enough to use just `strlen` instead? Further more seems the `if` checking itself not necessary as just add the `vs` string object without the checking.

nvme-print-json.c
cjp256: Fixed, thanks!

nvme-print-json.c
ikegami-t: The ns data including vs allocated by the `nvme_alloc()` as below then aligned and set zero so looks no problem for the reason.
```c
void *nvme_alloc(size_t len)
{
	void *p;

	len = ROUND_UP(len, 0x1000);
	if (posix_memalign((void *)&p, getpagesize(), len))
		return NULL;

	memset(p, 0, len);
	return p;
}
```
By the way the vs value is printed to stdout by using `d()` therefore can we use `d_json()` to print to json?
```c
static void stdout_id_ns(struct nvme_id_ns *ns, unsigned int nsid,
			 unsigned int lba_index, bool cap_only)
{
...
	if (vs && !cap_only) {
		printf("vs[]:\n");
		d(ns->vs, sizeof(ns->vs), 16, 1);
	}
```

igaw: @ikegami-t the problem is that the whole allocated buffer is used without the the NULL byte. This means we can't rely on the fact that it's a NULL terminated string. The memset in nvme_alloc doesn't help when len is a multiple of 0x1000. So using strnlen is the correct thing to check for the max output len.

But yes, we should use `d_json`  which handles the length correctly.

ikegami-t: I see. Just I thought currently only `strlen()` is used by nvme-cli and libnvme so seemed okay as same but understood as other code also should be fixed to use `strnlen()` as same if same case. Thank you.

cjp256: Sorry about the delay, I see the follow up PRs.  What is the intended behavior for d_json?  To print all bytes for the full length of the vs structure?

igaw: Good question. From you initial comment I gather that some implementation of `vs` are exposing NULL terminated strings but this is not a requirement. If it's a string I'd say just output the string. In the other case I don't see any other option than outputting everything. Is there a way to detect these two situations? 

After looking at the `stdout` implementation, it always prints the whole data structure. I suppose we should keep this in sync.

igaw: I decided to keep the `stdout` output and `json` output in sync. So both are using the `d` helpers. If this is not what you need, we should modify these helpers so we get consistent output everywhere.

igaw:
Merged https://github.com/linux-nvme/nvme-cli/pull/2148 instead

PR #2148, opened 2023-12-11
nvme-print-json: include vs for identify namespace

igaw:
While the spec may allow for arbitrary use of vendor-specific data, some implementations are simply strings.  JSON will encode the string, allowing for safe use of control characters.

The only requirement is that the there be a null-byte in the vendor-specific field terminating the string.  Data beyond the null-byte will be ignored.

[dwagner: use d_json instead obj_add_str]

Supersedes: #2143


igaw:
I think we can't use d_json as we don't encode all values instead we just replace the non printable values as '.'.

PR #2149, opened 2023-12-11
nvme: Add support for dsm command latency option

ikegami-t:


ikegami-t:
Now I am thinking to add the latency command option for all other nvme commands also.

igaw:
I am not so keen on adding performance and debugging code everywhere. Makes the code base bigger/... for small gain. Someone pointed out to me that we could try to make use of user trace libraries such as lttng-ust which then lives outside of the main code base. Haven't really spend time playing with this yet. But I think this would better for these cases someone really wants to trace things.

ikegami-t:
The purpose is not debugging but adding performance since currently only added `passthru()` (called by io_passthru() and admin_passthru()) and `submit_io()` (called by compare(), read_cmd() and write_cmd()) so tried to add dsm command also. By the way I thought the option can be implemented as same with the `verbose` and `output-format` as common option. Still should we use the tool lttng-ust or etc.?

ikegami-t:
For the build check needed libnvme update to add add nvme_cli_set_latency().

ikegami-t:
Thank you so much for your advice! Just confirmed the tool bpftrace works as expected as below.
1. dsm command latency option: 3,480 us
```sh
root@tokunori-desktop:/home/tokunori/nvme-cli-2# .build/nvme dsm /dev/nvme0n1 -b 0x100000,0x100000,0x100000,0x100000 -s 0x0,0x100000,0x200000,0x300000 -L -d
ioctl: 4e40 (ID) latency: 3 us
IO Command opcode: 09 (Dataset Management) latency: 3480 us
NVMe DSM: success
```
2. bpftrace: 3,456 us
```sh
root@tokunori-desktop:/home/tokunori# bpftrace -e 'kprobe:nvme_submit_user_cmd { @start = nsecs; } tracepoint:nvme:nvme_complete_rq /@start != 0/ { printf("latency: %u usec\n", (nsecs - @start) / 1e3); @start = 0; }'
Attaching 2 probes...
latency: 3456 usec
^C

@start: 0
```

Do you mean we can add the bpftrace tool bt script file for the latency checking instead of the option? By the way seems the tool depended on the kernel version environment but okay for using to check the nvme-cli latency? Also currently the latecny option is used by the commonds calling submit_io() and passthru() but okay to remain them?

igaw:
> Do you mean we can add the bpftrace tools bt script file for the latency checking instead of the option? 

Yes, that is what I had in mind.

> By the way seems the tool depended on the kernel version environment but okay for using to check the nvme-cli latency?

I know bpftrace has a bit of dependencies. But for who is this feature anyway? What is the use case? I don't think you need this feature for embedded device.

> performance statisics purpose but for the nvme device performence stastics

I assume this is primarily interesting for the device developers. I am pretty sure they have way better tools to measure their device latencies and measuring the ioctl includes the whole kernel stack with all the noise it introduces.  Also, measuring admin commands is not so interesting. The read/writes commands are interesting ones and nvme-cli is the wrong tool for this.

igaw:
With #2234 merged, you can get this info with increasing the verbosity level.

PR #2150, opened 2023-12-11
build: Disable muon bestline for CentOS 7 build error

ikegami-t:


ikegami-t:
This was suggested by the muon list below.
<https://lists.sr.ht/~lattis/muon/%3C623b3378-cbce-4e7a-9ca7-4a0a4586763e%40yahoo.co.jp%3E>

igaw:
Thanks!

PR #2154, opened 2023-12-14
nvme: restric hmac options for gen-tls-key

martin-gpy:
During nvme gen-tls-key generation, the permitted hmac options is 1 for SHA-256 & 2 for SHA-384 respectively for the retained key. But nvme-cli mistakenly permits an additional option 3 too which defaults to SHA-256 itself. Rectify this.


igaw:
Thanks!

PR #2158, opened 2023-12-16
fabrics: fix invalid output format error during nvme connect

martin-gpy:
Nvme connect currently fails with an "Invalid output format" error if no output-format is specified to the nvme connect command. Fix this by initializing the format string.


igaw:
Thanks!

PR #2159, opened 2023-12-16
fabrics: fix connect error if hostid file does not exist

martin-gpy:
Currently one sees a connect error during a nvme discover/connect if the hostnqn file exists, but not the hostid file. Fix this by ensuring the hostid is taken from the corresponding hostnqn for such scenarios.


igaw:
Thanks!

PR #2178, opened 2024-01-04
completions: added Solidigm plugin to autocomplete script

amashiana-sd:


igaw:
Could you also update the zsh completion as well? Thanks!

amashiana-sd:
> Could you also update the zsh completion as well? Thanks!

Am I correct in my understanding that zsh completion is the file called _nvme?

I see individual commands in there but nothing related to plugins. Do I just need to make sure all commands under Solidigm plugin are in that file?

igaw:
Yes, the `_nvme` file contains the tab completion for zsh. So far the only other plugin added supported is the `ocp`, though it doesn't look it is really implementing any option completion :( Well at least it support for the tab completion for the command. 

In the long run I'd like to see zsh also implementing support for the all the plugins, that's why I ask for it :)

igaw:
Very cool. Thanks!

PR #2181, opened 2024-01-11
nvme: print inserted tls key for check-tls-key

martin-gpy:
nvme check-tls-key currently misses out successfully printing the inserted tls key, so print the same. And while we are it, fix a small typo while printing the error message related to the failure of generating the identity key.


igaw:
Thanks!

PR #2183, opened 2024-01-12
nvme: Parse telemetry da in create telemetry log host

hmi-jeon:
when create telemetry log host, parse telemetry da

Reported-by: Ki ryoung <kiryong1.kim@samsung.com>
Signed-off-by: Minsik jeon <hmi.jeon@samsung.com>

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -701,8 +701,11 @@ static int __create_telemetry_log_host(struct nvme_dev *dev,
 	if (err)
 		return -errno;
 
-	*size = NVME_LOG_TELEM_BLOCK_SIZE;
-	return get_log_telemetry_host(dev, NVME_LOG_TELEM_BLOCK_SIZE, buf);
+	err = parse_telemetry_da(dev, da, log, size);
+	if (err)
+		return err;
+
+	return get_log_telemetry_host(dev, *size, buf);
 }
 
 static int __get_telemetry_log_ctrl(struct nvme_dev *dev,


hmi-jeon:
Hi @igaw 

I checked your refactor fix.
https://github.com/linux-nvme/nvme-cli/commit/0040a905783a5118216520ba85aaee64f1320326

But is there a reason why the size is fixed to 512 when telemetry log creating?
Previously, I was looking at the entir log while createing, and I modified it and would like to receive a review from you.

igaw:
> But is there a reason why the size is fixed to 512 when telemetry log creating?

I don't know this part too well (just touch the code). Though I found this in the base spec:

```
5.16.1.8 Telemetry Host-Initiated (Log Identifier 07h)

This log consists of a header describing the log and zero or more Telemetry Data Blocks (refer to section
8.24). All Telemetry Data Blocks are 512 bytes in size. 
```

Has something changed?

hmi-jeon:
I think each telemetry data block size are 512 bytes.
(zero or more Telemetry Data Blocks)
I think the number of data block should be output.

I checked the specifications as follows.
```
1023:512
Telemetry Host-Initiated Data Block 1: Contains Telemetry Data Block 1 for the Telemetry Host-Initiated Log.
1535:1024
Telemetry Host-Initiated Data Block 2: Contains Telemetry Data Block 2 for the Telemetry Host-Initiated Log.
…
…
(n*512)+511:(n*512)
Telemetry Host-Initiated Data Block n: Contains Telemetry Data Block n for the Telemetry Host-Initiated Log.
```

Thank you.

igaw:
After reading up a bit more on this, I am sure I just didn't understand it correctly when I refactored it. Thus I mixed up the size arguments. Your patch is restoring the old (correct) behavior. Let me add also a Fixed tag to the commit message.

igaw:
Thanks!

PR #2191, opened 2024-01-23
Implement support for secure concatenation

hreinecke:
Add support for option '--concat' to enable secure concatenation after DH-HMAC-CHAP.


igaw:
Thanks!

PR #2197, opened 2024-01-31
kbusch/fdp fix dir receive offsets

keithbusch:
Fix up directive receive identify offsets for saved properties.


igaw:
The same wrong offset calculation exists in the json output.

keithbusch:
> The same wrong offset calculation exists in the json output.

Good catch, fixed up now.

igaw:
Thanks!

PR #2199, opened 2024-02-01
doc: Fix short option name for cfg-file

felixonmars:
It was renamed to "-J" but some docs were left out.


igaw:
Thanks! I've added a bunch of more fixups and added your Signed-off-by. Hope this is okay.

felixonmars:
> Thanks! I've added a bunch of more fixups and added your Signed-off-by. Hope this is okay.

Sure, it's okay. Thanks!

igaw:
Thanks!

PR #2200, opened 2024-02-01
doc: Fix config-schema.json's URL

felixonmars:
The old URL returns 404.


felixonmars:
Updated to change only *.txt files and add signoff. Thanks!

igaw:
Thanks a lot!

PR #2201, opened 2024-02-01
Remove unused cfg argument from NVME_ARGS() macro

calebsander:
The `NVME_ARGS()` macro takes a `c` argument but doesn't use it. Remove it from the definition and all use sites to simplify the code.


igaw:
Just added subject prefix.

igaw:
Thanks!

Issue #2198, opened 2024-02-01
Consider disable NBFT when ACPI is not present


felixonmars:
On my riscv64 devices, the path /sys/firmware/acpi/tables doesn't exist so that `nvme connect-all` fails with:

```
Failed to open /sys/firmware/acpi/tables.
```

This renders the provided systemd services hard to use because I can't seem to pass `--no-nbft` in the config file.

Would it be a good idea to disable NBFT automatically when it's not available?

igaw:
Absolutely, we should just ignore it silently

PR #2202, opened 2024-02-02
Do not print error message if ACPI tables are no present

igaw:
Fixes: #2198


igaw:
@mwilck @tbzatek are you okay with this change?

mwilck (approved):
Fine with me.

igaw:
Thanks!

tbzatek (commented):
Looks good to me!

PR #2210, opened 2024-02-07
don't include newlines in already wrapped text

keithbusch:
The descriptions are already line wrapped as appropriate. Don't insert random newline characters in inappropriate places. It just makes the output look rediculous.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -128,9 +128,9 @@ static struct program nvme = {
 	.name = "nvme",
 	.version = nvme_version_string,
 	.usage = "<command> [<device>] [<args>]",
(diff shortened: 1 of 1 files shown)


keithbusch:
checkpatch doesn't like multi-line strings??? WTF?

keithbusch:
> checkpatch doesn't like multi-line strings??? WTF?

It is with purposeful decision to write it this way. checkpatch is wrong in this case.

igaw:
checkpatch.pl is good to catch some of the formatting issue. But as you pointed out it is not always correct. The good thing is we can just ignore it :D

Issue #2205, opened 2024-02-02
Fix code scanning alert - Time-of-check time-of-use filesystem race condition


igaw:
https://github.com/linux-nvme/nvme-cli/blob/79abafe689432647b97a441b34714ab2b2708d4d/plugins/micron/micron-nvme.c#L332-L332

Often it is necessary to check the state of a file before using it. These checks usually take a file name to be checked, and if the check returns positively, then the file is opened or otherwise operated upon.

Tracking issue for:
- [ ] https://github.com/linux-nvme/nvme-cli/security/code-scanning/4

ikegami-t:
This issue also can be closed.

PR #2212, opened 2024-02-11
plugins/micron: Fix code scanning alert

ikegami-t:
Fix the time-of-check time-of-use filesystem race condition.

### DIFF
--- a/plugins/micron/micron-nvme.c
+++ b/plugins/micron/micron-nvme.c
@@ -226,7 +226,6 @@ static int SetupDebugDataDirectories(char *strSN, char *strFilePath,
 	int length = 0;
 	int nIndex = 0;
 	char *strTemp = NULL;
-	struct stat dirStat;
 	int j;
 	int k = 0;
 	int i = 0;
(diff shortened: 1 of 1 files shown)


ikegami-t:
Fix for the issue #2205.

ikegami-t:
Sorry I am not able to check locally if the fix resolves the alert actually.

igaw:
Please use a more descriptive subject and also a bit of explanation how it is addressed. It is not obvious to me why the setguid() fixes the given alert. Or is this a well known thing?

ikegami-t:
Updated the commit description to explain the changes more detail.

igaw:
- According the link you posted, you need to set the effective user before checking and later. 
- Commit title is still not reflecting what it does. 'Fix code scanning alert' could be anything.

plugins/micron/micron-nvme.c
```diff
@@ -303,8 +303,14 @@ static int SetupDebugDataDirectories(char *strSN, char *strFilePath,
...
 
+	if (setuid(getuid())) {
+		err = -1;
+		goto exit_status;
+	}
+
 	j = 1;
-	while (!stat(strMainDirName, &dirStat)) {
+	while (!(fd = open(strMainDirName, 0))) {
+		close(fd);
```
igaw, line 313: This doesn't make any sense. It's the same. The problem is that we check if strMainDirName exists and then we do later a rmdir. This has to be tied together.

ikegami-t: Sorry the change not correct and should be fixed as below.
```diff
- 	while (!(fd = open(strMainDirName, 0))) {
+ 	while (!!(fd = open(strMainDirName, 0))) {
```
But changed to fix by unusing stat() and it can be checked by the mkdir errno instead if strMainDirName exists.
The code behavior itself same with the original code scanned the alert but the alert itself can be resolved by the changes. (The alert itself can be resolved as same by the previous changes also.)
Since seems the scanner just checks if using stat() and rmdir() only.
Also the issue not only for strMainDirName but seems same for strOSDirName and strCtrlDirName but just not detected by the scanner.

igaw: Yes, using mkdir directly to check if the patch exists, is the canonical approach for this. Let's go with it.

igaw:
Thanks

PR #2220, opened 2024-02-19
sed: Confirm entered password before initialization.

gjoyce-ibm:
Ask for password twice in case there is a typo in entering it.


igaw:
Thanks!

Issue #2203, opened 2024-02-02
Fix code scanning alert -  Potential use after free


igaw:
CodeQL complains about a UAF:

```
Memory may have been previously freed by  


ccan/ccan/strset/strset.c:196
			}

			/* Sew empty string back so remaining logic works */
			free(n->u.n);
			n->u.s = empty_str;
			break;
		}


```

https://github.com/linux-nvme/nvme-cli/blob/79abafe689432647b97a441b34714ab2b2708d4d/ccan/ccan/strset/strset.c#L211-L211

Tracking issue for:
- [ ] https://github.com/linux-nvme/nvme-cli/security/code-scanning/57

igaw:
Maybe we should just merge your fix. Upstream is not really responding. We did it the same for the include/define fix for bigendian too.

ikegami-t:
Thank you. Noted.

ikegami-t:
So can we close the issue now?

igaw:
Let's keep it open to track upstream. I still hope it gets eventually addressed.

ikegami-t:
I see. Thank you.

igaw:
Upstream is not responding, let's close it then.

PR #2222, opened 2024-02-23
ccan: Add freed pointer checking to delete strset member

ikegami-t:
Add to set freed entry pointer to NULL then check it if NULL.


ikegami-t:
Fix for the issue #2203.

igaw:
This should go upstream and then we just backport the change.

ikegami-t:
Okay if changed the upstream then will backport by a separate patch so close the PR.

PR #2224, opened 2024-02-24
nvme-print: print more details in ns-descs verbose output

martin-gpy:
Print additional details such as loc, nidt, nidl & type in the ns-descs verbose output for parity with the corresponding json output.

### DIFF
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -2714,7 +2714,7 @@ static void stdout_cmd_set_independent_id_ns(struct nvme_id_independent_id_ns *n
 static void stdout_id_ns_descs(void *data, unsigned int nsid)
 {
 	int pos, len = 0;
-	int i;
+	int i, verbose = stdout_print_ops.flags & VERBOSE;
 	__u8 uuid[NVME_UUID_LEN];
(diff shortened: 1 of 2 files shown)


nvme-print-stdout.c
```diff
@@ -2724,13 +2724,23 @@ static void stdout_id_ns_descs(void *data, unsigned int nsid)
...
+			printf("nidt    : %d\n", (int)cur->nidt);
+			printf("nidl    : %d\n", (int)cur->nidl);
+		}
+
 		switch (cur->nidt) {
 		case NVME_NIDT_EUI64:
 			memcpy(eui64, data + pos + sizeof(*cur), sizeof(eui64));
+			nidt_type = "eui64";
+			if (verbose)
+				printf("type    : %s\n", nidt_type);
```
calebsander, line 2743: Why not just `printf("type : eui64\n")` (or `puts()`)? That would allow you to get rid of the `nidt_type` variable.

calebsander: Also seems like this is redundant with the `printf("eui64 : ");` printed immediately afterwards?

igaw:
Thanks!

PR #2225, opened 2024-02-25
nvme-print-stdout: Use persistent event log RCI definitions

ikegami-t:
Change the hardcoded the register bit shift to use the definitions.


ikegami-t:
Pushed some commits depended on the libnvme PR changes linux-nvme/libnvme#785 added.

igaw:
Once again, please have the libnvme.wrap change as separate patch and not part of any other changes. This makes back porting and merging really difficult.

ikegami-t:
Sorry for the miss. Also noted for your mention so next will do separate the libnvme.wrap change. Thank you.

PR #2188, opened 2024-01-21
nvme: Add support for get-reg command to output single register

ikegami-t:
Note: Currently only stdout print supported.

PR #2226, opened 2024-02-28
Revert "nvme-print-json: remove obj_print helper"

ikegami-t:
This reverts commit b854cbda04a19b7dcfcbeb569146cf9ab79fd5f3.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -179,6 +179,12 @@ static void json_print(struct json_object *r)
 	json_free_object(r);
 }
 
+static void obj_print(struct json_object *o)
+{
+	if (!json_r)
+		json_print(o);
+}
+
 static bool human(void)
 {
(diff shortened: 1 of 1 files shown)


ikegami-t:
The helper function is used the `json_r` is set by `json_show_init()` then the function does not print but it is done by `json_show_finish()` instead. The get-feature command using `nvme_show_init()` and `nvme_show_finish()` those call `json_show_init()` and `json_show_init()` json print functions so the command json output does not work correctly currently. Also the init and finish functions will be used by the PR #2188 get-reg command also. (The issue found by the implementation working.)

igaw:
Could you add the comment  from above to the commit message? When someone is looking at the git history, it makes simpler to figure out why this was reverted.

ikegami-t:
Added the commit description as mentioned. Thank you.

PR #2227, opened 2024-03-03
nvme: Add show-regs and get-property commands NSSD print outputs

ikegami-t:
Note: Partially duplicated with the PR #2188 changes so merged later.


igaw:
The 'note' is something for this pull request. It doesn't add any useful information to the git commit itself. A git commit should describe the commit itself and should not just reference outside resources.  Adding a link or so for better understanding is okay, but as said it's additional information. Thanks.

ikegami-t:
Okay understood the comment for `note`. Thank you.

PR #2228, opened 2024-03-04
nvme-print: Fix PEL SEB print functions

ikegami-t:
This is for the issue mentioned by https://github.com/linux-nvme/nvme-cli/pull/1962#discussion_r1213398564 before.


igaw:
Thanks.

PR #2230, opened 2024-03-04
ocp: Fix timestamp displayed by fw-activate-history command

jeff-lien-wdc:
The current code is displaying all 8 bytes of the timestamp field for the timestamp.  The timestamp value is only in the first 6 bytes. The 7th byte contains the origin and synch fields and the 8th byte is reserved.

### DIFF
--- a/plugins/ocp/ocp-fw-activation-history.c
+++ b/plugins/ocp/ocp-fw-activation-history.c
@@ -66,7 +66,7 @@ static void ocp_fw_activation_history_normal(const struct fw_activation_history 
 		printf("      %-22s%d\n", "activation count:",
 		       le16_to_cpu(entry->activation_count));
 		printf("      %-22s%"PRIu64"\n", "timestamp:",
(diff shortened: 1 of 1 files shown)


jeff-lien-wdc:
@igaw I noticed my commits for this PR and PR #2229 show up as "Unverified" and was wondering why?  I do have a gpg key setup so when I commit changes to my forked repo, they show up as "Verified".  Do I need to setup a gpg key or do something else in the main repo to make my commits verified?  Is this a problem?

igaw:
It looks like when I click the 'rebase & merge' button github is not able to track the signature:

https://github.com/orgs/community/discussions/11639

jeff-lien-wdc:
Thanks Dan.  I've subscribed to that discussion.  Hopefully it will be fixed soon.

igaw:
We could obviously just go back to the merge PR workflow but I really don't like it. It's just a lot of noise in the git history and this a simple project and not the linux kernel where this additional information makes sense.

PR #2234, opened 2024-03-06
Refactor and extend logging

igaw:
Introduce a global log level which then can be forwarded to libnvme. Also provide our own low level passthru commands so that we extend it with extensive logging outputs. While it also support the 64 bit version of it.


tbzatek (approved):
Looks good to me. I'll cover `nbft.c` separately.

PR #2235, opened 2024-03-10
doc: Add get-feature cdw11 and set-feature cdw12 parameters

ikegami-t:
Those implemented but missed to described by the documentations.


igaw:
Thanks!

PR #2237, opened 2024-03-11
nvme: Add support for dsm command latency option

ikegami-t:
The purpose to add is for adding performance statistics. Currently the latency option is only added for passthru() and submit_io(). The verbose option also outputs the latency but other debuging info. So add for the dsm command also to output only the latency info.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -6402,6 +6402,8 @@ static int dsm(int argc, char **argv, struct command *cmd, struct plugin *plugin
 	__u32 nlbs[256] = {0,};
 	__u64 slbas[256] = {0,};
 	int err;
+	struct timeval start_time;
+	struct timeval end_time;
 
 	struct config {
 		__u32	namespace_id;
@@ -6412,6 +6414,7 @@ static int dsm(int argc, char **argv, struct command *cmd, struct plugin *plugin
 		bool	idw;
 		bool	idr;
 		__u32	cdw11;
+		bool	latency;
 	};
 
 	struct config cfg = {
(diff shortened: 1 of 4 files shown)


nvme.c
```diff
@@ -6473,7 +6477,11 @@ static int dsm(int argc, char **argv, struct command *cmd, struct plugin *plugin
 		.timeout	= NVME_DEFAULT_IOCTL_TIMEOUT,
 		.result		= NULL,
 	};
+	gettimeofday(&start_time, NULL);
 	err = nvme_dsm(&args);
+	gettimeofday(&end_time, NULL);
+	if (cfg.latency)
+		printf(" latency: %s: %llu us\n", __func__, elapsed_utime(start_time, end_time));
```
martin-belanger, line 6484: It may be more desirable to call `gettimeofday()` only when it's needed.
```
	if (cfg.latency)
		gettimeofday(&start_time, NULL);
	err = nvme_dsm(&args);
	if (cfg.latency) {
		gettimeofday(&end_time, NULL);
		printf(" latency: %s: %llu us\n", __func__, elapsed_utime(start_time, end_time));
	}
```

ikegami-t: Thanks for your comment. Just fixed as mentioned.  By the way both `submit_io()` and `passthru()` also may be needed to fix as same.

igaw:
This is already available via increasing the verbosity level to debug. All `ioctl` are intercepted and when the verbosity level is set to debug it measures the the latency.

PR #2238, opened 2024-03-12
nvme-print-stdout: Add hexadecimal value 0x prefix

ikegami-t:
Also unify to use %#x or %#"PRIx64" instead of 0x%x and 0x%"PRIx64".


igaw:
I'm a bit worried about the changing the format from `%x` to `%#x`, because it changes the output format slightly. It might break users. Not sure if it's worth the trouble. I am fine with changing `0x%x` to `%#x` to make these parts consistent.

ikegami-t:
Just reverted to `%x` from `%#x` and only change `0x%x` to `%#x`. Thank you.

igaw:
Thanks!

PR #2239, opened 2024-03-12
wdc:  Fix timestamp displayed by vs-firmware-activate-history command

jeff-lien-wdc:
The actual timestamp value is in bytes 0:5 of the 8 byte field. The code was using all 8 bytes so a change was made to mask off the top 2 bytes.  The timestamp calculations were also simplified to be more maintainable.

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -1383,6 +1383,11 @@ struct __packed wdc_fw_act_history_log_format_c2 {
 	__u8		log_page_guid[WDC_C2_GUID_LENGTH];
 };
 
+static __u8 ocp_C2_guid[WDC_C2_GUID_LENGTH] = {
+	0x6D, 0x79, 0x9A, 0x76, 0xB4, 0xDA, 0xF6, 0xA3,
+	0xE2, 0x4D, 0xB2, 0x8A, 0xAC, 0xF3, 0x1C, 0xD1
+};
+
 #define WDC_OCP_C4_GUID_LENGTH              16
 #define WDC_DEV_CAP_LOG_BUF_LEN           4096
 #define WDC_DEV_CAP_LOG_ID                0xC4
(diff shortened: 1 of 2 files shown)


plugins/wdc/wdc-nvme.c
```diff
@@ -5791,33 +5779,24 @@ static void wdc_print_fw_act_history_log_json(__u8 *data, int num_entries,
 			json_object_add_value_int(root, "Entry",
 			    le16_to_cpu(fw_act_history_entry->entry[entryIdx].fw_act_hist_entries));
 
+			uint64_t timestamp = (0x0000FFFFFFFFFFFF &
+					le64_to_cpu(fw_act_history_entry->entry[entryIdx].timestamp));
+			__u64 timestamp_sec = le64_to_cpu(timestamp)/1000;
```
martin-belanger, line 5784: Same comment as before.

plugins/wdc/wdc-nvme.c
```diff
@@ -9112,29 +9093,46 @@ static int wdc_get_fw_act_history_C2(nvme_root_t r, struct nvme_dev *dev,
...
-			/* get the FW customer id */
-			if (!wdc_is_sn861(device_id)) {
-				cust_id = wdc_get_fw_cust_id(r, dev);
-				if (cust_id == WDC_INVALID_CUSTOMER_ID) {
-					fprintf(stderr,
-						"%s: ERROR: WDC: invalid customer id\n",
-						__func__);
-					ret = -1;
-					goto freeData;
+		for (i = 0; i < 16; i++) {
```
martin-belanger, line 9099: For the loop you could use `WDC_C2_GUID_LENGTH` instead of `16`.  Or replace the whole loop with `memcmp()`:
```
c2GuidMatch = !memcmp(ocp_C2_guid, fw_act_history_log->log_page_guid, WDC_C2_GUID_LENGTH);
```

plugins/wdc/wdc-nvme.c
```diff
@@ -5599,48 +5603,32 @@ static void wdc_print_fw_act_history_log_normal(__u8 *data, int num_entries,
 				memcpy(new_fw, null_fw, 8);
 
 			printf("%5"PRIu16"", (uint16_t)le16_to_cpu(fw_act_history_entry->entry[entryIdx].fw_act_hist_entries));
+
+			uint64_t timestamp = (0x0000FFFFFFFFFFFF &
+					le64_to_cpu(fw_act_history_entry->entry[entryIdx].timestamp));
+			__u64 timestamp_sec = le64_to_cpu(timestamp) / 1000;
```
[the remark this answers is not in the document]
jeff-lien-wdc: Removed extra le64_to_cpu calls.

plugins/wdc/wdc-nvme.c
jeff-lien-wdc: Removed extra le64_to_cpu calls.

plugins/wdc/wdc-nvme.c
jeff-lien-wdc: Changed to use memcmp.

jeff-lien-wdc:
@martin-belanger  @igaw Fixed build and checkpatch issues with the latest commits.  Also addressed review comments from Martin.  Thanks

PR #2240, opened 2024-03-13
logging: Split to output ioctl latency by log info level

ikegami-t:
This is for only output latency but not output other debugging info.


igaw:
Alright, if this makes things simpler. So far the info level is not used. So we have our first user.

PR #2242, opened 2024-03-14
ocp: support OCP 2.5 Set Telemetry Profile feature

bpaupore-wdc:
Adds a new OCP plugin function to set the device telemetry profile, configuring what debug data shall be collected by the device.


bpaupore-wdc:
Checkpatch review warnings seem appropriate to ignore in this case, but I can make adjustments here if needed.

igaw:
From my point of view, it looks good. 

@arthurshau

arthurshau:
Agreed, LGTM

PR #2243, opened 2024-03-14
ocp: fix for power-state argument checking

bpaupore-wdc:
Noticed this as part of supporting set-telemetry-profile, it seems that the power-state option is not being detected properly. With this fix the function is able to enter set_dssd_power_state when given the option or its alias.


igaw:
Thanks!

PR #2244, opened 2024-03-15
nvme-print: Add PEL EHAI port identifier type printf function

ikegami-t:
Change to use NVME_PEL_EHAI_PIT definitions.


ikegami-t:
Depended on the libnvme PR changes linux-nvme/libnvme#789.

igaw:
Thanks.

PR #2245, opened 2024-03-15
nvme-print-stdout: Use CHECK_BIT macro for PEL seb print

ikegami-t:
Also use event string pointer to check and print.


ikegami-t:
Okay understood. Thanks for your reviewing and comment. By the way as far as I checked the similar bit manipulation is only implemented for both stdout and json PEL seb prints only.

Issue #2048, opened 2023-09-13
show-regs has excess dependencies 


NateThornton:
The show-regs command has some additional dependencies that the drive is operational and responding to an Identify admin command. For a drive that is truly unresponsive, say in a `Controller Fatal Status = True` state, this means one cannot inspect the controller registers.

Call stack:
```
show_registers
  nvme_scan
    nvme_scan_topology
      nvme_scan_subsystem
        nvme_subsystem_scan_namespace
          nvme_subsystem_scan_namespace
            __nvme_scan_namespace
              nvme_ns_open
                nvme_ns_init
                  nvme_ns_identify
```

Doing a namespace scan seems unnecessary for the show-regs command, and I'm wondering if this can somehow be removed so the show-regs is still functional in fatal situations. That would likely include an enhancement to libnvme.

If someone can provide guidance on the necessary code changes, I can either work on the PR or help with validation.

Thanks

keithbusch:
That does seem excessive just to get some registers. I think part of this might be trying to establish what transport the device is since there are different ways to access the registers depending on the interface. For pci, though, we don't even really need a device handle, which I suspect is the transport you're interested in.

Since it's possible a device handle doesn't even exist for a failed controller, we could add a toggle parameter to the `show-regs` command to take a `/sys/bus/pci/devices/.../resource0` argument instead of a `/dev/nvme0` and skip all the identification steps.

NateThornton:
Thanks Keith, the change ended up being pretty trivial and I just needed a dummy root object.

```c++
	if (!cfg.no_scan)
		r = nvme_scan(NULL);
	else
		r = nvme_create_root(NULL, DEFAULT_LOGLEVEL);
```

I can get a PR prepared.

Is using `--no-scan` with help text "skip NVMe scan and use the provided device directly" acceptable naming/description? We can visit that on the PR too.

keithbusch:
That will still scan the controller and/or namespace in `mmap_registers()` though, right? This change will just avoid scanning every nvme device in the system, which honestly doesn't appear to provide any use in this path, so your suggestion should just be the default. But even that looks like it's doing to much; we just want to the sysfs directory of the requested device/namespace. We don't need all the extra device properties that are being queried for show-registers.

NateThornton:
I checked via strace and no ioctls were observed with the added `--no-scan` behavior when a controller is found. Otherwise you are correct, we'd still do a `nvme_scan_namespace` if the controller is _not_ found.
[...]

PR #2246, opened 2024-03-16
nvme: Scan NVMe topology without nvme identify namespace

ikegami-t:
Since the show-regs command needed this to scan for PCI.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5322,7 +5322,7 @@ static int show_registers(int argc, char **argv, struct command *cmd, struct plu
 	if (err)
 		return err;
 
-	r = nvme_scan(NULL);
+	r = nvme_scan_no_id(NULL);
 	err = validate_output_format(output_format_val, &flags);
 	if (err < 0) {
 		nvme_show_error("Invalid output format");


ikegami-t:
Depended on the libnvme PR changes https://github.com/linux-nvme/libnvme/pull/790 and this can resolved the issue #2048 also.

igaw:
Why is this necessary? What is the problem you are trying to solve?

ikegami-t:
Since the show-regs command  hanged up when CC.EN set to zero then reset the controller by timeout so the register values not able to print before the reset. This caused by the nvme identify namespace but not necessary for the command output. Later will do include the reason into the commit message.

hreinecke:
I'd phrase it differently: why do we need to call 'nvme_scan()' here? It feels a bit too heavy-handed to me. We should just drop it and use the device name directly, much like we already do eg for 'nvme reset'.

igaw:
I suspect the scan is needed if we want to list the fabrics properties. Still we could just do the scan only for the fabric devices and not for PCI devices.

hreinecke:
For get_regs ? Does that even work on fabrics devices?

igaw:
I thought that `nvme_get_properties` might rely on `nvme_root_t` but nope. I think we don't need the scan at all.

igaw:
It's `mmap_registers` which wants `nvme_root_t`. Could we get away without it?

hreinecke:
Like this?
`diff --git a/nvme.c b/nvme.c
index 40496b54..8ffce55b 100644
--- a/nvme.c
+++ b/nvme.c
@@ -5249,32 +5249,10 @@ static int nvme_get_properties(int fd, void **pbar)
        return err;
 }
 
-static void *mmap_registers(nvme_root_t r, struct nvme_dev *dev)
+static void *mmap_registers(int fd)
 {
-       nvme_ctrl_t c = NULL;
-       nvme_ns_t n = NULL;
-
-       char path[512];
        void *membase;
-       int fd;
-
-       c = nvme_scan_ctrl(r, dev->name);
-       if (c) {
-               snprintf(path, sizeof(path), "%s/device/resource0",
-                       nvme_ctrl_get_sysfs_dir(c));
-               nvme_free_ctrl(c);
-       } else {
-               n = nvme_scan_namespace(dev->name);
-               if (!n) {
-                       nvme_show_error("Unable to find %s", dev->name);
-                       return NULL;
-               }
-               snprintf(path, sizeof(path), "%s/device/device/resource0",
-                        nvme_ns_get_sysfs_dir(n));
-               nvme_free_ns(n);
-       }
 
-       fd = open(path, O_RDONLY);
        if (fd < 0) {
                if (log_level >= LOG_DEBUG)
                        nvme_show_error("%s did not find a pci resource, open failed %s",
@@ -5290,8 +5268,6 @@ static void *mmap_registers(nvme_root_t r, struct nvme_dev *dev)
                }
                membase = NULL;
        }
-
-       close(fd);
        return membase;
 }
 
@@ -5334,7 +5310,7 @@ static int show_registers(int argc, char **argv, struct command *cmd, struct plu
        if (cfg.human_readable)
                flags |= VERBOSE;
 
-       bar = mmap_registers(r, dev);
+       bar = mmap_registers(dev_fd(dev));
        if (!bar) {
                err = nvme_get_properties(dev_fd(dev), &bar);
                if (err)
`

ikegami-t:
Yes the nvme_scan only used for the mmap_registers to get the path so will do consider to get the path from libnvme without the scan. I think still the path information required for the mmap as mentioned by https://github.com/linux-nvme/nvme-cli/issues/2048#issuecomment-1719716455 also.

igaw:
> Like this?

I can't really the exact reason for the `resource0` path lookups. We added it for some reason but I can't remember. I'll try to figure out why we added it initially.

igaw:
Ah after digging into it, I think we can do something very simple:

```c
sprintf(path, "/sys/class/nvme/%s/device/resource0", dev->name);
```

because the fallback (via namespace) was needed prio v4.0 but since then the `resource0` file is available via the `/sys/class/nvme` path. So no need to be fancy here.

d9020321e66a ("Register read fix for kernel 4.0")
eb5d180c6dbf ("Fixed versioning for pre 1.2 devices")
08290ca94abc ("NVMe: Add show registers command")

PR #2247, opened 2024-03-17
nvme-print-stdout: Use NVME_SET_FEAT_EVENT register definitions

ikegami-t:
Change it instead of hardcoded register mask and shift values.


ikegami-t:
Depended on the libnvme PR changes linux-nvme/libnvme#791.

igaw:
just a rebase

igaw:
Thanks.

PR #2249, opened 2024-03-17
nvme-print-stdout: Use NVME_BOOT_PARTITION_INFO field definitions

ikegami-t:
Change it instead of hardcoded field mask and shift values.


ikegami-t:
Depended on the libnvme PR changes https://github.com/linux-nvme/libnvme/pull/792.

igaw:
just a rebase

igaw:
Thanks.

Issue #2241, opened 2024-03-14
Commands fail on nvme-cli 2.x when nvme_core module isn't loaded, regression from nvme-cli 1.x


calebsander:
On `nvme-cli` 1.x, `nvme list`, `nvme disconnect-all`, etc. are no-ops when the `nvme_core` module isn't loaded:
```
[root@init100-17-1 ~]# lsmod | grep nvme
[root@init100-17-1 ~]# nvme version
nvme version 1.16
[root@init100-17-1 ~]# nvme list
Node                  SN                   Model                                    Namespace Usage                      Format           FW Rev
--------------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
[root@init100-17-1 ~]# nvme disconnect-all
```
But on `nvme-cli` 2.x, they fail when trying to open the nonexistent `/sys/class/nvme`. It would be nice for the commands to be no-ops again.
```
[root@init100-17 ~]# lsmod | grep nvme
[root@init100-17 ~]# nvme version
nvme version 2.2.1 (git 2.2.1)
libnvme version 1.2 (git 1.2)
[root@init100-17 ~]# nvme list
Failed to scan topology: No such file or directory
[root@init100-17 ~]# nvme disconnect-all
Failed to scan topology: No such file or directory
```

igaw:
Is still the case with v2.8? I thought it should not complain anymore.

calebsander:
Yes, I just tried it with 2.8:
```
[root@init100-17 nvme-cli]# nvme version
nvme version 2.8 (git 2.8-38-g2af3218)
libnvme version 1.8 (git 1.8-22-gd96e92a)
[root@init100-17 nvme-cli]# modprobe -r nvme_core
[root@init100-17 nvme-cli]# nvme disconnect-all
Failed to scan topology: No such file or directory
```

PR #2250, opened 2024-03-19
nvme: do not issue warning when nvme_core module is not loaded

igaw:
nvme 1.x did not issue an warning when the nvme_core module was not loaded. Make 2.x behave in the same way.

Fixes: #2241

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -728,8 +728,9 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 		nvme_root_set_application(r, context);
 	ret = nvme_scan_topology(r, NULL, NULL);
 	if (ret < 0) {
-		fprintf(stderr, "Failed to scan topology: %s\n",
-			 nvme_strerror(errno));
+		if (errno != ENOENT)
+			fprintf(stderr, "Failed to scan topology: %s\n",
+				nvme_strerror(errno));
 		nvme_free_tree(r);
 		return ret;
 	}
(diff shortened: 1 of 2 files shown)


fabrics.c
```diff
@@ -728,8 +728,9 @@ int nvmf_discover(const char *desc, int argc, char **argv, bool connect)
 		nvme_root_set_application(r, context);
 	ret = nvme_scan_topology(r, NULL, NULL);
 	if (ret < 0) {
-		fprintf(stderr, "Failed to scan topology: %s\n",
-			 nvme_strerror(errno));
+		if (errno != ENOENT)
+			fprintf(stderr, "Failed to scan topology: %s\n",
+				nvme_strerror(errno));
```
calebsander, line 733: Sorry if I was unclear in #2241. I don't really have a problem with the warning message. The main issue is that the commands fail on `nvme-cli` 2.x where it succeeds on 1.x, which I don't think this change addresses. I think the correct fix is for `nvme_scan_topology()` not to return an error in the `ENOENT` case.

calebsander: I think it would actually be preferable to keep the error message. If the command fails, the user should know why. I posted https://github.com/linux-nvme/libnvme/pull/795 to avoid `nvme_scan_topology()` returning this error in the first place when `nvme_core` isn't loaded. So I think we can revert this change.

igaw: And I think https://github.com/linux-nvme/libnvme/pull/795 is just the wrong thing to do

PR #2251, opened 2024-03-19
nvme: do not scan topology when mmaping registers

igaw:
For accessing the registers we don't need to scan the topology. With the
device name we already know where to lookup the resource0 file in sysfs.

We also don't need to do the fallback lookup via the namespaces because
Linux kernel v4.0 has introduced the link in /sys/class/nvme/nvme%d. And
nvme-cli 2.x depends on Linux kernel >= 4.15, so it's safe to drop the
fallback.

This avoids all the topology scanning which is excessive and also might
issue an 'id command' via the nuse sysfs entry. So let's not do it.

Fixes: #2246 #2048


hreinecke (approved):
Looks good to me.

PR #2252, opened 2024-03-19
Implement persistent keys

hreinecke:
Add a new utility 'tls-key' to import or export keys to or from the kernel keystore.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -8540,9 +8540,8 @@ static int gen_tls_key(int argc, char **argv, struct command *command, struct pl
 	const char *insert = "Insert only, do not print the retained key.";
 
 	_cleanup_free_ unsigned char *raw_secret = NULL;
-	char encoded_key[128];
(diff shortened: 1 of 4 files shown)


nvme.c
```diff
@@ -8776,6 +8776,124 @@ static int check_tls_key(int argc, char **argv, struct command *command, struct
...
+		  OPT_STR("keyfile",	'f', &cfg.keyfile,	keyfile),
+		  OPT_FLAG("import",	'i', &cfg.import,	import),
+		  OPT_FLAG("export",	'e', &cfg.export,	export));
+
+	err = argconfig_parse(argc, argv, desc, opts);
+	if (err)
+		return err;
+
+	if (cfg.keyfile) {
+		fd = fopen(cfg.keyfile, "r");
```
igaw, line 8836: For export, shouldn't this file be opened with in `write` mode?

igaw:
Looks good to me. Let's wait a bit and see if there are some more comments coming in.

PR #2253, opened 2024-03-19
meson buildtarget + some warning fixes

tbzatek:
The meson 'debug' buildtype defaults to '-O0 -g' which misses some important compiler warnings (as of gcc-13). While there's     no universal solution, the 'debugoptimized' buildtype supplies '-O2 -g' that appears to be a reasonable compromise for having    important compiler warnings by default.
    
Either way, the specified warning_level=1 always supplies '-Wall' that is unfortunately not enough with '-O0'.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@ project(
     version: '2.8',
     default_options: [
       'c_std=gnu99',
-      'buildtype=debug',
+      'buildtype=debugoptimized',
       'prefix=/usr/local',
       'warning_level=1',
       'sysconfdir=etc',
--- a/nbft.c
+++ b/nbft.c
@@ -87,7 +87,7 @@ int discover_from_nbft(nvme_root_t r, char *hostnqn_arg, char *hostid_arg,
 	nvme_ctrl_t c;
 	int ret, i;
 	struct list_head nbft_list;
-	struct nbft_file_entry *entry;
+	struct nbft_file_entry *entry = NULL;
 	struct nbft_info_subsystem_ns **ss;
 	struct nbft_info_hfi *hfi;
 
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -1335,7 +1335,7 @@ struct json_object* json_effects_log(enum nvme_csi csi,
 static void json_effects_log_list(struct list_head *list)
 {
 	struct json_object *r = json_create_array();
-	nvme_effects_log_node_t *node;
+	nvme_effects_log_node_t *node = NULL;
 
 	list_for_each(list, node, node) {
 		struct json_object *json_page =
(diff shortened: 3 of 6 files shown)


tbzatek:
I've been hit by silenced compiler warnings a couple of times since the default buildtype `debug` sets `-O0` that makes gcc to avoid some code checks. 

Now as a result of my experiments the following set of CFLAGS would report various (and valid) warnings, not seen with other combinations:
* `CC='clang' CFLAGS='-O1 -flto=auto -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer'`
  * Fedora 40 defaults except of `-O1`
  * this will scream about unused `-specs=` switches, removing those would however mysteriously silence some warnings
  * tested with `clang version 18.1.0`

* `CC='gcc' CFLAGS='-O1 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer'`
  * switching to `-O2` would miss some important warnings
  * switching to `-O0` would miss essentially all valid warnings
  * tested with `gcc version 14.0.1 20240228`

igaw:
FWIW, there is also `scripts/build.sh` which could be extended and use it the CI build environment. I could also build a Fedora build container if you want. This could catch some more errors.

tbzatek:
> FWIW, there is also `scripts/build.sh` which could be extended and use it the CI build environment. I could also build a Fedora build container if you want. This could catch some more errors.

Yes, that script was a source of my inspiration when looking for differences between CI and my system. The default buildtarget there is 'release', i.e. `-O2`. There's a big difference between `-O0`, `-O1` and `-O2` with `-O1` reporting some false warnings as a bonus (they're gone with `-O2`). The Fedora CFLAGS didn't make that much difference and we're covered by our internal CI.

PR #2255, opened 2024-03-22
nvme: Use libnvme sysfs function to get NVMe controller directory

ikegami-t:
Also delete the SYS_NVME definition unused instead.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5256,7 +5256,7 @@ static void *mmap_registers(struct nvme_dev *dev)
 	void *membase;
 	int fd;
 
-	sprintf(path, "/sys/class/nvme/%s/device/resource0", dev->name);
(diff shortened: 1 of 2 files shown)


ikegami-t:
Depended on the libnvme PR changes https://github.com/linux-nvme/libnvme/pull/801.

igaw:
As I wrote in [PR](https://github.com/linux-nvme/libnvme/pull/801) I think we need a good reason for making sysfs configurable in nvme-cli. libnvme has unit tests which rely on it. But for nvme-cli we don't have those and for this particular function here it's not going to work because of the mmap operation.

ikegami-t:
Okay I see. Understood as the sysfs functions purpose is differrent for the patch since I though just change nvme-cli not use the sysfs paths directly but libnvme provides them. I will do consider this more later but anyway let me close the PR. Thank you for your reviewing and comments.

PR #2257, opened 2024-03-24
nvme-print: Use EOM optional data present macros changed

ikegami-t:
Also use the EOM optional data present reserved field macro added.


ikegami-t:
Depended on the libnvme PR changes https://github.com/linux-nvme/libnvme/pull/802.

igaw:
Thanks!

PR #2261, opened 2024-03-25
plugins/memblaze: code-format on leading indent and align

jinhuahuang2021:
1. Set leading indent to tab
2. Set proper alignment as code len limitation


igaw:
Thanks!

PR #2262, opened 2024-03-25
nvme: Fix to cleanup free raw_secret memory allocated

ikegami-t:
The memory is allocated by the gen-tls-key and check-tls-key commands.


igaw:
Thanks!

PR #2263, opened 2024-03-26
plugins/memblaze: code refactor on definitions and structs

jinhuahuang2021:
1. apply __cleanup_nvme_dev__ method for nvme_dev.
2. apply __packed for more structs.
3. change a definition error on struct add-smart-log.


igaw:
Generally, it's better to have a patch per logical change. E.g. in your case you're listening three bullet points which are all changes which are independent of each other, thus each could go into a separate patch.

jinhuahuang2021:
OK

PR #2264, opened 2024-03-26
nbft: Refactor connection attempt out

tbzatek:
With planned NBFT discovery work, a common connection function would be handy. Besides, the DHCP fallback mostly duplicated the connection code anyway, let's refactor that.


igaw:
Thanks!

PR #2265, opened 2024-03-26
ocp: support OCP DSSD Async Event Config feature

bpaupore-wdc:
Adds two new OCP plugin functions for getting and setting this feature, based on the 2.5 OCP spec.

### DIFF
--- a/Documentation/nvme-ocp-get-dssd-async-event-config.txt
+++ b/Documentation/nvme-ocp-get-dssd-async-event-config.txt
@@ -1,0 +1,53 @@
+nvme-ocp-get-dssd-async-event-config(1)
+=======================================
+
+NAME
+----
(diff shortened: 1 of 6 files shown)


bpaupore-wdc:
I added the second commit here to prefer tabs to spaces over the few places in the file where that wasn't the case, but I could remove that commit and just fix the new functions if that would be preferred.

Otherwise the remaining checkpatch warnings seem not applicable.

bpaupore-wdc:
@igaw Any thoughts on this PR, or the comment above?

igaw:
Sorry I didn't CC @arthurshau for a review and thus I didn't merge it and complete forgot.

arthurshau:
Yeah, the implementation LGTM.

igaw:
Thanks!

PR #2266, opened 2024-03-26
build: bump libnvme wrap

ikegami-t:
Include latest libnvme changes.


igaw:
I've bump the ref today with a newer ref. So this one is not needed anymore. Thanks  anyway.

PR #2267, opened 2024-03-27
plugins/memblaze: code refactor on using cleanup_nvme_dev

jinhuahuang2021:
using cleanup_nvme_dev to open/close device


jinhuahuang2021:
@igaw 
It contains only one change about using cleanup_nvme_dev.

igaw:
Nice. Thanks!

PR #2269, opened 2024-03-27
plugins/memblaze: code refactor on applying __packed on more structs

jinhuahuang2021:
applying __packed on more structs


jinhuahuang2021:
@igaw
It contains only one change about applying __packed on some structs.

PR #2270, opened 2024-03-27
plugins/memblaze: fix code mis-type error

jinhuahuang2021:
Change the definition from "flash_error_media_count" to "inflight_write_io_cmd".


jinhuahuang2021:
@igaw
It contains only one change about a code mis-type error.

igaw:
Thanks for splitting the patches. This makes easy to review. Excellent work!

BTW, you don't have to create for each patch a PR. You can also submit several patches/commits in PR.

PR #2273, opened 2024-03-27
sed: update SED password when initializing

gjoyce-ibm:
The SED key and kernel keyring were not being updated by the initialization operation. Add password update after the other initialization tasks are completed.


igaw:
Thanks!

PR #2274, opened 2024-03-28
nvme-print-json: Bugfix status json key of sanitize-log

hmi-jeon:
There is a history of combining duplicated json keys before.
Fixes: 1d7c092f9dac ("nvme-print-json: Combine duplicated json key and val string variables")
Fixes: 59a0f125cb08 ("nvme-print-json: Delete static const char string global variables")
But there is a local variable with the same name, so the key and value are the same.
    
Reviewed-by: Sangkuk Lee <sangkuk0.lee@samsung.com>
Reported-by: Kiryong Kim <kiryong1.kim@samsung.com>
Signed-off-by: Minsik Jeon <hmi.jeon@samsung.com>
Co-authored-by: Steven Seungcheol Lee <sc108.lee@samsung.com>


igaw:
Thanks!

PR #2275, opened 2024-03-30
nvme-print-stdout: use FDP configuration attributes definitions

ikegami-t:
Replace the hard coded shift and mask values in the print function.


igaw:
Thanks!

PR #2276, opened 2024-03-30
nvme-print-stdout: use NVME_MAJOR and NVME_MINOR definitions

ikegami-t:
~~Replase~~ **Replace** hard coded shift and mask values.


igaw:
Thanks!

PR #2277, opened 2024-03-31
nvme-print-stdout: use controller configuration definitions

ikegami-t:
~~Replase~~ **Replace** hard corded register value, shift and mask values.


igaw:
Thanks!

PR #2279, opened 2024-03-31
nvme-print-stdout: use admin queue attributes definitions

ikegami-t:
Replace hard coded register shift and mask values instead.


igaw:
Thanks!

PR #2282, opened 2024-04-03
nvme: print encoded key for gen-tls-key's -i option too

martin-gpy:
Printing the encoded key is useful with gen-tls-key's -i option as well.


igaw:
Thanks!

PR #2284, opened 2024-04-03
nvmf-connect: systemd hardening effort

igaw:
Apply the recommended hardening settings as recommended by openSUSE project but without "ProtectKernelTunables=true" as we need write access to /sys.

Links: https://en.opensuse.org/openSUSE:Security_Features#Systemd_hardening_effort

This configuration is in use for a couple of years and all seems to work fine. 

@tbzatek any objections?


tbzatek (approved):
No objections, in fact there's a similar initiative proposed for the next Fedora release: https://fedoraproject.org/wiki/Changes/SystemdSecurityHardening

I see `nvmefc-boot-connections.service` and `nvmf-autoconnect.service` in the proposal.

I haven't tested this change on my side though.

igaw:
I went through the list from the Fedora wiki and added all missing ones except:
- `PrivateTmp=yes`: we can't use this because it adds a dependency to all mounted units and we need to run the service before local-fs-pre.target. (Enabling this setting has the side effect of adding Requires= and After= dependencies on all mount units necessary to access /tmp/ and /var/tmp/. Moreover an implicitly After= ordering on [systemd-tmpfiles-setup.service(8)](https://www.freedesktop.org/software/systemd/man/latest/systemd-tmpfiles-setup.service.html#) is added.) https://www.freedesktop.org/software/systemd/man/latest/bootup.html
- `ProtectKernelTunables=yes`: we need write access to sysfs.
- `PrivateDevices=yes`: we need to access the /dev/nvme* devices
- `PrivateNetwork=yes`: we need to use TCP for nvme-tcp
- `User=`: we need to run as root.
- `CapabilityBoundingSet=`: not really sure what to add, we are running as root?
- `DevicePolicy=closed`: we need to read/write /dev devices
- `KeyringMode=private`: auth/tls nees to access global keyring
- `PrivateUsers=yes`: again we run as root
- `RestrictNamespaces=yes`: we need to run in main namespace
- `RestrictSUIDSGID=yes`: adds dependency on DynamicUser which we can't use.

I'll try to test this now

igaw:
I had to drop these below for getting it working with systemd 246

- `ProtectClock=yes`
- `NoNewPrivileges=yes`
- `SystemCallFilter=@system-service,@network-io`
- `SystemCallArchitectures=native`

No idea why.

igaw:
The current list is
```ProtectSystem=full
ProtectHome=true
ProtectHostname=true
ProtectKernelModules=true
ProtectKernelLogs=true
ProtectControlGroups=true
ProtectProc=invisible
RestrictRealtime=true
LockPersonality=yes
MemoryDenyWriteExecute=yes
RemoveIPC=yes
RestrictAddressFamilies=AF_INET AF_INET6
```

igaw:
@mwilck  could you have a look and check if I don't break the nbft bits?

tbzatek:
> The current list is
> 
> ```
> ProtectHome=true
> ProtectHostname=true
> ProtectKernelModules=true
> ProtectKernelLogs=true
> ProtectControlGroups=true
> ProtectProc=invisible
> RestrictRealtime=true
> LockPersonality=yes
> MemoryDenyWriteExecute=yes
> RemoveIPC=yes
> RestrictAddressFamilies=AF_INET AF_INET6
> ```

That looks reasonable to me. At least nothing spurious jumped at me while reading the `systemd.exec` manpage.


> @mwilck could you have a look and check if I don't break the nbft bits?

I think all the nbft bits are still within the usual bounds like the rest of the project. I.e. ipv4, ipv6 networking, resolver access, sysfs access (`/sys/firmware/acpi`), `/etc/nvme` readonly access.

mwilck:
> > @mwilck could you have a look and check if I don't break the nbft bits?
> 
> I think all the nbft bits are still within the usual bounds like the rest of the project. I.e. ipv4, ipv6 networking, resolver access, sysfs access (`/sys/firmware/acpi`), `/etc/nvme` readonly access.

I agree. Nothing catches the eye, at least.

tbzatek:
> So let's see what explodes :D

Thanks, I've pushed the patch in Fedora rawhide in the meantime.

PR #2286, opened 2024-04-06
nvme: check-tls-key fails despite having hostnqn file

martin-gpy:
Nvme check-tls-key requires both hostnqn and subsysnqn to be specified along with the keydata. Ideally, it should be able to fetch the hostnqn string from the hostnqn file if no hostnqn option is passed, irrespective of whether the insert option is specified or not. Also passing the subsysnqn option is required not just for identity v1, but for v0 as well, irrespective of the insert option. Fix accordingly.


igaw:
Thanks!

PR #2288, opened 2024-04-09
Initialize logging for commands

igaw:
We currently only initialize the libnvme logging for those commands which use the `nvme_root_t` object. In order to get the logging going for all commands we have to initialize the logging for every command.

This is a first attempt to solve this. Maybe it would be better to initialize the libvnme logging just once for all commands.

For this draft I'v just made it work for the `read` and `write` command (submit_io).

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -3269,7 +3269,7 @@ static bool nvme_match_device_filter(nvme_subsystem_t s,
 static int list_subsys(int argc, char **argv, struct command *cmd,
 		struct plugin *plugin)
 {
-	nvme_root_t r = NULL;
+	_cleanup_nvme_root_ nvme_root_t r = NULL;
 	enum nvme_print_flags flags;
 	const char *desc = "Retrieve information for subsystems";
 	nvme_scan_filter_t filter = NULL;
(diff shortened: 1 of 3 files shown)


ikegami-t:
Seems `nvme_create_root()` also can be moved into `parse_args()`. If this PR merged will do rebase #2285. Thank you.

igaw:
> Seems nvme_create_root() also can be moved into parse_args().

Yes this is a possibility. Let me play with it and see how this works out.

igaw:
depends on https://github.com/linux-nvme/libnvme/pull/824

nvme.c
```diff
@@ -9498,7 +9504,7 @@ static int show_topology_cmd(int argc, char **argv, struct command *command, str
 	const char *desc = "Show the topology\n";
 	const char *ranking = "Ranking order: namespace|ctrl";
 	enum nvme_print_flags flags;
-	nvme_root_t r;
+	_cleanup_nvme_root_ nvme_root_t r = NULL;
```
ikegami-t, line 9507: The `nvme_free_tree()` call in this function can be deleted.

util/cleanup.h
```diff
@@ -34,4 +34,10 @@ static inline void close_file(int *f)
 }
 #define _cleanup_file_ __cleanup__(close_file)
 
+static inline void cleanup_nvme_root(nvme_root_t *r)
+{
+	nvme_free_tree(*r);
```
ikegami-t, line 39: Seems `*r ` should be checked if NULL.

PR #2292, opened 2024-04-14
nvme-print-stdout: Use NVME_CMBLOC definitions

ikegami-t:
Use libnvme register definitions instead.


igaw:
FTR, depends on https://github.com/linux-nvme/libnvme/pull/814

PR #2293, opened 2024-04-15
fabrics: fix mem leak in nvmf_check_hostid_and_hostnqn()

martin-gpy:
Valgrind revealed a mem leak in nvmf_check_hostid_and_hostnqn() during nvme discover & connect. Leak traced to both hostid pointers not getting freed for cases where the hostid value matches the strings from both these respective hostid pointers. Fix the same.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -643,7 +643,8 @@ char *nvmf_hostid_from_hostnqn(const char *hostnqn)
 
 void nvmf_check_hostid_and_hostnqn(const char *hostid, const char *hostnqn, unsigned int verbose)
 {
-	char *hostid_from_file, *hostid_from_hostnqn;
+	_cleanup_free_ char *hostid_from_file = NULL;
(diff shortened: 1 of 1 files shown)


fabrics.c
```diff
@@ -643,7 +643,7 @@ char *nvmf_hostid_from_hostnqn(const char *hostnqn)
 
 void nvmf_check_hostid_and_hostnqn(const char *hostid, const char *hostnqn, unsigned int verbose)
 {
-	char *hostid_from_file, *hostid_from_hostnqn;
+	char *hostid_from_file = NULL, *hostid_from_hostnqn = NULL;
```
igaw, line 646: Let's  use the `_cleanup_free_` helpers here as well.

martin-gpy: Sure, will do.

igaw:
Thanks!

PR #2299, opened 2024-04-16
nvme: fix mem leak in multiple key related functions

martin-gpy:
Valgrind revealed mem leaks in multiple key related functions including gen_tls_key() and check_tls_key() during insert, and gen_dhchap_key(). This was traced to the respective cfg pointer not getting freed after fetching the hostnqn string from the hostnqn file. Fix the same.


igaw:
Thanks!

PR #2302, opened 2024-04-17
nvme-print-stdout: Use NVME_CMIC definitions

ikegami-t:
Use libnvme definitions to print CMIC data.


ikegami-t:
Depended on the libnvme PR changes linux-nvme/libnvme#817.

igaw:
Thanks!

PR #2303, opened 2024-04-18
add plugins to get addtional smart log for ssstc ventor

kobyou:
Add plugins to get addtional smart log for ssstc ventor


igaw (changes requested):
This PR wants to change all the file permissions for all existing files. I assume you are developing Windows. Please be careful with this.  Please drop all the file permission changes in this PR accordingly.

PR #2306, opened 2024-04-19
plugins/ocp: Adjusted variable type size carrying UUID index to 8 bit.

lgdacunh:
Changed uuidx type to __u8 in the related function calls, because that's the type that will feed struct nvme_get_log_args.


igaw:
Thanks!

PR #2307, opened 2024-04-21
nvme-print-stdout: Add CSTS.ST register stdout print

ikegami-t:
Added by NVMe revision 2.0 multiple controller shutdown enhancement.


ikegami-t:
Depended on the libnvme PR changes linux-nvme/libnvme#818.

igaw:
Thanks!

PR #2309, opened 2024-04-22
nvme: fix security buffer allocation size

twilfredo:
`sizeof(*sec_buf)` should return 1, which would mean `sec_buf` is not properly allocated to receive data from the target device. This fixes that.


keithbusch:
Ha! I guess this means no one has ever run this command before.

PR #2311, opened 2024-04-22
nvme: fix error path if write() fails in get_telemetry_log()

maurizio-lombardi:
Print an error message and return the error code to the caller.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -913,7 +913,9 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd,
 	while (data_remaining) {
 		data_written = write(output, data_ptr, data_remaining);
 		if (data_written < 0) {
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -913,7 +913,9 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd,
 	while (data_remaining) {
 		data_written = write(output, data_ptr, data_remaining);
 		if (data_written < 0) {
-			data_remaining = data_written;
+			err = errno;
```
igaw, line 916: Either do `err = -errno` or just set `err = -1`.

maurizio-lombardi: Done

igaw:
Thanks!

PR #2314, opened 2024-04-24
add a new plugin for SSSTC driver

kobyou:
add a new plugin for SSSTC driver

Signed-off-by: kobyou <kobyou@126.com>


igaw:
@kobyou I'll take it from here. I'll fix the final problems myself.

PR #2315, opened 2024-04-25
nbft: Discovery support

tbzatek:
This brings support for performing discovery and connect-all from Discovery Descriptor List records.

Needs https://github.com/linux-nvme/libnvme/pull/820 to parse URIs from NBFT.

Tested with static IPv4 and IPv6 against linux soft target. ~~For now, persistent discovery controllers have not been tested.~~


igaw:
Thanks!

PR #2317, opened 2024-04-26
The correct frl1/2/3 value not printed in the output of `zns id-ns`.

sharpLiu:


igaw:
Can I add your Signed-off-by tag?

sharpLiu:
> Can I add your Signed-off-by tag?

OK,Thanks

igaw:
Thanks!

PR #2319, opened 2024-04-29
nvme-print-json: Fix linux kernel check patch errors

ikegami-t:
Note: The long line string and strncpy() warnings are not fixed.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -33,7 +33,7 @@
 
 static const uint8_t zero_uuid[16] = { 0 };
 static struct print_ops json_print_ops;
-static struct json_object *json_r = NULL;
+static struct json_object *json_r;
 
 static void json_feature_show_fields(enum nvme_features_id fid, unsigned int result,
 				     unsigned char *buf);
@@ -132,6 +132,7 @@ static void obj_add_int_secs(struct json_object *o, const char *k, int v)
 static void obj_add_result(struct json_object *o, const char *v, ...)
 {
 	va_list ap;
+
 	va_start(ap, v);
 	char *value;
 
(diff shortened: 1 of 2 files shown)


ikegami-t:
Depended on the libnvme PR changes https://github.com/linux-nvme/libnvme/pull/828.

nvme-print-json.c
```diff
@@ -69,15 +69,15 @@ static void obj_add_uint_0x(struct json_object *o, const char *k, __u32 v)
...
+	sprintf(str, "%#x", v);
 	obj_add_str(o, k, str);
 }
 
 static void obj_add_uint_02x(struct json_object *o, const char *k, __u32 v)
 {
 	char str[STR_LEN];
 
-	sprintf(str, "0x%02x", v);
+	sprintf(str, "%#0*x", v ? 4 : 2, v);
```
igaw, line 80: This looks a bit wierd. The idea of this function is to print always two digits with the prefix `0x`. 

```c
        v = 0;
	printf("%#0*x\n", v ? 4 : 2, v);
	v = 456;
	printf("%#0*x\n", v ? 4 : 2, v);
```

gives

```
00
0x1c8
```

ikegami-t: In the past a related change done by the commit 4c7b19a also since `0x` included as field length as below.
```c
	v = 1;
	printf("%#0*x\n", v ? 4 : 2, v);
```
gives
```
0x1
```
but expected
```
0x01
```

ikegami-t:
Rebased the PR with latest master and fixed the issue mentioned by the comment https://github.com/linux-nvme/nvme-cli/pull/2321#discussion_r1589071418.

nvme-print-json.c
igaw: after playing around with `#` I think we should just stick with prefixing the string ourself. Too much troubles for something so trivial and I really would like to avoid to have these `v ? 4 : 2` around.

igaw:
Thanks!

PR #2320, opened 2024-04-29
sed_opal: Only build when we really have sed_opal headers

bsdimp:
There's already code to not build the sed opal plugin when the necessary headers aren't present. However, it doesn't work, since HAVE_SED_OPAL is always defined to be either 0 or 1, so the meson has() return true, even when the value is 0, causing the sed-opal plugin to build, even when the header files it needs are missing.

Note: This came up in the context of my updating the FreeBSD port to the latest. Meson makes it a lot easier than before to do a minimal change port, but this is one of the few actual bugs that I found and so should change upstream. This issue is independent of my port, even though it came up in that context.

Also updated to add a second commit that fixes a spelling error in meson.build while I'm in the neighborhood.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -160,7 +160,7 @@ conf.set10(
         name: 'linux/sed-opal.h'
 
     ),
-    description: 'Is linux/sed-opa.h include-able?'
+    description: 'Is linux/sed-opal.h include-able?'
 )
 conf.set10(
     'HAVE_KEY_TYPE',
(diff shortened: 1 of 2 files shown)


bsdimp:
Added SOBs, fixed email addr to be correct and add spelling fix to my original submission.
Should be good to go... It's super simple.

bsdimp:
All green, should be ready to review.

igaw:
Thanks a lot! Glad to hear that meson might your life easier.

PR #2321, opened 2024-04-30
Pel set feature

ikegami-t:


igaw:
Thansk!

nvme-print-json.c
```diff
@@ -73,14 +85,19 @@ static void obj_add_uint_0x(struct json_object *o, const char *k, __u32 v)
 	obj_add_str(o, k, str);
 }
 
-static void obj_add_uint_02x(struct json_object *o, const char *k, __u32 v)
+static void obj_add_uint_0nx(struct json_object *o, const char *k, __u32 v, int width)
 {
 	char str[STR_LEN];
 
 	sprintf(str, "0x%02x", v);
```
martin-belanger, line 92: Shouldn't `width` be used here?

igaw: yes, makes sense. @ikegami-t has another series which fixes this. I just saw it too late.

ikegami-t: Verry sorry for this was missed to implent to use the input parameter `width`. Probably mention the PR #2319 as fixed this so I will do rebase the PR #2319 since the PR #2321 merged already and fix this issue. Thank you.

PR #2322, opened 2024-04-30
nvme-print-stdout: fix hexadecimal field width

ikegami-t:
Correct "0x" prepended and zero padded hexadecimal field width. Fixes: 8f61eb4 ("nvme-cli: Add support set feature event in PEL") Fixes: 9b7c76c ("nvme-print-stdout: Add register print functions") Fixes: df11a0e ("nvme-print-stdout: Change hexadecimal value 0x prefix to use %#x")

### DIFF
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -487,8 +487,8 @@ static void stdout_persistent_event_log(void *pevent_log_info,
 			fid = NVME_GET(le32_to_cpu(set_feat_event->cdw_mem[0]), FEATURES_CDW10_FID);
 			cdw11 = le32_to_cpu(set_feat_event->cdw_mem[1]);
 
-			printf("Set Feature ID  :%#02x (%s),  value:%#08x\n", fid,
-				nvme_feature_to_string(fid), cdw11);
+			printf("Set Feature ID  :%#0*x (%s),  value:%#0*x\n", fid ? 4 : 2, fid,
+			       nvme_feature_to_string(fid), cdw11 ? 10 : 8, cdw11);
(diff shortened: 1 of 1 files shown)


igaw:
The `#` modifier doesn't work on my system correctly for `0`. I think we should just go with the old style and prefix our strings ourself. Way to complicated with these additional width strings.

```c
#include <stdio.h>

void test1(unsigned int v)
{
	printf("test0: v %d -> %#0*x\n", v, v ? 4 : 2, v);
}

void test2(unsigned int v)
{
	printf("test1: v %d -> 0x%02x\n", v, v);
}

{
	test1(0);
	test1(1);
	test1(300);
	test2(0);
	test2(1);
	test2(300);

	return 0;
}
```

produces

```shell
› ~/tmp/test
test0: v 0 -> 00
test0: v 1 -> 0x01
test0: v 300 -> 0x12c
test1: v 0 -> 0x00
test1: v 1 -> 0x01
test1: v 300 -> 0x12c
```

ikegami-t:
Okay let me close the PR changes. By the way about the `0` result is expected behavior as desribed by the man page as below.
<https://man7.org/linux/man-pages/man3/printf.3.html>
```
#      The value should be converted to an "alternate form". ...
              ...
              For x and X conversions, a nonzero result has the string
              "0x" (or "0X" for X conversions) prepended to it. ...
              ...
```

igaw:
Ah, good to know. I thought this is a bug but actually, it is the defined behavior. Not really working for us :(

ikegami-t:
Thanks for the confirmation.

PR #2323, opened 2024-04-30
Added new drive support

Inba-Vinayagam:
Added a new drive support in micron plugin. Please review

### DIFF
--- a/plugins/micron/micron-nvme.c
+++ b/plugins/micron/micron-nvme.c
@@ -56,7 +56,7 @@ enum eDriveModel {
 	M51CX,
 	M5407,
 	M5411,
-  M6003,
+	M6003,
 	UNKNOWN_MODEL
 };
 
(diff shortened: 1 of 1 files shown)


plugins/micron/micron-nvme.c
```diff
@@ -56,6 +56,7 @@ enum eDriveModel {
 	M51CX,
 	M5407,
 	M5411,
+  M6003,
```
igaw, line 59: The indention is off.

igaw (changes requested):
Please use subject prefix, e.g. 'plugins/micron: add support for new drive' and also add a commit message and a Signed-off-by.

Also use the correct indention. Thanks

igaw:
@hanumanthuh please also review. Thanks!

plugins/micron/micron-nvme.c
Inba-Vinayagam: plugins/micron: fixed the indention issue and update the PR signed off by Inba Vinayagam

igaw:
Alright, I'll close this PR for now. Please reopen or create a new one if you have an update. Thanks.

PR #2324, opened 2024-05-01
nvme-print-stdout: fix ASQB, ACQB and BMBBA prints

ikegami-t:
The fields needed be output only as the 52 most significant bits.


ikegami-t:
Depended on the libnvme PR changes https://github.com/linux-nvme/libnvme/pull/830.

PR #2325, opened 2024-05-01
Print json define

ikegami-t:


igaw:
Thanks! 

BTW, thanks a lot for doing these small commit changes, it makes the review way simpler. Good work!

PR #2326, opened 2024-05-01
nvme-wrap: remove unused nvme_cli_get_log_ana_groups()

calebsander:
`nvme_cli_get_log_ana_groups()` is not called anywhere. In preparation for a `libnvme` change that changes the signature of `nvme_get_log_ana_groups()`, just remove the function.


igaw:
Thanks!

PR #2331, opened 2024-05-03
plugins/solidigm: Automatic enabling Data Area 4 when retrieving Tele…

lgdacunh:
…metry.

vs-internal-log, stopped extracting Host Initiated Telemetry previous snapshot, simplified the type name of Telemetry snapshots to CIT and HIT, removed command options for setting namespace, and file prefix, and added command option to set output folder.


igaw:
Thanks!

PR #2332, opened 2024-05-04
Use C99 types for uint32_t

thesamesam:
<stdint.h> provides `uint32_t`, while `u_int_32` is an unofficial/internal typedef that glibc happens to provide. This fixes the build on musl.

Bug: https://bugs.gentoo.org/931194

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -34,6 +34,7 @@
 #include <inttypes.h>
 #include <locale.h>
 #include <stdio.h>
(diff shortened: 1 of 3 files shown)


thesamesam:
For completeness, I've submitted https://github.com/linux-nvme/libnvme/pull/838 too.

igaw:
added prefix to commit subject

igaw:
Thanks!

PR #2334, opened 2024-05-06
solidigm: Eliminate <linux/limits.h>

bsdimp:
ARG_MAX is defined in <limits.h>, per POSIX, but it's defined to be a variable in glibc. Instead, get rid of it entirely by using asprintf to construct the commands. This prevents us from trying a partially constructed command that might do something unintentional. For solidigm-market-log.c, there's no constants needed from <linux/limits.h>. With this we can delete it both places.

### DIFF
--- a/plugins/solidigm/solidigm-internal-logs.c
+++ b/plugins/solidigm/solidigm-internal-logs.c
@@ -12,7 +12,6 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <inttypes.h>
-#include <linux/limits.h>
 #include <time.h>
 
 #include "common.h"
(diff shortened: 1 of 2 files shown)


plugins/solidigm/solidigm-internal-logs.c
```diff
@@ -628,23 +627,26 @@
...
 			output_path = zip_name;
-			snprintf(cmd, sizeof(cmd), "rm -rf %s", cfg.dir_prefix);
+			free(cmd);
+			asprintf(&cmd, "rm -rf %s", cfg.dir_prefix);
 			printf("Removing %s\n", cfg.dir_prefix);
 			if (system(cmd) != 0)
 				perror("Failed removing logs folder");
+			free(cmd);
 		}
+		free(cmd);
```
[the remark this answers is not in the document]
bsdimp: Ah, that's correct... Fixed.

plugins/solidigm/solidigm-internal-logs.c
```diff
@@ -628,23 +627,25 @@ int solidigm_get_internal_log(int argc, char **argv, struct command *command,
 
 	if (log_count > 0) {
 		int ret_cmd;
-		char cmd[ARG_MAX];
+		char *cmd;
 		char *where_err = cfg.verbose ? "" : ">/dev/null 2>&1";
 
 		snprintf(zip_name, sizeof(zip_name), "%s.zip", cfg.dir_prefix);
-		snprintf(cmd, sizeof(cmd), "cd \"%s\" && zip -r \"../%s\" ./* %s", cfg.dir_prefix,
+		asprintf(&cmd, "cd \"%s\" && zip -r \"../%s\" ./* %s", cfg.dir_prefix,
```
HaroPanosyan, line 634: would be nice to check the return value -1 for asprintf() to make sure malloc internally worked and no other errors.

bsdimp: Updated, though I could also see not doing that since the prior code didn't check to see if snprintf overflowed...
Is the update good or too ugly?

HaroPanosyan: thanks

igaw:
rebase and merge conflict resolved

igaw:
Thanks!

Issue #2336, opened 2024-05-08
No output logs in fw-download


JohnGoogl:
Hi Team, 
While checking fw-download case, for invalid file size case it failed as expected with respective log, 
but, for valid file size case, its not showing any logs about command operation like success/failure.
And, after fw-download call for 'echo $?' command its printing '1' which tells non-positive with fw-download operation.

I tried with options like #progress(-p) #verbose(-v), but no luck in logs display. Any thoughts on this pls?

BR, Paul

ikegami-t:
Could you please try the PR #2337 changes to make sure since the `echo $?` output 1 caused by the error return as below and also only the memory allocation error does not output any message? The change is just for the error message output so the error itself not resolved.
```c
int main(int argc, char **argv)
{
...
	err = handle_plugin(argc - 1, &argv[1], nvme.extensions);
	if (err == -ENOTTY)
		general_help(&builtin);

	return err ? 1 : 0;
```

igaw:
I don't think it is a ENOMEM situation. And if so, then the print out of 'no mem' is highly likely to fail to, because printing also is involving allocation. I've just merged a feature which enables low level logging for all commands. So if you could try with the current libnvme and nvme-cli master and use `-vvv` we should hopefully see a bit more.

igaw:
You are likely not to observe any memory allocation with strace, because glibc uses buffers for streams.

ikegami-t:
`printf()` and `fprintf()` call `__vfprintf_internal()` (seems implemented as `vfprintf()`) and `vfprintf()` uses the buffer as below so is this failed if a ENOMEM situation?
<https://codebrowser.dev/glibc/glibc/stdio-common/vfprintf-internal.c.html#1554>
```c
  /* Set up the wrapping buffer.  */
  struct Xprintf (buffer_to_file) wrap;
  Xprintf (buffer_to_file_init) (buf: &wrap, fp: s);
```
Now I am thinking to close the PR #2337 but to make sure if possible I would like to confirm the detail if really highly likely to fail.

igaw:
How a C library implements these APIs is undefined. Embedded libc libraries usually try to hard to avoid to have dynamic allocation wherever possible. I haven't look at the glibc implementation (not it would help, their coding style is from another planet) but I would be very surprised if it would use a buffered implementation. So the question is, are you sure the buffer is not resized when we try to print 'no memory'?

A quick test shows we have a hidden allocation when using printf:

```shell
[...]

PR #2337, opened 2024-05-09
nvme: add -ENOMEM alloc error message output

ikegami-t:
Since for the error case no message output.


igaw:
As state [here](https://github.com/linux-nvme/nvme-cli/issues/2336#issuecomment-2166175151), there is not much we can't do safely when `ENOMEM` happens. Obviously, large allocation can and should be handled but all these hidden and small allocation are just not worth to handle. It makes the error code ugly and no one is testing it anyway. Rather just die with abort and be done with it.

PR #2338, opened 2024-05-09
ocp: Add Get DSSD Power State Feature (FID: C7h)

arthurshau:
Implements the Get DSSD Power State Feature (FID C7h). Also added an optional flag that you can pass to the command that allows the user to get all 3 select values at once, by calling get feature 3 times. 

Also fixed two typos that I failed to catch in previous reviews - one in the description of ```set-dssd-power-state-feature``` in ```ocp-nvme.h```, and one in the bash completion, where ```get-dssd-async-event-config``` was incorrectly named ```get-dssd-power-state-feature``` (which is what I'm implementing in this patch).

Signed-off-by: Arthur Shau <arthurshau@meta.com>


arthurshau:
@igaw Can you double check the ```checkpatch.pl``` and make sure those warnings are acceptable?

igaw:
All good. Thanks!

PR #2339, opened 2024-05-10
nvme: remove double free in persistent-event-log

francispravin5:
The pevent_collected structure uses the buffer address which is allocated using nvme_alloc_huge(). So, pevent_collected and mh.p has same address. Hence, removing _cleanup_free_ from pevent_collected to prevent double free.


igaw:
Thanks!

PR #2340, opened 2024-05-10
build: Update meson commands in Makefile

martin-belanger:
The `Makefile` is mainly used as a tool for people not familiar with `meson`. However, the `Makefile` is a good way to see how to invoke `meson` properly. This patch contains a few small fixes to match the preferred way of invoking `meson`.


igaw:
Thanks!

PR #2341, opened 2024-05-10
sed: only re-read partition table after unlock.

gjoyce-ibm:
The partition table was being re-read after both lock and unlock operations. The re-read would fail with an error message after a lock since the partiton was no longer readable.


igaw:
Thanks!

PR #2342, opened 2024-05-13
nvme: fix fw-commit MUD result message

francispravin5:
As per spec, 0th bit is set when the controller detects an overlapping firmware/boot partition image update command sequence due to processing a command from an Admin Submission Queue on a controller and the 1st bit is set due to processing the command from Management Endpoint. So, fixed the MUD result message.


ikegami-t:
Looks good.

igaw:
Thanks!

PR #2343, opened 2024-05-13
ocp: Switch OCP plugin to use semantic versioning

arthurshau:
Title - I probably should have done this a lot earlier.

I (sort of arbitrarily) chose 77f6c8fc3ad808422127c052cdbd40b62cb348d7 to be the starting point for version ```1.0.0```. From there, I worked forward in the history and more or less added 1 to the minor version for every command added to the OCP plugin. Happy to take suggestions if anyone has a better idea.


igaw:
Thanks!

PR #2344, opened 2024-05-13
docs: Add missing OCP plugin docs to meson

arthurshau:
Multiple commands for OCP plugin have associated doc text files, but weren't added to the meson.build, causing docs to not be generated.

Please let me know if I missed anything.


arthurshau:
@igaw Do you have any guidance on how to regenerate the docs, or should I leave that up to you?

igaw:
> @igaw Do you have any guidance on how to regenerate the docs, or should I leave that up to you?

All good. All the docs output files will be generated on release.

igaw:
Thanks!

PR #2345, opened 2024-05-14
plugins/nbft: Use default library logging

tbzatek:
Remove the workaround with creating empty root object and use the new libnvme API instead.

(reverts one part of #2215)


igaw:
Thanks!

PR #2348, opened 2024-05-16
nvme-print-stdout: fix format index in stdout_nvm_id_ns

francispravin5:
This fix for nvme spec 2.0 where FLBAS bits 6:5 indicates the most significant 2 bits of the Format Index
that was used to format the namespace. So, use nvme_id_ns_flbas_to_lbaf_inuse function to get the proper
format index of the namespace.


igaw:
Thanks!

PR #2349, opened 2024-05-17
plugins/solidigm: Added log pages to vs-internal-logs

lgdacunh:
Added some Identify pages, and some standard log, and some VU logs. Created abstract data type with common fields for internal logs. Better string buffer overflow avoidance.

### DIFF
--- a/plugins/solidigm/solidigm-internal-logs.c
+++ b/plugins/solidigm/solidigm-internal-logs.c
@@ -22,6 +22,7 @@
 #include "solidigm-util.h"
 
 #define DWORD_SIZE 4
+#define LOG_FILE_PERMISSION 0644
 
 enum log_type {
 	NLOG = 0,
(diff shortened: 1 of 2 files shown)


plugins/solidigm/solidigm-internal-logs.c
```diff
@@ -415,102 +425,376 @@ static int dump_nlogs(struct nvme_dev *dev, struct config cfg, int core)
...
+log_save_close_output:
+	close(output);
+	return err;
+}
+
+static int ilog_dump_identify_page(struct ilog *ilog, struct log *cns, __u32 nsid)
+{
+	__u8 data[NVME_IDENTIFY_DATA_SIZE];
+	__u8 *buff = cns->buffer ? cns->buffer : data;
+	char filename[sizeof(
```
igaw, line 495: Nitpicking: why not just using asprintf() with the cleanup helpers?

igaw:
Thanks! 

As there other places which use the fixed size buffer for string formatting, there was no point in asking to update this update. Maybe update the all the places in one go next time, if  you agree with my idea.

lgdacunh:
> Thanks!
> 
> As there other places which use the fixed size buffer for string formatting, there was no point in asking to update this update. Maybe update the all the places in one go next time, if you agree with my idea.

I agree asprintf() with cleanup is less error prone. I will figure out a strategy for this change.

PR #2350, opened 2024-05-20
sed: perform a tper revert after lsp revert

gjoyce-ibm:
The non-destructive revert was not sufficiently reverting the lsp. A TPER revert is required or the drive is only partially reverted.


igaw:
Thanks!

PR #2351, opened 2024-05-21
plugins/ssstc: Replace __uint16_t with uint16_t

kraj:
uint16_t is ISO defined and comes from stdint.h, makes it portable across glibc and musl on linux.


igaw:
Thanks!

PR #2356, opened 2024-05-28
nvme-print-stdout: fix to print the log line by line

francispravin5:
Fix to print the log within one line per each FID (LID - 12h) and NVMe-MI Command (LID - 13h) while printing in human readable form.


igaw:
Thanks!

PR #2357, opened 2024-06-03
nvme-print-stdout: print MEM bit of CTRATT field

francispravin5:
Print MDTS and Size Limits Exclude Metadata(MEM) bit of CTRATT field which added in TP-4099.


igaw:
Code looks fine, though a couple of question. Is the TP already released? I got the impression it is still in the 30 days review period. 

Second, the TP defines a few more bits, wouldn't it make sense to add them as well.  

And shouldn't we also update the json output as well?

francispravin5:
1. TP-4099 (MDTS Enhancement) is already ratified. 
2. There is only one bit field (MEM) is added as part of this TP (Fig-275).
3. As per the existing implementation, the entire value of ctratt filed is updated in json, not bit by bit.

Thanks!

igaw:
Thanks for the clarification. I must have looked at the wrong document. 

json: ah okay, I didn't check if we decode the field or not. Maybe we should?

igaw:
Thanks!

PR #2358, opened 2024-06-04
plugins/solidigm: Added extra VU fields to id-ctrl

lgdacunh:
Added 13 extra VU fields to Solidigm id-ctrl
including a buffer for correct offset


igaw:
Thanks!

PR #2360, opened 2024-06-04
plugins/solidigm: Added workload-tracker command

lgdacunh:
Added set up and continuous parse and consolidation of workload log.

### DIFF
--- a/plugins/solidigm/meson.build
+++ b/plugins/solidigm/meson.build
@@ -11,6 +11,7 @@ sources += [
   'plugins/solidigm/solidigm-temp-stats.c',
   'plugins/solidigm/solidigm-get-drive-info.c',
   'plugins/solidigm/solidigm-ocp-version.c',
+  'plugins/solidigm/solidigm-workload-tracker.c',
 ]
 subdir('solidigm-telemetry')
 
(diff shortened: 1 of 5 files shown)


lgdacunh:
@igaw, checkpatch is throwing an error when you add a new file. I think the new file makes sense in this case.

igaw:
> @igaw, checkpatch is throwing an error when you add a new file. I think the new files make sense in this case.

checkpatch is wrong in our case. We don't have a maintainer file, so it doesn't really make sense. Hmm we could though add one, but I don't know if this make sense or it is generally something the plugins maintainers are interested in.

igaw:
I've merge the update directly after resolving the merge conflict.

lgdacunh:
> I've merge the update directly after resolving the merge conflict.

Thanks!

lgdacunh:
> > @igaw, checkpatch is throwing an error when you add a new file. I think the new files make sense in this case.
> 
> checkpatch is wrong in our case. We don't have a maintainer file, so it doesn't really make sense. Hmm we could though add one, but I don't know if this make sense or it is generally something the plugins maintainers are interested in.
I am not aware of how the maintainer file works, but I would be interested if it streamline the automated checks.

PR #2361, opened 2024-06-05
plugins/ocp: Added ocp TCG Configuration Log Page

VigneshwaranSaravana:
Reviewed-by: Karthik Balan <karthik.b82@samsung.com>, Arunpandian J <arun.j@samsung.com>

### DIFF
--- a/Documentation/nvme-ocp-tcg-configuration-log.txt
+++ b/Documentation/nvme-ocp-tcg-configuration-log.txt
@@ -1,0 +1,44 @@
+nvme-ocp-tcg-configuration-log(1)
+=========================
+
+NAME
+----
+nvme-ocp-tcg-configuration-log - Retrieves tcg configuration of given OCP
+compliant device
+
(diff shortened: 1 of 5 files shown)


arthurshau:
I unfortunately don't have a drive to test the new feature, but as long as vendor feels like they need the functionality, I have nothing against merging it. After fixing the compiler warnings of course.

Documentation/nvme-ocp-tcg-configuration-log.txt
```diff
@@ -0,0 +1,44 @@
+nvme-ocp-tcg-configuration-log(1)
+=========================
```
igaw, line 2: This will not build. The two lines need to be exactly the same length.

igaw:
I've updated the commit message and dropped the html generated documentation(will be generated on release date). Please have a close look how this PR is supposed to look like (commit message), so I don't have to fix this up myself. Thanks.

PR #2363, opened 2024-06-08
documentation fix for format cmd

hasnpol:
Protection Information is not only 8 bytes of the metadata, it can also be 16 bytes such as in 32b and 64b guard protection formats.


igaw:
Thanks, I've updated the commit message and dropped the auto generated files. These will be regenerated on release.

igaw:
btw, usually the email address and the submission email should match

WARNING: From:/Signed-off-by: email address mismatch: 'From: Hasan Polat <hep1031@wildcats.unh.edu>' != 'Signed-off-by: Hasan Polat <hasnpol2002@gmail.com>'

PR #2369, opened 2024-06-17
build: update libnvme wrap

francispravin5:
Get PIF enum and the new field added in TP4141a.
* db3ff1a - types: add enum for Protection Information Format.
* 770f7bb - types: add new filed added in TP4141a .


igaw:
Thanks!

PR #2372, opened 2024-06-18
plugins/ocp: undeflow + index fixes for telemetry

bpaupore-wdc:
Printing the GUID was crashing due to the unsigned int never becoming less than 0, and fixing that triggered a warning about an incorrect max index when printing Event FIFO data.


igaw:
Thanks!

PR #2375, opened 2024-06-21
nvme-print: print the new fields added in TP4141a

francispravin5:
Print the newly added fields in TP4141a. Also, fixed the stcrs bit mask value.


igaw:
Thanks!

PR #2376, opened 2024-06-24
nvme-print-json: add missing va_end()

maurizio-lombardi:
Every va_start() should be matched by a corresponding va_end() to keep the static analyzers happy.


igaw:
Thanks!

PR #2377, opened 2024-06-24
Use ANA log page fixes from libnvme

calebsander:
Depends on linux-nvme/libnvme/pull/848

---

`nvme ana-log` only includes space for MNAN namespace IDs
in the allocated ANA log page buffer if bit 6 of ANACAP
in the Identify Controller structure isn't set.

But the specification says:
Bit 6 if set to '1', then the ANAGRPID field in the Identify Namespace
data structure (refer to the NVM Command Set Specification)
does not change while the namespace is attached to any controller

There is no suggestion that NSIDs won't be returned in the ANA log page.

However, we can exclude space for NSIDs from the allocated buffer
if the Get Log Page command is going to be sent with the RGO bit set,
i.e. the `--groups` argument is passed.

Therefore, use the `libnvme` helper `nvme_get_ana_log_len_from_id_ctrl()`
to calculate the maximum ANA log page length.
It ignores ANACAP and takes the RGO setting into account.

---

`nvme ana-log` currently uses `libnvme`'s `nvme{,_mi_admin}_get_log_ana()`
function to fetch the ANA log page.
As described in the commit adding `nvme_get_ana_log_atomic()` to `libnvme`,
this has the potential to overread the ANA log page
and may result in a torn result if the log page changes concurrently.
Use `nvme{,_mi_admin}_get_ana_log_atomic()` to only fetch up to the end
of the ANA log page and protect against concurrent log page changes.

### DIFF
--- a/nvme-wrap.c
+++ b/nvme-wrap.c
@@ -256,11 +256,11 @@ int nvme_cli_get_log_predictable_lat_event(struct nvme_dev *dev, bool rae,
 			   len, log);
 }
 
-int nvme_cli_get_log_ana(struct nvme_dev *dev,
-			 enum nvme_log_ana_lsp lsp, bool rae,
-			 __u64 offset, __u32 len, void *log)
+int nvme_cli_get_ana_log_atomic(struct nvme_dev *dev, bool rgo, bool rae,
+				unsigned int retries,
+				struct nvme_ana_log *log, __u32 *len)
 {
-	return do_admin_op(get_log_ana, dev, lsp, rae, offset, len, log);
+	return do_admin_op(get_ana_log_atomic, dev, rgo, rae, retries, log, len);
 }
 
 int nvme_cli_get_log_lba_status(struct nvme_dev *dev, bool rae,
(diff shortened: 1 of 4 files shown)


igaw:
Thanks, let's hope there are no big regressions. I don't expect it because the extensive testing code you provided. But one never knows :)

calebsander:
Agreed. I've tested it against our NVMe-oF controller (Pure Storage FlashArray) and it prints the same output as before:
```
[root@init21-24 nvme-cli]# nvme ana-log /dev/nvme1
Asymmetric Namespace Access Log for NVMe device: nvme1
ANA LOG HEADER :-
chgcnt	:	0
ngrps	:	1
ANA Log Desc :-
grpid	:	2194672188
nnsids	:	9
chgcnt	:	0
state	:	optimized
	nsid	:	1
	nsid	:	2
	nsid	:	3
	nsid	:	4
	nsid	:	5
	nsid	:	6
	nsid	:	7
	nsid	:	8
	nsid	:	9

[root@init21-24 nvme-cli]# nvme ana-log /dev/nvme1 --groups
Asymmetric Namespace Access Log for NVMe device: nvme1
ANA LOG HEADER :-
chgcnt	:	0
ngrps	:	1
ANA Log Desc :-
grpid	:	2194672188
nnsids	:	0
chgcnt	:	0
state	:	optimized

```

PR #2378, opened 2024-06-26
nvme: choose PIF from QPIF if QPIFS supports and PIF is QTYPE

francispravin5:
As per TP4141a:
"If the Qualified Protection Information Format Support(QPIFS) bit is set to 1 and the Protection Information Format(PIF) field is set to 11b (i.e., Qualified Type), then the pif is as defined in the Qualified Protection Information Format (QPIF) field."
So, choose PIF from QPIF if QPIFS supports and PIF is QTYPE.


igaw:
Thanks. Maybe we should move this into a helper eventually?

francispravin5:
Yes. We can create a helper function to get pif and sts value.

PR #2379, opened 2024-06-26
nvme: add helper function to get pif and sts value

francispravin5:
Add a helper function to get pif and sts value from elbaf filed.


igaw:
Thanks a lot!

PR #2380, opened 2024-06-28
nvme-print: Print PEL reserved event string

ikegami-t:
Previously PEL reserved event output as NULL string. This resolves to print as reserved event with the event type value.


igaw:
Thanks!

PR #2381, opened 2024-06-28
nvme: Add smart-log command fahrenheit option

ikegami-t:
Show temperatures in degrees fahrenheit by the option.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -767,6 +767,20 @@ void nvme_show_endurance_log(struct nvme_endurance_group_log *endurance_log,
 	nvme_print(endurance_log, flags, endurance_log, group_id, devname);
 }
 
+const char *nvme_degrees_string(long t, bool fahrenheit)
+{
+	static char str[STR_LEN];
+	long val = kelvin_to_celsius(t);
+
+	if (fahrenheit)
+		val = kelvin_to_fahrenheit(t);
+
+	sprintf(str, "%ld °%s", val, fahrenheit ? "F" : "C");
+
+
+	return str;
+}
+
 void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid,
 			 const char *devname, enum nvme_print_flags flags)
 {
--- a/nvme-print.h
+++ b/nvme-print.h
@@ -163,8 +163,8 @@ void nvme_show_supported_log(struct nvme_supported_log_pages *support,
 	const char *devname, enum nvme_print_flags flags);
 void nvme_show_error_log(struct nvme_error_log_page *err_log, int entries,
 	const char *devname, enum nvme_print_flags flags);
-void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid,
-	const char *devname, enum nvme_print_flags flags);
+void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid, const char *devname,
+			 enum nvme_print_flags flags);
 void nvme_show_ana_log(struct nvme_ana_log *ana_log, const char *devname,
 		       size_t len, enum nvme_print_flags flags);
 void nvme_show_self_test_log(struct nvme_self_test_log *self_test, __u8 dst_entries,
--- a/nvme-print.h
+++ b/nvme-print.h
@@ -322,4 +322,5 @@ void nvme_show_finish(void);
 bool nvme_is_fabrics_reg(int offset);
 bool nvme_registers_cmbloc_support(__u32 cmbsz);
 bool nvme_registers_pmrctl_ready(__u32 pmrctl);
+const char *nvme_degrees_string(long t, bool fahrenheit);
 #endif /* NVME_PRINT_H */
--- a/Documentation/nvme-smart-log.txt
+++ b/Documentation/nvme-smart-log.txt
@@ -9,7 +9,7 @@ SYNOPSIS
 --------
 [verse]
 'nvme smart-log' <device> [--namespace-id=<nsid> | -n <nsid>]
-			[--raw-binary | -b]
+			[--raw-binary | -b] [--fahrenheit | -f]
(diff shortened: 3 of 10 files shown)


nvme-print.c
```diff
@@ -763,10 +763,19 @@ void nvme_show_endurance_log(struct nvme_endurance_group_log *endurance_log,
 	nvme_print(endurance_log, flags, endurance_log, group_id, devname);
 }
 
+const char *nvme_degrees_string(long t, bool fahrenheit)
+{
+	static char str[STR_LEN];
+
+	sprintf(str, "%ld °%s", kelvin_to_degrees(t, fahrenheit), fahrenheit ? "F" : "C");
```
igaw, line 770: I think it would better to call the right kelvin_to_TYPE directly and get rid of kelvin_to_degrees, e.g.

```c

long val;

if (fahrenheit) 
  val = kelvin_to_fahrenheit(t);
else
  val = kelvin_to_celsius(t);

sprintf(str, "%ld °%s", val, fahrenheit ? "F" : "C");
```

nvme-print.h
```diff
@@ -163,8 +164,8 @@ void nvme_show_supported_log(struct nvme_supported_log_pages *support,
 	const char *devname, enum nvme_print_flags flags);
 void nvme_show_error_log(struct nvme_error_log_page *err_log, int entries,
 	const char *devname, enum nvme_print_flags flags);
-void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid,
-	const char *devname, enum nvme_print_flags flags);
+void nvme_show_smart_log(struct nvme_smart_log *smart, unsigned int nsid, const char *devname,
+			 enum nvme_print_flags flags, bool fahrenheit);
```
igaw, line 168: I don't like to pass this as additional argument down. What about using the flags for this? Obviously, this means we would need to use a simpler type for flags here and not `enum nvme_print_flags`. This would also address the strange handling of the flags where we treat half of the time as enum and the rest of the time as bit field. A bit of code churn but again I think if we name this argument `flags` it should also be right type. What do you think?

nvme-print.c
ikegami-t: Fixed as mentioned.

nvme-print.h
ikegami-t: Yes right and agreed. Fixed the changes but the flag type will be fixed separately later. Thank you.

igaw:
Thanks!

igaw:
I  think we should do this differently. Instead adding a new command line option, we should consult the locale settings as we do in other places already.

ikegami-t:
The locale settings look not clear to map the temperature scales.

igaw:
What's wrong with something like this?

```c
#include <locale.h>
#include <stdio.h>
#include <string.h>

const char *guess_temperature_unit(const char *country)
{
    const char* fahrenheit_countries[] = {"US", "BS", "BZ", "KY", "PW", NULL};
    int i;

    for (i = 0; fahrenheit_countries[i]; i++) {
        if (!strcmp(country, fahrenheit_countries[i]) )
            return "Fahrenheit";
    }

    return "Celsius";
}

{
    const char *locale, *tmp;
    char country[3];
    char *underscore;

    setlocale(LC_ALL, "");
    tmp = locale = setlocale(LC_ALL, NULL);

    underscore = strchr(locale, '_');
    if (underscore != NULL && strlen(underscore) > 1)
	    tmp += 3;

    strncpy(country, tmp, 2);
    country[2] = '\0';

    const char* temperature_unit = guess_temperature_unit(country);

    printf("Locale: %s\n", locale);
    printf("Country: %s\n", country);
    printf("Temperature Unit: %s\n", temperature_unit);

    return 0;
}
```
We could also look at the language settings if needed. I really don't see a problem with this.

ikegami-t:
Currently already both degrees celsius and kelvin values are output by the smart-log command and the -f option added outputs degrees fahrenheit and kelvin values.
```sh
tokunori@tokunori-desktop:~/nvme-cli$ nvme-build smart-log /dev/nvme0
...
temperature                             : 43 °C (316 K)
...
Temperature Sensor 1                    : 43 °C (316 K)
Temperature Sensor 2                    : 51 °C (324 K)
...
tokunori@tokunori-desktop:~/nvme-cli$ nvme-build smart-log /dev/nvme0 -f
...
temperature                             : 109 °F (316 K)
...
Critical Composite Temperature Time     : 0
Temperature Sensor 1                    : 109 °F (316 K)
Temperature Sensor 2                    : 125 °F (325 K)
...
```

ikegami-t:
About the fahrenheit countries is it enough as the following countries but just checked other countries for example "AS", "AQ" and "CY" etc. also seem the fahrenheit countries?
```c
    const char* fahrenheit_countries[] = {"US", "BS", "BZ", "KY", "PW", NULL};
```

igaw:
My search says that US is the only country which uses Fahrenheit as official temperature measurement.

> Fahrenheit is used in the United States, its territories and associated states (all served by the U.S. [National Weather Service](https://en.wikipedia.org/wiki/National_Weather_Service)), as well as the (British) [Cayman Islands](https://en.wikipedia.org/wiki/Cayman_Islands) and [Liberia](https://en.wikipedia.org/wiki/Liberia) for everyday applications. For example, U.S. weather forecasts, food cooking, and freezing temperatures are typically given in degrees Fahrenheit. Scientists, including meteorologists, use degrees Celsius or kelvin in all countries.

([Wikipedia](https://en.wikipedia.org/wiki/Fahrenheit))

There are a few more countries which seems to use Fahrenheit.  I'd say the above list looks okay. The last sentence from Wikipdia indicates that everyone else is able to deal with Celsius.

 And if someone is unhappy that a specific country is missing we can still add it.

igaw:
https://worldpopulationreview.com/country-rankings/countries-that-use-fahrenheit

ikegami-t:
Yes I also searched the page so will do add the 19 countries into the list. Thank you.

PR #2383, opened 2024-06-28
nvme-rpmb: send RPMB_REQ_READ_RESULT for authentication key programming

ikegami-t:
This follows the NVMe revision 2.0a authentication key data flow.

### DIFF
--- a/nvme-rpmb.c
+++ b/nvme-rpmb.c
@@ -602,9 +602,28 @@ static int rpmb_program_auth_key(int fd, unsigned char target,
 		goto out;
 	}
 
-	/* send request and read the result first */
-	rsp = rpmb_read_request(fd, req, req_size, rsp_size);
-	if (rsp == NULL || rsp->result != 0) {
+	/* send the request and get response */
+	err = send_rpmb_req(fd, req->target, req_size, req);
+	if (err) {
+		fprintf(stderr, "RPMB request 0x%04x for 0x%x, err: %d\n", req->type, req->target,
+			err);
+		goto out;
+	}
+
+	/* send the request to get the result and then request to get the response */
+	rsp = (struct rpmb_data_frame_t *)calloc(rsp_size, 1);
+	if (!rsp) {
+		fprintf(stderr, "failed to allocate response buffer memory\n");
+		err = -ENOMEM;
+		goto out;
+	}
+
+	rsp->target = req->target;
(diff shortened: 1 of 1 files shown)


yang8621:
Hi @ikegami-t 
My RPMB was programmed. I'm not sure a virgin one can be found, but I will try to seek one.
Before that, it would be great if the review could be done because key programming is a one-time operation (it can be tested only once).

igaw:
Okay, I'll see if I can wrap my head around the change for the review

igaw:
The first patch looks good.

The second one is not really necessary to fix the problem  and it doesn't really make sense just to update one function and leave the rest in the 'old' style. The third one is big and thus it easy to miss something which shouldn't be changed. A quick binary compare between old and new version (third patch) show that the resulting binary was not identically. So I am not really sure if we should touch this code without being able to test it.  In short, it's quiet a bit of a code churn without any real benefit. I would suggest we only do the minimal change as long we can't make sure the the resulting binary is identically and we know nothing has been broken by the change.

ikegami-t:
Thanks for your reviewing. Understood so deleted the second and third patches and rebased.

nvme-rpmb.c
```diff
@@ -602,9 +602,27 @@ static int rpmb_program_auth_key(int fd, unsigned char target,
...
+		fprintf(stderr, "failed to allocate response buffer memory\n");
+		err = -ENOMEM;
+		goto out;
+	}
+
+	rsp->target = req->target;
+	rsp->type = RPMB_REQ_READ_RESULT;
+	err = send_rpmb_req(fd, req->target, rsp_size, rsp);
+	if (err || rsp->result) {
+		fprintf(stderr, "Write-data read result 0x%x, error = 0x%x\n", rsp->result, err);
```
yang8621, line 625: Write-data -> Program auth key

ikegami-t: Thanks for the comment. Just fixed the patch as mentioned.

yang8621 (approved):

igaw:
@yang8621 did you find a test drive? I think it's fairly safe to merge it. If it really breaks something, we can fix it up when we get a bug report.

yang8621:
No.   But I think it can be merged. Thanks

igaw:
Thanks!

PR #2384, opened 2024-06-29
Nvme print flags

ikegami-t:


nvme.c
```diff
@@ -3437,7 +3439,8 @@ int __id_ctrl(int argc, char **argv, struct command *cmd, struct plugin *plugin,
 	NVME_ARGS(opts,
 		  OPT_FLAG("vendor-specific", 'V', &cfg.vendor_specific, vendor_specific),
 		  OPT_FLAG("raw-binary",      'b', &cfg.raw_binary,      raw_identify),
-		  OPT_FLAG("human-readable",  'H', &cfg.human_readable,  human_readable_identify));
+		  OPT_FLAG("human-readable",  'H', &cfg.human_readable,  human_readable_identify),
+		  OPT_FLAG("fahrenheit",      'f', &cfg.fahrenheit,      fahrenheit));
```
igaw, line 3443: I think we should use the locale information and not adding an explicit command line option.

ikegami-t:
As already merged split the first patch. Update the PR patch for the second patch only. The reformatting patches and the fahrenheit command line option I will do change and consider separately. Thank you.

PR #2386, opened 2024-06-30
nvme: update parse_args() return value handling

martin-gpy:
Treat non-zero return values from parse_args() as errors, and not just negative return values. Also makes this consistent with other functions in nvme.c where parse_args() is invoked.


ikegami-t:
Looks good.

igaw:
Thanks!

Issue #2365, opened 2024-06-11
Provide option to remove a TLS retained key from the kernel keyring


martin-gpy:
Just like we have a ```--insert``` option for ```nvme gen-tls-key``` & ```check-tls-key``` commands to insert a TLS retained key into the kernel keyring, it would help to have a ```--remove``` option too for both these commands, to remove a TLS retained key from the kernel keyring.

PR #2387, opened 2024-07-01
 nvme: add support to revoke TLS key

igaw:
Add support to nvme-cli to remove TLS keys from a keyring.

I've decided to add the remove option to the `tls-key` command because adding it to `gen-tls-key`
 or `check-tls-key` seems wrong. `tls-key` support import/export operation while the other two command name indicate for what they are should be used. The import option for `check-tls-key` is already wrong IMO. But let's keep it for backwards compatibility. 

Fixes: #2365

Depends: https://github.com/linux-nvme/libnvme/pull/856


Documentation/nvme-tls-key.txt
```diff
@@ -54,15 +54,73 @@ OPTIONS
 -e::
 --export::
 	Write the key data to the file specified by '--keyfile'
-	or stdou if not present.
+	or stdout if not present.
+
+-r <id>::
+--remove <id>::
```
ikegami-t, line 60: `=` needed and is `<id>` correct since seems better to be `<name>`, `<key>`, `<keyring>`, `<key-name>` or `<keyring-name>`.
```diff
- --remove <id>::
+ --remove=<name>::
```

Documentation/nvme-tls-key.txt
```diff
@@ -54,15 +54,73 @@ OPTIONS
 -e::
 --export::
 	Write the key data to the file specified by '--keyfile'
-	or stdou if not present.
+	or stdout if not present.
+
```
ikegami-t, line 58: The short option `-k` for `--keyfile` not correct but should be `-f`.
```diff
- -k <file>::
+ -f <file>::
```

nvme.c
```diff
@@ -9439,16 +9480,18 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
 	const char *keyfile = "File for list of keys.";
 	const char *import = "Import all keys into the keyring.";
 	const char *export = "Export all keys from the keyring.";
+	const char *remove = "Remove key from the keyring.";
 
 	FILE *fd;
```
ikegami-t, line 9485: `FILE` type can be changed to `_cleanup_file_ int` type.

Documentation/nvme-tls-key.txt
```diff
@@ -54,15 +54,73 @@ OPTIONS
...
+ 353599402 --alswrv      0 65534   \_ keyring: _uid.0
+ 475911922 ---lswrv      0     0   \_ keyring: .nvme
+ 649274894 --als-rv      0     0       \_ psk: NVMe0R01 hostnqn0 subsys0
+------------
+
+* Remove/revoke a key using the description as id and verifying with
+keyctl the operation
++
+------------
+# nvme tls-key --remove "NVMe0R01 hostnqn0 subsys0"
```
ikegami-t, line 102:
```diff
- # nvme tls-key --remove "NVMe0R01 hostnqn0 subsys0"
+ # nvme tls-key --remove="NVMe0R01 hostnqn0 subsys0"
```

ikegami-t (commented):
completions not added.

nvme.c
igaw: No, it's the wrong type, the cleanup expect an filedescriptor not a stream.

Documentation/nvme-tls-key.txt
igaw: The equal sign is not really needed and the rest of the examples in this file don't use it. So why here?

Documentation/nvme-tls-key.txt
igaw: Yep, tt's `desc` after reading a bit more on the keyring interface.

nvme.c
ikegami-t: I thought just to use the cleanup function instead of the free with the goto label out and also the file descriptor can be used instead of the stream (by using `read()`, `write()` or `fdopen()` etc.) but understood as not good as complicated. By the way seems a file stream cleanup function can be created instead I thought.

Documentation/nvme-tls-key.txt
ikegami-t: I see. By the way do you think the option description `--keyring=<name>` should be (or is better to be) `--keyring <name>` as not using the equal sign?

nvme.c
igaw: Yep, I was also thinking along this line, was just to lazy.

Documentation/nvme-tls-key.txt
igaw: Ah, okay that's where `=` is from. I don't really care which version we use in the docs, but we should try to keep it consistent. Need to check what the rest of the examples use.

util/cleanup.h
```diff
@@ -36,6 +36,13 @@ static inline void close_file(int *f)
 }
 #define _cleanup_file_ __cleanup__(close_file)
 
+static inline void close_stream(FILE **f)
+{
+	if (*f)
+		fclose(*f);
+}
+#define _cleanup_stream_ __cleanup__(close_stream)
```
igaw, line 44: after the cleanup.h sync in https://github.com/linux-nvme/nvme-cli/pull/2395 this should be 
```
static inline DEFINE_CLEANUP_FUNC(cleanup_file, FILE *, fclose)
#define _cleanup_file_ __cleanup__(cleanup_file)
```

igaw:
Well, there is some more to fix with the config.json connect all code path when there is only the config.json file around.

completions/bash-nvme-completion.sh
```diff
@@ -427,6 +427,11 @@ nvme_list_opts () {
 		"show-hostnqn")
 		opts+=$NO_OPTS
 			;;
+		"tls-key")
+		opts+=" --output-format= -o --verbose -v --keyring -k \
+			--keytype= -k --keyfile= -f --import -i \
+			--export= -e --revoke= -r"
```
ikegami-t, line 433:
```diff
- 			--export= -e --revoke= -r"
+ 			--export -e --revoke= -r"
```

nvme.c
```diff
@@ -9463,77 +9504,61 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
...
+	if (cnt != 1) {
+		nvme_show_error("Must specify either --import, --export or --revoke");
+		return -EINVAL;
 	}
 
-out:
-	if (cfg.keyfile)
-		fclose(fd);
+	if (cfg.export) {
+		err = nvme_scan_tls_keys(cfg.keyring, __scan_tls_key, fd);
```
ikegami-t, line 9554: Seems better to output any error message if failed.

nvme.c
```diff
@@ -9463,77 +9504,61 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
...
+	} else if (cfg.import) {
+		err = import_key(cfg.keyring, fd);
```
ikegami-t, line 9556: The function import_key() outpus the error messages in the function but the function nvme_scan_tls_keys() does not output any error message but returns the error codes only. (Seems not same for the error message output.)

nvme.c
igaw: Yep, added this part as well.

PR #2388, opened 2024-07-01
nvme-print: Use 'unsigned int' instead of 'unsigned'

ikegami-t:
Fix for the checkpatch review error below.
  Error: WARNING: Prefer 'unsigned int' to bare use of 'unsigned'


igaw:
Thanks!

Issue #1956, opened 2023-05-25
hostnqn is not used from config.json file


igaw:
`nvme connect-all` will only use the `hostnqn` from `/etc/nvme/hostnqn`. 

With given configuration `nvme connect-all` will not connect to anything:

```shell
# cat /etc/nvme/hostnqn
nqn.2014-08.org.nvmexpress:uuid:31333937-3136-584d-5135-323430365637
```

```json
# cat config.json 
[
  {
    "hostnqn":"nqn.2014-08.org.nvmexpress:uuid:1a9e23dd-466e-45ca-9f43-a29aaf47cb21",
    "hostid":"1a9e23dd-466e-45ca-9f43-a29aaf47cb21",
   ...
  }
]
```

Only if we provide the hostnqn *and* hostid it will connect:

```shell
# nvme connect-all --hostnqn nqn.2014-08.org.nvmexpress:uuid:1a9e23dd-466e-45ca-9f43-a29aaf47cb21 --hostid 1a9e23dd-466e-45ca-9f43-a29aaf47cb21
```

I am not sure if this is the desired behavior. What do people expect with such a configuration?

igaw:
@martin-gpy @hreinecke @johnmeneghini

johnmeneghini:
> I am not sure if this is the desired behavior. What do people expect with such a configuration?

nvme-cli has always taken the host-nqn and hostid from /etc/nvme/hostnqn and /etc/nvme/hostid and historically users were not required to add either one to the command line.  My preferred way to handle this problem would be:

1. take host-nqn and hostid from /etc/nvme files
2. override host-nqn or hostid with values from discovery.conf
3. override host-nqn or hostid with values from json.config
4. override host-nqn or hostid with values from the command line

I think this would provide a fully backwards compatible method of supporting the previous precedence.

igaw:
Make sense to me. 

BTW, should we insist on that the user providing both `hostnqn` and `hostid`? 

The current code expects that the user provides both. And with the recent discussion/decision upstream on this [topic](https://lore.kernel.org/linux-nvme/20230512154155.38694-1-mgurtovoy@nvidia.com/), newer kernel will only allow a 1:1 mapping. So should we be smart here as well and warn if they do not match and if only `hostnqn` is provided we figure out `hostid`?

martin-gpy:
> I think this would provide a fully backwards compatible method of supporting the previous precedence.

Yes, this approach looks good.

[...]

PR #2394, opened 2024-07-03
Use libnvme helper to retrieve hostnqn/hostid

igaw:
There are several  places to lookup the default hostnqn/hostid. Add a helper which implements a canonical way to do this.

Fixes: #https://github.com/linux-nvme/nvme-cli/issues/1956
Depends: https://github.com/linux-nvme/libnvme/pull/857


igaw:
Well, there is some more to fix with the config.json connect all code path when there is only the config.json file around.

PR #2395, opened 2024-07-04
nvme: change to use libnvme cleanup functions

ikegami-t:
Delete duplicated functions and change _cleanup_file_ to _cleanup_fd_.


ikegami-t:
Depended on the linux-nvme/libnvme#858.

igaw:
I don't want to export the private `cleanup.h` file from libnvme. We can update version in nvme-cli using the same names as in libnvme though.

ikegami-t:
Just updated the nvme-cli version to use the same names as in libnvme.

igaw:
Thanks!

PR #2396, opened 2024-07-04
nvme-print: check locale to use temperatures in degrees fahrenheit

ikegami-t:
Delete the fahrenheit option separately.

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <time.h>
 #include <sys/stat.h>
+#include <locale.h>
 
 #include "nvme.h"
 #include "libnvme.h"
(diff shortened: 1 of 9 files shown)


nvme-print.c
```diff
@@ -767,10 +769,45 @@ void nvme_show_endurance_log(struct nvme_endurance_group_log *endurance_log,
...
+	const char *locale;
+	char country[3];
+	char *underscore;
+
+	setlocale(LC_ALL, "");
+	locale = setlocale(LC_ALL, NULL);
+
+	underscore = strchr(locale, '_');
+	if (underscore != NULL && strlen(underscore) > 1)
+		locale += 3;
```
igaw, line 798: Hmm, we might a bit more smarter here. IIRC, there are locales which have more than two chars before the underscore. Maybe `locale += strlen(understore)`?

nvme-print.c
```diff
@@ -767,10 +768,45 @@ void nvme_show_endurance_log(struct nvme_endurance_group_log *endurance_log,
...
+{
+	const char *locale;
+	char country[3] = { 0 };
+	const char *underscore;
+
+	setlocale(LC_ALL, "");
+	locale = setlocale(LC_ALL, NULL);
+
+	underscore = strchr(locale, '_');
+	if (underscore && strlen(underscore) > 1)
```
igaw, line 797: This should be `strlen(underscore) >= 3` because of the `memcpy`.

igaw: I've fixed this one up.

igaw:
Thanks!

PR #2397, opened 2024-07-04
common.h: Avoid using unsupported load/store instructions in arm64 VMs

oupton:
Using nvme show-regs within a VM on arm64 can sometimes lead to VM termination.

To answer why this happens: one of the deficiencies of the Arm architecture is that there exists a range of load/store instructions that have insufficient decode information for traps taken to the hypervisor. KVM, for example, may raise an external abort or outright terminate the VM depending on the configuration.

This is a known problem on the kernel side, and is fixed by using assembly MMIO accessors w/ 'safe' load/store instructions. So do exactly that, providing arm64-specific accessors and falling back to plain old volatile pointer accesses for other architectures.

Reported-by: William Butler <wab@google.com>

### DIFF
--- a/common.h
+++ b/common.h
@@ -17,43 +17,81 @@
 #define __packed __attribute__((__packed__))
 #endif /* __packed */
 
+/*
+ * VMs on arm64 can only use a subset of instructions for MMIO that provide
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #2399, opened 2024-07-05
ocp: Update Plugin Version

jeff-lien-wdc:
Update ocp plugin version to 2.9.0


igaw:
I am fine with this change. Not sure what the version strategy of this plugin is. @arthurshau what's your thoughts on this?

arthurshau:
Yes, that's fine.

igaw:
Thanks!

PR #2401, opened 2024-07-07
ocp-nvme: add get-error-injection command

ikegami-t:
The set-error-injection command will be added separately.


igaw:
@arthurshau please have a look.

arthurshau:
LGTM!

igaw:
Thanks!

PR #2406, opened 2024-07-08
nvme: fix verbose logging for certain commands

martin-gpy:
Verbose logging for certain nvme commands like id-ctrl, id-ns, smart-log, sanitize-log, etc. does not print anything additional or verbose compared to the regular output. These commands already implement a human-readable option which works fine, but it is the verbose option that is broken. So fix the verbose option here to ensure it is on par with the human-readable option. Later if required, we can eliminate any one of these options for these commands to avoid redundancy.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -578,7 +578,7 @@ static int get_smart_log(int argc, char **argv, struct command *cmd, struct plug
 	if (cfg.raw_binary)
 		flags = BINARY;
 
-	if (cfg.human_readable)
+	if (cfg.human_readable || argconfig_parse_seen(opts, "verbose"))
 		flags |= VERBOSE;
 
 	if (cfg.fahrenheit)
@@ -1091,7 +1091,7 @@ static int get_effects_log(int argc, char **argv, struct command *cmd, struct pl
 	if (cfg.raw_binary)
 		flags = BINARY;
 
-	if (cfg.human_readable)
+	if (cfg.human_readable || argconfig_parse_seen(opts, "verbose"))
 		flags |= VERBOSE;
 
 	list_head_init(&log_pages);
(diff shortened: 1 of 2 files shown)


ikegami-t:
The verbose option shows detail information if twice input as `-vv` for example. (If only once specificed the command latency only ouput.) Some commands options may be duplicated with the human-readable and verbose options but basically the verbose option is for the command detail information output but not human readable I think.

martin-gpy:
> The verbose option shows detail information if twice input as `-vv` for example. (If only once specificed the command latency only ouput.) 

Yes, I know. But the current ```-v``` invocation is a no-op (other than printing additional latency info along with the regular output) for these commands, and that is what this patch attempts to address here. If ```-vv``` is specified, it will continue to display additional ioctl structure values as before. This patch doesn't hinder that. So we are still good here.

> Some commands options may be duplicated with the human-readable and verbose options but basically the verbose option is for the command detail information output but not human readable I think.

As described above, this patch only fixes the ```-v``` invocation which is currently a no-op for these commands. Doesn't alter anything else like the ```-vv``` option. And it fixes this ```-v``` option by making it on par with the human readable option. As described in the commit, this does have the downside that you now have two redundant options but that anyways is better than having a no-op ```-v``` option in the first place, isn't it? And that could be cleaned later if need be.

igaw:
I like the idea to use `-v` as synonym for `-H`. If we go this route I think we should move the latency output to the next level of verbose (`-vv`). TBH, I've never liked it to have it split away from the debug output.

martin-gpy:
> I like the idea to use `-v` as synonym for `-H`. If we go this route I think we should move the latency output to the next level of verbose (`-vv`). TBH, I've never liked it to have it split away from the debug output.

Ah, ok. So basically you are asking to:

- Make ```-v``` synonymous with ```-H``` (which this patch already does)
- Additionally display the latency & ioctl command values in the debug (i.e. ```-vv```) output. 

Will give that a try here.

igaw:
Yes. I agree with your reasoning to retire `-H` eventually and use the more generic `-v` flag for all commands.

IIRC, the latency output is printed at INFO level instead of DEBUG.

igaw:
Thanks!

Issue #2405, opened 2024-07-08
nvme effects-log shows nothing when set to binary


rwk-unil:
The output format "binary" shows nothing when chosen :

```
sudo nvme effects-log -o binary /dev/nvme0
```

Neither does :

```
sudo nvme effects-log -b /dev/nvme0
```

The other formats work :

```
sudo nvme effects-log -o normal /dev/nvme0
NVM Command Set Log Page
--------------------------------------------------------------------------------
Admin Commands
ACS0     [Delete I/O Submission Queue     ] 00000001
ACS1     [Create I/O Submission Queue     ] 00000001
ACS2     [Get Log Page                    ] 00000001
ACS4     [Delete I/O Completion Queue     ] 00000001
ACS5     [Create I/O Completion Queue     ] 00000001
ACS6     [Identify                        ] 00000001
ACS8     [Abort                           ] 00000001
ACS9     [Set Features                    ] 00000001
ACS10    [Get Features                    ] 00000001
ACS12    [Asynchronous Event Request      ] 00000001
ACS16    [Firmware Commit                 ] 00000001
ACS17    [Firmware Image Download         ] 00000001
ACS20    [Device Self-test                ] 00000001
ACS128   [Format NVM                      ] 00020003
ACS129   [Security Send                   ] 00020003
ACS130   [Security Receive                ] 00010001

I/O Commands
IOCS0    [Flush                           ] 00010001
IOCS1    [Write                           ] 00000003
IOCS2    [Read                            ] 00000001
IOCS4    [Write Uncorrectable             ] 00000003
IOCS5    [Compare                         ] 00000001
IOCS8    [Write Zeroes                    ] 00000003
IOCS9    [Dataset Management              ] 00010003
```

```
sudo nvme effects-log -o json /dev/nvme0
[
  {
    "command_set_identifier":0,
    "admin_cmd_set":{
      "ACS_0 (Delete I/O Submission Queue)":1,
      "ACS_1 (Create I/O Submission Queue)":1,
      "ACS_2 (Get Log Page)":1,
      "ACS_4 (Delete I/O Completion Queue)":1,
      "ACS_5 (Create I/O Completion Queue)":1,
      "ACS_6 (Identify)":1,
      "ACS_8 (Abort)":1,
      "ACS_9 (Set Features)":1,
      "ACS_10 (Get Features)":1,
      "ACS_12 (Asynchronous Event Request)":1,
      "ACS_16 (Firmware Commit)":1,
[...]

PR #2409, opened 2024-07-09
nvme-print-binary: add effects-log command output missed

ikegami-t:
The csi data added before effects log page output.

### DIFF
--- a/nvme-print-binary.c
+++ b/nvme-print-binary.c
@@ -297,6 +297,16 @@ static void binary_discovery_log(struct nvmf_discovery_log *log, int numrec)
 	      numrec * sizeof(struct nvmf_disc_log_entry));
 }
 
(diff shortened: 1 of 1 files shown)


ikegami-t:
Fix for the isseu #2405.

igaw:
The 1.16 code base only prints the effects log in binary mode. Is this something we should/need to keep?

ikegami-t:
The effects log was changed by 155fbebfe to add CSI but looks not applied the changes in the 1.16 code. So currently multiple CSIs effects logs are output then if not added the CSI data then seems difficult to know which CSI effects log from the binary output. But still should we keep the output before?

igaw:
I suppose the main use case is to store it into a file and do some operation on it. I was wondering if we mix different section into the output it makes it hard to use. But the CSI section and the effect section have a fixed size.

Alright, let's go with this.

PR #2413, opened 2024-07-12
nvme: check MD size with PI size when PRACT set to 1

francispravin5:
As per NVM Command Set Specification, Rev 1.0d, Sec-5.2.2 PRACT Bit: "If PRACT bit is set to 1 and Metadata size equals to the Protection Information size, the protection information is stripped or inserted by the controller."
Currently NVMe supports 8 and 16 byte size of Protection Information. So, use pi_size instead of 8 byte.


igaw:
Thanks!

PR #2419, opened 2024-07-15
wdc: Update and refactor the C0h log page parsing

jeff-lien-wdc:
Add OCP 2.5 fields
Create common functions to get and parse the
different variations of the C0 log page.


igaw:
Thanks!

PR #2421, opened 2024-07-16
argconfig refactoring

calebsander:


igaw:
Thanks a lot! Nice cleanup!

PR #2422, opened 2024-07-17
util: remove unnecessary NULL check in cleanup_nvme_root()

calebsander:
`cleanup_nvme_root()` is checking whether the passed pointer is non-NULL, but since the pointer is to a local variable, this will always be true. Therefore, remove the check and always call `nvme_free_tree()`.


igaw:
Thanks!

PR #2423, opened 2024-07-17
nvme: avoid unnecessary dup() + close() in io_mgmt_send()

calebsander:
`_cleanup_fd_` doesn't close `STD{IN,OUT,ERR}_FILENO`, so don't bother creating a duplicate fd.

Fixes: 7e1f6799f8cd (`nvme: use cleanup helper to close file descriptor`)


igaw:
Ah, yes, keep forgetting this. Thanks!

PR #2424, opened 2024-07-17
Changed latency_monitor_feature_enable value to 01h as per OCP spec v2.0

mh2022-meta:
Changed latency_monitor_feature_enable value to 01h as per OCP spec v2.0

image


mh2022-meta:
Approved

igaw:
Was this a typo, or is this a change in the spec=

igaw:
@arthurshau not sure what to with this one. Could you have a look? Thanks!

igaw:
Argh, was too fast again with hitting the merge button. Anyway, the fix is in now.

PR #2425, opened 2024-07-17
fabrics: remove unused _discover_from_json_config_file() argument

calebsander:
`_discover_from_json_config_file()` takes a `nvme_subsystem_t` argument but doesn't use it.
Drop the unnecessary argument.

Fixes: b3b9963396c5 (`fabrics: refactore discover from json config`)

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -502,7 +502,7 @@ next:
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #2426, opened 2024-07-17
fabrics: update printing an error for ENOENT too

martin-gpy:
Many nvme commands such as discover, connect, connect-all, etc. simply return to the prompt without printing any useful error message when the respective nvme driver modules are not loaded in the kernel. This is because nothing gets printed when nvme_scan_topology() returns an ENOENT due to missing nvme system files stemming from not loading the appropriate nvme modules. Fix the same.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -1106,6 +1106,14 @@ int nvmf_disconnect(const char *desc, int argc, char **argv)
 	nvme_root_skip_namespaces(r);
 	ret = nvme_scan_topology(r, NULL, NULL);
 	if (ret < 0) {
+		/*
+		 * Do not report an error when the modules are not
+		 * loaded, this allows the user to unconditionally call
+		 * disconnect.
+		 */
+		if (errno == ENOENT)
+			return 0;
+
 		fprintf(stderr, "Failed to scan topology: %s\n",
 			nvme_strerror(errno));
 		return -errno;
@@ -1174,9 +1182,16 @@ int nvmf_disconnect_all(const char *desc, int argc, char **argv)
 	nvme_root_skip_namespaces(r);
 	ret = nvme_scan_topology(r, NULL, NULL);
 	if (ret < 0) {
-		if (errno != ENOENT)
-			fprintf(stderr, "Failed to scan topology: %s\n",
-				nvme_strerror(errno));
+		/*
+		 * Do not report an error when the modules are not
+		 * loaded, this allows the user to unconditionally call
+		 * disconnect.
+		 */
+		if (errno == ENOENT)
+			return 0;
+
+		fprintf(stderr, "Failed to scan topology: %s\n",
+			nvme_strerror(errno));
 		return -errno;
 	}
 
--- a/fabrics.c
+++ b/fabrics.c
@@ -45,6 +45,7 @@
 #include "nbft.h"
 #include "nvme-print.h"
 #include "fabrics.h"
+#include "util/cleanup.h"
 #include "util/logging.h"
 
 #define PATH_NVMF_DISC		SYSCONFDIR "/nvme/discovery.conf"
@@ -1064,7 +1065,7 @@ static void nvmf_disconnect_nqn(nvme_root_t r, char *nqn)
 int nvmf_disconnect(const char *desc, int argc, char **argv)
 {
 	const char *device = "nvme device handle";
-	nvme_root_t r;
+	_cleanup_nvme_root_ nvme_root_t r = NULL;
 	nvme_ctrl_t c;
 	char *p;
 	int ret;
@@ -1107,8 +1108,7 @@ int nvmf_disconnect(const char *desc, int argc, char **argv)
 	if (ret < 0) {
 		fprintf(stderr, "Failed to scan topology: %s\n",
 			nvme_strerror(errno));
-		nvme_free_tree(r);
-		return ret;
+		return -errno;
 	}
 
 	if (cfg.nqn)
@@ -1125,7 +1125,6 @@ int nvmf_disconnect(const char *desc, int argc, char **argv)
 			if (!c) {
 				fprintf(stderr,
 					"Did not find device %s\n", p);
-				nvme_free_tree(r);
 				return -errno;
 			}
 			ret = nvme_disconnect_ctrl(c);
@@ -1135,16 +1134,15 @@ int nvmf_disconnect(const char *desc, int argc, char **argv)
 					p, nvme_strerror(errno));
 		}
 	}
-	nvme_free_tree(r);
 
 	return 0;
 }
 
 int nvmf_disconnect_all(const char *desc, int argc, char **argv)
 {
+	_cleanup_nvme_root_ nvme_root_t r = NULL;
 	nvme_host_t h;
 	nvme_subsystem_t s;
-	nvme_root_t r;
 	nvme_ctrl_t c;
 	int ret;
 
@@ -1179,8 +1177,7 @@ int nvmf_disconnect_all(const char *desc, int argc, char **argv)
 		if (errno != ENOENT)
 			fprintf(stderr, "Failed to scan topology: %s\n",
 				nvme_strerror(errno));
-		nvme_free_tree(r);
-		return ret;
+		return -errno;
 	}
 
 	nvme_for_each_host(r, h) {
@@ -1200,7 +1197,6 @@ int nvmf_disconnect_all(const char *desc, int argc, char **argv)
 			}
 		}
 	}
-	nvme_free_tree(r);
 
 	return 0;
 }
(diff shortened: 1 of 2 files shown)


igaw:
IIRC, this is done on purpose. I agree it would make sense to print any error though.

@keithbusch do you remember why it is this way?

calebsander:
This is essentially reverting #2250. As I [commented there](https://github.com/linux-nvme/nvme-cli/pull/2250/files#r1530578816), I agree suppressing the warning message is not an improvement and am in favor of this revert. #2241 was supposed to be about fixing `nvme disconnect-all` not to error out if `nvme_core` isn't loaded, rather than making it error out with no helpful output.

martin-gpy:
So what are the next steps here? 

It is indeed confusing for an end user to see a blank output for these nvme commands during these ENOENT scenarios. Printing an error is definitely helpful/intuitive for such a case...

igaw:
If I get this right, we have to decide between two opposing behavior If modules are not loaded:

1) don't report anything including no error code 

or

2) report it and set error code

I get the arguments for both ways. Personally, I tend to say reporting an error is more user friendly. We could obviously added a global options which like `--no-error-if-modules-are-missing`...

martin-gpy:
> If I get this right, we have to decide between two opposing behavior If modules are not loaded:
> 
>     1. don't report anything including no error code
> 

What exactly is the argument for this behavior? We did have multiple instances from the field where this was reported, and it actually took some time to figure out that the respective modules not loaded was the culprit. Printing a simple error could have saved all that effort..

> 
> or
> 
>     2. report it and set error code
> 
> 
> I get the arguments for both ways. Personally, I tend to say reporting an error is more user friendly. We could obviously added a global options which like `--no-error-if-modules-are-missing`...

I honestly don't prefer a global option for this case. Seems like an overkill for something so trivial...

calebsander:
Agree. My original ask was for `nvme disconnect-all` not to fail if `nvme_core` is unloaded, so it could be always be called to "disconnect all NVMe-oF devices" rather than making the caller check if `nvme_core` is loaded, or if there are connected devices, or always load `nvme_core` first.
But `nvme connect` can't possibly work without the module loaded. Either it should fail if the module isn't loaded with a helpful error message, or it should automatically load the module if it's not already.

martin-gpy:
Had chatted with @igaw on this. Seems the concern is primarily with ```nvme disconnect-all``` here. So we could probably suppress this ENOENT error for ```disconnect-all``` alone, and keep it for the remaining commands. Will make that change now.

igaw:
Looks good to me. IIRC, we still return an error case though for the `nvme disconnect` case. I think we should set it to `0` when `errno` says `ENOENT`.

martin-gpy:
> Looks good to me. IIRC, we still return an error for the `nvme disconnect` case. I think we should set it to `0` when `errno` says `ENOENT`.

Fine, will make that change too.

igaw:
I've refactored the error path slightly and splitted the change for disconnect case out into a separate patch. What do you think?

martin-gpy:
> I've refactored the error path slightly and splitted the change for disconnect case out into a separate patch. What do you think?

Other than a few minor review comments, this looks good to me.

fabrics.c
```diff
@@ -1107,11 +1106,17 @@ int nvmf_disconnect(const char *desc, int argc, char **argv)
...
-				nvme_strerror(errno));
-		nvme_free_tree(r);
-		return ret;
+		/*
+		 * Do not report an error when the modules are not
+		 * loaded, this allows the user to uncodinitally call
+		 * disconnect.
+		 */
+		if (errno == ENOENT)
+			return 0;
```
calebsander, line 1115: This looks great for `nvme disconnect-all` and `nvme disconnect --nqn`.
However, for `nvme disconnect --device`, perhaps we should report an error if the module is unloaded? Because whatever device(s) the user is requesting to disconnect definitely don't exist.

nvme.c
```diff
@@ -3331,8 +3331,7 @@ static int list_subsys(int argc, char **argv, struct command *cmd,
 
 	err = nvme_scan_topology(r, filter, (void *)devname);
 	if (err) {
-		if (errno != ENOENT)
-			nvme_show_error("Failed to scan topology: %s", nvme_strerror(errno));
+		nvme_show_error("Failed to scan topology: %s", nvme_strerror(errno));
```
calebsander, line 3334: Personally feel like `nvme list-subsys` and `nvme list` should succeed with empty output if the modules are unloaded. But if others disagree, I am fine with keeping it this way.

fabrics.c
```diff
@@ -1213,11 +1221,12 @@ int nvmf_config(const char *desc, int argc, char **argv)
 	char *subsysnqn = NULL;
 	char *transport = NULL, *traddr = NULL;
 	char *trsvcid = NULL, *hostnqn = NULL, *hostid = NULL;
-	char *hnqn = NULL, *hid = NULL;
+	_cleanup_free_ char *hnqn = NULL;
+	_cleanup_free_ char *hid = NULL;
```
calebsander, line 1225: Pre-existing, but these variables look redundant with `hostnqn` and `hostid`, since they are assigned at the same time. Why do we need both?

fabrics.c
```diff
@@ -1317,12 +1324,6 @@ int nvmf_config(const char *desc, int argc, char **argv)
 	if (dump_config)
 		nvme_dump_config(r);
 
-out:
-	if (hid)
-		free(hid);
-	if (hnqn)
-		free(hnqn);
-	nvme_free_tree(r);
 	return -errno;
```
calebsander, line 1327: This could probably be `return 0` now

calebsander (approved):
Looks good, just a few comments. Thank you for making `nvme disconnect-all` work when the module is unloaded!

fabrics.c
martin-belanger: > Because whatever device(s) the user is requesting to disconnect definitely don't exist.

One could argue that if a user tries to disconnect a device that doesn't exist, then the operation is a success. In the end, the user is really trying to remove the device. The end result is achieved whether the device existed or not when the disconnect was issued.

calebsander: Fair. I am fine with it this way. It's certainly simpler to handle all the disconnect commands the same way.

fabrics.c
```diff
@@ -1106,6 +1106,14 @@ int nvmf_disconnect(const char *desc, int argc, char **argv)
 	nvme_root_skip_namespaces(r);
 	ret = nvme_scan_topology(r, NULL, NULL);
 	if (ret < 0) {
+		/*
+		 * Do not report an error when the modules are not
+		 * loaded, this allows the user to uncodinitally call
```
martin-gpy, line 1111: Nit: Typo. This should read as ```unconditionally```.

fabrics.c
```diff
@@ -1174,9 +1182,16 @@ int nvmf_disconnect_all(const char *desc, int argc, char **argv)
 	nvme_root_skip_namespaces(r);
 	ret = nvme_scan_topology(r, NULL, NULL);
 	if (ret < 0) {
-		if (errno != ENOENT)
-			fprintf(stderr, "Failed to scan topology: %s\n",
-				nvme_strerror(errno));
+		/*
+		 * Do not report an error when the modules are not
+		 * loaded, this allows the user to uncodinitally call
```
martin-gpy, line 1187: Nit: Typo. This should read as ```unconditionally```.

nvme.c
martin-gpy: Really? I can understand making an exception for ```disconnect-all```, but do we want this for  ```list``` and ```list-subsys``` too? I'd prefer not...

fabrics.c
martin-gpy: I think that's because ```nvmf_hostnqn_from_file()``` & ```nvmf_hostid_from_file()``` allocates memory while fetching the respective hostnqn & hostid strings. And while exiting this current function, one would ideally want to deallocate this memory but retain these string values at the same time, and that is accomplished by using these redundant variables ```hnqn``` & ```hid``` respectively.

nvme.c
calebsander: Yes, that's why I suggested doing this in the implementation of `nvme_scan_topology()` so we wouldn't have to modify all the call sites. If the module isn't loaded, then the topology should be empty.
But as I said, I don't feel strongly about `nvme list`/`list-subsys`. I think it would be more consistent to print an empty list, but we're not relying on it in our workflow.

fabrics.c
calebsander: I'm not following. `hid` and `hnqn` aren't used anywhere and their values match `hostid` and `hostnqn`. So why not remove `hid` and `hnqn` and instead mark `hostid` and `hostnqn` as `_cleanup_free_`?

fabrics.c
calebsander: I'm not following. `hid` and `hnqn` aren't used anywhere and their values match `hostid` and `hostnqn`. So why not remove `hid` and `hnqn` and instead mark `hostid` and `hostnqn` as `_cleanup_free_`?

fabrics.c
igaw: This was also my thinking. In the end, the user wanted that the device is disconnected and it is.

nvme.c
igaw: nvme-cli is not the only user of the library. I generally try to avoid filtering out errors in the middle of a call stack unless we can resolve them. But in this case some users want to see the error others not. See this very discussion.

fabrics.c
igaw: `hostid` and `hostnqn` are origin from `argc`, while `nvme_hostid_from_file` and `nvme_hostnqn_from_file` allocated from the heap. We need to free the heap allocation, thus we have to track it via additional variables. Not nice, but unless we start to allocate all arguments on the heap we have to distinguish between the type of memory we operate on.

igaw:
Thanks!

fabrics.c
calebsander: > `hostid` and `hostnqn` are origin from `argc`

Ah, I missed the `NVMF_ARGS()` macro magic. Thanks.

PR #2427, opened 2024-07-17
nvme-print: Added print for two new fields for HMB feature

nitin-sao:
As per TP4104, added print for two new fields for HMB feature HMNARE (Host Memory Non-operational Access Restriction Enable) and HMNAR (Host Memory Non-operational Access Restricted) with 1 bit each. Bit 1 is not used for Get Feature command and will be 0. And added print for HMBR field in stdout_id_ctrl_ctratt.

Reviewed-by: Mohit Kapoor <mohit.kap@samsung.com>


igaw:
Thanks!

PR #2429, opened 2024-07-18
nvme: use proper mask to get correct lbafu value

francispravin5:
The NVME_NS_FLBAS_HIGHER/LOWER_MASK is used to get the format index value from Formatted LBA Size (FLBAS) field of Identify Namespace Data Structure. But, We are not getting the proper lbafu value while using the same macro on user passed format index value. So, use the proper mask to get the correct lbafu value.


igaw:
Thanks!

PR #2430, opened 2024-07-18
nvme: use argconfig_parse_seen to check the parameter

LasunaCn:
The value of NVME_FEAT_FID_TIMESTAMP cannot be set to 0 through 'set-feature -f 0xe -v 0'
Using argconfig_parse_seen() can fix this issue

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -6415,7 +6415,8 @@ static int set_feature(int argc, char **argv, struct command *cmd, struct plugin
 		 * should use the buffer method if the value exceeds this
 		 * length.
 		 */
-		if (cfg.feature_id == NVME_FEAT_FID_TIMESTAMP && cfg.value) {
+		if (cfg.feature_id == NVME_FEAT_FID_TIMESTAMP &&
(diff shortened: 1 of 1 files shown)


LasunaCn:
When the command "set-feature /dev/nvmeX -f 0xe -v 0" is used, the command should be executed immediately, and the value of FID_TIMESTAMP should be set to 0. 
The current situation, is that the command waits to read data from stdin, and then, sets the read data to FID_TIMESTAMP

igaw:
The line is a bit long. Could you move condition to the next line and align it with `(`.

```
WARNING: line length of 103 exceeds 100 columns
#22: FILE: nvme.c:6418:
+ if (cfg.feature_id == NVME_FEAT_FID_TIMESTAMP && argconfig_parse_seen(opts, "value")) {
```

LasunaCn:
the parameter check code has been moved  to the next line

LasunaCn:
modified the commit description

igaw:
Thanks!

PR #2431, opened 2024-07-18
nvme: return ret error value to open device if errno not set

ikegami-t:
The APIs may not set errno set errno to 0 before the device open.


igaw:
Yes, that's way better.

Thanks!

PR #2432, opened 2024-07-19
wdc: OCP 2.5 Log Page Updates

jeff-lien-wdc:
Add panic count and prev panic id to Error Recovery log page (0xC1)
Add debug telemetry log size to Latency Monitor log page (0xc3)


igaw:
Sorry for the delay. Currently working on an side project :)

PR #2438, opened 2024-07-25
nvme-print: add feature fields struct definition to print feild values

ikegami-t:
Add arbitration, power management, LBA range type and temperature threshold features definitions at first so still remained features.

### DIFF
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -4410,8 +4410,10 @@ static void stdout_directive_show(__u8 type, __u8 oper, __u16 spec, __u32 nsid, 
 
 static void stdout_lba_status_info(__u32 result)
 {
-	printf("\tLBA Status Information Poll Interval (LSIPI)  : %u\n", (result >> 16) & 0xffff);
-	printf("\tLBA Status Information Report Interval (LSIRI): %u\n", result & 0xffff);
+	printf("\tLBA Status Information Poll Interval (LSIPI)  : %u\n",
+	       NVME_FEAT_LBAS_LSIPI(result));
+	printf("\tLBA Status Information Report Interval (LSIRI): %u\n",
+	       NVME_FEAT_LBAS_LSIRI(result));
 }
 
 void stdout_d(unsigned char *buf, int len, int width, int group)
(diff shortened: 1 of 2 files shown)


nvme-print.h
```diff
@@ -129,6 +129,33 @@ struct nvme_bar_cap {
...
+		struct {
+			__u32 num:6;
+			__u32 rsvd6:26;
+		} lba_range_type;
+		struct {
+			__u32 tmpth:16;
+			__u32 tmpsel:4;
+			__u32 thsel:2;
+			__u32 rsvd22:10;
+		} temp_thresh;
```
igaw, line 155: Bitfields should be avoided, as they are not endian safe. There are more reason for it but the big one is the endianess problem . 

It's better to work with shift and mask operations.

ikegami-t: Sorry probably this was mentioned before also. Before to fix the patch as mentioned to make sure let me check the endian problem behavior detail later.

ikegami-t: Still I could not prepare the big endian environment to check the endian problem behavior detail. I seems that the problem can be avoided to define the reserved data correctly as for example rsvd3, rsvd8, rsvd6 and rsvd22. Is this understanding not correct?

igaw: For bitfielts it's necessary to define a big-endian and little-endian version (not to mention there is also another endianess but that's irrelevant these days 'middle-endian'). Anyway, just use SHIFT operators and this all goes away, because SHIFT operators a well defined.

ikegami-t: Thanks for your explanatoin. Just fixed the patch to use the shift operatiors.

nvme-print-stdout.c
```diff
@@ -4474,92 +4476,99 @@ static void stdout_feature_show_fields(enum nvme_features_id fid,
...
+		printf("\tHigh Priority Weight   (HPW): %u\n", NVME_FEAT_ARB_HPW(result) + 1);
+		printf("\tMedium Priority Weight (MPW): %u\n", NVME_FEAT_ARB_MPW(result) + 1);
+		printf("\tLow Priority Weight    (LPW): %u\n", NVME_FEAT_ARB_LPW(result) + 1);
 		printf("\tArbitration Burst       (AB): ");
-		if ((result & 0x00000007) == 7)
+		if (NVME_FEAT_ARB_BURST(result) == NVME_FEAT_ARBITRATION_BURST_MASK)
 			printf("No limit\n");
 		else
-			printf("%u\n",  1 << (result & 0x00000007));
+			printf("%u\n", NVME_FEAT_ARB_BURST(result));
```
igaw, line 4486: This is changing the output value. No idea if this is correct or not.

igaw:
The rest looks good. I just checked what we have in libnvme. Hopefully those definitions are correct :)

nvme-print-stdout.c
ikegami-t: Sorry for this error then just fixed the changes. Thank you.

igaw:
Thanks!

PR #2440, opened 2024-07-27
nvme-print-stdout: fix persistent-event-log set feature event output

xyzhang-up:
Refer to FDP Events Set Feature Data Structure:
the number of FDP Event Types is only Dword11 bits [23:16].
the buffer only contains FDP Event Type.

The value of mem_buf is incorrect:
the pointer 'set_feat_event' should be converted to 'unsigned char *' first


igaw:
Thanks!

PR #2442, opened 2024-07-30
plugin/ocp: Resolved the Segmentation fault issue while printing tele…

VigneshwaranSaravana:
plugin/ocp: Resolved the Segmentation fault issue while printing telemetry string log page(C9h)

Reviewed-by: Karthik Balan <karthik.b82@samsung.com>, 
Reviewed-by: Arunpandian J <arun.j@samsung.com>

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -902,6 +902,7 @@ static int eol_plp_failure_mode(int argc, char **argv, struct command *cmd,
 /// Telemetry Log
 //global buffers
 static __le64 total_log_page_sz;
+static __u8 *header_data;
 static struct telemetry_str_log_format *log_data;
 
 __u8 *ptelemetry_buffer;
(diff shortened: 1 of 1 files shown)


igaw:
Your commit message is not following the style rules. Check the kernel documentation on this subject:

https://www.kernel.org/doc/html/v6.10/process/submitting-patches.html#the-canonical-patch-formatt

Especially the  `Signed-off-by` is wrong.

VigneshwaranSaravana:
@igaw - Resolved the commit message issue

plugins/ocp/ocp-nvme.c
```diff
@@ -902,6 +902,7 @@ static int eol_plp_failure_mode(int argc, char **argv, struct command *cmd,
 /// Telemetry Log
 //global buffers
 static __le64 total_log_page_sz;
+__u8 *header_data;
```
igaw, line 905: You should use static here to avoid leaking into the global namespace.

But I have to say these globals are just wrong IMO. The getter functions should return directly and not via globals. 

The fix itself is correct but I really don't like where this goes with these globals. But I don't maintain this so I don't care too much.

igaw: I've updated it myself.

igaw:
Thanks!

PR #2452, opened 2024-08-06
nvme-print-stdout: Added print for new field CSER (TP4167)

nitin-sao:
As per TP4167, added print for a new field CSER (Command Submission
and Execution Relaxations) [15:14] added in Commands Supported and
Effects Data Structure.

Signed-off-by: Nitin Sao <nitin.sao@samsung.com>
Reviewed-by: Mohit Kapoor <mohit.kap@samsung.com>


igaw:
Thanks!

PR #2453, opened 2024-08-07
plugins/dapustor: dapustor smart-log-add update

xyzhang-up:
Extended Additional SMART log is optional.
If the collection fails, the command dose not stop.

Add the manual of dapustor plugin.
*.1 and *.html can be generated automatically,
so I only added the nvme-dapustor-smart-log-add.txt file.

Update tab completion.


igaw:
Thanks! 

BTW, there is also a zsh completion file... just saying :)

xyzhang-up:
I missed it.
Thanks for the reminder.

PR #2455, opened 2024-08-08
completions: add the zsh completion of the dapustor plugin

xyzhang-up:
add the zsh completion of the dapustor plugin


igaw:
Thanks a lot!

PR #2457, opened 2024-08-09
nvme-print: use Completion Condition enum

francispravin5:
Use the Completion Condition enum for LBA Status Descriptor.
Also, modified the print statement of Completion Condition.


igaw:
Thanks!

Issue #2456, opened 2024-08-09
nvme smart-log does not show temps in Celsius


Strykar:
Temps in Fahrenheit and Kelvin but not Celsius?
```
$ sudo nvme smart-log /dev/nvme0
Smart Log for NVME device:nvme0 namespace-id:ffffffff
critical_warning			: 0
temperature				: 91 °F (306 K)
available_spare				: 100%
available_spare_threshold		: 5%
percentage_used				: 1%
endurance group critical warning summary: 0
Data Units Read				: 42,210,392 (21.61 TB)
Data Units Written			: 20,246,705 (10.37 TB)
host_read_commands			: 2,247,662,721
host_write_commands			: 1,572,136,596
controller_busy_time			: 150
power_cycles				: 111
power_on_hours				: 5,221
unsafe_shutdowns			: 78
media_errors				: 0
num_err_log_entries			: 1,124
Warning Temperature Time		: 0
Critical Composite Temperature Time	: 0
Thermal Management T1 Trans Count	: 0
Thermal Management T2 Trans Count	: 0
Thermal Management T1 Total Time	: 0
Thermal Management T2 Total Time	: 0

$ sudo nvme version 
nvme version 2.10.2 (git 2.10.2)
libnvme version 1.10 (git 1.10)

$ uname -a
Linux r912 6.10.3-arch1-2 #1 SMP PREEMPT_DYNAMIC Tue, 06 Aug 2024 07:21:19 +0000 x86_64 GNU/Linux
```

igaw:
The unit type is chosen depending of your locale settings, so if you use `US` locales you get Fahrenheit.

Strykar:
> The unit type is chosen depending of your locale settings, so if you use `US` locales you get Fahrenheit.

Thanks for clearing that up, setting `LC_MEASUREMENT=metric` in `/etc/locale.conf` works along side `LANG=en_US.UTF-8`,

igaw:
Oh, we might also need to extend our check, though. I don't think we are actually looking at `LC_MEASURMENT` instead just on the country. This is wrong I think.

wdoekes:
@Strykar : I'm not so sure `LC_MEASUREMENT` takes a `metric` argument. But it does take `en_GB.UTF-8` for instance.

And the first hit on google for `LC_MEASUREMENT=metric` is the [Arch Wiki](https://wiki.archlinux.org/title/Locale#toc-Metric_measurements_with_US_locale) which references this post!

[...]

PR #2458, opened 2024-08-11
nvme-print: use LC_MEASUREMENT to check fahrenheit temperature

ikegami-t:
Previously used LC_ALL so not checked correctly if LC_MEASUREMENT set.


ikegami-t:
Fix for the issue #2456.

igaw:
Thanks for following up on this one!

PR #2459, opened 2024-08-12
wdc: Add Support for new SN-861 PCI device id

jeff-lien-wdc:
Support for a new form factor (EDSFF E3.S) of the SN-861
drive was added to the wdc plugin code.


igaw:
Thanks!

PR #2462, opened 2024-08-14
innogrit get-eventlog method change for new project support

innogritkenny:


igaw:
I've applied the commit directly on master after updating the commit message slightly. Thanks!

igaw:
I suppose you also need to update the tab completion (bash/zsh) and the documentation that the command `smart-log-add` is not available anymore.

PR #2463, opened 2024-08-16
nvme-print: print the new fields added in TP4142

francispravin5:
Print the new fields added in Identify Controller Data Structure.
Also, print the field added in Temperature Threshold feature.


igaw:
Thanks!

PR #2473, opened 2024-08-21
fixing CNS 11h to generate identify per allocated namespace.

HaroPanosyan:
Fixing generation of CNS 11h identify namespace data structure generation for each allocated namespace instead of attached namespace.

### DIFF
--- a/plugins/solidigm/solidigm-internal-logs.c
+++ b/plugins/solidigm/solidigm-internal-logs.c
@@ -4,6 +4,7 @@
  *
  * Authors: leonardo.da.cunha@solidigm.com
  * shankaralingegowda.singonahalli@solidigm.com
+ * haro.panosyan@solidigm.com
  */
 
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


HaroPanosyan:
> Fixing generation of CNS 11h identify namespace data structure generation for each allocated namespace instead of attached namespace.

Please ignore this mess. It would be nice if I can remove this Pull request and then will create new one clean.

igaw:
Deleting pull request is not so easy. Only the plattform admins can do this. But it's not that bad you can fix your branch locally with all sort of git operation (rebasing, squashing etc) and just force push it to this branch. This will fix it.

HaroPanosyan:
> Deleting pull request is not so easy. Only the plattform admins can do this. But it's not that bad you can fix your branch locally with all sort of git operation (rebasing, squashing etc) and just force push it to this branch. This will fix it.

Thanks, will do, I have a good help here.

HaroPanosyan:
Cleaned up the mess with 3 commits. Now only one clean commit to be reviewed.

igaw:
Looks good. Just your Signed-off-by tag is missing. I can add it myself if you are okay with it.

HaroPanosyan:
> Looks good. Just your Signed-off-by tag is missing. I can add it myself if you are okay with it.

Thanks for your help... I rebased and added Signed-off-by tag.

igaw:
Thanks!

PR #2475, opened 2024-08-23
nvme-print: added new fields for nvm_id_ns

ankitvvsoni:
Added new fields from I/O Command Set Specific Identify Namespace Data Structure CNS5 as per NVM Command Set Specification(Rev 1.1).


igaw:
Thanks!

PR #2476, opened 2024-08-23
case NVME_CSI_KV implemented

laiaferrer:
Before this commit it used to print Unknown when executing ./nvme effects-log <device> --csi = 1. After this commit we are able to see as a tittle KV Command Set Log Page.


igaw:
I've merged your other PR: https://github.com/linux-nvme/nvme-cli/commit/6dd1c5817287ce0b9579e51a579fce886a1cba8d

PR #2480, opened 2024-08-26
NVME_CSI_KV implemented

laiaferrer:
type: KV Command Set Log Page

### DIFF
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -3872,6 +3872,10 @@ static void stdout_effects_log_page(enum nvme_csi csi,
 		printf("NVM Command Set Log Page\n");
 		printf("%-.80s\n", dash);
 		break;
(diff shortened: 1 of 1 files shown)


nvme-print-stdout.c
```diff
@@ -3873,7 +3873,7 @@ static void stdout_effects_log_page(enum nvme_csi csi,
 		break;
 	default:
 		printf("Unknown Command Set Log Page\n");
-		printf("%-.80s\n", dash);
+		printf("%-.80s\n", dash);
```
igaw, line 3876: this doesn't change anything, it just  adds an empty space at the end of the line.

igaw (changes requested):
This should be one commit with a proper commit message.

igaw:
I've merged the three commits into one and updated the commit message.

Thanks for your contribution!

PR #2481, opened 2024-08-28
fabrics: fix incorrect access filename check

martin-gpy:
Config JSON file handling is currently broken due to an incorrect access filename check. Fix it.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -669,8 +669,8 @@ static int nvme_read_volatile_config(nvme_root_t r)
 
 static int nvme_read_config_checked(nvme_root_t r, const char *filename)
 {
-	if (!access(filename, F_OK))
-		return -ENOENT;
+	if (access(filename, F_OK))
+		return -errno;
 	if (nvme_read_config(r, filename))
 		return -errno;
 	return 0;


igaw:
Can you elaborate what you are doing. I am pretty sure I added this on purpose during my testing. So what did I miss?

martin-gpy:
> Can you elaborate what you are doing. I am pretty sure I added this on purpose during my testing. So what did I miss?

Well, I was just passing the JSON config file using the ```-J``` option to the ```nvme connect-all``` command, which ended up failing:

```nvme connect-all -J <filename>```

Turns out ```access``` returns a zero if the filename exists & non-zero if otherwise. So this current check actually returns an ```ENOENT``` if the JSON file exists:

```if (!access(filename, F_OK)) return -ENOENT;```

So all one needs to do is switch the check to make this work, and that's what this patch does:

```if (access(filename, F_OK)) return -ENOENT;```

igaw:
Ah, I think I figured out why I didn't see any problem during my testing. I only tested with `connect` which doesn't use this function for checking if there is config. So this one makes sense. 

We might also return `errno`; I'll update the patch accordingly.

PR #2484, opened 2024-08-30
Ocp hwcomp

ikegami-t:

PR #2485, opened 2024-09-01
util: extern uint128_t_to_double() function

ikegami-t:
Use the function to calculate with the uint128_t value.


ikegami-t:
Cherry picked from the PR #2484 since not depended on the changes.

igaw:
Thanks!

PR #2486, opened 2024-09-01
logging: add print_info function

ikegami-t:
This is for debugging print.


ikegami-t:
Cherry picked this also from the PR #2484 to reduce the PR commits.

igaw:
Thanks!

PR #2487, opened 2024-09-01
nvme-print-json: extern json object add functions

ikegami-t:
Some additinal functions introduced and fix obj_add_uint_0nx error.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -25,7 +25,6 @@
 #define obj_add_array json_object_add_value_array
 #define obj_add_int json_object_add_value_int
 #define obj_add_obj json_object_add_value_object
-#define obj_add_str json_object_add_value_string
 #define obj_add_uint json_object_add_value_uint
 #define obj_add_uint128 json_object_add_value_uint128
 #define obj_add_uint64 json_object_add_value_uint64
@@ -75,7 +74,7 @@ static void obj_add_uint_x(struct json_object *o, const char *k, __u32 v)
 	obj_add_str(o, k, str);
 }
 
-static void obj_add_uint_0x(struct json_object *o, const char *k, __u32 v)
+void obj_add_uint_0x(struct json_object *o, const char *k, __u32 v)
 {
 	char str[STR_LEN];
 
@@ -83,15 +82,15 @@ static void obj_add_uint_0x(struct json_object *o, const char *k, __u32 v)
 	obj_add_str(o, k, str);
 }
 
-static void obj_add_uint_0nx(struct json_object *o, const char *k, __u32 v, int width)
+void obj_add_uint_0nx(struct json_object *o, const char *k, __u32 v, int width)
 {
 	char str[STR_LEN];
 
-	sprintf(str, "0x%02x", v);
+	sprintf(str, "0x%0*x", width, v);
 	obj_add_str(o, k, str);
 }
 
-static void obj_add_uint_02x(struct json_object *o, const char *k, __u32 v)
+void obj_add_uint_02x(struct json_object *o, const char *k, __u32 v)
 {
 	obj_add_uint_0nx(o, k, v, 2);
 }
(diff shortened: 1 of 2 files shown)


ikegami-t:
Separated the commit from the PR #2484.

nvme-print-json.c
```diff
@@ -4787,3 +4786,34 @@ struct print_ops *nvme_get_json_print_ops(nvme_print_flags_t flags)
...
+	value = calloc(1, (len + 1) * 2 + 1);
+
+	if (!value) {
+		obj_add_str(o, k, "Could not allocate string");
+		return;
+	}
+
+	sprintf(value, "0x");
+	for (i = 0; i < len; i++)
+		sprintf(&value[strlen(value)], "%02x", buf[len - i - 1]);
```
calebsander, line 4808: This is a quadratic algorithm for what should be linear. Can you just keep track of the current length in a local variable?

Reversing the bytes also seems a little weird, but I guess I don't know what the intended use case is.

ikegami-t: > This is a quadratic algorithm for what should be linear. Can you just keep track of the current length in a local variable?

Just fix this as mentioned.

> Reversing the bytes also seems a little weird, but I guess I don't know what the intended use case is.

Unfortunately I do not any idea to resovle this. This outputs the little endian bytes endian as a big endian string as below for example.
```c
buf[0] = 0xaa;
buf[1] = 0xbb;
buf[2] = 0xcc;
buf[3] = 0xdd;
value = "0xbbccbbaa"
```
Thank you for your reviewing.

igaw: Not really looked into it, so just a stupid comment from my side on the output order. Shouldn't this be output how the spec defines it? IIRC everything as little endian? (if this statement is untrue, I blame my vacation...)

calebsander: Agree with @igaw. It's hard to say without knowing how this function is going to be used. (I would expect the functions to be added in the same patch set as users of the functions.) But for generic "print bytes" functionality, it seems like it would make sense to print the bytes in the order they appear. If the function is interpreting the bytes as an integer, it might make more sense to have "unit" in the function name.

nvme-print-json.c
```diff
@@ -4787,3 +4786,34 @@ struct print_ops *nvme_get_json_print_ops(nvme_print_flags_t flags)
...
 	return &json_print_ops;
 }
+
+void obj_add_byte_array(struct json_object *o, const char *k, unsigned char *buf, int len)
+{
+	int i;
+
+	_cleanup_free_ char *value = NULL;
+
+	if (!o || !k || !buf || !len)
```
calebsander, line 4796: Is it necessary to handle all these cases? I don't see the other JSON value add functions checking their arguments like this. Seems like this would be papering over a bug in the caller...

nvme-print-json.c
```diff
@@ -4787,3 +4786,34 @@ struct print_ops *nvme_get_json_print_ops(nvme_print_flags_t flags)
...
+		return;
+
+	value = calloc(1, (len + 1) * 2 + 1);
+
+	if (!value) {
+		obj_add_str(o, k, "Could not allocate string");
+		return;
+	}
+
+	sprintf(value, "0x");
```
calebsander, line 4806: Isn't this being overwritten by the hex characters of the first byte?

nvme-print-json.c
ikegami-t: Hi,
For example the existing fw-activate-history command outputs the loge page GUID value as below as a big endian string.
```sh
tokunori@tokunori-desktop:~/nvme-cli$ nvme-build ocp fw-activate-history /dev/nvme0
Firmware History Log:
  log identifier:           194
  valid entries:            0
  entries:
  log page version:         1
  log page guid:            0xd11cf3ac8ab24de2a3f6dab4769a796d
```
Also the OCP specification is also describes same as below as D11CF3AC8AB24DE2A3F6DAB4769A796Dh.
Screenshot from 2024-09-09 23-13-59
So seems the ocp commands just follow the specification descritions.

nvme-print-json.c
ikegami-t: Which do you suggest?
1. Ignore the invalid value with the checking then the function caused a null pointer exception or any unexpected error. (I do not think this function behavior correct.)
2. Change the void function to return any error by the checking?

nvme-print-json.c
ikegami-t: Right. Sorry the fix was not correct. Thanks for your suggestion.

ikegami-t: Fixed the overwritten error.

igaw:
Thanks!

PR #2488, opened 2024-09-01
nvme-print-stdout: add print_array function

ikegami-t:
To reduce the repeated print code for arrays.


ikegami-t:
Separated from the PR #2484 and moved into nvme-print-stdout from nvme-ocp.

igaw:
Thanks!

PR #2489, opened 2024-09-01
nvme: update nvme_insert_tls_key_versioned() return handling

martin-gpy:
Treat nvme_insert_tls_key_versioned() return value of zero as errors too.


igaw:
Thanks!

PR #2490, opened 2024-09-01
ocp: fix to set return value to get c9 log page data

ikegami-t:
The value not set but checked the ret variable after the call.


igaw:
Thanks!

PR #2491, opened 2024-09-02
plugins/solidigm: fix use after free.

maurizio-lombardi:
the "output" file descriptor is automatically closed when the function returns; therefore we shouldn't explicitely call close() against it.


igaw:
Thanks!

PR #2503, opened 2024-09-17
Ocp wo json

ikegami-t:


igaw:
I was wondering if we should move the json code into a separate files and only include them when the json library is available, like we have with `nvme-print-stdout.c` and `nvme-print-json.c`.

ikegami-t:
Okay will try to do it. Thank you.

plugins/ocp/meson.build
```diff
@@ -6,5 +6,9 @@ sources += [
   'plugins/ocp/ocp-fw-activation-history.c',
   'plugins/ocp/ocp-telemetry-decode.c',
   'plugins/ocp/ocp-hardware-component-log.c',
+  'plugins/ocp/ocp-print.c',
+  'plugins/ocp/ocp-print-stdout.c',
+  'plugins/ocp/ocp-print-json.c',
```
igaw, line 11: The json file should only be added when the json library is available. And then you could update the `plugins/meson.build` file to always build the ocp plugin.

The rest looks good!

ikegami-t: Update the meson.build files by the additional commit. Thank you.
  Note: Still some exisitng ocp commands files needed to split json code.

igaw: So do you want me to wait with merging this and wait until you finished the json code splitting?

ikegami-t: No need to wait for me. Still working on this but I am okay to do it by additional PRs. Thank you.

ikegami-t:
All OCP command json codes execpt the ocp-telemetry-decode functions are split into the ocp-print-json.c.
Note: The ocp-telemetry-decode functions used by the internal command are mixed stdout and json outputs implementation so not moved into the ocp-print-json.c but using the CONFIG_JSONC to disable intead.

ikegami-t:
Just rebased and fixed the conflicts.

igaw:
I've noticed that the stdout plugin does

```c
static struct ocp_print_ops stdout_print_ops = {
	.hwcomp_log = stdout_hwcomp_log,
	.fw_act_history = stdout_fw_activation_history,
	.smart_extended_log = stdout_smart_extended_log,
	.telemetry_log = stdout_telemetry_log,
	.c3_log = (void *)stdout_c3_log,
	.c5_log = (void *)stdout_c5_log,
	.c1_log = stdout_c1_log,
	.c4_log = stdout_c4_log,
	.c9_log = (void *)stdout_c9_log,
	.c7_log = (void *)stdout_c7_log,
};
```

I think the cast should be removed by updating the function or updating the callback function signature. Either way, cast should go away eventually. Though no reason not go apply it now. Thanks for the great work! Really appreciated!

ikegami-t:
Thank you so much for your reviewing and comments. Just fixed for the comment by the PR #2541. Thank you.

PR #2505, opened 2024-09-19
plugins/solidigm: Added Workload Tracker Triggers and Wall Time

lgdacunh:
Added Trigger configurations, Stopped converting timestamps to millisec, added optional property to display wall clock column. Fixed zsh misplaced auto-complete of ocp set-telemetry-profile.


igaw:
I've splitted the patch and commit it directly. Thanks!

PR #2509, opened 2024-09-21
nvme-print: fix error information log page endianness error

ikegami-t:
Convert the status and phase tag fields to native byte order at first.


igaw:
Good catch! Thanks!

PR #2510, opened 2024-09-22
nvme: fix to convert metadata size to native byte order

ikegami-t:
The changes for create-ns, compare, read and write commands.


igaw:
Thanks!

PR #2511, opened 2024-09-22
nvme: fix uninitialized value in error-log

martin-gpy:
Valgrind complained about an "uninitialized value created by a stack allocation" error while running the error-log command. Fix the same.


igaw:
Thanks!

PR #2513, opened 2024-09-25
nvme-print-json: print CMBEBS and CMBSWTP in json format

francispravin5:
1. Print the CMBEBS and CMBSWTP register in json format
2. Print the Controller Register values in order based on the offset.


igaw:
Thanks!

PR #2515, opened 2024-09-27
ocp: fix GUID output

Ted1642:
Some OCP commands are missing a high-order zero in each byte of the output GUID.

This commit fixes the format specifier for the printf function.


ikegami-t:
Looks good. (If you like we can use the print function print_array() instead.)

igaw:
Thanks!

PR #2516, opened 2024-09-30
nvme-print: print NSSES field of CAP register

francispravin5:
Print the newly added NSSES field of CAP register.


igaw:
Thanks!

PR #2518, opened 2024-10-01
ocp: fix option handling in internal-log

Ted1642:
When the noraml format is specified without -t option to ocp internal-log command, the command ends abnormally with segmentation fault.
This commit fixes the option handling.


igaw:
Thanks!

PR #2521, opened 2024-10-02
nvme: Support show-regs for nvmeof

hmi-jeon:
Get properties only at offsets that support nvme fabric registers.
Since crto was not found in the nvmeof spec, it is removed.
Also, NSSR is only optional register. If the device is not supported by the optional register, the error is ignored.

Reviewed-by: Steven Seungcheol <sc108.lee@samsung.com>
Signed-off-by: Minsik Jeon <hmi.jeon@samsung.com>

### DIFF
--- a/nvme-print.c
+++ b/nvme-print.c
@@ -406,7 +406,18 @@ bool nvme_is_fabrics_reg(int offset)
 	case NVME_REG_CC:
 	case NVME_REG_CSTS:
 	case NVME_REG_NSSR:
-	case NVME_REG_CRTO:
+		return true;
+	default:
+		break;
+	}
+
+	return false;
+}
+
+bool nvme_is_fabrics_optional_reg(int offset)
+{
+	switch (offset) {
+	case NVME_REG_NSSR:
 		return true;
 	default:
 		break;
--- a/nvme.c
+++ b/nvme.c
@@ -5285,6 +5285,11 @@ static int nvme_get_single_property(int fd, struct get_reg_config *cfg, __u64 *v
 	if (!err)
 		return 0;
 
+	if (cfg->fabrics && nvme_is_fabrics_optional_reg(cfg->offset)) {
+		*value = -1;
+		return 0;
+	}
+
 	if (!cfg->fabrics &&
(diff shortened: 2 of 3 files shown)


nvme.c
```diff
@@ -5380,12 +5388,12 @@ static int show_registers(int argc, char **argv, struct command *cmd, struct plu
 
 	_cleanup_nvme_dev_ struct nvme_dev *dev = NULL;
 	nvme_print_flags_t flags;
-	bool fabrics = false;
 	void *bar;
 	int err;
 
 	struct get_reg_config cfg = {
 		.human_readable	= false,
+		.fabrics = false,
```
igaw, line 5396: This change doesn't seem necessary. `nvme_get_properties` operates only on offsets. Anyway, the whole logic is already a bit complex so I'd rather avoid this type of additional info passed into the getters.

nvme-print.c
```diff
@@ -406,7 +406,18 @@ bool nvme_is_fabrics_reg(int offset)
...
+		break;
+	}
+
+	return false;
+}
+
+bool nvme_is_fabrics_optional_reg(int offset)
+{
+	switch (offset) {
+	case NVME_REG_NSSR:
```
igaw, line 420: Is the `NSSR` the only optional register? If so please update the commit message accordingly

hmi-jeon: Yes. NSSR is optional register. I update the commit message.
Thanks.

nvme.c
hmi-jeon: Yes.
But in the nvme_get_single_property, If it is cfg.fabrics, check optional register.
Is there a better fix?

Thanks.

igaw: I see, we already use it. This is so horrible code. I shouldn't have merged it.

igaw:
Thanks!

We should though rethink this part of the code. Really messy how it currently is.

nvme.c
hmi-jeon: Okay. I understood it. 
As you said, it looks like refactoring is needed.

PR #2523, opened 2024-10-02
wdc: Add Support for SN5100S

jeff-lien-wdc:
This commit will add support for the SN51000S drive which includes the following pci device ids:
0x5061, 0x5062, 0x5063.


igaw:
Thanks!

PR #2524, opened 2024-10-03
nvme-print-json: update JSON verbose output for nvm-id-ctrl

martin-gpy:
Update the nvme-id-ctrl JSON verbose output for parity with this command's normal verbose output.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -3072,6 +3072,15 @@ static void json_nvme_id_ctrl_nvm(struct nvme_id_ctrl_nvm *ctrl_nvm)
 	obj_add_uint64(r, "dmsl", le64_to_cpu(ctrl_nvm->dmsl));
 	obj_add_uint(r, "aocs", le16_to_cpu(ctrl_nvm->aocs));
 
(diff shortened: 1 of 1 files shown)


nvme-print-json.c
```diff
@@ -3072,6 +3072,15 @@ static void json_nvme_id_ctrl_nvm(struct nvme_id_ctrl_nvm *ctrl_nvm)
 	obj_add_uint64(r, "dmsl", le64_to_cpu(ctrl_nvm->dmsl));
 	obj_add_uint(r, "aocs", le16_to_cpu(ctrl_nvm->aocs));
 
+	if (json_print_ops.flags & VERBOSE) {
```
igaw, line 3075: I wonder if we should always print everything. For me the verbose flag makes only sense for the stdout output.

igaw: My point is that the json output could always print everything, no need make it short. Anyway, your change is fine, let's go with it, I ponder a bit if we should just kill the 'verbose' flag in the json output.

igaw:
Thanks!

PR #2525, opened 2024-10-04
doc: show where self-test results can be found

paumr:
As a first-time user I was wondering on where to find the `device-self-test` results.

I think adding this information to the `device-self-test` documentation would be quite useful.

### DIFF
--- a/Documentation/nvme-device-self-test.txt
+++ b/Documentation/nvme-device-self-test.txt
@@ -22,6 +22,8 @@ device (ex: /dev/nvme0), or a namespace block device (ex: /dev/nvme0n1).
 
 On success, the corresponding test is initiated.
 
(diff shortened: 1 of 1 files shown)


igaw:
Looks good. 

Just a nit: Could you add your Signed-off-by to the commit message? Or I can add it if it's okay for you.

Thanks!

paumr:
Thanks! Sure, I signed the commit - thanks for the review.

igaw:
We follow the Linux kernel mailing list workflow, thus commit message itself should be something like

```
doc: show where self-test results can be found

Point out where find results when running the self tests.

Signed-off-by: My Name/alias <email.address.foo>
```

paumr:
Done, sorry for the delay, I did not receive a notification about your most recent message.

igaw:
Thanks!

PR #2526, opened 2024-10-05
plugins/solidigm: Automatic retry smaller telemetry chunk size.

lgdacunh:
Retry to retrieve telemetry with smaller chunk size, because some systems are failing to retrieve telemetry in 256KB chunks.

### DIFF
--- a/plugins/solidigm/solidigm-util.c
+++ b/plugins/solidigm/solidigm-util.c
@@ -37,3 +37,19 @@ int sldgm_get_uuid_index(struct nvme_dev *dev, __u8 *index)
 
 	return sldgm_find_uuid_index(&uuid_list, index);
 }
+
+int sldgm_dynamic_telemetry(int dev_fd, bool create, bool ctrl, bool log_page, __u8 mtds,
+			    enum nvme_telemetry_da da, struct nvme_telemetry_log **log_buffer,
+			    size_t *log_buffer_size)
+{
+	int err;
+	size_t max_data_tx = (1 << mtds) * NVME_LOG_PAGE_PDU_SIZE;
+
+	do {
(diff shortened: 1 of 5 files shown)


plugins/solidigm/solidigm-util.c
```diff
@@ -37,3 +37,19 @@ int sldgm_get_uuid_index(struct nvme_dev *dev, __u8 *index)
 
 	return sldgm_find_uuid_index(&uuid_list, index);
 }
+
+int sldgm_dynamic_telemetry(int dev_fd, bool create, bool ctrl, bool log_page,
+			    size_t max_data_tx, enum nvme_telemetry_da da,
+			    struct nvme_telemetry_log **log_buffer,
+			    size_t *log_buffer_size)
```
igaw, line 44: Shouldn't you also check the max_data_tx argument to be at least 4k?

lgdacunh: max_data_tx is meant to derive from id_ctrl.mdts, and should be at least 4K. I will experiment moving max_data_tx calculation into this function, and see if the rest of the code don't get too messy.

igaw: It's not that important, just wanted to point out, that the caller could use something smaller than 4k. In this case  nvme_get_telemetry_log will fail anyway, IIRC.

lgdacunh: Well, I moved things around and hit another log using large chunks based on  id_ctrl.mdts. I will amend this commit with all the changes.

lgdacunh: Done!

igaw:
Thanks!

PR #2527, opened 2024-10-07
ocp-nvme: remove ocp log page version checking

sc108-lee:
For supporting ocp 2.5 and legacy version
if we want to check the log version,
we need to get DSSD(OCP) Spec version from ocp smart.

example) Latency Monitor (Log Identifier C3h)
OCP 2.0 Log Page version : 0x1
OCP 2.5 Log Page version : 0x4


igaw:
@arthurshau please have a look. Thanks.

igaw:
Thanks!

PR #2528, opened 2024-10-07
wdc: Fix for Reading WDC C2 Vendor Unique Log Page

jeff-lien-wdc:
In cases where nvme drives are connected via a PCIe swtich, the pci vendor and device id are unable to be read.  This causes a failure when reading values in the WDC C2 log page needed for the get-drive-status wdc plugin command.


igaw:
Thanks!

PR #2530, opened 2024-10-09
plugins/dapustor: smart-log-add fix

xyzhang-up:
Output Extended Additional SMART raw binary data if it exists. 
Add the missing 'temp_since_bootup' field for json output.

### DIFF
--- a/plugins/dapustor/dapustor-nvme.c
+++ b/plugins/dapustor/dapustor-nvme.c
@@ -312,16 +312,16 @@ static void show_dapustor_ext_add_smart_log_jsn(struct nvme_extended_additional_
 	json_object_add_value_object(dev_stats, "temp_since_born", entry_stats);
 
 	entry_stats = json_create_object();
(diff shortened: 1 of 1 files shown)


plugins/dapustor/dapustor-nvme.c
```diff
@@ -312,16 +312,16 @@ static void show_dapustor_ext_add_smart_log_jsn(struct nvme_extended_additional_
 	json_object_add_value_object(dev_stats, "temp_since_born", entry_stats);
```
igaw, line 312: I assume you want to keep the `temp_since_born` here and not `temp_since_bootup`?

xyzhang-up: both 'temp_since_born' and 'temp_since_bootup' need to be displayed.
the previous problem was that 'temp_since_born' was duplicated and 'temp_since_bootup' was missing.

igaw:
Thanks!

PR #2531, opened 2024-10-09
nvme-print-stdout: print VERS bit of SANICAP field

francispravin5:
Print the newly added VERS bit of SANICAP field in Identify Controller Data Structure. Also, changed the variable names as per spec.


igaw:
Sorry, I missed to merge this. Thanks!

PR #2529, opened 2024-10-08
ocp-nvme: Add LMDATA-37 for Latency Monitor Log

sc108-lee:
Datacenter NVMe SSD Specification v2.5
Log Identifier C3h
[447:436] Latency Monitor Debug Telemetry Log Size
This is the number of Dwords in the Latency Monitor Debug Log. This value is in Dwords.
little endian order print

Signed-off-by: Hyuntae Kim <h1219.kim@samsung.com>
Signed-off-by: Steven Seungcheol Lee <sc108.lee@samsung.com>

This PR based on PR #2527

PR #2535, opened 2024-10-15
ocp-nvme: ocp plugin version update #2527 #2529

sc108-lee:
Add missing version update after adding changes for ocp spec 2.5


igaw:
Just dropped the github references in the subject. No point in making the git history depended on github. We never know how long a service is around, thus the commit message should contain everything to figure out what's going on (which it does in this case :))

PR #2537, opened 2024-10-16
Minor misc fixes in tls_key()

martin-gpy:


nvme.c
```diff
@@ -9523,10 +9523,20 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
...
+
 		err = nvme_scan_tls_keys(cfg.keyring, __scan_tls_key, fd);
-		if (err)
+		if (err < 0) {
 			nvme_show_error("Export of TLS keys failed with '%s'",
 				nvme_strerror(errno));
+			return err;
+		}
+
+		fprintf(stderr, "exporting to %s\n", cfg.keyfile);
```
igaw, line 9538: Please don't print anything in success per default. The default behavior for Unix tools is no output/no error on success.

nvme.c
```diff
@@ -9538,7 +9538,19 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
...
+		}
+
 		err = import_key(cfg.keyring, fd);
+		if (err) {
+			nvme_show_error("Import of TLS keys failed with '%s'",
+					nvme_strerror(errno));
+			return err;
+		}
+
+		fprintf(stderr, "importing from %s\n", cfg.keyfile);
```
igaw, line 9553: Same here, no output on default. I'm fine with having this with the verbose output enabled.

martin-gpy: Ok, will move this under the verbose flag.

nvme.c
```diff
@@ -9553,9 +9553,13 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
...
 	} else {
 		err = nvme_revoke_tls_key(cfg.keyring, cfg.keytype, cfg.revoke);
-		if (err)
+		if (err) {
 			nvme_show_error("Failed to revoke key '%s'",
 					nvme_strerror(errno));
+			return err;
+		}
+
+		fprintf(stderr, "revoking key\n");
```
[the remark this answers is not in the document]
martin-gpy: Sure, will print this to stdout itself.

nvme.c
```diff
@@ -9523,10 +9523,22 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
 		nvme_show_error("Must specify either --import, --export or --revoke");
 		return -EINVAL;
 	} else if (cfg.export) {
+		if (!cfg.keyfile) {
```
igaw, line 9526: This will prevent to print it to stdout, which is very helpful when debugging. Also `nvme tls --export > file` doesn't work anymore.

nvme.c
```diff
@@ -9540,7 +9540,20 @@ static int tls_key(int argc, char **argv, struct command *command, struct plugin
 
 		return 0;
 	} else if (cfg.import) {
+		if (!cfg.keyfile) {
```
igaw, line 9543: same here, this will prevent to read from stdin.

nvme.c
martin-gpy: Ah ok...I didn't realize those were applicable. I will go ahead and remove this check now. Also, I think it makes sense now to consolidate these minor patches to a single one patch itself.

igaw: Thanks a lot!

igaw:
Thanks!

PR #2538, opened 2024-10-17
nvme-print-json: display only verbose output

martin-gpy:
Having separate regular & verbose outputs makes more sense for the stdout outputs, and not necessarily for the JSON outputs. So let's make things simpler by having the JSON output everything, and nothing short.

Suggested-by: Daniel Wagner <dwagner@suse.de>


igaw:
Thanks!

PR #2539, opened 2024-10-21
doc: added commit conventions to contribution guidelines

paumr:
This PR is a result of https://github.com/linux-nvme/nvme-cli/pull/2525#issuecomment-2394228985

Hopefully this prevents such additional workload for the maintainers in the future.


paumr:
Just for clarification: gpg-signing the commits is a non-requirement right?

igaw:
Excellent docs update! I was about to write something similar! Thanks!

No, gpg  signing is not a requirement. I usually review everything based on what github is giving me.  So the gpg signing is likely not going to help as far I understand. But I am a pretty ignorant when it comes to security.

paumr:
Thanks! Happy to help!

PR #2542, opened 2024-10-24
nvme: add EMVS support to sanitize command

francispravin5:
Add Enter Media Verification State (EMVS) support to sanitize command.
TP4152 - Post-Sanitize Media Verification 2024.04.01 Ratified.


igaw:
Thanks!

PR #2543, opened 2024-10-25
ocp: combine to use GUID length definitions

ikegami-t:
This reduces the duplicated GUID length 16 bytes definitions.


igaw:
Thanks!

PR #2544, opened 2024-10-25
doc: Add sanitize command emvs option

francispravin5:
Add Enter Media Verification State (EVMS) option of Sanitize command.


igaw:
Thanks!

PR #2545, opened 2024-10-27
Add workflow to run tests on real hardware

MaisenbacherD:
This PR introduces a GitHub workflow that runs all test cases under the
`tests` directory on real hardware.
 
In preparation, the existing test cases are fixed, such that the new
`run-tests` workflow completes successfully (see actions on my fork's master
branch).
 
The infrastructure that provides the self-hosted GitHub runner with real nvme
devices attached to it are provided by Western Digital Corporation.
On a high-level overview, this infrastructure contains multiple storage nodes
that form a Kubernetes cluster. Within this Kubernetes cluster, KubeVirt
virtual machines can be spawned on demand with different hardware
configurations. Those VMs are owned by different users. One user group that
is separated in its own namespace is self-hosted GitHub runners.
 
The overall goal of this infrastructure is to provide device access to open
source projects for testing purposes. The first two pioneering candidates to
make use of this service are nvme-cli and ZenFS (see https://github.com/westerndigitalcorporation/zenfs/pull/294).
 
The new self-hosted GitHub runner IaC is to be open-sourced.
 
If people want to see other nvme devices for testing in GitHub workflows, we
are happy to discuss details on how to move forward with integrating new
test devices.
 
**Initial requirements for this self-hosted runner**:
- Share NVMe devices (both NVM and ZNS command sets. Western Digital
  Ultrastar® SN640 and Western Digital Ultrastar® ZN540 are provided)
- Update the VM kernel nightly with Linus' master branch (through cron job)
- Security:
  - Running in a dedicated KubeVirt VM, one per repository.
  - Restricted cluster network access: KubeVirt masquerade interface that only opens
    necessary ports.
  - Privileged code execution within the VM ('gh-runner' user). The VM will be reset
    after each workflow run.
  - Running in the gh-runner namespace with NetworkPolicies where
    ssh is just allowed from local IPs, https is just allowed to and from
    external addresses (no access to internal cluster services), access to VM
    image repository access just from local IPs.
  - /etc/hosts.allow and /etc/hosts.deny rules for the gh-runner instances that
    only allow sshd access from cluster local IPs into the VM.
 
 
**IMPORTANT: Configuration required by the GitHub repo that uses the self-hosted runner**:
- A single self-hosted runner instance (Kubevirt VM) is only allowed to serve
  one repository
- Optionally: Prevent a group of people from allowing actions:
Repo -> Settings -> Actions -> General -> Actions permissions
- Prevent PR's from executing code on the self-hosted runner before getting approved:
Repo -> Settings -> Actions -> General -> 'Require approval for all outside
collaborators' -> Save
- Restrict write access to the repository with the GITHUB_TOKEN:
Repo -> Settings -> Actions -> General -> 'Read repository content and packages
permissions' -> Save
- Preventing GitHub Actions from creating or approving pull requests through the
  GITHUB_TOKEN:
Repo -> Settings -> Actions -> General -> DISABLE 'Allow GitHub Actions to
create and approve pull requests' -> Save
- In reviews watch out for code injection and secret leaks within workflows
  (https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions)
 
 
In a GitHub workflow that uses a container on the self-hosted instance, the
block device has to be passed into the container:
```
...
    container:
      image: ghcr.io/igaw/linux-nvme/debian.python:latest
      options: '--device=/dev/nvme0n1:/dev/nvme0n1'
...
``` 

Before this PR gets merged we should make sure that the self-hosted runner is
added to the repository, which requires a token exchange on a side channel. :)

### DIFF
--- a/.github/workflows/run-nightly-tests.yml
+++ b/.github/workflows/run-nightly-tests.yml
@@ -1,0 +1,93 @@
+---
+name: run-nightly-tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 01 * * *'
+
+jobs:
+  nightly-tests:
+    runs-on: nvme-nvm
+    steps:
+      - name: Output kernel version
+        run: |
+          uname -a
+      - name: Clean up test device
+        run: |
+          #BDEV0 is an environment variable of the self-hosted runner instance
+          #that contains a valid nvme ctrl name which is capable of the nvm
+          #command set.
+          CONTROLLER=$(echo /dev/${BDEV0} | sed 's/n[0-9]*$//')
+          sudo nvme delete-ns $CONTROLLER -n 0xffffffff
+          sudo nvme format $CONTROLLER -n 0xffffffff -l 0 -f
+          SIZE=$(sudo nvme id-ctrl $CONTROLLER --output-format=json | jq -r '{tnvmcap} | .[]' | awk '{print $1/512}')
+          sudo nvme create-ns -s $SIZE -c $SIZE -f 0 -d 0 --csi=0 $CONTROLLER
+          sudo nvme attach-ns $CONTROLLER -n 1 -c 0
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y \
+          meson gcc pkg-config git libjson-c-dev libssl-dev libkeyutils-dev \
+          libdbus-1-dev libpython3-dev pipx python3-dev swig xz-utils
+          pipx ensurepath
+          sudo PIPX_BIN_DIR=/usr/local/bin pipx install nose2
+          sudo PIPX_BIN_DIR=/usr/local/bin pipx install flake8
+          sudo PIPX_BIN_DIR=/usr/local/bin pipx install mypy
+          sudo PIPX_BIN_DIR=/usr/local/bin pipx install autopep8
+          sudo PIPX_BIN_DIR=/usr/local/bin pipx install isort
+      - name: Build and install nvme-cli
+        run: |
+          scripts/build.sh -b release -c gcc
+          sudo meson install -C .build-ci
+          sudo ldconfig /usr/local/lib64
+      - name: Overwrite test config
+        run: |
+          CONTROLLER=$(echo /dev/${BDEV0} | sed 's/n[0-9]*$//')
+          cat > tests/config.json << EOF
+          {
+            "controller" : "$CONTROLLER",
+            "ns1": "/dev/${BDEV0}",
+            "log_dir": "tests/nvmetests/"
+          }
+          EOF
+      - name: Run on device tests
+        run: |
+          sudo nose2 --verbose --start-dir tests \
+          nvme_attach_detach_ns_test \
+          nvme_compare_test \
+          nvme_copy_test \
+          nvme_create_max_ns_test \
+          nvme_ctrl_reset_test \
+          nvme_dsm_test \
+          nvme_error_log_test \
+          nvme_flush_test \
+          nvme_format_test \
+          nvme_fw_log_test \
+          nvme_get_features_test \
+          nvme_get_lba_status_test \
+          nvme_id_ctrl_test \
+          nvme_id_ns_test \
+          nvme_lba_status_log_test \
+          nvme_read_write_test \
+          nvme_smart_log_test \
+          nvme_verify_test \
+          nvme_writeuncor_test \
+          nvme_writezeros_test
+      - name: Upload logs
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: logs files
+          path: |
+            ./tests/nvmetests/**/*.log
+      - name: Clean up test device
+        if: always()
+        run: |
+          CONTROLLER=$(echo /dev/${BDEV0} | sed 's/n[0-9]*$//')
+          sudo nvme delete-ns $CONTROLLER -n 0xffffffff
+          sudo nvme format $CONTROLLER -n 0xffffffff -l 0 -f
+          SIZE=$(sudo nvme id-ctrl $CONTROLLER --output-format=json | jq -r '{tnvmcap} | .[]' | awk '{print $1/512}')
+          sudo nvme create-ns -s $SIZE -c $SIZE -f 0 -d 0 --csi=0 $CONTROLLER
+          sudo nvme attach-ns $CONTROLLER -n 1 -c 0
--- a/tests/nvme_format_test.py
+++ b/tests/nvme_format_test.py
@@ -37,6 +37,7 @@ Namespace Format testcase :-
            - Delete Namespace.
 """
 
+import math
 import subprocess
 import time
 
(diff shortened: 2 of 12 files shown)


tests/nvme_test.py
```diff
@@ -160,8 +160,10 @@ def get_ctrl_id(self):
                                 encoding='utf-8')
         err = proc.wait()
         self.assertEqual(err, 0, "ERROR : nvme list-ctrl failed")
-        line = proc.stdout.readline()
-        ctrl_id = line.split(":")[1].strip()
+        lines = proc.stdout.readlines()
+        self.assertTrue(len(lines) > 1, "ERROR : nvme list-ctrl could not find ctrl")
+        # Skipping fist stdout line "num of ctrls present: <num_ctrls>"
+        ctrl_id = lines[1].split("]:")[1].strip()
```
igaw, line 190: I think we should start using the json output of the commands, this would make all the crazy parsing unnecessary.

tests/nvme_format_test.py
```diff
@@ -63,10 +63,13 @@ class TestNVMeFormatCmd(TestNVMe):
...
-        self.flbas = 0               # ns formattes logical block settings
-        self.nsze = 0x1400000        # ns size
-        self.ncap = 0x1400000        # ns capacity
+        self.dps = 0
+        self.flbas = 0
+        # Assuming run_ns_io with 4KiB * 10 writes and at
+        # minimum lba format 512B -> (4096B * 10)/512B = 80 = 0x50
+        ncap = 0x50
+        self.ncap = ncap
+        self.nsze = ncap
```
igaw, line 72: Would it possible to query the numbers and use them as input here? I am fine with going with a simple version first. Just wondered if we should be a bit more smarter eventually, e.g. different format setting (512, 4k)?

.github/workflows/run-nightly-tests.yml
```diff
@@ -0,0 +1,93 @@
...
+          uname -a
+      - name: Clean up test device
+        run: |
+          #BDEV0 is an environment variable of the self-hosted runner instance
+          #that contains a valid nvme ctrl name which is capable of the nvm
+          #command set.
+          CONTROLLER=$(echo /dev/${BDEV0} | sed 's/n[0-9]*$//')
+          sudo nvme delete-ns $CONTROLLER -n 0xffffffff
+          sudo nvme format $CONTROLLER -n 0xffffffff -l 0 -f
+          SIZE=$(sudo nvme id-ctrl $CONTROLLER | grep tnvmcap | sed 's/^tnvmcap   : //g' | sed 's/,//g' | awk '{print $1/512}')
```
igaw, line 24: `nvme id-ctrl /dev/nvme0 --output json | jq  -r '{tnvmcap} | .[]'`

.github/workflows/run-nightly-tests.yml
```diff
@@ -0,0 +1,93 @@
...
+          sudo nvme create-ns -s $SIZE -c $SIZE -f 0 -d 0 --csi=0 $CONTROLLER
+          sudo nvme attach-ns $CONTROLLER -n 1 -c 0
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y \
+          meson gcc pkg-config git libjson-c-dev libssl-dev libkeyutils-dev \
+          libdbus-1-dev libpython3-dev pipx python3-dev swig xz-utils
```
igaw, line 33: I think we should prebuild these containers, since in the past this step fails way to often due network SNAFUs.

igaw:
Looks really good already. I think we should start using the json output for the new code and transform the old code afterwards when the CI is up and running. But if you think you would like first to get it running as it is and then change the tests accordingly, that's also fine with me.

tests/nvme_test.py
MaisenbacherD: I agree, the code is much cleaner like that. :)
For now, I start using the json output on the lines that I am touching within the commits.
I will submit a chore patchset for some cleanups where I transition the rest to use json outputs.

tests/nvme_test.py
```diff
@@ -478,7 +478,7 @@ def get_id_ctrl(self, vendor=False):
         if not vendor:
             id_ctrl_cmd = "nvme id-ctrl " + self.ctrl
         else:
-            id_ctrl_cmd = "nvme id-ctrl -v " + self.ctrl
+            id_ctrl_cmd = "nvme id-ctrl -V " + self.ctrl
```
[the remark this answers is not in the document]
MaisenbacherD: I agree, I will start using the long format on the lines that I am touching right now.
This is also a task for the chore patchset.

tests/nvme_format_test.py
MaisenbacherD: Yes :) I changed the setUp step.

.github/workflows/run-nightly-tests.yml
MaisenbacherD: This is installing the dependencies directly on the VM (without a container in between) which gets reset everytime a workflow finishes.
I could install those additional packages when spawning the self-hosted runner.
Otherwise, a cleaner option could be to start using a prebuild container in the VM to run those tests on.
What do you think - should I run the tests additionally in the container within the VM? :)

MaisenbacherD:
> Looks really good already. I think we should start using the json output for the new code and transform the old code afterwards when the CI is up and running. But if you think you would like first to get it running as it is and then change the tests accordingly, that's also fine with me.

Thanks for the review :) I changed the sections you commented on. Letting the tests run on my branch now.
I think it makes sense to use a container within the VM instead of running directly in the VM.
I will now experiment a bit with that before pushing this update here.

.github/workflows/run-nightly-tests.yml
igaw: Now I wonder why this worked... I though it should be `--output-format`...

.github/workflows/run-nightly-tests.yml
igaw: I think we can start with this approach and address it later. The install step usually works for a couple of month followed by few days nothing works. That's why I think we are going to need the pre build containers eventually. But for the time being it's okay to go with this. Let's get it working first before going into too many rabbit holes.

tests/nvme_test.py
```diff
@@ -147,22 +147,23 @@ def nvme_reset_ctrl(self):
...
     def get_ctrl_id(self):
-        """ Wrapper for extracting the controller id.
+        """ Wrapper for extracting the first controller id.
             - Args:
                 - None
             - Returns:
                 - controller id.
         """
-        get_ctrl_id = "nvme list-ctrl " + self.ctrl
+        get_ctrl_id = f"nvme list-ctrl {self.ctrl} --output json"
```
igaw, line 180: hmm, should be `--output-format` IIRC. Does `--output` work for you? I remember we did some changes in the option parsing section. Did we break something again?

igaw: interesting `--output` works as well. Maybe the parser is a bit funky and matches it with `--output-format`. Anyway, use `--output-format` and we need look into why `--output` is also working. Yay another bug!

MaisenbacherD: I changed it to `--ouput-format=json` for all occasions :)
I can start looking at the faulty parsing next.

MaisenbacherD:
The checkpatch is failing with the following:
```
-------------------------------------------------------------
[13/13] Check commit - b75fb594d323ca1f5facf47c9062037969037bb4
-------------------------------------------------------------
WARNING: added, moved or deleted file(s), does MAINTAINERS need updating?
#18:
new file mode 100644
```
I am not sure if this is applicable since I can't spot a MAINTAINERS file 🤔

.github/workflows/run-nightly-tests.yml
MaisenbacherD: Sounds good! I will include a commit for this in the next PR to the test infrastructure :)

igaw:
You can ignore that checkpatch failure. We don't do the MAINTAINER file thing, so any new file added to project will trigger this message.

igaw:
BTW, I've just remembered, that we also have a Python binding for the library which provides the constants. Maybe we could use these in future as well. Just as idea.

igaw:
Now, I have to figure out how to enable it :)

MaisenbacherD:
Thanks for the review and merge :)
I will contact you to set up the VM and exchange the config token.

PR #2547, opened 2024-10-28
nvme-print: Add Sanitize Media Verification Event in PEL log

francispravin5:
Add Sanitize Media Verification Event in Persistent Event Log.
TP4152 - Post-Sanitize Media Verification 2024.04.01 Ratified.


igaw:
Could you also update `nvme-print-json.c` accordingly? At least it looks like `json_pevent_entry` does also listing each event individual. Thanks!

francispravin5:
Hi @igaw, the `Sanitize Media Verification Event` don't have the event data. I added this print statement just to ignore `Reserved Event` print. In `json_pevent_entry` we are not logging as reserved event.
Thanks!

PR #2549, opened 2024-10-28
ocp: fix firmware activation history entry endian

ikegami-t:
Fix the entry data fields to little endian format. Basically print functions convert the fields data to cpu format. But still some errors then fixed the print functions also.


igaw:
@arthurshau  another endian fix from @ikegami-t to review

arthurshau:
LGTM. Thanks!

PR #2551, opened 2024-10-29
nvme-id-domain(1): dom{ia => ai}n

nabijaczleweli:
$ sed -i 's/domian/domain/g' $(git grep -l domian)


nabijaczleweli:
applied

igaw:
I've applied the patch directly after adding a prefix to subject title. Thanks!

PR #2553, opened 2024-10-30
docs: add howto to setup TLS connections

igaw:
There are couple of step necessary to get TLS working nicely. Document it how this is done.

### DIFF
--- a/TLS.md
+++ b/TLS.md
@@ -1,0 +1,280 @@
+# How to Set Up TLS for NVMe-TCP
+
+Enabling TLS for the NVMe-TCP transport requires a few configuration
+steps for both the kernel and userland.
+
+## Kernel Configuration
+
+To support TCP authentication and TLS encryption, enable the following
+kernel options:
+
+- For DHCHAP authentication:
+  `CONFIG_NVME_HOST_AUTH`
+
+- For TLS transport encryption:
+  `CONFIG_NVME_TCP_TLS`
+
+These configuration option depend on another config option but these
+will be auto selected.
+
+## Userland
+
+For the userland configuration two components need to be configured.
+First, the tlshd TLS handshake daemon needs to be running and TLS keys
+need to be loaded into the kernel keystore.
+
+### Setting Up `tlshd`
+
+For TLS protocol support, which handles authentication and encryption,
+the kernel handles data encryption only, so userland support is required
+for the TLS handshake. The `tlshd` daemon implements the handshake
+process.
+
+#### Requirements
+
+Ensure `tlshd` includes the commit `311d9438b984` ("tlshd: always link
+.nvme default keyring into the session") - likely in `ktls-utils` version
+0.12. Alternatively, you can set the keyring manually in
+`/etc/tlshd.conf`:
+
+```ini
+[authenticate]
+keyrings = .nvme
+```
+
+#### Enable/start tlshd
+
+No additional configuration is necessary for `tlshd`; simply start it as
+a daemon:
+
+```bash
+systemctl enable --now tlshd
+```
+
+### Loading Keys on Boot or Module Load
+
+When the kernel is establishing a TCP connection with TLS, the NVMe
+subsystem loads keys from the kernel keystore. This means these keys
+must be available in the keystore before establishing a connection.
+
+nvme-cli provides command line interfaces to create, import and export
+keys into the kernel keystore. Though it's not the only way to
+import/export keys. If there is another system component managing the
+keys, the following steps for creating and making the keys persistent
+over boot cycles are not necessary.
+
+To stress this point, the nvme-cli is explicitly trying to avoid handling
+the keys, the only requirement is that the keys are present in the
+keystore.
+
+#### Creating a New Key
+
+```bash
+nvme gen-tls-key \
+  --hostnqn nqn.2014-08.org.nvmexpress:uuid:befdec4c-2234-11b2-a85c-ca77c773af36 \
+  --subsysnqn nqn.io-1 --hmac 1 --identity 1 --insert --keyfile /etc/nvme/tls-keys
+```
+
+This command creates a new host key, inserts it into the kernel keyring,
+and appends the derived TLS PSK to the keyfile (`/etc/nvme/tls-keys`).
+
+### Inserting an Existing Key
+
+```bash
+nvme check-tls-key \
+  --hostnqn nqn.2014-08.org.nvmexpress:uuid:befdec4c-2234-11b2-a85c-ca77c773af36 \
+  --subsysnqn nqn.io-1 --identity 1 \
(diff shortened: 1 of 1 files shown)


TLS.md
```diff
@@ -0,0 +1,194 @@
...
+This command creates a new host key, inserts it into the kernel keyring, and appends the derived TLS PSK to the keyfile (`/etc/nvme/tls-keys`).
+
+### Inserting an Existing Key
+
+```bash
+nvme check-tls-key \
+  --hostnqn nqn.2014-08.org.nvmexpress:uuid:befdec4c-2234-11b2-a85c-ca77c773af36 \
+  --subsysnqn nqn.io-1 --identity 1 \
+  --keydata NVMeTLSkey-1:01:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACtVQoZ: \
+  --insert --keyfile /etc/nvme/tls-keys
```
martin-gpy, line 55: Looks like the ```--keyfile``` option is currently not available for ```check-tls-key```, but only for ```gen-tls-key```.

TLS.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+This command inserts the configured key (`--keydata`) into the kernel keyring and appends the derived TLS PSK to the keyfile.
+
+### Loading Keys on Boot or Module Load
+
+The kernel keyring does not persist keys, so userland must import keys into the keyring upon each boot or module load (for NVMe-TCP). To load keys:
+
+```bash
+nvme tls --import /etc/nvme/tls-keys
```
martin-gpy, line 65: This will hang waiting for stdin input. This should probably be replaced with ```nvme tls --import --keyfile=/etc/nvme/tls-keys``` or ```nvme tls < /etc/nvme/tls-keys``` here.

TLS.md
igaw: I just added this new option yesterday :)

TLS.md
igaw: I should have copied these commands verbatim from my test script...

igaw:
I've added some more content to it, e.g. the debugging tips. I've spend some time to figure them all out. Maybe it helps others to shortcut if something isn't working.

TLS.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Loading Keys on Boot or Module Load
+
+The kernel keyring does not persist keys, so userland must import keys
+into the keyring upon each boot or module load (for NVMe-TCP). The
+nvme-tcp module provides the `psk` type keystore, thus only when the
+nvme-tcp module is available it possible to load keys into the keystore:
+
+```bash
+nvme tls --import /etc/nvme/tls-keys --keyfile /etc/nvme/tls-keys
```
martin-gpy, line 102: ```/etc/nvme/tls-keys``` is repeated here. Just keep it as ```nvme tls --import --keyfile /etc/nvme/tls-keys```.

igaw:
@martin-gpy do miss something or is my wording difficult to understand? I tried to write down everything I learned in the last few weeks on this topic. But don't expect too much from me, I am not a technical doc writer :)

TLS.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+### Loading Keys on Boot or Module Load
+
+When the kernel is establishing a TCP connection with TLS, the NVMe
+subsystem loads keys from the kernel keystore. This means these keys
+must be available in the keystore before establishing a connection.
+
+nvme-cli provides command line interfaces to create, import and export
+keys into the kernel keystore. Though it's not necessary the only way to
```
martin-gpy, line 61: Nit: Replace ```necessary``` with ```necessarily```

TLS.md
```diff
@@ -0,0 +1,280 @@
...
+will load keys from `/etc/nvme/tls-keys` automatically.
+
+```udev
+ACTION=="add", SUBSYSTEM=="module", KERNEL=="nvme_tcp", TEST=="@SYSCONFDIR@/tls-keys", RUN+="@SBINDIR@/nvme tls --import --keyfile @SYSCONFDIR@/tls-keys"
+```
+
+### Recommendation for Handling TLS Keys
+
+The `nvme connect` command also allows passing a TLS key directly via the
+command line or a JSON config file. Avoid this method in production
```
martin-gpy, line 116: Are we saying passing the TLS key via the config JSON file is also a security risk, and should be avoided in production environments?

TLS.md
```diff
@@ -0,0 +1,280 @@
...
+- If several keys available in the keystore which match up to the `<PSK digest>`
+the first match will be used. If this is the wrong key, it can be revoked by
+```bash
+nvme tls --revoke <identity>
+```
+
+- It's possible to provide a TLS key directly via the `nvme connect --tls
+--tls-key` command. If only the key is provided, nvme-cli assumption it
+is a configured PSK and thus does all the key transformation and creates
+the identity automatically. If the `--tls-psk-identity` is also present
```
martin-gpy, line 273: I don't see a ```--tls-psk-identity``` option available for the nvme connect command. Am I missing something here?

igaw: Typo, should be `--tls-key-identity`

TLS.md
igaw: I decided to drop the word completely. It's not adding anything :)

TLS.md
igaw: I think we should state this. Generally, mixing configuration with authentication/security tokens is discourage by the security folks.  This doesn't mean it is insecure per se. If no one has access to config file except root and nvme-cli, it's propably okay? I heard that the aim should be to remove/hide the tokens from the FS after it has been used. So given this I'd say it better to say it's not recommended to do so.

TLS.md
```diff
@@ -0,0 +1,280 @@
...
+`<PSK>` is the derive TLS PSK and not the retained nor the configured PSK.
+
+- If several keys available in the keystore which match up to the `<PSK digest>`
+the first match will be used. If this is the wrong key, it can be revoked by
+```bash
+nvme tls --revoke <identity>
+```
+
+- It's possible to provide a TLS key directly via the `nvme connect --tls
+--tls-key` command. If only the key is provided, nvme-cli assumption it
```
martin-gpy, line 271: Nit: Replace ```assumption``` with ```assumes```

TLS.md
```diff
@@ -0,0 +1,280 @@
...
+is a configured PSK and thus does all the key transformation and creates
+the identity automatically. If the `--tls-key-identity` is also present
+nvme-cli assumes it is a derived TLS PSK and does not transformation on
```
martin-gpy, line 274: Nit: Replace ```does not transformation``` with ```does not attempt transformation```

martin-gpy:
> @martin-gpy do miss something or is my wording difficult to understand? I tried to write down everything I learned in the last few weeks on this topic. But don't expect too much from me, I am not a technical doc writer :)

Just had a few minor review comments. This looks quite good now. Thanks for getting this done.

PR #2554, opened 2024-10-31
build: Add token for coverage test

MaisenbacherD:
The code coverage action is silently failing partly because the required Codecov token is missing.
Reading the token from the GitHub secrets.

The missing gpg package must be fixed in
https://github.com/linux-nvme/ci-containers.

Furthermore, fail the coverage workflow if the codecov-action errors out.

Please add the `CODECOV_TOKEN` to the GitHub secrets :)


MaisenbacherD:
The missing gpg package was fixed in https://github.com/linux-nvme/ci-containers/pull/3

PR #2556, opened 2024-11-06
nvme: change NVME_ARGS -v and -o options ordering to back

ikegami-t:
This reverts the short options -o and -v changes as before. Before the short option -o used for admin-passthru command opcode option. But currently changed the short option -o for output-format option.

Fixes: 0ded9387b ("nvme: Change short option -o and -v duplicated to upper case")

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -879,7 +879,7 @@ static int get_telemetry_log(int argc, char **argv, struct command *cmd,
 	};
 
 	NVME_ARGS(opts,
-		  OPT_FILE("output-file",     'O', &cfg.file_name, fname),
+		  OPT_FILE("output-file",     'o', &cfg.file_name, fname),
 		  OPT_UINT("host-generate",   'g', &cfg.host_gen,  hgen),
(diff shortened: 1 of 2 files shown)


igaw:
I am not happy to change the command line options again. Though I haven't looked into closely yet.

ikegami-t:
Yes I know you mentioned this before by the PR comments https://github.com/linux-nvme/nvme-cli/pull/2285#issuecomment-2166109178 then the short options including `-o` were not changed. But actually the `-o` option issue has just reported again so I thought to try to fix the options again. But still should we keep the short options `-o` and `-v`? Note: Currently the script failed by using `-o` option is also going to be changed to use `--opcode` instead as same with the linux-nvme mailing list mentioned by the PR comment https://github.com/linux-nvme/nvme-cli/pull/2285#issuecomment-2074877250. Thank you.

ikegami-t:
Thank you so much for your explanation. Understood so let me close the PR.

PR #2558, opened 2024-11-07
nvme-print: use NVME_GET in sanitize log

francispravin5:
Used NVME_GET macro in sanitize log. Also, updated the print statement as per NVM Express Base Specification Revision 2.1.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -1375,19 +1375,16 @@ static void json_sanitize_log(struct nvme_sanitize_log_page *sanitize_log,
 	struct json_object *r = json_create_object();
 	struct json_object *dev = json_create_object();
 	struct json_object *sstat = json_create_object();
+	__u16 status = le16_to_cpu(sanitize_log->sstat);
 	const char *status_str;
 	char str[128];
-	__u16 status = le16_to_cpu(sanitize_log->sstat);
 
 	obj_add_int(dev, "sprog", le16_to_cpu(sanitize_log->sprog));
-	obj_add_int(sstat, "global_erased", (status & NVME_SANITIZE_SSTAT_GLOBAL_DATA_ERASED) >> 8);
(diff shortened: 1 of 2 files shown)


nvme-print-json.c
```diff
@@ -1375,20 +1375,17 @@ static void json_sanitize_log(struct nvme_sanitize_log_page *sanitize_log,
...
 	char str[128];
-	__u16 status = le16_to_cpu(sanitize_log->sstat);
 
 	obj_add_int(dev, "sprog", le16_to_cpu(sanitize_log->sprog));
-	obj_add_int(sstat, "global_erased", (status & NVME_SANITIZE_SSTAT_GLOBAL_DATA_ERASED) >> 8);
-	obj_add_int(sstat, "no_cmplted_passes",
-		    (status >> NVME_SANITIZE_SSTAT_COMPLETED_PASSES_SHIFT) &
-		    NVME_SANITIZE_SSTAT_COMPLETED_PASSES_MASK);
+	obj_add_int(sstat, "global_data_erased", NVME_GET(status, SANITIZE_SSTAT_GLOBAL_DATA_ERASED));
+	obj_add_int(sstat, "overwrite_passes_completed", NVME_GET(status, SANITIZE_SSTAT_COMPLETED_PASSES));
```
igaw, line 1384: well, the problem with the renaming is that this is will break any existing user. I have not problem with the renaming when we doing major version update. We could also just add the correct one and leave the old around and remove it when we do a major version upgrade (it's in the pipeline...). In this case please add a comment like 'deprecated, remove it on major update' or even open up an issue to track it, so we don't forget it.

francispravin5: @igaw Shall I change the name to old for now?

francispravin5: Changed the names to old one to stay backward compatible.
Thanks!

PR #2560, opened 2024-11-08
nvme: Live Migration

NateThornton:
Implementation of TP 4159 PCIe Infrastructure for Live Migration plugin. Includes command support for Track Send, Migration Receive, Migration Send, and Controller Data Queue; Identify Controller LM related fields; Bash and ZSH completions.

Changes are isolated to the User Data Migration subset, with Track Memory functionality deferred to a future commit.

[TP4159 PCIe Infrastructure for Live Migration 2024.07.30 Ratified.pdf](https://github.com/user-attachments/files/17682852/TP4159.PCIe.Infrastructure.for.Live.Migration.2024.07.30.Ratified.pdf)

### DIFF
--- a/plugins/lm/lm-nvme.c
+++ b/plugins/lm/lm-nvme.c
@@ -1,0 +1,669 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., LTD.
+ *
+ * Authors: Nate Thornton <n.thornton@samsung.com>
+ */
+
+#include <assert.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <linux/fs.h>
+#include <inttypes.h>
+#include <asm/byteorder.h>
+#include <sys/mman.h>
+#include <sys/shm.h>
+#include <sys/sysinfo.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <dirent.h>
+#include <time.h>
+
+#include "common.h"
+#include "nvme.h"
+#include "nvme-print.h"
+#include "libnvme.h"
+#include "plugin.h"
+#include "linux/types.h"
+#include "nvme-wrap.h"
+#include "util/cleanup.h"
+
+#define CREATE_CMD
+#include "lm-nvme.h"
+
+#include "lm-print.h"
+
+static inline const char *arg_str(const char * const *strings, size_t array_size, size_t idx)
+{
+	if (idx < array_size && strings[idx])
+		return strings[idx];
+	return "unrecognized";
+}
+
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
+#define ARGSTR(s, i) arg_str(s, ARRAY_SIZE(s), i)
+
+static int lm_create_cdq(int argc, char **argv, struct command *command, struct plugin *plugin)
+{
+	const char *desc = "Create Controller Data Queue for controller of specific type and size";
+	const char *sz = "CDQ Size (in dwords)";
+	const char *cntlid = "Controller ID";
+	const char *qt = "Queue Type (default: 0 = User Data Migration Queue)";
+	const char *consent = "I consent this will not work and understand a CDQ cannot be mapped "
+			      "to user space. If I proceed with the creation of a CDQ, the device "
+			      "will write to invalid memory, inevitably leading to MMU faults or "
+			      "worse.";
+
+	_cleanup_huge_ struct nvme_mem_huge mh = { 0, };
+	_cleanup_nvme_dev_ struct nvme_dev *dev = NULL;
+	struct lba_migration_queue_entry_type_0 *queue = NULL;
+	int err = -1;
+
+	struct config {
+		__u32	sz;
+		__u16	cntlid;
+		__u8	qt;
+		bool	consent;
+		char	*file;
+	};
+
+	struct config cfg = {
+		.sz = 0,
+		.cntlid = 0,
+		.qt = 0,
+		.consent = false,
+		.file = NULL,
+	};
+
+	OPT_ARGS(opts) = {
+		OPT_UINT("size",	's', &cfg.sz,		sz),
+		OPT_SHRT("cntlid",	'c', &cfg.cntlid,	cntlid),
+		OPT_BYTE("queue-type",	'q', &cfg.qt,		qt),
+		OPT_FLAG("consent",	  0, &cfg.consent,	consent),
+		OPT_END()
+	};
+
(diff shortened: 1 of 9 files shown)


plugins/lm/lm-nvme.c
```diff
@@ -0,0 +1,988 @@
...
+	DELETE_CONTROLLER_DATA_QUEUE = 1
+};
+
+static int lm_create_cdq(int argc, char **argv, struct command *command, struct plugin *plugin)
+{
+	const char *desc = "Create Controller Data Queue for controller of specific type and size";
+	const char *sz = "CDQ Size (in dwords)";
+	const char *cntlid = "Controller ID";
+	const char *qt = "Queue Type (default: 0 = User Data Migration Queue)";
+	const char *consent = "I consent this will not work and understand a CDQ cannot be mapped "
```
igaw, line 91: nice one!

plugins/lm/lm-nvme.c
```diff
@@ -0,0 +1,988 @@
...
+	};
+
+	err = nvme_submit_admin_passthru(dev_fd(dev), &cmd, NULL);
+	if (!err)
+		printf("Migration Send (%s) Successful\n",
+		       ARGSTR(lm_migration_send_select_argstr, cfg.sel));
+	else if (err > 0)
+		nvme_show_status(err);
+	else
+		nvme_show_error("ERROR: nvme_submit_admin_passthru() failed %s", strerror(errno));
```
igaw, line 458: I think this part should be in libnvme like the other helpers.

igaw: and with this in mind, the same comment is for the other places above. This is in the base spec, so we should also properly support it in the library and not have all the logic in the frontend.

plugins/lm/lm-nvme.c
```diff
@@ -0,0 +1,988 @@
...
+struct nvme_io_submission_queue_data {
+	__u64 prp1;
+	__u16 qsize;
+	__u16 qid;
+	__u16 cqid;
+	__u16 attrs;
+	__u16 hp;
+	__u16 tp;
+	__u8  rsvd[4];
+};
```
igaw, line 473: these things really should be in libnvme

NateThornton:
As we placed the new lm iocts in libnvme version 1.12, I think this change is still pending the next libnvme release.

Then one follow-up change to meson.build is required to grab that new release.

completions/_nvme
```diff
@@ -617,6 +618,118 @@ _nvme () {
 				;;
 			esac
 			;;
+		(lm)
```
NateThornton, line 621: @NateThornton Self-reminder these need updating too

subprojects/libnvme.wrap
```diff
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/linux-nvme/libnvme.git
-revision = 8efcec730603ffcc379ba0bf821e4c24ec87715b
+revision = 27ed889d0b1917919763816c97a49bb6e29d4547
```
igaw, line 3: btw, could split the libnvme build dependency into a separate patch and have it first? This makes any backporting way simpler.

nvme-print-json.c
```diff
@@ -434,6 +434,16 @@ void json_nvme_id_ctrl(struct nvme_id_ctrl *ctrl,
...
 
+	obj_add_uint(r, "cmmrtd", le16_to_cpu(ctrl->cmmrtd));
+	obj_add_uint(r, "nmmrtd", le16_to_cpu(ctrl->nmmrtd));
+	obj_add_uint(r, "minmrtg", ctrl->minmrtg);
+	obj_add_uint(r, "maxmrtg", ctrl->maxmrtg);
+	obj_add_uint(r, "trattr", ctrl->trattr);
+	obj_add_uint(r, "mcumq", le16_to_cpu(ctrl->mcudmq));
+	obj_add_uint(r, "mcmr", le16_to_cpu(ctrl->mcmr));
+	obj_add_uint(r, "nmcmr", le16_to_cpu(ctrl->nmcmr));
+	obj_add_uint(r, "mcdqpc", le16_to_cpu(ctrl->mcdqpc));
```
igaw, line 445: the nvme-print*.c updates should be a separate patch.

completions/_nvme
igaw: and while at, also split the tab completions changes into a separate patch. don't forget to update the docs too :)

plugins/lm/lm-nvme.c
```diff
@@ -0,0 +1,894 @@
...
+		nvme_show_status(err);
+	else
+		printf("Migration Send (%s) Successful\n",
+		       ARGSTR(lm_migration_send_select_argstr, cfg.sel));
+
+
+	return err;
+}
+
+static void json_controller_state_data(struct nvme_lm_controller_state_data *data, size_t len)
```
igaw, line 420: So it might make sense to support builds without json-c dependencies for this new plugin. Most plugins depend on json-c to be available.  @ikegami-t started to split the plugins so they work also when json-c is missing (see the ocp plugin). I think it would be good to have this from the beginning as it is currently small code base. Though the LM is something in the base spec, so not unlikely to change or get udpated soonish :)

subprojects/libnvme.wrap
NateThornton: Created #2616

nvme-print-json.c
NateThornton: Glad you pointed this out because it's already included in #2607. Surprised I didn't get a merge conflict.

subprojects/libnvme.wrap
igaw: Usually you just add a patch to a PR, no need for an extra PR just to update the dependency.

plugins/lm/lm-nvme.c
NateThornton: Refactored following ocp plugin as an example.

igaw:
The commit subject should be prefix, e.g. `plugins/lm: introduce live migration plugin`, `plugins/ls: add print files`, ...
Commit message shouldn't contain version information such as 'reverted ...'. After it gets merged no one cares about this things. If we really want to preserve the development history we would need to change a few things anyway.

Reverted nvme-print-stdout.c changes: still something in the first patch. As I said, I don't mind that nvme-print-stdout.c and nvme-print-json.c get updated, it  just deserves it's own patch.

Also why patch 2 removing stuff added previously. These patches should really be consisting out of new code being added and a tiny change in the meson.build file.

And the build is unhappy due to printf formatting issues.

NateThornton:
I should have clarified my intent to squash the commits after they were individually reviewed. Everything now is in one commit and the build is passing.

I think the remaining warnings are acceptable, but if something needs to be done to clean them up LMK.

As we placed the new lm_* ioctls in release 12 of libnvme, does this merge need to wait until that is released?

igaw:
I've dropped the changes in `nvme-print-stdout.c`. These were removing the output for the LM feature. 

Thanks!

PR #2561, opened 2024-11-11
plugins/netapp: add include of libgen.h for basename(3) prototype

dne:
Fixes build with musl libc.


igaw:
I'll see that we get a musl build setup in our workflows. We should catch such things earlier. Thanks!

PR #2563, opened 2024-11-12
ocp-nvme: Recommit removed patch #2529

sc108-lee:
ocp-nvme: Add LMDATA-37 for Latency Monitor Log
3cc286067954ebed1e4588989498e6377246f3f0 commit removed


igaw:
Could update the commit message so that it doesn't contain a github reference. The rev is not a valid (it's not part of this repository). Also use something like  `git show -s --pretty='format:%h (\"%s\")\n' --abbrev=12`  (I have this as alias in `.gitconfig` 'one', so I can do `git one $rev`) to give the rev a bit of a meaning:  `5572c9072b7e ("tests: Use raw string for compiling regex")`

Thanks

sc108-lee:
Thanks for your feedback, Updated as you recommanded !

PR #2564, opened 2024-11-12
netapp-ontapdev: add verbose output option

martin-gpy:


Documentation/nvme-netapp-ontapdevices.txt
```diff
@@ -22,6 +22,10 @@ OPTIONS
 	Set the reporting format to 'normal' (default), 'column', or
 	'json'. Only one output format can be used at a time.
 
+-v::
+--verbose::
+	Display additional information of ONTAP devices on the host.
+
```
igaw, line 28: Could also update the section above `[verse]`? And just update the `txt` file, the rest is auto generated by release. Thanks.

martin-gpy: Done. I've updated the ```txt``` file alone now. Thanks for reviewing.

PR #2574, opened 2024-11-15
sed: add '--read-only' to 'sed initialize'

hreinecke:
By default 'sed initialize' will set the entire disk to read/write locked, ie no access is possible. That is all fine if the BIOS supports TCG Opal, as then the BIOS can unlock the drive before access. But for BIOS _not_ supporting TCG Opal this results in an inaccessible drive, and requires the user to use a different drive to boot and unlock the SED device.
This patch adds a flag '--read-only' to 'nvme sed initlialize' such that the 'read-lock enable' flag is not set for the locking range, and the device continues to be readable by the BIOS for booting.

### DIFF
--- a/plugins/sed/sed.c
+++ b/plugins/sed/sed.c
@@ -23,6 +23,12 @@ OPT_ARGS(no_opts) = {
 	OPT_END()
 };
 
+OPT_ARGS(init_opts) = {
+	OPT_FLAG("read-only", 'r', &sedopal_lock_ro,
+		 "Set locking range to read-only"),
+	OPT_END()
+};
+
 OPT_ARGS(key_opts) = {
 	OPT_FLAG("ask-key", 'k', &sedopal_ask_key,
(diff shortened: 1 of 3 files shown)


igaw:
looks good to me. 

@gjoyce-ibm could have a look too? thanks!

gjoyce-ibm:
> looks good to me.
> 
> @gjoyce-ibm could have a look too? thanks!

@igaw The code changes look good. My only question would be on the utility of this:

1. SED is supposed to protect data at rest. If the drive is locked read only the drive could be moved to another system and the data would be exposed.
2. Can Linux by default boot on a read only boot device? Even if boot succeeds, the drive would need to changed to read/write at some point.
3. This might also suggest that "lock" and "unlock" sub-commands need to support read/write rather than "initialize".

The other possibility for unlocking at boot time is to use a shadow mbr and boot something that can prompt for or otherwise provide the SED password to unlock the drive.

PR #2570, opened 2024-11-13
ocp-nvme: fix the error display of Hardware Component Log

YUEyuan310:
The display of Hardware Component (Log Identifier C6h) failures to comply with protocol OCP2.5

Issue: https://github.com/linux-nvme/nvme-cli/issues/2566

PR #2580, opened 2024-11-19
Ocp2.6 Hardware Component Log 

sc108-lee:
Based #2570 spec modification based ocp2.6

Log Identifier C6h
Hardware Component Log Size (bytes)
Component Identifier added
0x000C Born on Date. The date on which the device was manufactured.
ASCII string format is MMDDYYYY.


plugins/ocp/ocp-hardware-component-log.c
```diff
@@ -154,6 +154,8 @@ const char *hwcomp_id_to_string(__u32 id)
 		return "Country of Origin";
 	case HWCOMP_ID_HW_REV:
 		return "Global Device Hardware Revision";
+	case HWCOMP_ID_BORN_ON_DATE:
+		return "Born on Date";
```
ikegami-t, line 158: The command option comp-id is not added.
```c
	OPT_VALS(id) = {
		VAL_LONG("asic", HWCOMP_ID_ASIC),
		...
		VAL_LONG("hw-rev", HWCOMP_ID_HW_REV),
		VAL_LONG("vendor", HWCOMP_ID_VENDOR),
		VAL_END()
	};
```

ikegami-t (commented):
By the way only changed the hardware component log size and added the component ID born on date for the log page by OCP verion 2.6? Is there any documentation or information for the changes version 2.6? Should we compare the documentation with the older version documentation? (I did this before to check the OCP 2.5 changes..)

sc108-lee:
https://www.opencompute.org/events/past-events/2024-ocp-storage-tech-talk

I could find a documentation, but I don't think this doc checks everything. (since there is no Hardware Component Log change)
check it out.

ikegami-t:
The changes look good. Thank you so much for your information. Just watched the OCP video for several 10 minutes and very interesting so will check more later since the play time is long.

igaw:
@sc108-lee could you rebase this PR? Looks like it conflicts with the merge from #2570. Thanks

sc108-lee:
@igaw  rebased :)

PR #2581, opened 2024-11-20
ocp-nvme: Add Error Injection type from ocp2.6

sc108-lee:
4.15.3.1 Error Injection Entry Data Structure
ERRIE-3 Error Injection Type

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -104,6 +104,17 @@ enum erri_type {
 	ERRI_TYPE_HW_MALFUNCTION,
 	ERRI_TYPE_NO_MORE_NAND_SPARES,
 	ERRI_TYPE_INCOMPLETE_SHUTDOWN,
+	ERRI_TYPE_METADATA_CORRUPTION,
+	ERRI_TYPE_CRITICAL_GC,
+	ERRI_TYPE_LATENCY_SPIKE,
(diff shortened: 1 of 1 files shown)


plugins/ocp/ocp-nvme.c
```diff
@@ -104,6 +104,16 @@ enum erri_type {
...
+	ERRI_TYPE_METADATA_CORRUPTION,
+	ERRI_TYPE_CRITICAL_GC,
+	ERRI_TYPE_LATENCY_SPIKE,
+	ERRI_TYPE_IO_CMD_FAILURE,
+	ERRI_TYPE_IO_CMD_TIMEOUT,
+	ERRI_TYPE_ADMIN_CMD_FAILURE,
+	ERRI_TYPE_ADMIN_CMD_TIMEOUT,
+	ERRI_TYPE_THERMAL_THROTTLE_ENGAGED,
+	ERRI_TYPE_THERMAL_THROTTLE_DISENGAGED,
+	ERRI_TYPE_CRITICAL_TEMPERATURE_EVENT,
```
igaw, line 116: The spec v2.6 also lists `Die Offline`. Shouldn't it also be listed?

sc108-lee: Yes you are right, I saw documentation comparison doc from OCP official site when they release the 2.6.
But ratified doc has Die offline!.
I just added it

Thanks for the comment :)

PR #2583, opened 2024-11-24
ocp: set UUID index to get hwcomp log

ikegami-t:

### DIFF
--- a/plugins/ocp/ocp-hardware-component-log.c
+++ b/plugins/ocp/ocp-hardware-component-log.c
@@ -11,6 +11,7 @@
 #include "nvme-print.h"
 #include "ocp-hardware-component-log.h"
 #include "ocp-print.h"
+#include "ocp-utils.h"
 
 //#define HWCOMP_DUMMY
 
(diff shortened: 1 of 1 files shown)


YUEyuan310:
When memory was not allocated to desc during the first acquisition, the nvme_get_log_page function was called, which requires a high degree of SSD-side compatibility. Can the nvme_get_log_simple function be optimized to include uuidx?

YUEyuan310:
The desc in log has not been allocated memory, and there should be special handling within the nvme_get_log_simple function. Directly using nvme_get_log_page may cause issues.

ikegami-t:
The desc is not used at the first nvme_get_log_page calling but passed to use the function parameter log instead as below so no need to allocate the memory for the call. Yes at first I also considered to change the function nvme_get_log_simple itself but seems better to keep the current API function behavior for other commands and callers. By the way is there any error caused actually by the changes? Thank you.
```diff
                .timeout = NVME_DEFAULT_IOCTL_TIMEOUT,
                .lid = LID_HWCOMP,
                .nsid = NVME_NSID_ALL,
+               .log = log,
+               .len = desc_offset,
        };
```

YUEyuan310:
> The desc is not used at the first nvme_get_log_page calling but passed to use the function parameter log instead as below so no need to allocate the memory for the call. Yes at first I also considered to change the function nvme_get_log_simple itself but seems better to keep the current API function behavior for other commands and callers. By the way is there any error caused actually by the changes? Thank you.
> 
> ```diff
>                 .timeout = NVME_DEFAULT_IOCTL_TIMEOUT,
>                 .lid = LID_HWCOMP,
>                 .nsid = NVME_NSID_ALL,
> +               .log = log,
> +               .len = desc_offset,
>         };
> ```

According to my previous NVMe feature code, using the modified CLI, the attempt to retrieve content from the hardware component failed.Modifying nvme_get_log_simple to accept uuidx should make it compatible with other calls, defaulting to 0 should be fine.

ikegami-t:
Thank you so much for your confirmation and also very sorry for the error. Just confirmed the cause and fixed the error so could you please retry the changes?

ikegami-t:
Thank you so much for your verification!

igaw:
Thanks a lot!

PR #2586, opened 2024-11-27
plugins/ocp: fix telemetry string log output format

Karthikmct:
This PR will take care of fixing the output format for "telemetry-string-log" command in ocp plug in

Fixed the code to handle json output crash and proper display of ASCII Table content
Disabling debug logs output while parsing telemetry-string-log data

Signed-off-by: Karthik Balan [karthik.b82@samsung.com](mailto:karthik.b82@samsung.com),

Reviewed-by: Arunpandian J [arun.j@samsung.com](mailto:arun.j@samsung.com),

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -2502,7 +2502,11 @@ static int get_c9_log_page(struct nvme_dev *dev, char *format)
 		return ret;
 	}
 
-	ret = get_c9_log_page_data(dev, 1, 0);
+	if (fmt == BINARY)
+		ret = get_c9_log_page_data(dev, 0, 1);
+	else
+		ret = get_c9_log_page_data(dev, 0, 0);
+
 	if (!ret) {
 		ocp_c9_log(log_data, pC9_string_buffer, total_log_page_sz, fmt);
 	} else {
@@ -2529,7 +2533,8 @@ static int ocp_telemetry_str_log_format(int argc, char **argv, struct command *c
 	};
 
 	OPT_ARGS(opts) = {
-		OPT_FMT("output-format", 'o', &cfg.output_format, "output Format: normal|json"),
+		OPT_FMT("output-format", 'o', &cfg.output_format,
+				"output Format:normal|json|binary"),
 		OPT_END()
 	};
 
--- a/plugins/ocp/ocp-print-json.c
+++ b/plugins/ocp/ocp-print-json.c
@@ -569,7 +569,7 @@ static void json_c9_log(struct telemetry_str_log_format *log_data, __u8 *log_dat
 	json_object_add_value_int(root, "Log Page Version",
 				  le16_to_cpu(log_data->log_page_version));
 
-	memset((__u8 *)res, 0, 15);
+	memset((__u8 *)res, 0, 48);
 	for (j = 0; j < 15; j++)
 		res += sprintf(res, "%d", log_data->reserved1[j]);
 	json_object_add_value_string(root, "Reserved", res_arr);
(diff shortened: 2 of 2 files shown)


igaw:
Generally, if you commit message already contains bullet points which this change does, these bullet point changes should usually be separate patches.

As far I read the ascii output change, it's a output format change, right? Not sure how you want to handle this. @arthurshau what are you thoughts on this?

Karthikmct:
>Generally, if you commit message already contains bullet points which this change does, these bullet point changes should usually be separate patches.
@igaw , patch is about fixing json output format for telemetry sting log, one is handling a crash and aligning the ASCII table as per specification requirement. These two fixes goes hand in hand, so raised as one PR. 

>As far I read the ascii output change, it's a output format change, right? Not sure how you want to handle this.
@igaw , It's about change in ascii table display in json output format. Not a change to output format itself. fix is to align with specification expected template.

Thanks.

igaw:
First point, okay. Didn't look too closely. Just thought, it's unrelated. 

Second point. If your change doesn't change the field names, than all is good. My argument is that changing the json field names (e.g. renaming) is a regression. The json output should be stable, so that any parser doesn't have to be touched just because we rename something.

Karthikmct:
> Second point. If your change doesn't change the field names, then all is good. My argument is that changing the json field names (e.g. renaming) is a regression. The json output should be stable, so that any parser doesn't have to be touched just because we rename something.

Got it. No changes are there for the field names. Its still noted 'ASCII Table'. The change is instead of printing the complete ascii data as a single dump, organized it in Byte wise ASCII data (as expected by specification)

Karthikmct:
@igaw , To ease the review and merge. I modified this pull request into 2 part. This part is to fix the json output format crash and disabling the debug prints only. 

Will take up the ASCII Table alignment in a separate pull request. please check further :)

plugins/ocp/ocp-print-json.c
```diff
@@ -569,10 +569,10 @@ static void json_c9_log(struct telemetry_str_log_format *log_data, __u8 *log_dat
 	json_object_add_value_int(root, "Log Page Version",
 				  le16_to_cpu(log_data->log_page_version));
 
-	memset((__u8 *)res, 0, 15);
+	memset((__u8 *)res, 0, 48);
 	for (j = 0; j < 15; j++)
 		res += sprintf(res, "%d", log_data->reserved1[j]);
-	json_object_add_value_string(root, "Reserved", res_arr);
+	json_object_add_value_string(root, "Reserved[15:01]", res_arr);
```
igaw, line 575: Printing the reserved fields in the json output is problematic. IMO. I kind of get it when printing on the console. The JSON output is for machines to parse. So updating the field name is a regression and when we add new fields the output might change as well. I think we should just not output these at all.

Obviously, this reserved field is already here and we should not break parsers. So I'd say just leave it as it is, don't rename it.

igaw:
Looks good, just the reserved field renaming thing

plugins/ocp/ocp-print-json.c
Karthikmct: I get your point. Since, this usage of reserved key had corrupted data added in its value field. 
However, the intention is not to break the parsers already available and machines don't care about reserved value. I will leave it as its , without renaming

igaw: One thing we could do is that we just append the 'reserved' field with no data at all. I don't expect that `res_arr` should be exposed.

Karthikmct: Oh, I just retained reserved and it anyway has non key value (just zero). 
Updated the pull request with the latest code. please do check.

Karthikmct:
@igaw , Could you please update further on the patch upstream?
Do let me know on clarification.

igaw:
Looks good to me. Thanks!

plugins/ocp/ocp-nvme.c
```diff
@@ -2502,7 +2502,11 @@ static int get_c9_log_page(struct nvme_dev *dev, char *format)
 		return ret;
 	}
 
-	ret = get_c9_log_page_data(dev, 1, 0);
+	if (fmt == BINARY)
+		ret = get_c9_log_page_data(dev, 0, 1);
```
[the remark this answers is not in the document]
Karthikmct: Ok @igaw , i will check and work on it.

PR #2588, opened 2024-11-29
nvme-netapp: misc fixes

martin-gpy:


plugins/netapp/netapp-nvme.c
```diff
@@ -276,16 +276,18 @@ static void netapp_ontapdevice_json(struct json_object *devices, char *devname,
...
 	device_attrs = json_create_object();
 	json_object_add_value_string(device_attrs, "Device", devname);
 	json_object_add_value_string(device_attrs, "Vserver", vsname);
 	json_object_add_value_string(device_attrs, "Namespace_Path", nspath);
 	json_object_add_value_int(device_attrs, "NSID", nsid);
 	json_object_add_value_string(device_attrs, "UUID", uuid);
-	json_object_add_value_uint64(device_attrs, "LBA_Data_Size", lba);
-	json_object_add_value_uint64(device_attrs, "Namespace_Size", nsze);
-	json_object_add_value_uint64(device_attrs, "UsedBytes", nuse);
+	json_object_add_value_uint64(device_attrs, "LBA_Size", lba);
```
igaw, line 288: This obviously introduces a regression, as the current field name `LBA`_Data_Size` is missing afterwards. Not sure if anyone will notice. We can add it and see if someone reports a regression. In this case can add it back. Though not so happy when we change field names in the json output. It's the 'stable' output format.

PR #2593, opened 2024-12-02
ocp: TCG Configuration Log field added ocp v2.6

sc108-lee:
Add Number of Namespace Provisioned Locking Objects Extended(TCGL-23)
remove reserved area prints (Shall be cleared to zero)
reserved area field name modified by byte address

### DIFF
--- a/plugins/ocp/ocp-print-json.c
+++ b/plugins/ocp/ocp-print-json.c
@@ -806,6 +806,7 @@ static void json_c7_log(struct nvme_dev *dev, struct tcg_configuration_log *log_
 	char *guid = guid_buf;
 	char res_arr[458];
 	char *res = res_arr;
+	__u16 log_page_version = le16_to_cpu(log_data->log_page_version);
 
 	root = json_create_object();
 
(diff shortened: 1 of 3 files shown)


plugins/ocp/ocp-print-json.c
```diff
@@ -829,13 +823,12 @@ static void json_c7_log(struct nvme_dev *dev, struct tcg_configuration_log *log_
 				  le32_to_cpu(log_data->pro_rlc));
 	json_object_add_value_int(root, "TCG Error Count", le32_to_cpu(log_data->tcg_ec));
 
-	memset((__u8 *)res, 0, 458);
-	for (j = 0; j < 458; j++)
-		res += sprintf(res, "%d", log_data->rsvd3[j]);
-	json_object_add_value_string(root, "Reserved3", res_arr);
```
igaw, line 835: Not happy that the json output changes. Adding is not an issue but changing is causing regressions. Though I think adding the reserved fields into json is wrong. But this ship has already sailed.

igaw:
@arthurshau could you have a look. Thanks.

igaw:
Thanks!

Issue #2582, opened 2024-11-21
ocp uuidx


YUEyuan310:
Why do some OCP commands specify UUID, while others do not？

ocp_get_uuid_index(dev, &uuid_index);？

For example：
The command fw activate history specifies uuidx
The command hardware-component-log doesn't specify uuidx.

YUEyuan310:
Should the uuid function be completed?

ikegami-t:
Because it is required by the OCP specification. For example the error injection set features required it as below by the OCP version 2.6.
![ocp-error-inject-set](https://github.com/user-attachments/assets/7727bcf3-0480-4d5c-88d3-98a0105e90d5)
But it was not required by the OCP version 1.0.3 as below.
![ocp-error-inject-set-v1 0 3](https://github.com/user-attachments/assets/11f368a7-de2d-4be2-bc75-1a477580303a)

YUEyuan310:
[...]

PR #2595, opened 2024-12-03
ocp: set UUID index to get extended smart log

ikegami-t:


igaw:
Thanks!

ikegami-t:
Fix for the issue https://github.com/linux-nvme/nvme-cli/issues/2582. Thank you.

PR #2598, opened 2024-12-04
nvme-print-stdout: print MAXWZD and NSZS bits of ONCS field

francispravin5:
Print the newly added MAXWZD and NSZS bits of ONCS field in Identify Controller Data Structure.
TP4160 - WZSL Limit Modification


igaw:
Thanks!

PR #2599, opened 2024-12-05
Telemetry mcda

sc108-lee:


igaw:
just reordered the patches, first the lib dependency update and then the change. This makes it possible to bissect nvme-cli without breakings.

igaw:
Thanks!

PR #1896, opened 2023-04-06
plugins/ocp: Fix LM log command, Add set LM feature command, Add telemetry-log command

arthurshau:
Large PR which is meant to upstream the work I've done on Meta's internal OCP plugin (which is based off an older version of nvme-cli) to nvme-cli 2.3. 

Changes include:
- Reorganizing ocp-nvme.c to group related structs and functions
- Fixing typos in and updating the Latency Monitor Log functions to be OCP 2.0 compliant
- Adding the set latency monitor feature command
- Adding the telemetry log command

See commit messages for more details.

PR #2602, opened 2024-12-05
ocp: set UUID index to get telemetry log

ikegami-t:

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -832,6 +832,8 @@ static int extract_dump_get_log(struct nvme_dev *dev, char *featurename, char *f
 			.ot = false,
 		};
 
+		ocp_get_uuid_index(dev, &args.uuidx);
+
 		err = nvme_get_log(&args);
 		if (err) {
 			if (i > 0)
@@ -1217,6 +1219,8 @@ static int get_telemetry_log_page_data(struct nvme_dev *dev, int tele_type)
 		.rae = true,
 		.ot = false,
 	};
+
+	ocp_get_uuid_index(dev, &args.uuidx);
 
 	err = nvme_get_log(&args);
 	if (err < 0)


ikegami-t:
@arthurshau The OCP telemetry command implemented by the PR #1896 commit 49ad948b3. By the way which OCP documentation version and section is the command specification based? If possible I would like to confirm to make sure if the following telemetry type none, host 0 and host 1 is really correct and if is there any changes the type host and controller from the NVMe base specification.
```c
enum TELEMETRY_TYPE {
	TELEMETRY_TYPE_NONE       = 0,
	TELEMETRY_TYPE_HOST       = 7,
	TELEMETRY_TYPE_CONTROLLER = 8,
	TELEMETRY_TYPE_HOST_0     = 9,
	TELEMETRY_TYPE_HOST_1     = 10,
};
```

ikegami-t:
Fix for the issue #2582.

HaroPanosyan:
> The index is required by the OCP specification.

I am trying to understand more about this change and it would be nice if you can post the OCP spec link and paragraph/page where this is specified.

I was looking at OCP 2.5 spec and it talks about in general for both log pages and get/set features. Maybe this is what you are referring to:
**UUID-4 The Get Log Page command shall support UUID Index functionality.**

But telemetry being base NVMe log, the device should support it always at "default" uuid index 0, hence passing OCP uuid index to get the telemetry log seems to me a bit odd. Also, if we are making this change, then every single base NVMe log should be doing the same, meaning passing OCP uuid index to get a base log. But like I said, I could be wrong and would be nice if someone can shed a light. 
Thanks.

ikegami-t:
Thanks for your review comments. Just confirmed again the OCP v2.6.2 section 4.4.3 (UUID-4, UUID-5) and 4.8.4 below then understood as mentioned the OCP telemetry log command not needed to set the UUID index since not the vendor specific log page but the standard log pages so let me close this PR later.
  <https://www.opencompute.org/documents/datacenter-nvme-ssd-specification-v2-6-2-pdf>
- 4.4.3 UUID for DSSD Specific Information
ocp-v2 6 2-s4 4 3
- UUID-4, UUID-5
ocp-v2 6 2-s4 4 3-uuid4-5
- 4.8.4 DSSD Log Page Requirements
ocp-v2 6 2-s4 8 4

By the way still it seems the OCP telemetry command not correct for me as below.
- 4.8.1 Standard Log Page Requirements: The OCP telemetry command types (log identifiers) 0 and 10 are not described. Also the telemetry types 7 and 8 are described but those are standard log page indentifiers so not OCP command.
ocp-v2 6 2-s4 8 1
- STD-LOG-17: The telemetry type (log identifier) 9 is described but it is not the telemetry type host 0 but the endurance group information.
ocp-v2 6 2-s4 8 1-std-log-17

So can we delete the OCP telemetry command from the OCP plugin feature since the nvme telemetry-log command is already supported and duplicated. How do you think about this? @arthurshau and @HaroPanosyan. Thank you.

ikegami-t:
About the UUID specification it is described by the NVMe base specification revision 2.1 as below so it seems that to set UUID index itself correct behavior I think (then this patch seems still not incorrect but the OCP API to get UUID index checking OCP UUID index only). @HaroPanosyan Do you have any comment or advice about this?
- 5.1.12 Get Log Page command
nvme-base-r2 1-s5 1 12
- 8.1.28.2 UUIDs for Vendor Specific Information Requirements
nvme-base-r2 1-s8 1 28 2

---

@arthurshau Thanks for your comment. Sorry I have just realized as the OCP telemetry command decords the telemetry data to print the dump but the nvme telemetry-log command does not but just writes the telemetry data into the output file specified so the OCP telemetry command looks useful basically. Still I am not sure about the telemetry type 0: none, 9: host 0  and 10: host 1 if correct so needed to study more.

HaroPanosyan:
Sorry guys, I tried to spend some time through the source code but could not figure out the details in short time. But if the code change in this PR is for (Log Identifier) LID 0xC9, then yes, we need to pass the UUID index for OCP. Any get/set feature or get log command needs to pass correct UUID index if the FID/LID >= 0xC0. If we are getting basic LID 7 or 8 for telemetry, then default 0 index should go to DWORD 14, and no need for OCP UUID Index. LID 0xC9 is "Telemetry String Log" required by OCP and we must pass correct UUID index for this log. Same for Feature Identifier C8h (Set Telemetry Profile).

I am also unable to understand what it means "OCP telemetry command". Could be referring to Set Telemetry Profile?

ikegami-t:
Thank you so much for your support. Let me close this PR. But still I will do study more below separately.
1. Support UUID for standard log page
2. Delete OCP telemetry command telemetry type 0: none, 9: host 0 and 10: host 1 
FYI: Just confirmed again the NVMe base specification revision 2.1 as some standard log page supported UUID but standard feature not supported UUID as below.
nvme-base-r2 1-s5 1 12-2
nvme-base-r2 1-s5 1 12-3
nvme-base-r2 1-s5 1 25

HaroPanosyan:
Thanks. The logs with 5 in the exponent (all below 0xC0) must have some vendor specific info, hence "Selection of UUID may be supported". For those logs, running generic NVMe command to dump binaries, users can specify anything they want through current CLI implementation, and I assume current implementation allows users to specify UUID index or DWORD 14. As for parsing, vendors can provide specific parsers in their plugins. For OCP specific Logs and Get/Set features, we are the vendor, it's all publicly documented and we have to pass correct OCP UUID index for both parsing and dumping. And OCP LIDs/FIDs are all in vendor specific space C0h to FFh. Hope there are no vendor specific fields for any of OCP logs or get/set features :). But if there are, then vendors can provide plugins.

PR #2605, opened 2024-12-06
ns_descs: prevent buffer overrun and memory corruption

maurizio-lombardi:
nsdescs should point to an allocated memory of size NVME_IDENTIFY_DATA_SIZE bytes, not "sizeof(*void)" bytes.


igaw:
Thanks!

PR #2608, opened 2024-12-09
nvme: add Namespace Zeroes support to Write Zeroes command

francispravin5:
Added the Namespace Zeroes (NSZ) support to Write Zeroes command. Also, fixed the dtype limitation (DTYPE - 22:20 bits).
TP4160 - WZSL Limit Modification

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -6900,6 +6900,7 @@ static int write_zeroes(int argc, char **argv, struct command *cmd, struct plugi
 	_cleanup_nvme_dev_ struct nvme_dev *dev = NULL;
 	__u8 sts = 0, pif = 0;
 	__u16 control = 0;
+	__u32 result = 0;
 	int err;
 
 	const char *desc =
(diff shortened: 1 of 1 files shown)


nvme.c
```diff
@@ -7030,15 +7037,22 @@ static int write_zeroes(int argc, char **argv, struct command *cmd, struct plugi
...
-	else
+	else {
 		printf("NVME Write Zeroes Success\n");
+		if (cfg.nsz) {
+			if (result & 0x1)
+				printf("All logical blocks in the entire namespace cleared to zero\n");
+			else
+				printf("%d logical blocks cleared to zero\n", cfg.block_count);
+		}
+	}
```
igaw, line 7055: I think it would be better to be silent on success and print the success message only if the `--verbose` flag is provided. So the usual unix tool behavior. Rest looks good.

francispravin5: @igaw Addressed the review comment.
Thanks!

igaw:
Thanks!

PR #2609, opened 2024-12-09
ocp: fix telemetry string log output file handling

Karthikmct:
ocp: fix telemetry string log output file handling

Fixed the code to avoid binary file creation by tool for ocp

Signed-off-by: Karthik <karthik.b82@samsung.com>

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -2502,10 +2502,7 @@ static int get_c9_log_page(struct nvme_dev *dev, char *format)
 		return ret;
 	}
 
-	if (fmt == BINARY)
-		ret = get_c9_log_page_data(dev, 0, 1);
-	else
-		ret = get_c9_log_page_data(dev, 0, 0);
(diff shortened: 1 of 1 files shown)


Karthikmct:
@igaw , This fix is addressing your comment on get_c9_log_page_data in  [https://github.com/linux-nvme/nvme-cli/pull/2586]

igaw:
This drops the support for the binary output into a file completely. I'd say it would be nice to get all the use cases properly working instead just dropping this feature. Someone might want to us it.

In `nvme.c` we have couple example how it could be done. `get_log` for example has the `--raw-output` argument. But that was introduces before we had the global `--output-format=binary`. Hmm, so I think the most canonical way would be to get `--output-format=binary` properly working for all commands in `ocp` and use the shell redirect to write into a file. For those commands which want to write the output into a file an additional command line argument could be added. I thinke we have this for the `submit_io` based command.

Anyway, my point is that I'd like to see the ocp plugin to adapt the same type of input/output handling pattern we have in the main code base. Having two different philosophy how to things is just confusing.

PR #2610, opened 2024-12-10
tests: various improvements

MaisenbacherD:
This PR introduces various improvements to the python test cases.

Before merging this commit, we would need a new self-hosted GitHub runner
instance assigned to this repository, as the device clean up routine is now
handled by the runner instance on the `ACTIONS_RUNNER_HOOK_JOB_COMPLETED`. 

The new self-hosted runner should be created and assigned to this repository via
a different channel, as this requires a token exchange.

### DIFF
--- a/.github/workflows/run-nightly-tests.yml
+++ b/.github/workflows/run-nightly-tests.yml
@@ -9,39 +9,25 @@ on:
 jobs:
   nightly-tests:
     runs-on: nvme-nvm
+    container:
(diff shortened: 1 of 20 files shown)


MaisenbacherD:
A new self-hosted GitHub runner instance was added. Thanks @igaw :)
The old instance was paused until the new one runs successful for a couple of days.

igaw:
Are these changes ready to merge?

igaw:
There we go. Thanks!

PR #2611, opened 2024-12-10
ocp: add ocp_get_log_simple() to set UUID index

ikegami-t:


ikegami-t:
Fix for the issue https://github.com/linux-nvme/nvme-cli/issues/2582.

igaw:
@sc108-lee @HaroPanosyan  any comments?

igaw:
Thanks!

PR #2614, opened 2024-12-11
nbft: Report proper error codes from discover_from_nbft()

tbzatek:
In case of a connection failure of any SSNS records that are not marked as 'unavailable', report proper error code so that nvme-cli callers may retry the operation.

The semantics of most functions is to return 0 zero for success and negative errno numbers in case of an error.


igaw:
Thanks!

PR #2616, opened 2024-12-12
build: bump libnvme wrap to pick up latest for Live Migration

NateThornton:


igaw:
As pointed out, usually you would add this type of dependency update directly into the PR which depends on it. I'll take it as there is no harm to do so.

PR #2617, opened 2024-12-12
print: OACS Host Managed Live Migration Support (HMLMS)

NateThornton:
Additional HMLMS bit in OACS field (stdout)


igaw:
Thanks!

PR #2619, opened 2024-12-13
nvme: plugins/fdp, check endurance group identifier

nj-shetty:
For FDP options such as stats and events, default endurance identifier 0 is used which results in error.
This patch returns error for missing endurance group identifier from user.


igaw:
Thanks!

PR #2621, opened 2024-12-13
ocp: delete internal-log command telemetry type none value definition

ikegami-t:
Since the type none value never set by the command handling.


igaw:
Thanks!

HaroPanosyan:
Thanks a lot, this clarified my understanding.

PR #2622, opened 2024-12-14
plugins: build plugins not depended on json without json

ikegami-t:
Some plugins do not use json so no need to check the build dependency.


igaw:
Thanks!

PR #2624, opened 2024-12-17
ocp-nvme: Add a different formatting for JSON output

### LINKED ISSUE #2577: Field descriptors in OCP SMART Log Page are different to those in standard SMART Log Page
There is a lack of consistency between the field descriptors in the standard SMART log page and the OCP SMART extended log page. This is rather annoying but also poses a bit of a challenge when trying to scrape the OCP metrics via a [Prometheus node-exporter text collector][ref-node-exporter].

I *think* it would be lovely to alter the format of the OCP output (both stdout and JSON) but I am concerned that might cause an issue for people who have already written tools that parse the existing log page output. So I am creating this issue so we can discuss if such a change is welcome or not.
```
$ nvme smart-log /dev/nvme0n1 -o json
{
  "critical_warning":0,
  "temperature":323,
<snip>
}
```
versus
```
$ nvme ocp smart-add-log /dev/nvme0n1 -o json
{
<snip>
  "Bad user nand blocks - Raw":0,
  "Bad user nand blocks - Normalized":0,
<snip>
}
```
[ref-node-exporter]: https://github.com/sbates130272/node-exporter-textfile-collector-scripts/tree/dev/stephen/nvme-ocp-update

sbates130272:
The current OCP JSON format for the SMART extended log page is not condusive to metric collection via tools like Prometheus. So we add a new output mode that uses all lower case and underscores (instead of spaces). This should help with metric collection. At the same time we clean up some of the field names. We add a new argument (--output-format-version) to allow us to select which output version we want. Documentation updated to reflect this change and mark this as experimental.

Fixes #2577.

### DIFF
--- a/Documentation/nvme-ocp-smart-add-log.txt
+++ b/Documentation/nvme-ocp-smart-add-log.txt
@@ -9,7 +9,7 @@ compliant device
 SYNOPSIS
 --------
 [verse]
-'nvme ocp smart-add-log' <device> [--output-format=<fmt> | -o <fmt>]
+'nvme ocp smart-add-log' <device> [--output-format=<fmt> | -o <fmt>] [--output-format-version=<version>]
 
 DESCRIPTION
 -----------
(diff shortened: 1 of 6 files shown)


igaw:
Fixed up the checkpatch complain. Thanks!

PR #2625, opened 2024-12-18
doc: Add write zeroes command namespace-zeroes option

francispravin5:
Add Namespace Zeroes (NSZ) option of Write Zeroes command.


igaw:
Thanks!

PR #2628, opened 2024-12-18
nvme: disable output-format option json value without json-c

ikegami-t:
Since only json print codes work with the json-c library.


igaw:
Great that is exactly what I had in mind!

PR #2629, opened 2024-12-19
dapustor: enable dapustor-nvme.c build without json-c dependencies ch…

ikegami-t:
…eckin

### DIFF
--- a/plugins/dapustor/dapustor-nvme.c
+++ b/plugins/dapustor/dapustor-nvme.c
@@ -96,6 +96,7 @@ struct nvme_extended_additional_smart_log {
 	struct nvme_additional_smart_log_item	inflight_write_io_cmd;
 };
 
+#ifdef CONFIG_JSONC
 static void show_dapustor_add_smart_log_jsn(struct nvme_additional_smart_log *smart,
 					    struct json_object *dev_stats)
 {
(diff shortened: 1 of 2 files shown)


plugins/dapustor/dapustor-nvme.c
```diff
@@ -526,7 +532,9 @@ static int dapustor_additional_smart_log(int argc, char **argv, struct command *
 	OPT_ARGS(opts) = {
 		OPT_UINT("namespace-id", 'n', &cfg.namespace_id, namespace),
 		OPT_FLAG("raw-binary",   'b', &cfg.raw_binary,   raw),
+#ifdef CONFIG_JSONC
 		OPT_FLAG("json",         'j', &cfg.json,         json),
+#endif /* CONFIG_JSONC */
```
igaw, line 537: just thinking loud: if we introduce something like

```
#ifdef CONFIG_JSONC
#define OPT_FLAG_JSON OPT_FLAG("json",         'j', &cfg.json,       "Output In JSON format")
#else
#define OPT_FLAG_JSON 
#endif
```
on a global level we don't need these ugly `ifdefs` everywhere.

igaw:
Thanks!

plugins/dapustor/dapustor-nvme.c
ikegami-t: Good suggestion. Fixed this by the PR #2637. Thank you.

PR #2630, opened 2024-12-19
dell: build plugin not depended on json without json

ikegami-t:
dell plugin does not use json so no need to check the build dependency.


igaw:
Thanks!

PR #2631, opened 2024-12-19
fdp: build plugin not depended on json without json

ikegami-t:
fdp plugin does not use json so no need to check the build dependency.


igaw:
Thanks!

PR #2632, opened 2024-12-19
huawei: enable huawei-nvme.c build without json-c dependencies checking

ikegami-t:


igaw:
Thanks!

PR #2633, opened 2024-12-19
intel: enable intel-nvme.c build without json-c dependencies checking

ikegami-t:

### DIFF
--- a/plugins/intel/intel-nvme.c
+++ b/plugins/intel/intel-nvme.c
@@ -74,6 +74,7 @@ struct nvme_vu_id_ctrl_field { /* CDR MR5 */
 	__u8			mic_fw[4];
 };
 
+#ifdef CONFIG_JSONC
 static void json_intel_id_ctrl(struct nvme_vu_id_ctrl_field *id,
 	char *health, char *bl, char *ww, char *mic_bl, char *mic_fw,
 	struct json_object *root)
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

HaroPanosyan:
Intel sold it's SSD business 3 years ago to SK Hynix and they started a new company called Solidigm which I work for. We have Solidigm plugin and our plan is to eventually deprecate Intel plugin. We are thinking of having warning message so anyone using Intel plugin to consider start using Solidigm plugin.
Any help/suggestion how we can properly do this would be very nice. Thanks.

ikegami-t:
At first seems better create an issue also for the @HaroPanosyan mention since the PR already merged. Sorry I am not sure about that but seems good to delete the plugin on newer version nvme-cli and then mentioed it by the release note cleary. (By the way for example the toshiba plugin looks as long time not maintained. Also Toshiba SSD business separated from Toshiba to Toshiba memory before then now change the company name also to Kioxia. So seems the plugin also can be deleted.)

HaroPanosyan:
Thanks. There could be still some users, so I was thinking instead of deleting, allow some time by just showing "soon to be deprecated, instead use Solidigm plugin" message or something like that.

ikegami-t:
Okay I see. By the way the intel plugin is older but the newer version solidigm plugin will support the same command feature. Is this really correct? Or is the command specification basically same for the old features on both intel and solidgm SSDs?

HaroPanosyan:
Correct, we already should cover everything intel plugin does and more. We have newer drives with newer Vendor Unique (VU) logs/features and we plan Solidigm plugin to be where everything goes and eventually deprecate/delete Intel plugin.

ikegami-t:
Noted. Thank you.

PR #2639, opened 2024-12-20
scaleflux: enable sfx-nvme.c build without json-c dependencies checking

ikegami-t:


ikegami-t:
Only remaining a plugin for solidigm to enable on muon build without json but it seems as difficult to do this since basically the plugin based on to use json for example to parse json configuration file. So at this moment all plugins excluding solidgm enabled to build without json.

igaw:
Looks good. Thanks for your effort!

PR #2640, opened 2024-12-22
ocp: replace to use json_object_* macros to obj_* macros

ikegami-t:
Since currently mixed to use them in ocp-print-json.c then unify.


igaw:
I think we should not export the `obj_add_*` definition from `nvme-print-json.c`. Also we should move the existing `obj_add_*` definitions out of nvme-print.h and add them to `util/json.h` and use the same prefix, e.g. `json_object_add_*`.

It's a lot of code churn and adds an additional indirection without a lot of gain.

ikegami-t:
Thanks for your review comments. Just fixed the patches as suggested.

igaw:
Looks good. Thanks!

PR #2641, opened 2024-12-29
Ocp lid fid

ikegami-t:


plugins/ocp/ocp-hardware-component-log.c
```diff
@@ -193,16 +196,27 @@ static int get_hwcomp_log_data(struct nvme_dev *dev, struct hwcomp_log *log)
...
+		log_bytes *= sizeof(__le32);
+
+	args.len = 0;
+	if (log_bytes > desc_offset)
+		args.len = log_bytes - desc_offset;
+
+	if (!args.len) {
+		print_info_error("error: ocp: invalid hwcomp log size bytes: %.0Lf\n", log_bytes);
+		return -EINVAL;
+	}
```
igaw, line 217: Usually it's better style to test and bail out directly, so

```
if (log_bytes <= desc_offset)
  return err;

args.len = log_bytes - desc_offset;
```

This keeps the logic in order, that is check if valid, then do the operation if it all is good.

ikegami-t: Yes right so just fixed the changes as mentioned and rebased with the latest master code. Thank you.

igaw:
Thanks!

PR #2642, opened 2024-12-31
nvme-print-json: update register print code to use libnvme definition

ikegami-t:
Use libnvme register definitions instead of the shift operator.


igaw:
Thanks!

PR #2643, opened 2025-01-02
Print stdout pel

ikegami-t:


nvme-print-stdout.c
```diff
@@ -239,7 +239,7 @@ static void stdout_add_bitmap(int i, __u8 seb)
 {
 	for (int bit = 0; bit < CHAR_BIT; bit++) {
 		if (nvme_pel_event_to_string(bit + i * CHAR_BIT)) {
-			if ((seb >> bit) & 0x1)
+			if (BIT_CHECK(seb, bit))
```
igaw, line 242: Let's leave as it is. The macros always hide what is happening and it means when you don't know what it does you need to look it up. The original code is not that bad :)

ikegami-t: Okay understood then delete the commit to use BIT_CHECK() macro and rebased the branch with the latest master. Thank you. (But still I am not sure what is good cases to use macros and bad cases.. Since already nvme-cli using macros NVME_GET() and NVME_SET(), etc. in libnvme and also in linux kernel there are macros BIT(), BIT_MASK() and GEN_MASK(), etc. so if those macros are not good seems needed to be deleted to unify.)

igaw: I hear you and yes, it's a bit of muddy water. My personal taste is to avoid macros if possible due to the mentioned arguments above. Adding indirection/abstraction can be hard to maintain. In the end it's also personal preference here.

Though when the the code gets difficult to follow or we want to keep all the definition in one place, it makes sense to have macros like  NVME_GET/NVME_SET around. 

For simple bit operations I think the C API is simple enough to read. 

BTW, the BIT_CHECK macro in the kernel has fans and haters in the kernel. So not all maintainers want to see them in their subsystem :)

nvme-print-stdout.c
```diff
@@ -246,15 +246,14 @@ static void stdout_add_bitmap(int i, __u8 seb)
 	}
 }
 
-static void stdout_persistent_event_log_fdp_events(unsigned int cdw11,
-						   unsigned int cdw12,
-						   unsigned char *buf)
+static void stdout_persistent_event_log_fdp_events(struct nvme_feat_fdp_events_cdw11 *cdw11,
+						   unsigned int cdw12, unsigned char *buf)
```
igaw, line 250: Uhh, didn't know there was `struct nvme_feat_fdp_events_cwd11` although I merged it...

I don't think we should have this struct around in the first place. This is a CWD thus we should have just SHIFTS and MASKS definition as we have for all other commands. And the you can use NVME_GET(cdw11, ...) as you do for the cwd12.

And please mark `struct nvme_feat_fdp_events_cwd11` as deprecated. 

@birkelund

ikegami-t: For the review comment just created the libnvme PR https://github.com/linux-nvme/libnvme/pull/940.

nvme-print-stdout.c
ikegami-t: Thank you so much for your explanation. Now I could understand much more before.

nvme-print-stdout.c
ikegami-t: Just fixed the patch to use NVME_GET instead. Also rebased with the PR #2650 changes since the libnvme bump required and the build warings errors caused without the PR changes. Thank you.

igaw:
Thanks!

PR #2647, opened 2025-01-06
ocp: Fixes for OCP 2.5 Telemetry Stats Parsing

jeff-lien-wdc:
The OCP 2.5 Telemetry log statistics parsing code was not checking for the end of the stats data and therefore getting stuck in an infinite while loop.
Also added code to parse out the statistic specific data for the Max Die, Max NAND Channel, and Min NAND Channel Bad Block.

### DIFF
--- a/plugins/ocp/ocp-telemetry-decode.c
+++ b/plugins/ocp/ocp-telemetry-decode.c
@@ -1204,9 +1204,14 @@ int parse_statistic(struct nvme_ocp_telemetry_statistic_descriptor *pstatistic_e
 		    struct json_object *pstats_array, FILE *fp)
 {
 	if (pstatistic_entry == NULL) {
-		nvme_show_error("Input buffer was NULL");
+		nvme_show_error("Statistics Input buffer was NULL");
 		return -1;
 	}
+
+	if (pstatistic_entry->statistic_id == STATISTICS_RESERVED_ID)
+		/* End of statistics entries, return -1 to stop processing the buffer */
+		return -1;
+
 
 	unsigned int data_size = pstatistic_entry->statistic_data_size * SIZE_OF_DWORD;
 	__u8 *pdata = (__u8 *)pstatistic_entry +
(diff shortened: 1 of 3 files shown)


plugins/ocp/ocp-telemetry-decode.c
```diff
@@ -1204,10 +1204,14 @@ int parse_statistic(struct nvme_ocp_telemetry_statistic_descriptor *pstatistic_e
 		    struct json_object *pstats_array, FILE *fp)
 {
 	if (pstatistic_entry == NULL) {
-		nvme_show_error("Input buffer was NULL");
+		nvme_show_error("Statistics Input buffer was NULL");
 		return -1;
 	}
 
+	if (pstatistic_entry->statistic_id == STATISTICS_RESERVED_ID) {
+		return -2;
```
igaw, line 1212: Do these negative numbers have meaning? Then yous error codes instead, otherwise just return -1. Also braces are not necessary for single statement blocks.

plugins/ocp/ocp-telemetry-decode.c
```diff
@@ -1236,8 +1240,25 @@ int parse_statistic(struct nvme_ocp_telemetry_statistic_descriptor *pstatistic_e
 			pstatistic_entry->statistic_data_size);
 		json_add_formatted_u32_str(pstatistics_object, STR_RESERVED,
 			pstatistic_entry->reserved);
-		json_add_formatted_var_size_str(pstatistics_object, STR_STATISTICS_SPECIFIC_DATA,
-			pdata, data_size);
+		if (pstatistic_entry->statistic_id == MAX_DIE_BAD_BLOCK_ID) {
+			json_add_formatted_u32_str(pstatistics_object, STR_STATISTICS_WORST_DIE_PERCENT,
```
igaw, line 1244: just break the overlong line. 100 chars are wide enough for normal code. If the code block is hitting it all the time, it's usually a good indication that there is something wrong.

plugins/ocp/ocp-telemetry-decode.c
```diff
@@ -1236,8 +1240,25 @@ int parse_statistic(struct nvme_ocp_telemetry_statistic_descriptor *pstatistic_e
...
+					*(__u16 *)&pdata[2]);
+		} else if (pstatistic_entry->statistic_id == MIN_NAND_CHANNEL_BAD_BLOCK_ID) {
+			json_add_formatted_u32_str(pstatistics_object, STR_STATISTICS_BEST_NAND_CHANNEL_PERCENT,
+					pdata[0]);
+			json_add_formatted_u32_str(pstatistics_object, STR_STATISTICS_BEST_NAND_CHANNEL_RAW,
+					*(__u16 *)&pdata[2]);
+		} else {
+			json_add_formatted_var_size_str(pstatistics_object, STR_STATISTICS_SPECIFIC_DATA,
+						pdata, data_size);
+		}
```
igaw, line 1261: lot's of overlong lines.

plugins/ocp/ocp-telemetry-decode.c
jeff-lien-wdc: fixed long code lines.  
@igaw Should I squash these changes into 1 commit?

plugins/ocp/ocp-telemetry-decode.c
jeff-lien-wdc: Fixed overlong lines.

PR #2653, opened 2025-01-09
nvme: add the new Self-test Code of Device Self-test command

francispravin5:
Add the new Self-test Code (SCT) Host-Initiated Refresh operation support to Device Self-test command.
Also, add it in the Device Self-test Log Page.
TP4058 - Environmental Extremes Management.


igaw:
Thanks!

PR #2654, opened 2025-01-09
nvme-print-stdout: add human readable output for DSTO field

francispravin5:
Add human readable output for DSTO field of Identify Controller Data Structure.


igaw:
Thanks!

PR #2656, opened 2025-01-11
ocp-telemetry-decode: fix ascii string lookup

plmwd:
Lookup ascii string in VU Event Identifier String Table for vendor unique event classes. Fix not looking up spec-defined statistic id name. Fix typo ('static' => 'statistic').

### DIFF
--- a/plugins/ocp/ocp-telemetry-decode.c
+++ b/plugins/ocp/ocp-telemetry-decode.c
@@ -489,9 +489,9 @@ int get_telemetry_das_offset_and_size(
 	return 0;
 }
 
-int get_static_id_ascii_string(int identifier, char *description)
+int get_statistic_id_ascii_string(int identifier, char *description)
 {
-	if (pstring_buffer == NULL)
+	if (!pstring_buffer || !description)
 		return -1;
 
 	struct nvme_ocp_telemetry_string_header *pocp_ts_header =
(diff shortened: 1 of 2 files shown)


plugins/ocp/ocp-telemetry-decode.c
```diff
@@ -522,16 +522,18 @@ int get_static_id_ascii_string(int identifier, char *description)
...
-				       statistic_identifiers_map[identifier].description,
-				       peach_statistic_entry->ascii_id_length + 1);
 			return 0;
 		}
 	}
 
+	// If ASCII string isn't found, see in our internal Map
+	// for 2.5 Spec defined strings
+	if (identifier <= 0x1D)
+	{
```
igaw, line 532: Move the opening braket to the condition, e.g.

  `if (...) {`

and why don't we need to check for `!description` anymore?

plmwd: I believe removed it because 'description' isn't checked on line 522. However, you're right, it should be checked. I'll add a NULL check at the top of the function similar to 'pdescription'.

plmwd:
Hi @igaw, could you please check?

plmwd:
Fixed the issue and squashed. Can you please check @igaw?

igaw:
Thanks!

PR #2659, opened 2025-01-13
nvme: add mgmt-addr-list-log command

ikegami-t:
Since added the NVMe 2.1 log page.

PR #2660, opened 2025-01-13
Mal doc comp

ikegami-t:


ikegami-t:
The tab completion and the documentation updates for the command added by the PR #2659 and some exsiting tab compliton and documentation errors fixed for the ocp hardware-component-log command. Thank you.

igaw:
thanks!

PR #2664, opened 2025-01-18
nvme-print-stdout: use NVME_CTRL_OACS definitions

ikegami-t:
Still needed to add some remaining definitions in libnvme.


igaw:
Thanks!

Issue #2662, opened 2025-01-15
nvme flush command returns "invalid argument" when given -n options with nvme-cli 2.11, 2.10.2 and 1.16 


sukhi61166:
We are connected to our target over fabrics either via RoCEv2 and TCP protocol. We observed that the nvme flush command is not working as expected. It doesnt take -n 0xFFFFFFFF (even if the drive supports it) or -n 1 etc. options and returns invalid argument. If we pass individual name spaces then it works like /dev/nvme1n1. 

```
# nvme --version
    nvme version 2.10.2 (git 2.10.2)
    libnvme version 1.10 (git 1.10)
# nvme --version
nvme version 2.11 (git 2.11+)
libnvme version 1.11.1 (git 1.11.1)

Kernel : 6.8.12-u6.8.0-50.51
OS: Ubuntu 24.04 

$ sudo nvme list-ns /dev/nvme1
 [   0]:0x1
 [   1]:0x2

$sudo nvme id-ctrl /dev/nvme1 -H | grep Flush
[2:1] : 0x3      The Flush command supports NSID set to FFFFFFFFh

$ sudo nvme flush /dev/nvme1n1 -vv
opcode       : 00
flags        : 00
rsvd1        : 0000
nsid         : 00000001
cdw2         : 00000000
cdw3         : 00000000
data_len     : 00000000
metadata_len : 00000000
addr         : 0
metadata     : 0
cdw10        : 00000000
cdw11        : 00000000
cdw12        : 00000000
cdw13        : 00000000
cdw14        : 00000000
cdw15        : 00000000
timeout_ms   : 00000000
result       : 00000000
err          : 0
latency      : 171 us
NVMe Flush: success

$sudo nvme flush /dev/nvme1n2
NVMe Flush: success

$# sudo nvme flush /dev/nvme1 -n 0xFFFFFFFF -vv
opcode       : 00
flags        : 00
rsvd1        : 0000
nsid         : ffffffff
cdw2         : 00000000
cdw3         : 00000000
data_len     : 00000000
metadata_len : 00000000
addr         : 0
metadata     : 0
cdw10        : 00000000
cdw11        : 00000000
cdw12        : 00000000
cdw13        : 00000000
cdw14        : 00000000
cdw15        : 00000000
timeout_ms   : 00000000
result       : 00000000
err          : -1
latency      : 591 us
flush: Invalid argument
$# sudo nvme flush /dev/nvme1 -n 1 -vv
opcode       : 00
[...]

PR #2665, opened 2025-01-19
nvme: use blkdev for direct if namespace-id specified

ikegami-t:
Since IO commands rejected by the chardev if more than one namespace.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -299,7 +299,48 @@ static bool is_blkdev(struct nvme_dev *dev)
 	return S_ISBLK(dev->direct.stat.st_mode);
 }
 
-static int open_dev_direct(struct nvme_dev **devp, char *devstr, int flags)
+static bool is_nvme_dev(char *dev)
+{
+	int instance;
+	int head_instance;
+
+	return sscanf(basename(dev), "nvme%dn%d", &instance, &head_instance) == 1;
+}
+
+static __u32 get_nsid(struct argconfig_commandline_options *opts)
+{
+	__u32 *nsid = argconfig_get_value(opts, "namespace-id");
+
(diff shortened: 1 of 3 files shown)


ikegami-t:
@sukhi61166 If possible could you please test the PR chagnes since the changes for the issue #2662 but locally tested as the code itself works correctly but I do not have the device has more than one namespace for now?

nvme.c
```diff
@@ -7415,6 +7415,23 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
...
+		nvme_show_status(err);
+		return;
+	}
+
+	nvme_show_init();
+
+	nvme_show_error("flush: %s", nvme_strerror(errno));
+
+	if (is_chardev(dev))
+		nvme_show_result("chardev handled so needed to use blkdev instead");
```
igaw, line 7430: Maybe add something like "chardev handled so needed to use blkdev instead, e.g. /dev/nvme0n1" ?

ikegami-t: Changed the message as suggested. Thank you.

sukhi61166:
> @sukhi61166 If possible could you please test the PR chagnes since the changes for the issue #2662 but locally tested as the code itself works correctly but I do not have the device has more than one namespace for now?

I will checkout the latest and try it. Thanks

ikegami-t:
@sukhi61166 Unfortunately the PR changes reverted by the commmit e62ab3c for the issue #2668 introduced by the PR changes and also for the changes causing complicated as mentioned by the PR comment https://github.com/linux-nvme/nvme-cli/pull/2693#issuecomment-2648784345. So for now still we need to specify the blkdev for the flush command instead of the chardev mentioned by the issue #2662.

PR #2669, opened 2025-01-21
Print id ns

francispravin5:
Print the new fields of Identify Namespace Data Structure added in NVM Command Set Specification 1.1. Also, arranged few json prints to print based on the order.


francispravin5:
@igaw Could you please review this PR.
Thanks!

igaw:
Sorry for the delay. A bit busy :)

Thanks!

PR #2670, opened 2025-01-22
nvme: set eds to true if controller supports 128 bit hostid

bgurney-rh:
A controller that uses a 128-bit Host Identifier may result in the "nvme resv-report" command failing with a "Host Identifier Inconsistent Format" error that suggests the "simultaneous use of 64-bit and 128-bit Host Identifier values on different controllers".

This error can be avoided if the "--eds" option is used, to request the extended data structure.  However, the controller's ctratt value indicates whether the Host Identifier is 64 bits or 128 bits. Therefore, check the ctratt flag, and set eds to true if the controller indicates a 128-bit Host Identifier.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -7740,6 +7740,7 @@ static int resv_report(int argc, char **argv, struct command *cmd, struct plugin
 	const char *eds = "request extended data structure";
 
 	_cleanup_free_ struct nvme_resv_status *status = NULL;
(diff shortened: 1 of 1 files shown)


igaw:
Just one remark for my future me. The arg parser is not able to express bool command options to false. Thus the user can't explicitly disable it but explicitly enable `eds`. With this change if the controller supports 128 bits identifier it will always pick this one. I don't thinks this is a real issue.

igaw:
Thanks!

Issue #2668, opened 2025-01-21
attach/detach command line interface changes broke nightly tests


igaw:
It appears  https://github.com/linux-nvme/nvme-cli/pull/2665 breaks our nightly builds:

https://github.com/linux-nvme/nvme-cli/actions/runs/12878797474/job/35905289626#step:8:761

@ikegami-t  could you have a look?

BTW, be careful running these tests locally. These tests do write to the local hardware, e.g. /dev/nvme0n1

ikegami-t:
Just confirmed the log then not sure why the device /dev/nvme1 failed to detach as not botu chardev and blkdev and also /dev/nvme1 NS1 not failed but /dev/nvme1 NS2 failed but looks this not caused by the nvme-cli bug I think.
1. Build and install nvme-cli: nvme-cli 2.11 built and tested
<https://github.com/linux-nvme/nvme-cli/actions/runs/12878797474/job/35905289626#step:6:134>
``` sh
  Configuration
    git version   : "1.11.1-41-ga9f667a"

nvme-cli 2.11

  Paths
    prefixdir         : /usr/local
```
2. Run on device tests: Failed on detaching NS and also seems failed to detach NS2
<https://github.com/linux-nvme/nvme-cli/actions/runs/12878797474/job/35905289626#step:8:616>
```sh
##### Creating 109
##### Attaching 109
##### Running IOs in 109
Testcase main ... /dev/nvme1 is not a block or character device
Usage: nvme detach-ns <device> [OPTIONS]

Detach the given namespace from the given controller; de-activates the given
namespace's ID. A namespace must be attached to a controller before IO
commands may be directed to that namespace.

Options:
  [  --verbose, -v ]                    --- Increase output verbosity
  [  --output-format=<FMT>, -o <FMT> ]  --- Output format: normal|json|binary
  [  --namespace-id=<NUM>, -n <NUM> ]   --- namespace to attach
  [  --controllers=<LIST>, -c <LIST> ]  --- optional comma-sep controller id
                                            list
  [  --timeout=<NUM>, -t <NUM> ]        --- timeout value, in milliseconds
FAIL
...
##### Detaching 1
#### Deleting 1
##### Detaching 2

teardown: ctrl: /dev/nvme1, ns1: /dev/nvme1n1, default_nsid: 1, flbas: 0
```
3. Code: the error messege is output as the device /dev/nvme1 not both chardev and blkdev
[...]

PR #2671, opened 2025-01-22
nvme: restrict to use blkdev only for IO commands with NSID

ikegami-t:
Since NS detach command caused unexpected error by the blkdev.


ikegami-t:
The changes are to avoit the issue #2668.

PR #2672, opened 2025-01-22
nvmf-keys: udev rule should be using /etc/nvme

cleech:
Append nvme to @SYSCONFDIR@ in udev rule.

I'd think the TLS keyfile should be located at /etc/nvme/tls-keys and not just /etc/tls-keys.


igaw:
Thanks!

PR #2673, opened 2025-01-23
Nvme tests ns

ikegami-t:


ikegami-t:
The changes related to the issue #2668 but still the issue cause not sure and the blkdev checking fix may detect the cause. Also add verbose debugging option for NS management commands.

tests/nvme_test.py
```diff
@@ -392,6 +392,11 @@ def create_and_validate_ns(self, nsid, nsze, ncap, flbas, dps):
             err = subprocess.call(id_ns_cmd,
                                   shell=True,
                                   stdout=subprocess.DEVNULL)
+        if err == 0:
```
igaw, line 395: I suppose this should be indented, only test the return value from the `subprocess.call`?

tests/nvme_test.py
```diff
@@ -371,7 +371,7 @@ def create_ns(self, nsze, ncap, flbas, dps):
         """
         create_ns_cmd = f"{self.nvme_bin} create-ns {self.ctrl} " + \
             f"--nsze={str(nsze)} --ncap={str(ncap)} --flbas={str(flbas)} " + \
-            f"--dps={str(dps)}"
+            f"--dps={str(dps)} -v"
```
igaw, line 374: I don't know if we want the spam the output. Usually this stuff works just fine. I let @MaisenbacherD decide :)

tests/nvme_test.py
ikegami-t: Fixed the indentation. Thank you.

tests/nvme_test.py
ikegami-t: Understood. Thank you.

ikegami-t: Right so fixed to the long version option. Thank you.

MaisenbacherD (approved):
Looks good to me :)

PR #2675, opened 2025-01-25
nvme: add NVMe 2.1 changed-alloc-ns-list-log command

ikegami-t:
The command functions combined with the changed-ns-list-log command. The output almost same but different as lists attached or allocated NSs.


ikegami-t:
Note: Just realized to add the alloc option into the existing the changed-ns-list-log command instead of to add the changed-alloc-ns-list-log command.

nvme-wrap.c
```diff
@@ -208,12 +208,15 @@ int nvme_cli_get_log_fw_slot(struct nvme_dev *dev, bool rae,
 	return do_admin_op(get_log_fw_slot, dev, rae, fw_log);
 }
 
-int nvme_cli_get_log_changed_ns_list(struct nvme_dev *dev, bool rae,
-				     struct nvme_ns_list *ns_log)
+int nvme_cli_get_log_changed_ns_list(struct nvme_dev *dev, bool rae, __u32 len,
+				     struct nvme_ns_list *ns_log, bool alloc)
 {
+	if (alloc)
+		return nvme_get_log_changed_alloc_ns_list(dev->direct.fd, rae, len, ns_log);
```
igaw, line 215: I suppose the `do_admin_op` wrapper should be used here as well. Do we miss an API update in libnvme-mi?

nvme-print-json.c
```diff
@@ -591,10 +590,16 @@ void json_changed_ns_list_log(struct nvme_ns_list *log,
...
+
 	if (log->ns[0] == cpu_to_le32(0xffffffff))
 		return;
 
-	obj_add_str(r, "Changed Namespace List Log", devname);
+	if (asprintf(&k, "Changed %s Namespace List Log", alloc ? "Allocated" : "Attached") < 0)
+		k = NULL;
+
+	if (k)
+		obj_add_str(r, k, devname);
```
igaw, line 602: What about 

```c
if (asprintf(...) > 0)
    obj_add_str(...)
```
 ?

nvme-wrap.c
ikegami-t: Yes sorry for the missing so will update libnvme-mi to add APIs later. Thank you.

ikegami-t:
Rebased and fixed the conflict errors.

nvme-wrap.c
igaw: Please introduce a new wrapper function for `nvme_get_log_changed_alloc_ns_list`. These function here should really only switch between the standard API and MI API interfaces. There should be no logic besides this. Note, these wrappers will go away when we start with libnvme 2.x hopefully soonish.

ikegami-t: Understood and noted then fixed the changes as so. Thank you.

PR #2678, opened 2025-01-27
fabrics: make 'disconnect' accepting only -n or -d

hreinecke:
There is limited usability in specifying both -n and -d, so make both exclusive.


igaw:
Thanks!

PR #2680, opened 2025-01-28
Revert "tests: add blkdev checking after NS creation"

ikegami-t:
This reverts commit c276f3ddd780c55458e509b74d80633728828950.


igaw:
Thanks!

PR #2683, opened 2025-01-30
completions: add rotational-media-info-log command

ikegami-t:
Both bash and zsh completions updated for the command.


igaw:
Thanks!

PR #2684, opened 2025-01-31
udev-rules-ontap: switch to queue-depth iopolicy

martin-gpy:
The queue_depth based iopolicy suits ONTAP NVMe controllers better. So switch to the same.


igaw:
Thanks!

PR #2685, opened 2025-01-31
sed: add '--read-only' to lock/unlock commands

gjoyce-ibm:
This change allows a user to specify that a drive is to be locked read-only rather than just read-write. Also a drive that is locked read-write can be "unlocked" to locked read-only.

The flag '--read-only' is used so the locking type can be specified.


igaw:
Thanks!

PR #2687, opened 2025-02-02
Json key val

ikegami-t:


ikegami-t:
This is to check the NSID and NS if created correctly by the TP as mentioned by the issue comment https://github.com/linux-nvme/nvme-cli/issues/2668#issuecomment-2610212016.

igaw:
Sorry you lost me here. What's the purpose with these changes?

ikegami-t:
No problem. The purpose is to output the json nsid key and value as below then I think that the TP can check the NS if created correctly. Thank you.
```sh
tokunori@tokunori-desktop:~/nvme-cli$ nvme-build create-ns /dev/nvme0 -s -1 -c 1 -f 0 -o json
{
  "create-ns":"Success, created",
  "nsid":"1"
}
```

ikegami-t:
Rebased the commits and added the tests changes also to check the create-ns nsid value.

nvme-print-json.c
```diff
@@ -4640,6 +4640,21 @@ static void json_output_perror(const char *msg)
 	json_output_object(r);
 }
 
+static void json_output_key_value(bool space, bool eol, const char *key, const char *val,
```
igaw, line 4643: The two arguments `space` and `eol` should not be exposed to the user here. First, having function with several arguments using bools are always hard to read, secondly, this leaks the output layer features to the users side.

nvme.c
```diff
@@ -2822,26 +2822,42 @@ static int id_endurance_grp_list(int argc, char **argv, struct command *cmd,
...
-	ctrl = nvme_alloc(sizeof(*ctrl));
-	if (!ctrl)
-		return;
-
-	err = nvme_cli_identify_ctrl(dev, ctrl);
 	if (!err) {
-		if (!(le16_to_cpu(ctrl->oacs) & NVME_CTRL_OACS_NS_MGMT))
+		nvme_show_key_value(true, false, name, "Success, %s",
+				    !strcmp(cmd, "create") ? "created" :
+				    !strcmp(cmd, "delete") ? "deleted" : "");
```
igaw, line 2856: strcmp `cmd` seems a bit excessive, why not just write 'create command was sucessful' or something along the line. or just 'success'. `nvme create-ns` -> `success` should be good enough.

igaw:
The nightly builds are currently not running. I'd like to get them working again before merging this because of the last patch.

ikegami-t:
Fixed the changes as mentioned and also noted about the nightly builds situation. Thank you.

igaw:
FYI, @MaisenbacherD is updating/re-configuring the nightly CI runner environment. As all projects, it takes a bit longer than expected :)

ikegami-t:
Thank you. Noted it.

igaw:
I"ve rebased the series and refactored `is_ns_mgmt_support` slightly. Let's see what the nighly tests show.

igaw:
Nightly builds passed. So all good. Thanks everyone!

PR #2688, opened 2025-02-02
nvme-wrap: use rotational media information log libnvme-mi API

ikegami-t:
Since the libnvme-mi missed to add but will be added then use it.


ikegami-t:
This change depended on the libnvme PR https://github.com/linux-nvme/libnvme/pull/950 changes.

ikegami-t:
Add a commit to bump libnvme.

igaw:
I had to bump it another time, just merged a conflicting PR. No worries, that happens sometimes. 

Thanks!

PR #2690, opened 2025-02-05
ocp: fix to free hwcomp log desc memory allocated

ikegami-t:
Since the desc pointer used _cleanup_free_ only set NULL value.


plugins/ocp/ocp-hardware-component-log.c
```diff
@@ -256,15 +254,16 @@ static int get_hwcomp_log(struct nvme_dev *dev, __u32 id, bool list)
 	}
 
 	ret = get_hwcomp_log_data(dev, &log);
-	if (ret) {
+	if (ret)
 		print_info_error("error: ocp: failed get hwcomp log: %02X data, ret: %d\n",
 				 OCP_LID_HWCOMP, ret);
-		return ret;
```
igaw, line 262: if the command fails, `get_hwcomp_log_data` is in charge to free any allocated buffer. That means here the function should just return and after `osp_show_hwcomp_log` just always free the buffer. Note, `free` has a built in NULL check, so `if (ptr) free(ptr)` is not necessary anymore.

ikegami-t: The function nvme_get_log_page() called in get_hwcomp_log_data() but if failed the call then the allocated log->desc not freed so needed to free by the caller or the error return point.
```c
static int get_hwcomp_log_data(struct nvme_dev *dev, struct hwcomp_log *log)
{
...
	log->desc = calloc(1, args.len);
	if (!log->desc) {
		fprintf(stderr, "error: ocp: calloc: %s\n", strerror(errno));
		return -1;
	}

	args.log = log->desc,
	args.lpo = desc_offset,

#ifdef HWCOMP_DUMMY
	memcpy(log->desc, &hwcomp_dummy[desc_offset], args.len);
#else /* HWCOMP_DUMMY */
	ret = nvme_get_log_page(dev_fd(dev), NVME_LOG_PAGE_PDU_SIZE, &args);
	if (ret) {
		print_info_error("error: ocp: failed to get log page (hwcomp: %02X, ret: %d)\n",
				 OCP_LID_HWCOMP, ret);
		return ret;
	}
#endif /* HWCOMP_DUMMY */
````
> Note, free has a built in NULL check, so if (ptr) free(ptr) is not necessary anymore.
That is right. Thank you.

Comment (igaw) [reply plugins/ocp/ocp-hardware-component-log.c:262]: A well behaving function  should not change anything on error. So if a function allocates something and then fails later on the function is responsible to free the allocated resources. On the other hand if the function was successful and returns some resources to the caller, the caller is responsible for the resource now. 

This makes the error handling on the caller side way more sane.

In the above code path, the correct thing is 

```c

	ret = nvme_get_log_page(dev_fd(dev), NVME_LOG_PAGE_PDU_SIZE, &args);
	if (ret) {
		print_info_error("error: ocp: failed to get log page (hwcomp: %02X, ret: %d)\n",
				 OCP_LID_HWCOMP, ret);
		free(log->desc);
		return ret;
	}
```

igaw:
I've updated `get_hwcomp_log_data` so it frees the allocated buffer on error. So the only thing left is to free the buffer in the caller.

igaw:
Thanks!

plugins/ocp/ocp-hardware-component-log.c
ikegami-t: Understood well. Thank you for your advice and update the patch.

PR #2691, opened 2025-02-06
nvme: add dispersed-ns-participating-nss-log command

ikegami-t:
Since added the NVMe 2.1 log page.

### DIFF
--- a/Documentation/nvme-dispersed-ns-participating-nss-log.txt
+++ b/Documentation/nvme-dispersed-ns-participating-nss-log.txt
@@ -1,0 +1,55 @@
+nvme-dispersed-ns-participating-nss-log(1)
+==========================================
+
+NAME
+----
+nvme-dispersed-ns-participating-nss-log - Retrieve Dispersed Namespace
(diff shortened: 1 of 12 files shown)


ikegami-t:
The changes depended on the libnvme PR https://github.com/linux-nvme/libnvme/pull/950 changes.

keithbusch:
does anyone actually implement this log? Linux doesn't support the feature, so even if a device does support it, I don't think you get consistent enumeration in Linux.

hreinecke:
> does anyone actually implement this log? Linux doesn't support the feature, so even if a device does support it, I don't think you get consistent enumeration in Linux.

I have been frequently told that this is the case; most of the major storage vendors (claim to) support it. Not that we've actually ever tested this, being as it is a quite contentious topic with Linux.

ikegami-t:
Should we not added the log command for now as not supported the feature by Linux?

ikegami-t:
Just rebased and added a documentation and completions.

igaw:
Thanks!

PR #2693, opened 2025-02-07
Revert attach/detach command line interface changes

igaw:
These changes broke the nightly CI runs. We were unable to figure out in time what it causes the fail, thus revert these changes.
 
Introduced here with #2665 to address #2662

Fixes: #2668

### DIFF
--- a/.github/workflows/run-nightly-tests.yml
+++ b/.github/workflows/run-nightly-tests.yml
@@ -16,7 +16,7 @@ jobs:
       #BDEV0 is an environment variable of the self-hosted runner instance
       #that contains a valid nvme ctrl name which is capable of the nvm
       #command set.
-      options: '--privileged -e BDEV0'
+      options: '--privileged -v "/dev":"/dev":z -e BDEV0'
     steps:
       - name: Output kernel version
         run: |
--- a/nvme.c
+++ b/nvme.c
@@ -7465,23 +7465,6 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 	return err;
 }
 
-static void io_cmd_show_error(struct nvme_dev *dev, int err, const char *cmd)
-{
-	if (err > 0) {
-		nvme_show_status(err);
-		return;
-	}
-
-	nvme_show_init();
-
-	nvme_show_error("%s: %s", cmd, nvme_strerror(errno));
-
-	if (is_chardev(dev))
-		nvme_show_result("char device provided but blkdev is needed, e.g. /dev/nvme0n1");
-
-	nvme_show_finish();
-}
-
 static int flush_cmd(int argc, char **argv, struct command *cmd, struct plugin *plugin)
 {
 	const char *desc = "Commit data and metadata associated with\n"
@@ -7517,8 +7500,10 @@ static int flush_cmd(int argc, char **argv, struct command *cmd, struct plugin *
 	}
 
 	err = nvme_flush(dev_fd(dev), cfg.namespace_id);
-	if (err)
-		io_cmd_show_error(dev, err, "flush");
+	if (err < 0)
+		nvme_show_error("flush: %s", nvme_strerror(errno));
+	else if (err != 0)
+		nvme_show_status(err);
 	else
 		printf("NVMe Flush: success\n");
 
--- a/nvme.c
+++ b/nvme.c
@@ -300,48 +300,7 @@ static bool is_blkdev(struct nvme_dev *dev)
 	return S_ISBLK(dev->direct.stat.st_mode);
 }
 
-static bool is_nvme_dev(char *dev)
-{
-	int instance;
-	int head_instance;
-
-	return sscanf(basename(dev), "nvme%dn%d", &instance, &head_instance) == 1;
-}
-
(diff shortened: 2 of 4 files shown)


igaw:
The nightly CI still fails: https://github.com/linux-nvme/nvme-cli/actions/runs/13201431769

I've checked if we don't hunt a ghost, that is around 3  weeks ago the 6.14 bits have been merged into Linus' tree. I thought we are following this tree. Anyway, I didn't spot any changes which could explain this problem.

ikegami-t:
Agreed. Thank you.

igaw:
Something is a still off. I hoped a revert would help but the test run of the nightly CI doesn't pass.

MaisenbacherD:
> Can we retest the CI with the commit [cccefc4](https://github.com/linux-nvme/nvme-cli/commit/cccefc48a9aca3c0b3f0f2a2a0fe5d01ad9eb59a) that pushed before the blkdev changes and also the TP changes not applied?
> 
> ```shell
> tokunori@tokunori-desktop:~/nvme-cli$ git log --oneline -2 565b30e2
> 565b30e2 nvme: use blkdev for direct if namespace-id specified
> cccefc48 argconfig: add argconfig_get_value()
> ```

I just tested this, but the `nvme_create_max_ns_test` fails starting on commit 565b30e2. I am also getting a test fail for the revert branch on Daniel's fork. Already looking into this issue and see what is going on :)

ikegami-t:
Noted so looks the test fail on the revert branch caused by the TP changes. Thank you.

MaisenbacherD:
After looking into this issue, it turns out that 565b30e (nvme: use blkdev for direct if namespace-id specified)
initially caused the issue discovered by `nvme_create_max_ns_test`.
With this commit, devices are opened by the namespace (blkdev) if already existing rather the controller (chardev) and assigns it to the `nvme_dev` struct.

When trying to detach(/attach) a new namespace, `nvme_attach_ns` errors out if the given nvme_dev is a blkdev:
https://github.com/linux-nvme/nvme-cli/blob/724836d14ac7fcfdf3f7fa7ed30d8c36008667bc/nvme.c#L2955-L2959

So reverting this change also makes sense to me until we come up with a version that does not break this scenario.
@ikegami could you maybe explain what you tried to fix with commit 565b30e? :)

Reverting just the commits of https://github.com/linux-nvme/nvme-cli/pull/2665 was not enough to let the tests pass as the test cases were also modified, which uncovered some more errors.
f54f046e2a8c1393bcf5d843280769ef43700d14 introduced checking for the correct ns after attaching and was refined in 56cf51b1440b88182f08e1bbe9d07acc38404868.
Previously we only checked for the presence of n1 also when attaching namespaces > 1:
https://github.com/linux-nvme/nvme-cli/blob/1024886e0ddce829f7243d2f61e7b0e77513b137/tests/nvme_test.py#L407-L414
Somehow the block device for ns > 1 does not appear in the device tree waiting for it after attaching it to the controller.
I will still look a bit further into this issue as I have not found the solution yet. :)

ikegami-t:
> So reverting this change also makes sense to me until we come up with a version that does not break this scenario. 

Yes I can agree with the revert.

> @ikegami could you maybe explain what you tried to fix with commit [565b30e](https://github.com/linux-nvme/nvme-cli/commit/565b30e285934502e9fcaaf7d9b8bd81e5011d0f)? :)

The commit was to resolve the issue #2662 by using the blkdev instead of the chardev.

Note: The PR #2671 changes are to revert the blkdev changes partially to use the blkdev only for IO commands. If you have any time or chance to retest the PR changes also as if the test fail resolved it will be helpful for me to investigate the cause. But sorry not sure if the changes can be merged actually even if the works correctly.

igaw:
The whole open device logic is a convoluted and very hard to understand. I think it's time to redesign it. That's why I haven't pulled #2671, it just makes it even more complex.

ikegami:
> @ikegami could you maybe explain what you tried to fix with commit [565b30e](https://github.com/linux-nvme/nvme-cli/commit/565b30e285934502e9fcaaf7d9b8bd81e5011d0f)? :)

I am not Tokunori Ikegami <ikegami.t@gmail.com>

MaisenbacherD:
> > @ikegami could you maybe explain what you tried to fix with commit [565b30e](https://github.com/linux-nvme/nvme-cli/commit/565b30e285934502e9fcaaf7d9b8bd81e5011d0f)? :)
> 
> The commit was to resolve the issue #2662 by using the blkdev instead of the chardev.

I see. Thanks :)


> The whole open device logic is a convoluted and very hard to understand. I think it's time to redesign it. That's why I haven't pulled #2671, it just makes it even more complex.

Sounds good.

MaisenbacherD:
> > @ikegami could you maybe explain what you tried to fix with commit [565b30e](https://github.com/linux-nvme/nvme-cli/commit/565b30e285934502e9fcaaf7d9b8bd81e5011d0f)? :)
> 
> I am not Tokunori Ikegami [ikegami.t@gmail.com](mailto:ikegami.t@gmail.com)

I am so sorry. This probably isn't the first time this happened to you. I will pay attention next time :)

igaw:
Wow, that's a nasty one to debug! Great catch! Let me rebase this PR and add your change to it.

PR #2696, opened 2025-02-09
nvme-print: misc fixes to a few verbose outputs

martin-gpy:


igaw:
Thanks!

I just realized we need to fix the '-v' debug output for the file handles. Now I remember why I didn't like it in the first place :)

PR #2706, opened 2025-02-14
Improve tests

MaisenbacherD:
The discussion and testing in relation to https://github.com/linux-nvme/nvme-cli/issues/2668 discovered some bugs in the test cases that are addressed with this PR.

### DIFF
--- a/tests/nvme_test.py
+++ b/tests/nvme_test.py
@@ -67,6 +67,7 @@ class TestNVMe(unittest.TestCase):
         self.load_config()
         if self.do_validate_pci_device:
             self.validate_pci_device()
+        self.create_and_attach_default_ns()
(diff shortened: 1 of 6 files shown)


tests/nvme_test.py
```diff
@@ -36,6 +36,18 @@
...
 
 
+def hex_to_decimal(value):
+    """ Wrapper for converting hexadecimal numbers to decimal
+        - Args:
+            - value: A number either in hex or base 10
+        - Returns:
+            - Decimal integer
+    """
+    if str(value).startswith("0x"):
```
martin-belanger, line 46: You could set the base to 0 to do the same thing as checking if the string starts with 0x.
```
>>> print(f"{int('10', 0)}")
10
>>> print(f"{int('0x10', 0)}")
16
```

MaisenbacherD: Thanks, this is very neat!
Changed it :)

igaw:
Thanks!

PR #2707, opened 2025-02-14
Fdp feature enable

lgdacunh:


igaw:
Thanks! 

Could you also provide a update for the tab completions?

PR #2708, opened 2025-02-16
nvme-print-binary: fix to output reachability associations log length

ikegami-t:
Previously only the log header output since the log length is changeable. This changes to output the changeable log length correctly.


nvme-print-json.c
```diff
@@ -4701,14 +4708,16 @@ static void json_reachability_groups_log(struct nvme_reachability_groups_log *lo
...
+static void json_reachability_associations_log(struct nvme_reachability_associations_log *log,
+					       __u64 len)
 {
 	struct json_object *r = json_create_object();
 	__u16 i;
 	__u32 j;
 	char json_str[STR_LEN];
 	struct json_object *rad;
 
+	obj_add_debug(uint64, r, "len", len);
```
igaw, line 4721: I am not sure if we should add debug output to the json output. For debugging purposes we have the normal output.

nvme-print-binary.c
```diff
@@ -322,9 +322,10 @@ static void binary_dispersed_ns_psub_log(struct nvme_dispersed_ns_participating_
 	d_raw((unsigned char *)log, sizeof(*log));
 }
 
-static void binary_reachability_groups_log(struct nvme_reachability_groups_log *log)
+static void binary_reachability_groups_log(struct nvme_reachability_groups_log *log, __u64 len)
 {
-	d_raw((unsigned char *)log, sizeof(*log));
+	print_debug("len: %"PRIu64"\n", (uint64_t)len);
```
igaw, line 326: Same comment as with the json output. I don't know if we want to add to every output format debug information. I am fine with having the normal output format annotated for debugging purposes, but adding trippling the debug output seems add unnecessary maintenance burden.

nvme-print-json.c
ikegami-t: Fixed to delete the debug output but added to use UNUSED macro instead since the len parameter not used.

nvme-print-binary.c
ikegami-t: Yes agreed then delete the output.

igaw:
I've just updated the cross build containers, the newer tool chains finds a bunch of new bugs. Don't worry about them, I'll fix them up.

ikegami-t:
I see and noted. Thank you.

igaw:
The build works again. Could you rebase to the current HEAD? Thanks!

ikegami-t:
Sure done then all checks passed. Thank you.

igaw:
Thanks!

PR #2709, opened 2025-02-16
nvme: make get-feature JSON output print everything

martin-gpy:
The nvme JSON outputs now print everything verbose. But this was missed out for the get-feature command, so fix the same here.


igaw:
Thanks!

PR #2711, opened 2025-02-16
nvme: check for valid output format

martin-gpy:
Several commands such as nvme get-feature, get-property, id_iocs, etc. don't check for a valid output format unlike other nvme commands. Fix the same.


igaw:
Thanks!

PR #2712, opened 2025-02-17
nvme-print: print new fields of nvm_id_ctrl

francispravin5:
Print the new fields of I/O Command Set Specific Identify Controller Data Structure based on
NVM Command Set Specification 1.1.


igaw:
Thanks!

PR #2713, opened 2025-02-17
nvme-print: fix the result of I/O Command Set Profile feature

francispravin5:
The Get Features command submitted for I/O Command Set Profile feature will return the I/O Command Set Combination Index [08:00] in Dword0 of completion queue entry. So, fixed the mask of the result.

### DIFF
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -5034,7 +5034,7 @@ static void stdout_feature_show_fields(enum nvme_features_id fid,
 		printf("\tEndurance Group Critical Warnings  : %u\n", NVME_FEAT_EG_EGCW(result));
 		break;
 	case NVME_FEAT_FID_IOCS_PROFILE:
-		printf("\tI/O Command Set Profile: %s\n", result & 0x1 ? "True" : "False");
+		printf("\tI/O Command Set Profile: %u\n", NVME_GET(result, FEAT_IOCSP_IOCSCI));
 		break;
 	case NVME_FEAT_FID_SPINUP_CONTROL:
 		printf("\tSpinup control feature Enabled: %s\n", (result & 1) ? "True" : "False");


igaw:
I've just updated the cross build containers, thus the newer build chain is complaining. I am fixing this up, so don't worry abou these.

nvme-print-json.c
```diff
@@ -3518,7 +3518,7 @@ static void json_feature_show_fields_endurance_evt_cfg(struct json_object *r, un
 
 static void json_feature_show_fields_iocs_profile(struct json_object *r, unsigned int result)
 {
-	obj_add_str(r, "I/O Command Set Profile", result & 0x1 ? "True" : "False");
+	obj_add_uint(r, "I/O Command Set Profile", NVME_GET(result, FEAT_IOCSP_IOCSCI));
```
igaw, line 3521: I would leave the `True/False` in place to avoid any compatibility issues.

igaw:
> I've just updated the cross build containers, thus the newer build chain is complaining. I am fixing this up, so don't worry abou these.

I've rebased your PR, the build works again.

nvme-print-json.c
francispravin5: @igaw Do you want to keep both json prints or `True/False` only?

igaw: Yes, the output format should not change if possible, so we don't break any existing parser.

francispravin5: @igaw Addressed the review comment.
Thanks!

nvme-print-stdout.c
```diff
@@ -5034,7 +5034,7 @@ static void stdout_feature_show_fields(enum nvme_features_id fid,
 		printf("\tEndurance Group Critical Warnings  : %u\n", NVME_FEAT_EG_EGCW(result));
 		break;
 	case NVME_FEAT_FID_IOCS_PROFILE:
-		printf("\tI/O Command Set Profile: %s\n", result & 0x1 ? "True" : "False");
+		printf("\tI/O Command Set Profile: %u\n", NVME_GET(result, FEAT_IOCSP_IOCSCI));
```
igaw, line 5037: Please, keep the `True/False` in place. There other parts in the print function use these strings as well.

PR #2718, opened 2025-02-20
nvme-print: print Host Behavior Support feature fields

francispravin5:
Print the fields of Host Behavior Support feature (FID - 16h).
NVM Express Base Specification 2.1


francispravin5:
@igaw Addressed the review comment.
Thanks!

igaw:
Thanks!

Issue #2721, opened 2025-02-23
The LBA Format specified is not supported(0x410a) when creating last namespace


maxpain:
Hello. I have Samsung PM9A3 M.2 1.92TB and I want to create four 480G NVMe namespaces with 4K LBA.

```bash
# nvme create-ns /dev/nvme1 -S 480G -C 480G -flbas 1 -dps 0 -nmic 0
create-ns: Success, created nsid:1
# nvme create-ns /dev/nvme1 -S 480G -C 480G -flbas 1 -dps 0 -nmic 0
create-ns: Success, created nsid:2
# nvme create-ns /dev/nvme1 -S 480G -C 480G -flbas 1 -dps 0 -nmic 0
create-ns: Success, created nsid:3
# nvme create-ns /dev/nvme1 -S 480G -C 480G -flbas 1 -dps 0 -nmic 0
NVMe status: Invalid Format: The LBA Format specified is not supported(0x410a)
```

I also tried `-s 117187500 -c 117187500` instead.

After this, `nvme id-ctrl /dev/nvme1` shows:
```
tnvmcap   : 1920383410176
unvmcap   : 480382377984
```

However, creating single namespace, occupying all the 4K blocks, works:
```bash
nvme create-ns /dev/nvme1 -s 468843606 -c 468843606 -flbas 1 -dps 0 -nmic 0
```

I don't understand how to properly slice my NVMe disk to four even namespaces.

keithbusch:
480GiB is 515396075520 bytes, but you have 480382377984 available. I'm not sure the device's return status makes much sense, but your command shouldn't succeed.

maxpain:
> 480GiB is 515396075520 bytes, but you have 480382377984 available. I'm not sure the device's return status makes much sense, but your command shouldn't succeed.

`-S 480G -C 480G` is gigabytes, not gibibytes.

Anyway, as I said, I tried this as well:
```bash
# nvme create-ns /dev/nvme1 -s 117187500 -c 117187500 -flbas 1 -dps 0 -nmic 0
create-ns: Success, created nsid:1
# nvme create-ns /dev/nvme1 -s 117187500 -c 117187500 -flbas 1 -dps 0 -nmic 0
create-ns: Success, created nsid:2
# nvme create-ns /dev/nvme1 -s 117187500 -c 117187500 -flbas 1 -dps 0 -nmic 0
create-ns: Success, created nsid:3
[...]

PR #2722, opened 2025-02-25
nvme: enforce ns granularity on ns create

igaw:
When the device supports the namespace granularity feature, update the alignment requirements accordingly. Because this is likely to change the capacity of the namespace also issue a info what is going to happen.

Fixes: #2721


maurizio-lombardi:
Looks ok to me now

igaw:
Thanks!

PR #2728, opened 2025-02-27
nvme-print: add human readable output for IOCS Data Structure

francispravin5:
Add human readable output for Identify I/O Command Set Data Structure (CNS 1Ch).


igaw:
Thanks!

PR #2733, opened 2025-02-28
Makefile: Add new targets and update meson command syntax

martin-belanger:
This pull request simply updates the meson syntax for some of the existing targets and provides a few additional targets as follows:

- For the "install" target, use "--skip-subprojects" so that only nvme-cli (and not the subprojects) get installed.
- For the "purge" target, add "meson subprojects purge".
- Added "update-subprojects" target
- Added "test-strict" target. This is to limit testing to nvme-cli and not all the subprojects.


igaw:
Thanks!

PR #2735, opened 2025-02-28
micron: Add missing newline

bsdimp:
Add missing newline to error for cards that don't support the vs-drive-info command.


igaw:
Thanks!

PR #2739, opened 2025-03-08
wdc: On FreeBSD, get the timezone the same as with real glibc

bsdimp:
FreeBSD 15 and newer finally has the timezone variable (prior it that was a compat function leftover from 4BSD). Older versions are still supported. Both have tm_gmoff, which I've always preferred for better portability.

I also considered dropping the defined(glibc), but wasn't sure how many other targets there are these days and thought this would be better. I really didn't want any FreeBSD ifdefs, but I don't know how to avoid this.


igaw:
Thanks!

PR #2742, opened 2025-03-14
nvme: allow mmap_registers from block device

sc108-lee:
Below commands that use mmap_registers failed since a31081a3fa14 (v2.9)
nvme show-regs /dev/nvme0n1
nvme effects-log /dev/nvme0n1

### DIFF
--- a/Documentation/nvme-get-reg.txt
+++ b/Documentation/nvme-get-reg.txt
@@ -22,6 +22,9 @@ DESCRIPTION
 -----------
 Read and show the defined NVMe controller register.
 
+The <device> parameter is mandatory and must be the nvme admin character
+device (ex: /dev/nvme0).
+
 OPTIONS
 -------
 -O <offset>::
--- a/Documentation/nvme-set-reg.txt
+++ b/Documentation/nvme-set-reg.txt
@@ -22,6 +22,9 @@ DESCRIPTION
 -----------
 Writes and shows the defined NVMe controller register.
 
+The <device> parameter is mandatory and must be the nvme admin character
+device (ex: /dev/nvme0).
+
 OPTIONS
 -------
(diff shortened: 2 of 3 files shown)


igaw:
If I read it  correctly the change also implements the fallback for ns. Do we really need it:

> We also don't need to do the fallback lookup via the namespaces because
> Linux kernel v4.0 has introduced the link in /sys/class/nvme/nvme%d. And
> nvme-cli 2.x depends on Linux kernel >= 4.15, so it's safe to drop the
> fallback.

sc108-lee:
If I understand right, The commit comment assume that /sys/class/nvme/nvme0n1 also exist?
Currently, this modification do readlink to get nvme0 from nvme0n1. (since there is no nvme0n1 in /sys/class/nvme)

[root@localhost csk]# ls /sys/class/nvme/
nvme0

sc108-lee:
if we do not want to lookup via the namespaces
then should we just cutting the device name nvme0n1 -> nvme0 ?
so we can support same functionality for block device with older version of nvme-cli

igaw:
The documentation stays

`The <device> parameter is mandatory and must be the nvme admin character device (ex: /dev/nvme0)`

Thus it's correct that these commands fail. I think it would be better to issue an info instead having this magic.

sc108-lee:
Good point, I just concerned about backward compatibility.

After attempting to access the register via the fabric command and failing, I initially tried to check the block device and print an error message. However, I modified the implementation to allow only character devices, as intended by the documentation you mentioned.

igaw:
There will be some people which will be unhappy that it doesn't work with block devices anymore, but with the message in place that shouldn't be too hard to figure out what is happening. 

Thanks!

PR #2745, opened 2025-03-17
sed: add SED man pages

gjoyce-ibm:
man pages for SED plugin commands


igaw:
Sorry for the delay, was away.

Thanks!

PR #2746, opened 2025-03-19
netapp-ontapdev: add subsysname to the verbose output

martin-gpy:
Display the ONTAP subsystem name too in the verbose output.


igaw:
Thanks!

PR #2747, opened 2025-03-22
nvme-print-stdout: add performance characteristics feature output

ikegami-t:
The feature identifier 1Ch added by NVMe 2.1.


ikegami-t:
The changes depended on the libnvme PR https://github.com/linux-nvme/libnvme/pull/971 changes.

igaw:
Okay, let's get the libnvme changes then sorted out. I've reviewed the

PR #2748, opened 2025-03-24
Add new sndk plugin for Sandisk drives

jeff-lien-sndk:
With the split of Western Digital into a 2 companies, there is a need to provide a new plugin for Sandisk drives.  The existing wdc plugin will remain and continue to support the legacy Western Digital drives.  No future updates or enhancements will be made to the existing wdc commands though.  The new sndk plugin will contain support for the current Sandisk drives as well as updates, enhancements, and fixes for those drives and any new drives going forward.


igaw:
Thanks!

PR #2750, opened 2025-03-27
nvme-builtin: sort command list as alphabetical order

ikegami-t:
This is to find easily the sub-commands name from the help list.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -10911,14 +10911,14 @@ int main(int argc, char **argv)
 
 	nvme.extensions->parent = &nvme;
 	if (argc < 2) {
-		general_help(&builtin);
+		general_help(&builtin, NULL);
 		return 0;
 	}
 	setlocale(LC_ALL, "");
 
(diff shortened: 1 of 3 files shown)


igaw:
@MatiasBjorling updated the order of the zns commands so that most used come first. I don't know if we  should order them or not. I see value in both orders.

We could also make sections if this would help, like the main commands like `list`, `id-ctrl` etc. and  the plumbing commands go into their own section. And both could be alphabetically sorted.

ikegami-t:
Understood so I will do try to consider more if any option or etc,. to sort the order can be added or not. Thank you.

MatiasBjorling:
Thank you. If possible, I like to maintain a kind of "highlight" of the most used commands, making it easier to guide the user towards the commands that they're most likely to need.

ikegami-t:
Thank you for your advice. I will do consider later.

igaw:
Thanks!

PR #2751, opened 2025-03-27
nvme: add dry-run NVME_ARGS default option

ikegami-t:
The option is to show command instead of sending.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -8009,7 +8009,6 @@ unsigned long long elapsed_utime(struct timeval start_time,
 
 static int submit_io(int opcode, char *command, const char *desc, int argc, char **argv)
 {
-	struct timeval start_time, end_time;
 	void *buffer;
 	_cleanup_free_ void *mbuffer = NULL;
 	int err = 0;
@@ -8017,7 +8016,6 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	int flags, pi_size;
 	int mode = 0644;
 	__u16 control = 0, nblocks = 0;
-	__u32 dsmgmt = 0;
 	unsigned int logical_block_size = 0;
 	unsigned long long buffer_size = 0, mbuffer_size = 0;
 	_cleanup_huge_ struct nvme_mem_huge mh = { 0, };
@@ -8147,7 +8145,6 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	if (cfg.prinfo > 0xf)
 		return err;
 
-	dsmgmt = cfg.dsmgmt;
 	control |= (cfg.prinfo << 10);
 	if (cfg.limited_retry)
 		control |= NVME_IO_LR;
@@ -8161,7 +8158,6 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 			return -EINVAL;
 		}
 		control |= cfg.dtype << 4;
-		dsmgmt |= ((__u32)cfg.dspec) << 16;
 	}
 
 	if (opcode & 1) {
@@ -8285,27 +8281,6 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 		}
 	}
 
-	if (cfg.show || nvme_cfg.dry_run) {
-		printf("opcode       : %02x\n", opcode);
-		printf("nsid         : %02x\n", cfg.namespace_id);
-		printf("flags        : %02x\n", 0);
-		printf("control      : %04x\n", control);
-		printf("nblocks      : %04x\n", nblocks);
-		printf("metadata     : %"PRIx64"\n", (uint64_t)(uintptr_t)mbuffer);
-		printf("addr         : %"PRIx64"\n", (uint64_t)(uintptr_t)buffer);
-		printf("slba         : %"PRIx64"\n", (uint64_t)cfg.start_block);
-		printf("dsmgmt       : %08x\n", dsmgmt);
-		printf("reftag       : %"PRIx64"\n", (uint64_t)cfg.ref_tag);
-		printf("apptag       : %04x\n", cfg.app_tag);
-		printf("appmask      : %04x\n", cfg.app_tag_mask);
-		printf("storagetagcheck : %04x\n", cfg.storage_tag_check);
-		printf("storagetag      : %"PRIx64"\n", (uint64_t)cfg.storage_tag);
-		printf("pif             : %02x\n", pif);
-		printf("sts             : %02x\n", sts);
-	}
-	if (nvme_cfg.dry_run)
-		return 0;
-
 	struct nvme_io_args args = {
 		.args_size	= sizeof(args),
 		.fd		= dev_fd(dev),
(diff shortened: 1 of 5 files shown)


nvme.h
```diff
@@ -90,6 +91,7 @@ struct nvme_config {
 		OPT_FMT("output-format", 'o', &nvme_cfg.output_format, output_format), \
 		##__VA_ARGS__,                                                         \
 		OPT_UINT("timeout",      't', &nvme_cfg.timeout,       timeout),       \
+		OPT_FLAG("dry-run",      'd', &nvme_cfg.dry_run,       dry_run),       \
```
igaw, line 94: Let's use only the long option name for it. It's likely one command is already using `-d` and I'd like to avoid further confusion with the short options name.

util/logging.c
```diff
@@ -91,14 +97,17 @@ int nvme_submit_passthru(int fd, unsigned long ioctl_cmd,
 {
 	struct timeval start;
 	struct timeval end;
-	int err;
+	int err = 0;
 
-	if (log_level >= LOG_DEBUG)
+	if (log_level >= LOG_DEBUG || dry_run)
```
igaw, line 102: I'd rather not mix the log level with the `--dry-run` options. I suggest to update the documentation instead. 

Or if you don't agree, the `--dry-run` should then really only print the command nothing else.

ikegami-t: Yes right so fixed to output only command input data only without the result and err outputs for the `--dry-run` option.

ikegami-t:
Fixed as mentioned and just rebased. Thank you.

nvme.c
```diff
@@ -8295,7 +8293,7 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 		printf("pif             : %02x\n", pif);
```
igaw, line 8293: Wouldn't it make sense to move this part to `logging.c` eventually? I would love to see that we extend the logging there to print more of the commands. But this is just an idea for a new side project :)

nvme.c
```diff
@@ -8101,7 +8100,6 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 		  OPT_SHRT("dir-spec",          'S', &cfg.dspec,             dspec),
 		  OPT_BYTE("dsm",               'D', &cfg.dsmgmt,            dsm),
 		  OPT_FLAG("show-command",      'V', &cfg.show,              show),
-		  OPT_FLAG("dry-run",           'w', &cfg.dry_run,           dry),
```
igaw, line 8104: I think we should keep the `-w` short hand here, to stay compatible.

util/logging.c
```diff
@@ -91,7 +97,12 @@ int nvme_submit_passthru(int fd, unsigned long ioctl_cmd,
 {
 	struct timeval start;
 	struct timeval end;
-	int err;
+	int err = 0;
```
igaw, line 100: This is an unrelated change, not really needed.

nvme.c
ikegami-t: Right so okay I will do this by a separate patch. Thank you.

nvme.c
ikegami-t: Fixed as mentioned.

util/logging.c
ikegami-t: Yes so fixed.

util/logging.c
```diff
@@ -150,3 +150,51 @@ int nvme_submit_passthru64(int fd, unsigned long ioctl_cmd,
...
+	printf("reftag       : %"PRIx64"\n", (uint64_t)args->reftag_u64);
+	printf("apptag       : %04x\n", args->apptag);
+	printf("appmask      : %04x\n", args->appmask);
+	printf("storagetagcheck : %04x\n", args->control & NVME_IO_STC ? true : false);
+	printf("storagetag      : %"PRIx64"\n", (uint64_t)args->storage_tag);
+	printf("pif             : %02x\n", args->pif);
+	printf("sts             : %02x\n", args->sts);
+}
+
+int nvme_submit_io(struct nvme_io_args *args, __u8 opcode, bool show, bool latency)
```
igaw, line 180: I suggest instead adding more indirect code for the loggig, we could just look at the opcode in  `nvme_submit_passthru`, e.g.

```c
static void nvme_show_command(struct nvme_passthru_cmd *cmd, int err)
{
	switch (cmd->opcode) {
		case nvme_cmd_read:
		case nvme_cmd_write:
		case nvme_cmd_compare:
			nvme_show_io_command(cmd);
			break;
		default:
			nvme_show_common(cmd);
	}
	printf("result       : %08x\n", cmd->result);
	printf("err          : %d\n", err);
}
```

util/logging.c
```diff
@@ -150,3 +150,51 @@ int nvme_submit_passthru64(int fd, unsigned long ioctl_cmd,
...
+{
+	struct timeval start_time;
+	struct timeval end_time;
+	int err;
+
+	if (show || dry_run)
+		nvme_show_io_command(args, opcode);
+
+	if (dry_run)
```
igaw, line 189: and then the `dry_run` maybe should be

```c
int nvme_submit_passthru(int fd, unsigned long ioctl_cmd,
			 struct nvme_passthru_cmd *cmd, __u32 *result)
{
	struct timeval start;
	struct timeval end;
	int err = 0;

	if (log_level >= LOG_DEBUG)
		gettimeofday(&start, NULL);

	if (!dry_run)
		err = ioctl(fd, ioctl_cmd, cmd);

	if (log_level >= LOG_DEBUG) {
		gettimeofday(&end, NULL);
		nvme_show_command(cmd, err);
		nvme_show_latency(start, end);
	}

	if (err >= 0 && result)
		*result = cmd->result;

	return err;
}
```

ikegami-t:
@igaw Thanks for your review comments. Basically look good but let me confirm below before the fix.
1. The suggestion will change the some logging behaviors so are the changes okay for now? Or still should we keep the logging behavior basically?
2. Basically the `struct nvme_passthru_cmd` values `opcode` and `cdw`, etc. can be converted to `nvme_show_io_command`() outputs values `nlb` and `slba`, etc. but only the value `storage_tag` needed `pif` and `sts` values to pass and also 'args_size' needed probably. The nvme-cli converts `struct config` to `struct nvme_io_args` and the libnvme converts `struct nvme_io_args` to `struct nvme_passthru_cmd` then the logging feature converts the `struct nvme_passthru_cmd` to `struct nvme_io_args` again. Is this really required?

igaw:
I think this would be a very good thing to have in place for being able to add more tests and make the libnvme2 development a bit more less eventful. My idea is to tests for a bunch of command, e.g.

`nvme id-ctrl --dry-run --verbose /dev/nvme0 | nvme-check-passthru test-data/id-ctrl.log`

and nvme-check-passthru would compare if the command words are identically and obviously filter out the dynamic values such pointers.

I am fine with changing some of the logging behavior. This is something I don't feel it is so important to be stable. Let be give a go with your patches as I want to figure out how to get the testframe work done.

igaw:
I think the `-V` (show command) and `-t` (show latency) could just be mapped to the verbosity levels. This is not 100% the same output but as long these two options still do something it should be fine.

igaw:
I've got something here:
https://github.com/igaw/nvme-cli/commit/607f1306f3a1d645008523034e8042c7b6bf2b2f

Not sure if it is possible to write get a parser for getting the storage tag etc back from the command words, e.g. `STS`, `PIF`, etc. Would be nice if this is possible.

igaw:
On the other hand we could leave the submit_io logging in place as it shows the higher level view (e.g. storage tag, sts, etc) and the low level output for the command dwords could also be there... let me play with this.

ikegami-t:
Thanks for your explanations. I can agree with you.

> I've got something here: [igaw@607f130](https://github.com/igaw/nvme-cli/commit/607f1306f3a1d645008523034e8042c7b6bf2b2f)

Yes the commit looks good.

> Not sure if it is possible to write get a parser for getting the storage tag etc back from the command words, e.g. `STS`, `PIF`, etc. Would be nice if this is possible.
> On the other hand we could leave the submit_io logging in place as it shows the higher level view (e.g. storage tag, sts, etc) and the low level output for the command dwords could also be there... let me play with this.

Right and good suggetion to leave the higher level output also.

igaw:
I think we can leave the `submit_io` code where it is. One thing we might consider is to make it a bit more useful by "beautifying" the output. But this is a different task. 

I'll update my version and submit it properly for review.

ikegami-t:
Thank you. Noted it. Let me close this PR.

PR #2753, opened 2025-03-28
plugins/lm: Pass numd option to nvme_lm_migration_recv

DanceDance:
Forward the user-provided 'numd' value from the command-line options to the underlying nvme_lm_migration_recv library function call.

The `lm_migration_recv` function parsed the `--numd` option but failed to pass the value (`cfg.numd`) when calling the library function `nvme_lm_migration_recv`.

The library function `nvme_lm_migration_recv` requires this value to set command parameters (CDW15) and calculate the expected data transfer length. [libnvme/src/nvme/ioctl.c at master · linux-nvme/libnvme](https://github.com/linux-nvme/libnvme/blob/master/src/nvme/ioctl.c#L2406)


ikegami-t:
Looks good.

igaw:
Thanks!

PR #2756, opened 2025-03-29
plugins: move meson.build solidigm-nvme.c into solidigm/meson.build

ikegami-t:
Add as same with other solidigm sources added by solidigm/meson.build.


igaw:
Thanks!

PR #2757, opened 2025-03-30
ocp: replace no-uuid option help string to use no_uuid variable

ikegami-t:
Reduce duplicated string definitions.


igaw:
Thanks!

PR #2763, opened 2025-04-01
OCP Telemetry Data Statistics Parsing Fix

jeff-lien-sndk:
The function print_formatted_var_size_str (in util/utils.c) which is used to print the telemetry statistics additional data had a string array (used for printing out the data) with a hard coded length of 1024 chars.  The Stats data with 80 DW's of data would exceed the size of that array.  Therefore, the code was changed to allocate a buffer large enough to contain the full string data.


igaw:
Thanks!

PR #2764, opened 2025-04-02
netapp-ontapdev: add subsysname to regular output

martin-gpy:
Display the subsysname in the regular output as well.


igaw:
Thanks!

Issue #2761, opened 2025-03-31
The reference tag is parsed incorrectly, causing the PI test case to fail when the format is 512+8, with metadata transferred as a separate buffer


yalanliufio:
We recently upgraded our nvme cli version from 2.2.1 to 2.9.1, then encountered this problem.

Release 2.2.1 works correctly, but release 2.9.1 and the latest release 2.12, both have this issue.

When SSD drive is format as 512+8, and metadata is transferred as separate buffer, this is the id-ns output:

```
[root@localhost test-scripts]# nvme id-ns /dev/nvme1n1 
NVME Identify Namespace 1:
nsze    : 0x1bf1f72b0
ncap    : 0x1bf1f72b0
nuse    : 0x10
nsfeat  : 0x10
nlbaf   : 4
flbas   : 0x1
mc      : 0x3
dpc     : 0x13
dps     : 0x1
nmic    : 0
lbaf  0 : ms:0   lbads:9  rp:0 
lbaf  1 : ms:8   lbads:9  rp:0 (in use)
lbaf  2 : ms:0   lbads:12 rp:0 
lbaf  3 : ms:8   lbads:12 rp:0 
lbaf  4 : ms:64  lbads:12 rp:0 
```

When using nvme version 2.2.1, 
```
[root@localhost test-scripts]# nvme --version
nvme version 2.2.1 (git 2.2.1)
libnvme version 1.2 (git 1.2)
```

issue a write command and also enable the kernel trace:, this is the output

```
nvme write /dev/nvme1n1 -s 0x02000000 -c 2 -z 1536 -y 24 -r 0x02000000 -d write_D1D0027T.data -M write_meta_D1D0027T_slba02000000_nlb2.meta -p 0 -a 0
            nvme-29400   [004] ....... 235741.995845: nvme_setup_cmd: nvme1: disk=nvme1n1, qid=5, cmdid=768, nsid=1, flags=0x0, meta=0x1, cmd=(nvme_cmd_write slba=33554432, len=2, ctrl=0x0, dsmgmt=0, reftag=33554432)
          <idle>-0       [004] d..h1.. 235741.995853: nvme_sq: nvme1: disk=nvme1n1, qid=5, head=1, tail=1
          <idle>-0       [004] d..h1.. 235741.995854: nvme_complete_rq: nvme1: disk=nvme1n1, qid=5, cmdid=768, res=0x0, retries=0, flags=0x2, status=0x0
```

[...]

PR #2765, opened 2025-04-02
nvme: fix to set nvme_io_args regtag value

ikegami-t:
Since the value is used by the libnvme nvme_io() function.


ikegami-t:
Fix for the issue #2761.

igaw:
Thanks for taking care of this!

PR #2769, opened 2025-04-07
updating micron plugin

sivaprasad6541:
Updating micron plugin with OCP Telemetry support


igaw:
Thanks for your PR. Unfortunately, the code has a lot of style issues. Please see what checkpath.pl reports. Also it doesn't build with clang.

sivaprasad6541:
> Thanks for your PR. Unfortunately, the code has a lot of style issues. Please see what checkpath.pl reports. Also it doesn't build with clang.

Bi mistake I have pushed the files which are already available in OCP plugin. I will create a new PR with required changes only.

PR #2770, opened 2025-04-08
plugins/solidigm: Updated SMART PLL Lock Loss counters field

lgdacunh:
Broke down additional SMART PLL Lock Loss field into separate counters.


igaw:
Thanks!

PR #2771, opened 2025-04-08
plugins/lm: Refine Sequence Indicator description and macro usage

DanceDance:
Changed the command dword reference in the sequence indicator (seqind) help text from CDW11 to CDW10. This aligns with the spec.

In the spec, `mos` field is at bits 31:16 of CDW10, and `seqind` is at bits 01:00 of `mos`, which gives us the final location of `(CDW10[17:16])`.

Screenshot 2025-04-09 at 07 55 02

Screenshot 2025-04-08 at 20 32 07

Also changed the macro used for setting the sequence indicator bits in the `mos` field from LM_MIGRATION_SEND_MOS to LM_SEQIND.


igaw:
Thanks!

PR #2773, opened 2025-04-10
nvme-print: print new fields of nvme_id_independent_id_ns

francispravin5:
Print new fields of I/O Command Set Independent Identify Namespace Data Structure
based on NVM Express Base Specification 2.1.


igaw:
Thanks!

Issue #2760, opened 2025-03-31
Regression in "nvme connect-all -t tcp" in 2.12


bluca:
Hi,

Since version 2.12 appeared in various distributions, we have started to see occasional failures in the systemd integration test that uses nvme-cli:

```
[   10.378713] TEST-84-STORAGETM.sh[316]: + nvme connect-all -t tcp -a 127.0.0.1 -s 16858 --hostid=95fe8041-3f53-415b-bc40-1bbd8932e7e8
[   10.397892] nvme_tcp: queue 0: failed to receive icresp, error -4
[   10.397326] TEST-84-STORAGETM.sh[340]: failed to add controller, error failed to write to nvme-fabrics device
```

This test runs dozens if not hundreds of times every day and it was stable until 2.12. Are you aware of any potential issues/changes?

You can find the full test script here:

https://github.com/systemd/systemd/blob/main/test/units/TEST-84-STORAGETM.sh

bluca:
The kernel in this case is 6.13.8

Full (very verbose, including image build) log: https://productionresultssa11.blob.core.windows.net/actions-results/68ef4d36-6a89-4db7-ab5e-4b8d78357c5c/workflow-job-run-ab2a2163-2e0e-540b-4bd7-05f43b11d9c3/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-03-31T14%3A23%3A06Z&sig=T%2FcJ8HJd67kV%2Blub5TAL5EVe08Taa%2FHxeoqWbjSeyog%3D&ske=2025-04-01T00%3A00%3A35Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-03-31T12%3A00%3A35Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-01-05&sp=r&spr=https&sr=b&st=2025-03-31T14%3A13%3A01Z&sv=2025-01-05

This is not just a warning, but a full error that is failing the test case, and it was first observed with 2.12

igaw:
It's an error and not a warning. 

If you provide the very verbose option `-vv` the tool will print out which command string passes it's passing down to the kernel (`/dev/nvme-fabrics`). The diff between the two versions would help a lot.

Anyway, could you post the logs, I can't access them `AuthenticationFailed`

hreinecke:
[...]

PR #2775, opened 2025-04-16
logging: retry ioctl on EINTR return value

igaw:
Recent changes in the nvme-tcp transport started to return EINTR due to a signal while reading from a socket. The transport could retry the operation but so far this hasn't been implemented. Thus we retry in userpace the connect call when EINTR is returned.

Fixes: #https://github.com/linux-nvme/nvme-cli/issues/2760

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -183,6 +183,20 @@ else
   conf.set('fallthrough', 'do {} while (0) /* fallthrough */')
 endif
 
+if cc.has_function('TEMP_FAILURE_RETRY', prefix : '#include <errno.h>')
+  conf.set('TFR', 'TEMP_FAILURE_RETRY')
+else
+  conf.set('TFR(exp)', '''					\
+({								\
+	long int __result = 0; 					\
+	do {							\
+		__result = (long int)(exp);			\
+	} while ((__result == -1) && (errno == EINTR));		\
+	__result;						\
+})
+''')
+endif
+
 configure_file(
     output: 'config.h',
     configuration: conf
--- a/util/logging.c
+++ b/util/logging.c
@@ -7,6 +7,7 @@
 #include <sys/syslog.h>
 #include <sys/time.h>
 #include <linux/types.h>
+#include <errno.h>
 
 #include <libnvme.h>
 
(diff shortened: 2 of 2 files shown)


ikegami-t:
Just confirmed below the `EINTR` case but still the PR changes correct for the case alos?
1. `nvme_execute_rq()` returns `-EINTR` when `nvme_req(rq)->flags` set `NVME_REQ_CANCELLED`.
<https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/nvme/host/core.c#n1123>
```c
int nvme_execute_rq(struct request *rq, bool at_head)
{
	blk_status_t status;

	status = blk_execute_rq(rq, at_head);
	if (nvme_req(rq)->flags & NVME_REQ_CANCELLED)
		return -EINTR;
...
```
2. The `nvme_req(req)->flags` set `NVME_REQ_CANCELLED` if the controller deleting for the disable.
<https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/nvme/host/pci.c#n1476>
```c
static enum blk_eh_timer_return nvme_timeout(struct request *req)
{
...
	switch (nvme_ctrl_state(&dev->ctrl)) {
	case NVME_CTRL_CONNECTING:
		nvme_change_ctrl_state(&dev->ctrl, NVME_CTRL_DELETING);
		fallthrough;
	case NVME_CTRL_DELETING:
		dev_warn_ratelimited(dev->ctrl.device,
			 "I/O tag %d (%04x) QID %d timeout, disable controller\n",
			 req->tag, nvme_cid(req), nvmeq->qid);
		nvme_req(req)->flags |= NVME_REQ_CANCELLED;
```
Let me ask below questions since I am not sure the PR changes correct (but seems for some cases there are some risks and not good).
1. Is basically the `EINTR` error reuqired to retry? (Seems it is only for the `EGAIN` error basically.)
2. Is the nvme-tcp transport `EINTR` error return a special behavior? (Seems the nvme-tcp transport should return `EGAIN` error instead.)
3. For other nvme driver cases to return `EINTR` is there no risk to be continued the retry handling? (Concerning the nvme-cli command hangup for the controller reset or disable cases, etc.)
4. Is the `EINTR` only returned temporary case? (Since it means `Interrupted function call` so no concern for above questions? Concerning for some error cases if the `EINTR` is continued to return.)

PR #2779, opened 2025-04-19
nvme: add performance characteristic feature output

ikegami-t:
Only added the stdout output at first.


ikegami-t:
Note: This PR depeneded on the PR https://github.com/linux-nvme/libnvme/pull/994.

igaw:
Thanks!

PR #2780, opened 2025-04-22
udev: Add custom udev rule for the Vast Data Block array

sagigrimberg:
Vast Data array supports NVMe-oF block access, and it prefers the host to use round-robin path selection for performance benefits.

In addition, disable the ctrl_loss_tmo in case of prolonged maintenance operations.


igaw:
Thanks!

PR #2783, opened 2025-04-25
wdc:  Fixed check for supported log pages

jeff-lien-sndk:
Add checking of the supoorted log pages (log id 0) for currently supported log pages.  The previous code was only checking a vendor specific log page for this info. Also the uuid index value was added so the supported log pages can be checked for a specific uuid.


igaw:
Thanks!

PR #2784, opened 2025-04-27
plugin: translate -h into commands

ikegami-t:
As the help short option parsed -h by argconfig so follow plugin also.

### DIFF
--- a/plugin.c
+++ b/plugin.c
@@ -144,6 +144,7 @@ int handle_plugin(int argc, char **argv, struct plugin *plugin)
 	struct command **cmd = plugin->commands;
 	struct command *cr = NULL;
 	bool cr_valid = false;
+	int dash_count = 0;
 
 	if (!argc) {
 		general_help(plugin);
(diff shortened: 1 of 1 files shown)


plugin.c
```diff
@@ -156,11 +156,11 @@ int handle_plugin(int argc, char **argv, struct plugin *plugin)
 		sprintf(use, "%s %s %s <device> [OPTIONS]", prog->name, plugin->name, str);
 	argconfig_append_usage(use);
 
-	/* translate --help and --version into commands */
+	/* translate --help, -h and --version into commands */
 	while (*str == '-')
 		str++;
 
-	if (!strcmp(str, "help"))
+	if (!strcmp(str, "help") || (str == argv[0] + 1 && !strcmp(str, "h")))
```
igaw, line 163: Can you elaborate hat is `str == argv[0] + 1`  is doing?

ikegami-t: `str == argv[0] + 1` is checking if the dash number actually 1 as `-h`. Just updated the patch to make it clear.

igaw:
Thanks!

PR #2794, opened 2025-05-03
nvme: delete host identifier feature length checking

ikegami-t:
Since already checked in nvme_get_feature_length() calling. Deleted before by the commit: 18de3a6d but then incorrectly added.

Fixes: b1567901d9b7 ("nvme: split get_feature to get multiple feature ids")


igaw:
Thanks!

PR #2795, opened 2025-05-03
feat: add FEAT_ARGS to use instead of NVME_ARGS

ikegami-t:
Includes save and sel options for all nvme feat plugin commands.


ikegami-t:
Just changed to the save argument dafault value pointer to NULL.

igaw:
Thanks!

PR #2796, opened 2025-05-04
nvme: add output-format-version option

ikegami-t:
The JSON outputs changed to verbose by the format version 2.


nvme.c
```diff
@@ -6922,7 +6920,6 @@ static int dir_send(int argc, char **argv, struct command *cmd, struct plugin *p
 		  OPT_SHRT("dir-spec",       'S', &cfg.dspec,          dspec_w_dtype),
 		  OPT_BYTE("dir-oper",       'O', &cfg.doper,          doper),
 		  OPT_SHRT("endir",          'e', &cfg.endir,          endir),
-		  OPT_FLAG("human-readable", 'H', &cfg.human_readable, human_readable_directive),
```
igaw, line 6924: Best leave it the option around, just to avoid to break anyone. Couldn't we map it to verbose mode ?

igaw: But if there is no user left, just leave the option. You could add a help text like `deprecated; does nothing`

igaw:
Looks really good. Thanks a lot!

ikegami-t:
Fixed the commits as suggested. Thank you.

igaw:
Thanks!

Maybe we should also update the documentation on `-H` and mention to use the `--verbose` flag instead.

ikegami-t:
> Maybe we should also update the documentation on `-H` and mention to use the `--verbose` flag instead.

Noted so will do later. Thank you.

Issue #2781, opened 2025-04-23
blktests nvme/030 fails when `70-nvmf-autoconnect.rules` is active


igaw:
blktests nvme/030 test is likely to fail if the 70-nvmf-autoconnect.rules is active. In this case two discovery are running in parallel and nvme-cli/libnvme gets out of sync and it looks as nvme-cli operates on the wrong device:

```
nvme/030 (tr=fc) (ensure the discovery generation counter is updated appropriately) [failed]
    runtime    ...  1.254s
    --- tests/nvme/030.out      2023-08-30 08:39:08.428409596 +0000
    +++ /tmp/blktests/nodev_tr_fc/nvme/030.out.bad      2025-04-23 13:15:40.722450950 +0000
    @@ -1,2 +1,6 @@
     Running nvme/030
    +Failed to open ctrl nvme1, errno 11
    +Failed to open ctrl nvme1, errno 11
    +failed to get discovery log: Bad file descriptor
    +tests/nvme/rc: line 489: ((: 3 <=  : syntax error: operand expected (error token is "<=  ")
     Test complete
```

```
    [ 1434.556136] run blktests nvme/030 at 2025-04-10 10:56:03
    [ 1434.779180] nvmet: adding nsid 1 to subsystem blktests-subsystem-1
    [ 1434.858814] nvme nvme1: NVME-FC{0}: create association : host wwpn 0x20001100aa000001  rport wwpn 0x20001100ab000001: NQN "nqn.2014-08.org.nvmexpress.discovery"
    [ 1434.860275] (NULL device *): {0:0} Association created
    [ 1434.861773] nvmet: Created discovery controller 1 for subsystem nqn.2014-08.org.nvmexpress.discovery for NQN nqn.2014-08.org.nvmexpress:uuid:a94a718e-09b2-4b5b-898a-116845846761.
    [ 1434.864704] nvme nvme1: NVME-FC{0}: controller connect complete
    [ 1434.865082] nvme nvme1: NVME-FC{0}: new ctrl: NQN "nqn.2014-08.org.nvmexpress.discovery", hostnqn: nqn.2014-08.org.nvmexpress:uuid:a94a718e-09b2-4b5b-898a-116845846761
    [ 1434.871357] nvme nvme1: Removing ctrl: NQN "nqn.2014-08.org.nvmexpress.discovery"
    [ 1434.953940] nvmet: adding nsid 1 to subsystem blktests-subsystem-2
[...]

PR #2797, opened 2025-05-05
nvme: retry commands on EAGAIN and EINTR

igaw:
The kernel started to return EAGAIN and EINTR. It is quiet likely that this was possible before but it happened not so often. Thus for users of nvme-cli it is a behavior change.

The first attempt was to handle this in libnvme itself. But it turns out we can't just blindly retry. It is necessary to check if the user has pressed Ctrl-C to terminate the program. Adding a signal handler inside the library is a no go. Thus add the retry logic to nvme-cli itself.

Fixes: #2760 #2781 #2775

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -47,6 +47,7 @@
 #include "fabrics.h"
 #include "util/cleanup.h"
 #include "util/logging.h"
+#include "util/sighdl.h"
 
 #define PATH_NVMF_DISC		SYSCONFDIR "/nvme/discovery.conf"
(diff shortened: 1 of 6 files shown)


igaw:
And this also wants 
   
  https://github.com/linux-nvme/libnvme/commit/2b304892292d380beaeb241e51ab2a0fdaf5c05d

reverted.

ikegami-t:
Looks good. (By the way seems the commit changes can be split to some patches.)

igaw:
Oh, did something stupid with this PR and now I can't reopen it.

igaw:
Oh wait, I just need to add a comment to reopen...

PR #2801, opened 2025-05-08
nvme: retry commands on EAGAIN and EINTR (v2)

igaw:
The kernel started to return EAGAIN and EINTR. It is quiet likely that this was possible before but it happened not so often. Thus for users of nvme-cli it is a behavior change.

The first attempt was to handle this in libnvme itself. But it turns out we can't just blindly retry. It is necessary to check if the user has pressed Ctrl-C to terminate the program. Adding a signal handler inside the library is a no go. Thus add the retry logic to nvme-cli itself.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/2760 https://github.com/linux-nvme/nvme-cli/issues/2781 https://github.com/linux-nvme/nvme-cli/pull/2775

depends on https://github.com/linux-nvme/libnvme/pull/1010

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -47,6 +47,7 @@
 #include "fabrics.h"
 #include "util/cleanup.h"
 #include "util/logging.h"
+#include "util/sighdl.h"
 
 #define PATH_NVMF_DISC		SYSCONFDIR "/nvme/discovery.conf"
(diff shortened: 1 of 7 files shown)


igaw:
Oh man as soon I merged it I realized that this is will not work, the return value is always -1 and errno is set. Stupid POSIX APIs.

PR #2802, opened 2025-05-08
Handle add ctrl errors

igaw:
The previous 'fix' was not really working because the error status is in errno not in the return value.

fixes: #2801

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -179,12 +179,15 @@ static int nvme_add_ctrl(nvme_host_t h, nvme_ctrl_t c,
 
 retry:
 	ret = nvmf_add_ctrl(h, c, cfg);
-	if (ret == EAGAIN || (ret == EINTR && !nvme_sigint_received)) {
-		printf("nvmf_add_ctrl returned '%s'\n", strerror(ret));
+	if (!ret)
+		return 0;
+
+	if (errno == EAGAIN || (errno == EINTR && !nvme_sigint_received)) {
+		print_debug("nvmf_add_ctrl returned '%s'\n", strerror(errno));
 		goto retry;
 	}
 
-	return ret;
+	return -errno;
 }
 
 static nvme_ctrl_t __create_discover_ctrl(nvme_root_t r, nvme_host_t h,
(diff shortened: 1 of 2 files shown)


fabrics.c
```diff
@@ -1122,21 +1125,18 @@ int nvmf_connect(const char *desc, int argc, char **argv)
...
-	if (ret)
+	if (ret) {
 		fprintf(stderr, "could not add new controller: %s\n",
-			nvme_strerror(errno));
-	else {
-		errno = 0;
-		if (flags != -EINVAL)
-			nvme_show_connect_msg(c, flags);
+			nvme_strerror(-ret));
+	} else if (flags != -EINVAL) {
```
ikegami-t, line 1132: The existing flags checking not correct since never set `-EINVAL` to flags so always executed `nvme_show_connect_msg()` if `ret` set zero.

ikegami-t (commented):
Changes looks good but as commented the existing flags handling seems not correct. Sorry for the PR: #2797 errors as I could not review correctly. Thank you.

fabrics.c
igaw: Indeed, the flags check doesn't make sense. The flag can't be `EINVAL`, because we already filtered this out. It turns out that the device name is supposed to be printed on connect. I don't like it, IMO it should only printed with --verbose but it's too late to change this now. So lets keep it.

0ad49ade724e ("Print device name on connect")

igaw:
> Changes looks good but as commented the existing flags handling seems not correct. Sorry for the PR: #2797 errors as I could not review correctly. Thank you.

No worries. Since I wrote all the error handling code I should have spotted it. I had a very good reproducer
(blktests nvme/030 kept returning EAGAIN) but it went away for some reason. Thus I didn't really test it.

fabrics.c
igaw: It turns out we can cleanup the error code path even more. Thanks for pointing it out!

ikegami-t: Thank you. The changes look good.

PR #2803, opened 2025-05-08
nvme-print: print Boot Partition Write Protection Config feature field

francispravin5:
Print the fields of Boot Partition Write Protection Config feature (FID - 85h).


igaw:
Thanks!

PR #2804, opened 2025-05-09
CI: move to self-hosted GitHub ARC infrastructure

MaisenbacherD:
We are now running the self-hosted testing infrastructure with the help of GitHub's ARC. This allows us to be more considerate with the available resources as the VMs are now spawned on demand with the requested kernel.

For now, we have to print KERNEL_VERSION in a previous job to pick up the correct kernel. This should be fixed in the future by contributing a labeling mechanism to GitHub's ARC.


igaw:
Thanks!

PR #2805, opened 2025-05-09
updated micron plugin with new products support and aligned with kernal coding style

sivaprasad6541:
updated micron plugin with new products support and aligned with kernal coding style


sivaprasad6541:
Closing the PR as it has errors from each commit. We'll create a new one by resolving all the current erros.

igaw:
Okay, but you could also just update the patch(es) and force push it. github can handle this and on the plus side, it avoid having a lot of abandon PRs.

PR #2806, opened 2025-05-11
doc: update dir-send command documentation on -H

ikegami-t:
Since actually the option is not used by the command.


ikegami-t:
This updates for the comment: https://github.com/linux-nvme/nvme-cli/pull/2796#issuecomment-2865435986.

igaw:
Thanks!

PR #2807, opened 2025-05-12
logging: add nvme_mi_submit entry and exit functions

ikegami-t:
Only supported for the NVMe-MI message type NVMe admin command.

### DIFF
--- a/util/logging.c
+++ b/util/logging.c
@@ -11,12 +11,21 @@
 #include <linux/types.h>
 
 #include <libnvme.h>
+#include <libnvme-mi.h>
+
+#include <ccan/endian/endian.h>
 
 #include "logging.h"
 #include "sighdl.h"
 
+struct submit_data {
+	struct timeval start;
+	struct timeval end;
+};
+
 int log_level;
 static bool dry_run;
+static struct submit_data sb;
 
 int map_log_level(int verbose, bool quiet)
 {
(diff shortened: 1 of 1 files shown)


igaw:
LGTM.

@jk-ozlabs

util/logging.c
```diff
@@ -167,3 +176,103 @@ int nvme_submit_passthru64(int fd, unsigned long ioctl_cmd,
...
+		.data_len = data_len,
+		.cdw10 = le32_to_cpu(hdr->cdw10),
+		.cdw11 = le32_to_cpu(hdr->cdw11),
+		.cdw12 = le32_to_cpu(hdr->cdw12),
+		.cdw13 = le32_to_cpu(hdr->cdw13),
+		.cdw14 = le32_to_cpu(hdr->cdw14),
+		.cdw15 = le32_to_cpu(hdr->cdw15),
+	};
+
+	nvme_show_common(&cmd);
```
jk-ozlabs, line 199: I think we would want include the `dofst` / `dlen` here too, otherwise the data may not make much sense.

util/logging.c
```diff
@@ -167,3 +176,103 @@ int nvme_submit_passthru64(int fd, unsigned long ioctl_cmd,
...
+	}
+
+	return &sb;
+}
+
+static void nvme_show_resp_admin(const struct nvme_mi_admin_resp_hdr *hdr, size_t hdr_len,
+				 const void *data, size_t data_len)
+{
+	printf("result       : %08x\n", hdr->cdw0);
+	printf("err          : %d\n", hdr->status);
```
jk-ozlabs, line 242: I assume we don't want the data here?

jk-ozlabs (commented):
Looks good, just a couple of comments inline.

util/logging.c
ikegami-t: Currently nvme_submit_passthru() for PCIe does not print the data so at this moment not required as same but later possible to be added the data output also I think.

jk-ozlabs (approved):
I'm not fully across the logging patterns in general, but +1 from me on the MI-related parts.

igaw:
FWIW, I think it's safe to modify the logging later on. I don't consider this as a stable interface in any sorts. Let's start with this and see what works and what needs some more changes later on.

igaw:
Thanks!

PR #2808, opened 2025-05-12
sighdl: include stddef header file

ikegami-t:
The header file required for the muon build on CentOS.

### DIFF
--- a/util/sighdl.c
+++ b/util/sighdl.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include <signal.h>
 #include <errno.h>
+#include <stddef.h>
 
 #include "sighdl.h"
 


plugins/memblaze/memblaze-nvme.c
```diff
@@ -1616,7 +1616,7 @@ static int mb_get_smart_log_add(int argc, char **argv, struct command *cmd, stru
 
 	// Get log
 
-	struct smart_log_add log = {0};
+	struct smart_log_add log = {{{{{0}}}}};
```
igaw, line 1619: Uff, this looks rather unusual.  If this is really necessary because the compiler on CentOS can't handle this type of initialization, I suggest we rather use memset for this case.

ikegami-t: Changed to use memset as mentioned. Thank you.

ikegami-t:
The NULL definition error caused on Cent OS stream release 9.
```sh
[48/166] compiling c nvme.p/plugins/ocp/ocp-hardware-component-log.c.o
samu: job failed with status 1: gcc -std=gnu99 -g -Og -Wall -fdiagnostics-color=always -fomit-frame-pointer -D_GNU_SOURCE -include config.h -I . -I .. -I ccan -I ../ccan -I subprojects/libnvme/src -I ../subprojects/libnvme/src -MD -MQ nvme.p/util/sighdl.c.o -MF nvme.p/util/sighdl.c.o.d -o nvme.p/util/sighdl.c.o -c ../util/sighdl.c
../util/sighdl.c: 関数 ‘nvme_install_sigint_handler’ 内:
../util/sighdl.c:23:30: エラー: ‘NULL’ が宣言されていません (この関数内での最初の使用)
  if (sigaction(SIGINT, &act, NULL) == -1)
                              ^~~~
../util/sighdl.c:23:30: 備考: ‘NULL’ is defined in header ‘<stddef.h>’; did you forget to ‘#include <stddef.h>’?
../util/sighdl.c:6:1:
+#include <stddef.h>
 
../util/sighdl.c:23:30:
  if (sigaction(SIGINT, &act, NULL) == -1)
                              ^~~~
../util/sighdl.c:23:30: 備考: 未宣言の識別子は出現した各関数内で一回のみ報告されます
samu: subcommand failed
```

igaw:
I'd say drop the second patch (CentOS 7, 8 are EOL). CentOS Stream 9 is alive.

igaw:
Thanks!

PR #2812, opened 2025-05-13
Update: micron plugin with latest product support

sivaprasad6541:
Updated the Micron plugin to support the latest products and features. Changes follow the Kernel coding style guidelines for consistency.


igaw:
Just updated the commit message. Thanks for taking care of the Micron plugin!

sivaprasad6541:
Thank you @igaw  :-)

PR #2813, opened 2025-05-13
build: delete docs=disabled muon build option

ikegami-t:
Since currently the error output by the muon build.
  error invalid option: 'docs=disabled'


igaw:
Thanks!

PR #2815, opened 2025-05-15
nvme-print-json: reverse pel power on reset parameter order error

ikegami-t:
The parameters VSIL and EL order was incorrect so reverse the order.


igaw:
Thanks!

PR #2817, opened 2025-05-17
Json id ns lbaf

ikegami-t:


util/json.h
```diff
@@ -78,5 +78,6 @@ void json_object_add_byte_array(struct json_object *o, const char *k, unsigned c
 void json_object_add_nprix64(struct json_object *o, const char *k, uint64_t v);
 void json_object_add_uint_0nx(struct json_object *o, const char *k, __u32 v, int width);
 void json_object_add_0nprix64(struct json_object *o, const char *k, uint64_t v, int width);
+void json_object_add_format(struct json_object *o, const char *k, const char *format, ...);
```
igaw, line 81: I suggest to stick to the above pattern and use the name as final postfix, so in this case just `string`. There is also already one helper which is called `.._value_string`. So this would make more sense to me.

`json_object_add_string`

igaw:
I was wonder about the second patch how we deal with the verbose flag and json. it seems in most cases we are actually replace the output with something else. So the verbose flag is not extending the existing output. 

I think it would make more sense to handle it as we do it in the stdout case, where we leave the orignal output and just extend it. 

Though this is something we should do in a different patch and verify that we don't break anyone. 

I am fine with the second patch for the time being.

util/json.h
ikegami-t: Fixed as suggested. Thank you.

ikegami-t:
About the second patch understood your comment so for now just remaind the changes but just rebased with the first patch. Thank you.

igaw:
Thanks!

PR #2818, opened 2025-05-18
nvme: output copy command success message for json format

ikegami-t:
Use nvme_show_key_value() to output the message as json format.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -7601,7 +7601,7 @@ static int copy_cmd(int argc, char **argv, struct command *cmd, struct plugin *p
 	else if (err != 0)
 		nvme_show_status(err);
 	else
-		printf("NVMe Copy: success\n");
+		nvme_show_key_value("NVMe Copy", "success");
 
 	return err;
 }


util/logging.c
```diff
@@ -17,14 +17,15 @@
 
 #include "logging.h"
 #include "sighdl.h"
+#include "nvme-print.h"
```
igaw, line 20: I would like to avoid pulling in something from the parent directly into the util folder. Looking at what loggin.c does I suggest just move it to the parent folder.

util/logging.c
```diff
@@ -17,14 +17,15 @@
...
 
 struct submit_data {
 	struct timeval start;
 	struct timeval end;
 };
 
 int log_level;
-static bool dry_run;
+static bool dry_run_state;
```
igaw, line 28: ... and then you can just directly use the `dry_run`. so you can delete this definition.

util/logging.c
```diff
@@ -59,49 +60,49 @@ int map_log_level(int verbose, bool quiet)
 
 void set_dry_run(bool enable)
 {
-	dry_run = enable;
+	dry_run_state = enable;
 }
```
igaw, line 64: also we don't need this setter anymore.

ikegami-t:
Fixed as suggested. Thank you.

igaw:
Thanks!

PR #2823, opened 2025-05-21
plugins/micron: adding missing product support

sivaprasad6541:
Added missing micron product ID's into plugin


igaw:
Thanks!

sivaprasad6541:
Thank you @igaw :-)

PR #2825, opened 2025-05-24
nvme: add support for abort command

ikegami-t:
The abort command can be submitted by the admin-passthru command. But better to support by the nvme abort command directly then add.

### DIFF
--- a/nvme-builtin.h
+++ b/nvme-builtin.h
@@ -124,6 +124,7 @@ COMMAND_LIST(
 	ENTRY("io-mgmt-send", "I/O Management Send", io_mgmt_send)
 	ENTRY("nvme-mi-recv", "Submit a NVMe-MI Receive command, return results", nmi_recv)
 	ENTRY("nvme-mi-send", "Submit a NVMe-MI Send command, return results", nmi_send)
+	ENTRY("abort", "Submit an Abort command ", abort_cmd)
 );
 
 #endif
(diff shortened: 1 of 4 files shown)


ikegami-t:
The changes depended on the PR: https://github.com/linux-nvme/libnvme/pull/1017.

ikegami-t:
Just sent the commit below to get the SQID and CID informtion from the driver before the PR but is this still not safe? By the way just thought to change the driver to abort a command with the opcode or etc., used by the userspace also.
  <https://lore.kernel.org/linux-nvme/20250524091637.6938-1-ikegami.t@gmail.com/T/#u>

keithbusch:
> Just sent the commit below to get the SQID and CID informtion from the driver before the PR but is this still not safe?

Absolutely not. Outside broken controllers, that sysfs interface proposal just displays a racy snapshot, and I don't think it is appropriate for sysfs anyway as it looks like a debug interface rather than a device attribute. Nvme devices do millions of IOPs, there's no way the output from reading a sysfs file is accurate before the text even hits the output steam.

keithbusch:
The only reason the kernel is able to do this at all is because it holds a reference on the targeted qid:cid pair that locks the context of the outstanding IO. It can't possibly be a different context in this path, we know exactly what we're aborting. You can't do that from user space because it doesn't own the hardware abstractions. There is zero reason to make this an easily reachable command, just like we don't support create/delete cq/sq.

ikegami-t:
I could understood well so let me close the PRs. Thank you for your explanation.
> Outside broken controllers, that sysfs interface proposal just displays a racy snapshot, and I don't think it is appropriate for sysfs anyway as it looks like a debug interface rather than a device attribute.

Yes the sysfs interface is for the debugging purpose so I will do consider more later if any other way to confirm the command not completed.

PR #2827, opened 2025-05-26
nvme: update device error handling for list-subsys

martin-gpy:
The list-subsys command does not properly check if the nvme device passed to the <device> option is a valid device or not. Fix the same.


nvme.c
```diff
@@ -3459,12 +3460,10 @@ static int list_subsys(int argc, char **argv, struct command *cmd,
...
 	if (devname) {
-		int subsys_num;
+		err = parse_and_open(&dev, argc, argv, desc, opts);
+		if (err)
+			return err;
 
-		if (sscanf(devname, "nvme%dn%d", &subsys_num, &nsid) != 2) {
-			nvme_show_error("Invalid device name %s", devname);
-			return -EINVAL;
-		}
```
igaw, line 3467: IIRC, `devname` a filter name, why should we verify if it is valid?

igaw:
Hmm, still not explaining why the filter doesn't work. I'll look into it when I find some cycles to work on.

PR #2829, opened 2025-05-27
nvme: return earlier when we have bdev-ns for smart-log command

yizhanglinux:


nvme.c
```diff
@@ -564,6 +564,11 @@ static int get_smart_log(int argc, char **argv, struct command *cmd, struct plug
 	if (err)
 		return err;
 
+	if (is_blkdev(dev)) {
```
martin-gpy, line 567: This looks wrong. We should be able to get the smart log page data from a NVMe block device itself, right?

igaw: As far I understand the specs says SMART is only for the controller not a namespace.

igaw:
So indeed this is not correct as @martin-gpy pointed out. Let's close this one and fix the blktests nvme/023 bug correctly.

PR #2831, opened 2025-05-28
wdc: Fix for vs-smart-add-log wdc plugin command

jeff-lien-sndk:
Support for the 0xC0 log page needs to be added for the SNTMP device.
Removed support for parsing the 0xCA log page for
the SNTMP device.
Update the wdc plugin version to 2.14.0
Remove vs-drive-info support for SNTMP


igaw:
Thanks!

PR #2833, opened 2025-05-31
nvme-print-stdout: add format command relatives prints

ikegami-t:
The output changed as TBD by the commit 18de3a6 before.

### DIFF
--- a/nvme-print-stdout.c
+++ b/nvme-print-stdout.c
@@ -5973,6 +5973,65 @@ static void stdout_pull_model_ddc_req_log(struct nvme_pull_model_ddc_req_log *lo
 	d((unsigned char *)log->osp, osp_len, 16, 1);
 }
 
+static void stdout_relatives(nvme_root_t r, const char *name)
+{
+	struct nvme_resources res;
+	struct htable_ns_iter it;
+	bool block = true;
+	bool first = true;
+	nvme_ctrl_t c;
+	nvme_path_t p;
+	nvme_ns_t n;
+	int nsid;
+	int ret;
+	int id;
+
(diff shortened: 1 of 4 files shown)


nvme-print-stdout.c
```diff
@@ -5973,6 +5973,63 @@ static void stdout_pull_model_ddc_req_log(struct nvme_pull_model_ddc_req_log *lo
 	d((unsigned char *)log->osp, osp_len, 16, 1);
 }
 
+static void stdout_relatives(const char *name, nvme_root_t r)
+{
+	int id;
+	int nsid;
+	int ret = sscanf(name, "nvme%dn%d", &id, &nsid);
```
igaw, line 5980: Please avoid mixing declaration and assignment. Also order the variables so we get an inverse Christmas tree style.

igaw (changes requested):

nvme-print.c
```diff
@@ -490,9 +491,24 @@ void nvme_show_single_property(int offset, uint64_t value64, nvme_print_flags_t
...
+	int err = 0;
+
+	_cleanup_nvme_root_ nvme_root_t r = nvme_create_root(stderr, log_level);
+
+	if (!r) {
+		nvme_show_error("Failed to create topology root: %s", nvme_strerror(errno));
+		return;
+	}
+
+	err = nvme_scan_topology(r, NULL, NULL);
```
igaw, line 505: Instead creating a new root and scan sysfs again, just pass in the `nvme_root_t` object into `nvme_show_relatives`.

ikegami-t:
Updated the changes as commented. Thank you.

PR #2842, opened 2025-06-13
wdc: Fix uuid-index setting for WD 0xC2 log page

jeff-lien-sndk:
Determine the current uuid-index value used for the WDC 0xC2 log page.
Update wdc plugin version to 2.14.3


igaw:
Thanks! Sorry for the delay.

PR #2843, opened 2025-06-14
nvme: fix mem leak in nvme copy

martin-gpy:
Valgrind revealed a mem leak in copy_cmd() due to the copy pointer not getting freed after usage. Fix the same.


igaw:
Thanks!

PR #2851, opened 2025-06-30
list-ctrl: limit to JSON and normal output

guoguojenna:
Binary format is not implemented for list-ctrl, so limit output to normal and json formats only.


igaw:
Thanks!

PR #2855, opened 2025-06-30
nvme: output nvme_alloc() error message

ikegami-t:
Currently only the value -ENOMEM returned for the error. Also fixed some nvme_alloc() error handling.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -190,6 +190,21 @@ conf.set(
     ),
     description: 'Does struct tm have a tm_gmtoff field?'
 )
+conf.set(
+    'HAVE_WEAK_MALLOC',
+    cc.links(
+        '''#include <malloc.h>
+           void *malloc(size_t size) {
+               return NULL;
+           }
+           int main(void) {
+               malloc(1);
+           }
+        ''',
+        name: 'weak_malloc'
+    ),
+    description: 'Is malloc() weak function?'
+)
 
 if cc.has_function_attribute('fallthrough')
   conf.set('fallthrough', '__attribute__((__fallthrough__))')
--- a/util/mem.c
+++ b/util/mem.c
@@ -3,7 +3,9 @@
 #include <unistd.h>
 #include <malloc.h>
 #include <string.h>
+#include <dlfcn.h>
 #include <sys/mman.h>
+#include <ccan/likely/likely.h>
 
 #include "mem.h"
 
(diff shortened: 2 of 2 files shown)


nvme.c
```diff
@@ -577,8 +577,10 @@ static int get_smart_log(int argc, char **argv, struct command *cmd, struct plug
 		flags |= VERBOSE;
 
 	smart_log = nvme_alloc(sizeof(*smart_log));
-	if (!smart_log)
+	if (!smart_log) {
+		nvme_show_error("failed to allocate aligned memory");
 		return -ENOMEM;
+	}
```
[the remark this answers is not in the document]
ikegami-t: Seems nvme_alloc() itself also can be added into the macro. By the way is there any actual example for the macro in nvme-cli, libnvme or linux kernel do you know? Thank you.

igaw:
That `nvme_alloc_huge` fails is not 0%. This is due to the type of allocation we do with it to: larg linear buffer allocation. 

But in case of `nvme_alloc` this is very unlikely to fail. The Linux virtual memory manager normally does overcommiting (see `/proc/sys/vm/overcommit_memory`. That means malloc returns a pointer but when we try to use it, the OOT kills the process.

And we have many mallocs around where we don't print anything. I'd rather leave this as it is. 

An idea I am toying with is that we just fail on allocation failures in nvme-cli. There is absolut no point in trying to recover from such scenario.  E.g. we could use these hooks for malloc (https://www.gnu.org/software/libc/manual/html_node/Hooks-for-Malloc.html) and call abort. This would get rid of the error handling code, no one ever tests:

```c
    ptr = malloc(sizeof(*ptr));
    if (!ptr)
        return -ENOMEM;
```
  
Though for libnvme this is not an option.

meson.build
```diff
@@ -190,6 +190,21 @@ conf.set(
...
+           void *malloc(size_t size) {
+               return NULL;
+           }
+           int main(void) {
+               malloc(1);
+           }
+        ''',
+        name: 'weak_malloc'
+    ),
+    description: 'Does struct tm have a tm_gmtoff field?'
```
igaw, line 206: The description needs to be updated.

util/mem.c
```diff
@@ -107,3 +108,21 @@ void nvme_free_huge(struct nvme_mem_huge *mh)
...
+		malloc_sym = dlsym(RTLD_NEXT, "malloc");
+
+	if (malloc_sym)
+		result = malloc_sym(size);
+
+	CHECK_NULL_PTR(result);
+
+	return result;
+}
+#endif /* HAVE_WEAK_MALLOC */
```
igaw, line 128: I guess we need the same thing at least for `calloc` and `realloc`.

util/mem.h
```diff
@@ -5,6 +5,12 @@
 #include <stddef.h>
 #include <stdbool.h>
 
+#define CHECK_NULL_PTR(ptr) \
+	do { \
+		if (!ptr) \
+			abort(); \
+	} while (false)
+
```
igaw, line 13: I don't think we need this macro as API. The whole point if this would be that we nvme-cli aborts directly when the allocation fails. Nothing should call this macro.

util/mem.c
ikegami-t: Thanks for your comment. I will do for nvme-cli3.

PR #2856, opened 2025-07-01
wdc: invalid customer ID fix

brandon-paupore-sndk:
The return value of this function represents what commands are supported by the target, so returning a -1 here incorrectly enables all capability checks. Instead, this updates the behavior to just log the error and continue to report the actual device capabilities.

Reviewed-by: Jeffrey Lien <jeff.lien@sandisk.com>


igaw:
Thanks!

PR #2857, opened 2025-07-07
nvme: do not report error when nvme_core module is not loaded

Cav4ever:
On nvme-cli 1.x, nvme list is no-ops when the nvme_core module isn't loaded:
```
# nvme list
Node                  SN                   Model                                    Namespace Usage                      Format           FW Rev
--------------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
```

But on nvme-cli 2.x, it fails when trying to open the nonexistent /sys/class/nvme. It would be nice for the command to be no-ops again.
```
 # nvme list
Failed to scan topology: No such file or directory
```

commit 283585a3 fixes the issue, but commit f17afe1 reverts the change.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -3507,6 +3507,19 @@ static bool nvme_match_device_filter(libnvme_subsystem_t s,
 	return false;
 }
 
+static int handle_scan_topology_error(int err)
+{
+	/* Do not report an error when nvme_core module is not loaded */
+	if (err == -ENOENT) {
+		if (log_level >= LIBNVME_LOG_INFO)
+			nvme_show_error("nvme modules not loaded");
+		return 0;
+	}
+
+	nvme_show_error("Failed to scan topology: %s", libnvme_strerror(-err));
+	return err;
+}
+
 static int list_subsys(int argc, char **argv, struct command *acmd,
 		struct plugin *plugin)
 {
@@ -3558,10 +3571,8 @@ static int list_subsys(int argc, char **argv, struct command *acmd,
 	}
 
 	err = libnvme_scan_topology(ctx, filter, (void *)devname);
-	if (err) {
-		nvme_show_error("Failed to scan topology: %s", libnvme_strerror(err));
-		return -errno;
-	}
+	if (err)
+		return handle_scan_topology_error(err);
 
 	nvme_show_subsystem_list(ctx, nsid != NVME_NSID_ALL, flags);
 
@@ -3641,10 +3652,8 @@ static int list(int argc, char **argv, struct command *acmd, struct plugin *plug
 		return -ENOMEM;
 	}
 	err = libnvme_scan_topology(ctx, NULL, NULL);
-	if (err < 0) {
-		nvme_show_error("Failed to scan topology: %s", libnvme_strerror(-err));
-		return err;
-	}
+	if (err < 0)
+		return handle_scan_topology_error(err);
 
 	nvme_show_list_items(ctx, flags);
 
@@ -6647,7 +6656,7 @@ static void show_relatives(const char *name, nvme_print_flags_t flags)
 
 	err = libnvme_scan_topology(ctx, NULL, NULL);
 	if (err < 0) {
-		nvme_show_error("Failed to scan topology: %s", libnvme_strerror(-err));
+		handle_scan_topology_error(err);
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


igaw:
The commit message should contain the reason why this change is necessary. And given the amount of time this exact code has been touched by either someone complaining it should print a warning or not I have the suspicion this is going to cause another discussion.

Cav4ever:
@igaw 
> The commit message should contain the reason why this change is necessary. 
Thank you for your review. I have revised the commit message; kindly take another look.

To ensure consistency, it would be better if the behavior of "nvme list" in version 2.x matched that of version 1.x when the NVMe modules are not loaded.

igaw:
I have to dig into it again. But one thing is, that I am extremely hesitant to change the behavior 3 years after we have released 2.x.

tbzatek:
\<OT\> Yeah, it's the kind of annoyance when user needs to modprobe manually. Similar situation with `nvme connect` missing the `nvme-fabrics` module or with `nvmetcli` complaining about a missing `nvmet` module. In all these cases user gets rather cryptic message, not immediately pointing to a module to be loaded.

igaw:
Please discuss this out with @martin-gpy who added this feature  f17afe176926 ("fabrics: print an error for ENOENT too")

Obviously, you can't have both.

igaw:
As it is clear, it's not possible to have both behavior at the same time, let's make it configurable.

We the recent refactoring of the global options handling, it's possible to introduce a global command line options something like `rm` has with `--interactive`  (need a better name for this I suppose). The default behavior would be the one from 1.x and for those who want more info can just do a

```shell
$ alias nvme='nvme --interactive'
```

Would this work for everyone?

@maurizio-lombardi

maurizio-lombardi:
@igaw While it's not ideal, I agree that we can't have both behaviors at once. I don't have a better alternative at the moment, so this is okay with me.

Cav4ever:
@igaw  I think that when the nvme_core module is not loaded, commands like "nvme discover" and "nvme connect" should report an error so that users are informed that the required NVMe modules need to be loaded.

On the other hand, commands such as "nvme disconnect", "disconnect-all", "nvme list", "nvme list-subsys", "nvme show-topology", and "nvme show-relatives" should not report errors in this situation, because there is no action required from the user, such as loading additional NVMe modules.

Currently, "nvme disconnect" and "disconnect-all" already behave this way and do not report errors when nvme_core is not loaded. I think "nvme list", "nvme list-subsys", "nvme show-topology", and "nvme show-relatives" should follow the same behavior for consistency.

The code in this PR has been updated accordingly. What do you think?

igaw:
I am fine with the approach, though what about:

```c
	err = libnvme_scan_topology(ctx, NULL, NULL);
	if (err < 0) {
		if (errno == ENOENT) {
			if (log_level >= LIBNVME_LOG_VEBOSE)
				nvme_show_error("nvme modules not loaded");
			return 0;
		}

		nvme_show_error("Failed to scan topology: %s", libnvme_strerror(-err));
		return err;
	}
```

this could be a helper function, so we don't have redundant code.

Cav4ever:
> I am fine with the approach, though what about:
> 
> ```c
> 	err = libnvme_scan_topology(ctx, NULL, NULL);
> 	if (err < 0) {
> 		if (errno == ENOENT) {
> 			if (log_level >= LIBNVME_LOG_VEBOSE)
> 				nvme_show_error("nvme modules not loaded");
> 			return 0;
> 		}
> 
> 		nvme_show_error("Failed to scan topology: %s", libnvme_strerror(-err));
> 		return err;
> 	}
> ```
> 
> this could be a helper function, so we don't have redundant code.

@igaw Thanks, I agree this is cleaner.
I updated the patch accordingly and refactored the repeated `libnvme_scan_topology()` handling into a helper function.
The new version has been pushed, please take another look.

igaw:
I've fixed the test condition, it should test the `err` variable not the `errno`. Sorry I posted an bogus example. Anyway, I gave it a quick test:

```shell
root@localhost:/home/wagi/work/nvme-cli-upstream/.build-debian# lsmod
Module                  Size  Used by
root@localhost:/home/wagi/work/nvme-cli-upstream/.build-debian# ./nvme list
root@localhost:/home/wagi/work/nvme-cli-upstream/.build-debian# echo $?
0
root@localhost:/home/wagi/work/nvme-cli-upstream/.build-debian# ./nvme list -v
nvme modules not loaded
```

The next step would be to figure out how to get the other case working. I've tried a quick hack with the alias trick, but the command line parser doesn't like the order of the arguments...

```shell
root@localhost:/home/wagi/work/nvme-cli-upstream/.build-debian# alias nvme="nvme -v"
root@localhost:/home/wagi/work/nvme-cli-upstream/.build-debian# nvme list
ERROR: Invalid sub-command 'v'
```

Let me look into this one. It might be useful for other use cases as well.

igaw:
Thanks!

PR #2858, opened 2025-07-07
feat: add volatile-wc command

ikegami-t:
This is for Volatile Write Cache feature to get and set.

### DIFF
--- a/plugins/feat/feat-nvme.c
+++ b/plugins/feat/feat-nvme.c
@@ -47,6 +47,7 @@ static const char *hctm_feat = "host controlled thermal management feature";
 static const char *timestamp_feat = "timestamp feature";
 static const char *temp_thresh_feat = "temperature threshold feature";
 static const char *arbitration_feat = "arbitration feature";
+static const char *volatile_wc_feat = "volatile write cache feature";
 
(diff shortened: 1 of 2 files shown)


keithbusch:
There's a *lot* of features, i don't think any one of them needs to be a top level command. We already have the "get-feature" and "set-feature" which can easily access this. If you want to have symbolic names instead of having to remember the feature-id value, then sure, that might be helpful.

keithbusch:
oh, you all have already made a sub-command for all these. Well, I guess it's too late to go back now.

ikegami-t:
Thanks for your reviewing and comments.
> There's a _lot_ of features, i don't think any one of them needs to be a top level command. We already have the "get-feature" and "set-feature" which can easily access this. If you want to have symbolic names instead of having to remember the feature-id value, then sure, that might be helpful.
Yes will try to add the symbolic names later.
> oh, you all have already made a sub-command for all these. Well, I guess it's too late to go back now.
The feat plugin commands can be set the feature dword or data fields values directly.

ikegami-t:
Just pushed the PR: #2859 for the symbolic names.

PR #2859, opened 2025-07-08
nvme: add get and set feature commands symbolic names

ikegami-t:
The feature-id parameter can be set the symbolic names also.


igaw:
Thanks!

PR #2861, opened 2025-07-09
CI: adopt kubevirt-action for self-hosted ARC

MaisenbacherD:
Previously we used two independent ARC deployments where one compiled/announced the kernel and the second spun up a VM that executed our tests.

This setup caused concurrency related issues in other projects that use the same infrastructure. An ARC can assign work to any available resource (in our case the VM) and thus we could not guaranty which kernel was used for testing.

We are now using a single ARC that can build and spawn a VM. To spawn and run commands in the VM we are now using the kubevirt-action defined in linux-blktests/blktests-ci.

### DIFF
--- a/.github/workflows/run-nightly-tests.yml
+++ b/.github/workflows/run-nightly-tests.yml
@@ -7,80 +7,80 @@ on:
     - cron: '0 01 * * *'
 
 jobs:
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #2862, opened 2025-07-10
nvme: add get-log command symbolic names

ikegami-t:
The log-id parameter can be set the symbolic names also.


igaw:
Nice one! Thanks!

PR #2865, opened 2025-07-14
udev: Add custom udev rules for HPE Alletra Storage

rmvasuki:
HPE Alletra Storage arrays support NVMe-oF block access.
1. The "round-robin" iopolicy is preferred for performance benefits.
2. Setting ctrl_loss_tmo to -1 for NVMe/TCP controllers enables indefinite reconnect attempts after a path loss, and disables purging of the path on the host.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -267,6 +267,7 @@ udev_files = [
   '65-persistent-net-nbft.rules',
   '70-nvmf-autoconnect.rules',
   '70-nvmf-keys.rules',
+  '71-nvme-hpe.rules',
   '71-nvmf-netapp.rules',
   '71-nvmf-vastdata.rules',
 ]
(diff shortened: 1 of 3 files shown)


igaw:
I'll merged it for now. If someone complains, it's easy to change :)

rmvasuki:
> I don't have any problems with adding this rule, but are you sure about `round-robin`? Did you have a look at `queue-depth`? `round-robin` has severe limits, e.g. it doesn't check the utilization of a link, if one link fails, all others are taken out as well.

Sorry, I missed responding to this. I will investigate queue-depth further and report back here.
Could you elaborate a bit on the single link failure taking down other links? We haven’t seen that kind of behaviour in our testing so far.

thanks!

igaw:
Ah, no worries. 

It's been a while since I last played with round-robin, but IIRC, the failing path significantly impacted the good paths for a while. It might also be related to the transport used, e.g., FC. Anyway, I can't remember the details, but generally, we (SUSE) advise using queue-depth as it results in better performance and handles error cases more effectively. @mwilck has a bit more experience with this.

PR #2872, opened 2025-07-22
nvme-cli: sanitize free() handling

martin-gpy:
As per the POSIX manual, free() is NULL-safe i.e. we could call it without first checking whether the respective pointer is non-NULL. So apply that here.

Suggested-by: Martin Belanger <martin.belanger@dell.com>


igaw:
Thanks!

Issue #2870, opened 2025-07-17
nvme cli v2.14: wdc/sndk plug in commands not working for SN655 drives, 2.11 working fine.


Sayoojntly:
**Steps**

-  Power ON NVMe OF enclosure with SN655.
- connect the drives to the host.
- check the plug in commands .

**Observations**
nvme cli v2.14: wdc/sndk plug in commands not working for CB-II drives
nvme cli 2.11 working fine.

**Sandisk plugin in nvme cli v2.14**

```
#nvme version
nvme version 2.14 (git 2.14)
libnvme version 1.14 (git 1.14)
nvme sndk get-drive-status /dev/nvme0
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
wdc_get_fw_cust_id: ERROR: WDC: 0xC2 Log Page entry ID 0x15 not found
wdc_get_enc_drive_capabilities: ERROR: WDC: invalid customer id
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: 0xC2 Log Page not supported, uuid_index: 0
```

**WDC plugin in nvme cli v2.14**

```
root@W6:~# nvme wdc get-drive-status /dev/nvme0
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
[...]

PR #2874, opened 2025-07-22
wdc: Fix for get-drive-status wdc plugin command

jeff-lien-sndk:
With fabric attached drives, the PCI device id is not accessible.  Therefore, capability checks based on device id are not possible.  Added a default else leg that should catch these cases.


igaw:
Fixes: #2870

PR #2878, opened 2025-07-24
Add support for retrieving data area 4

jeff-lien-sndk:
This change will check if data area 4 is supported and set the etdas bit in feature id 0x16 if supported.


igaw:
Thanks!

PR #2880, opened 2025-07-25
udev: rename 71-nvme-hpe.rules to 71-nvmf-hpe.rules

igaw:
The rules file naming is inconsistent with the other rules. Use the nvmf- prefix.

@rmvasuki


rmvasuki:
Ah, sorry, don't know how I missed that one.

PR #2881, opened 2025-07-28
sndk: Add support for the Sandisk UUID

jeff-lien-sndk:
This change will check for the Sandisk UUID first and fallback to the WDC UUID if the Sandisk UUID is not found.


igaw:
Thanks!

Issue #2876, opened 2025-07-24
Replace AppImage artifcat with the static build artifact


igaw:
The AppImage build started to fail because of broken dependencies. Let's replace this with the static build artifact which is way simpler to handle from our side.

ikegami-t:
Is the actual build failure mentioned below for the PR #2875 appimage check failed?
  <https://github.com/linux-nvme/nvme-cli/actions/runs/16491703761/job/46627720769?pr=2875>
```sh
INFO:root:Downloading: https://github.com/AppImage/AppImageKit/releases/download/continuous/runtime-x86_64
Traceback (most recent call last):
  File "/usr/local/bin/appimage-builder", line 8, in <module>
    sys.exit(__main__())
  File "/usr/local/lib/python3.8/dist-packages/appimagebuilder/__main__.py", line 50, in __main__
    invoker.execute(commands)
  File "/usr/local/lib/python3.8/dist-packages/appimagebuilder/invoker.py", line 29, in execute
    command()
  File "/usr/local/lib/python3.8/dist-packages/appimagebuilder/commands/create_appimage.py", line 27, in __call__
    self.primer.prime()
  File "/usr/local/lib/python3.8/dist-packages/appimagebuilder/modules/prime/appimage_primer.py", line 43, in prime
    self._get_appimage_kit_runtime()
  File "/usr/local/lib/python3.8/dist-packages/appimagebuilder/modules/prime/appimage_primer.py", line 101, in _get_appimage_kit_runtime
    request.urlretrieve(url, self.carrier_path)
  File "/usr/lib/python3.8/urllib/request.py", line 247, in urlretrieve
    with contextlib.closing(urlopen(url, data)) as fp:
  File "/usr/lib/python3.8/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.8/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/lib/python3.8/urllib/request.py", line 640, in http_response
    response = self.parent.error(
  File "/usr/lib/python3.8/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
[...]

Issue #2398, opened 2024-07-05
Request: please add the appimage to releases.


Samueru-sama:
I had been building nvme-cli statically for a very long time, and it wasn't until right now that I realized that you guys actually have a working appimage build, but it is only done for testing instead of being in the releases section. 

Is there something that's holding that back?

igaw:
No not really. Just someone has to do the work. This is not very high on my TODO list. It needs a bit of tinkering with the github workflows and actions which is always very time consuming.

PR #2882, opened 2025-07-29
build: replace AppImage with static build binary

igaw:
The AppImage build was introduced to give users access to a test binary. Before the switch to meson, the static build was used for this. In the meantime we figured out how to build a static build with meson, so there is no need for the AppImage workflow.

Start to add to the release versions of nvme the tag as postfix. Furthermore, upload the artifacts to github.

Fixes: #2876 #2398


MaisenbacherD:
I created a PR with some smaller improvements: https://github.com/linux-nvme/nvme-cli/pull/2884
The main difference is the version tag, queried by `git describe`.

PR #2884, opened 2025-07-29
build: replace AppImage with static build binary

MaisenbacherD:
The AppImage build was introduced to give users access to a test binary. Before the switch to meson, the static build was used for this. In the meantime we figured out how to build a static build with meson, so there is no need for the AppImage workflow.

Start to add to the release versions of nvme the tag as postfix. Furthermore, upload the artifacts to github.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/2876 https://github.com/linux-nvme/nvme-cli/issues/2398


scripts/build.sh
```diff
@@ -108,16 +107,6 @@ config_meson_coverage() {
         "${BUILDDIR}"
 }
 
-config_meson_appimage() {
-    CC="${CC}" "${MESON}" setup                 \
-        --werror                                \
-        --buildtype="${BUILDTYPE}"              \
-        --force-fallback-for=libnvme            \
-        --prefix=/usr                           \
```
igaw, line 116: We should to add this prefix to the static build I think

igaw: Come to think to it maybe we should use the build.sh static build for this as well and not the Makefile one.

MaisenbacherD:
From the build log in https://github.com/linux-nvme/nvme-cli/actions/runs/16595459723/job/46940875602:
```
/usr/bin/ld: subprojects/openssl-3.0.8/libcrypto.a(crypto_bio_bio_addr.c.o): in function `BIO_lookup_ex':
bio_addr.c:(.text+0xb07): warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

PR #2890, opened 2025-08-11
plugin/solidigm: Added EXTENDED vs-internal-log option.

lgdacunh:
EXTENDED option to extract logs that slows things down during vs-internal-log.


igaw:
Thanks!

PR #2898, opened 2025-08-21
Add stats support for EC2 local storage instance type

lingkai0110:
This patch is a follow up of (https://github.com/linux-nvme/nvme-cli/pull/2559).

It adds the stats support for EC2 local storage on top of EBS.

For local stroage, it supports a `--details / -d` option. In this mode it will print out detail io histograms for different block size ranges.


igaw:
Could you tweak the commit message slightly. Use a prefix for the commit title, e.g. 

` plugins/amzn: ...`

Also the link to github in the first sentence should be removed. The simple reason is that we don't want to tight in this platform, so every commit message should be complete in itself. Pointer to external sources is sometimes okay, but usually those links have a limited life time. I think you can just remove the first sentence.

igaw:
```
 /usr/bin/arm-linux-gnueabihf-gcc -Invme.p -I. -I.. -Iccan -I../ccan -Isubprojects/libnvme/src -I../subprojects/libnvme/src -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Werror -std=gnu99 -O3 -fomit-frame-pointer -D_GNU_SOURCE -include config.h -MD -MQ nvme.p/plugins_amzn_amzn-nvme.c.o -MF nvme.p/plugins_amzn_amzn-nvme.c.o.d -o nvme.p/plugins_amzn_amzn-nvme.c.o -c ../plugins/amzn/amzn-nvme.c
../plugins/amzn/amzn-nvme.c: In function ‘get_stats’:
../plugins/amzn/amzn-nvme.c:538:51: error: macro "amzn_print_json_stats" passed 2 arguments, but takes just 1
  538 |                 amzn_print_json_stats(&log, detail);
      |                                                   ^
../plugins/amzn/amzn-nvme.c:455: note: macro "amzn_print_json_stats" defined here
  455 | #define amzn_print_json_stats(log)
      | 
../plugins/amzn/amzn-nvme.c:538:17: error: ‘amzn_print_json_stats’ undeclared (first use in this function); did you mean ‘amzn_print_io_stats’?
  538 |                 amzn_print_json_stats(&log, detail);
      |                 ^~~~~~~~~~~~~~~~~~~~~
      |                 amzn_print_io_stats
```

lingkai0110:
Thanks. Will look into those suggestions.

lingkai0110:
@igaw updated and fixed the issue. Please have a look. Thank you.

PR #2900, opened 2025-08-26
ocp: Fix telemetry string log json format parsing

jeff-lien-sndk:
The parsing of the ocp 2.5 telemetry string log fails with buffer overflow errors.  The overflows are caused by arrays that aren't long enough to contain the data once converted to strings.  This change will fix the overflow failures.


igaw:
Fixed merge conflict.

Thanks!

PR #2903, opened 2025-08-29
plugins/ocp: Add --all-ns option to command set-error-injection.

lgdacunh:
Different spec versions ask for different valid nsid values.


igaw:
Oh, fun stuff....

Thanks!

PR #2915, opened 2025-09-24
CI: add nightly blktests

MaisenbacherD:
This introduces blktests which require nvme-cli to the nightly tests.

### DIFF
--- a/.github/workflows/run-nightly-tests.yml
+++ b/.github/workflows/run-nightly-tests.yml
@@ -16,7 +16,56 @@ jobs:
           repository: "linux-blktests/blktests-ci"
       #We don't have to build the kernel here, as we have a cron job running on
       #the k8s cluster that builds the linus-master target nightly.
-      - name: Run in VM
+      - name: Run blktests in VM
+        uses: ./.github/actions/kubevirt-action
+        with:
+          kernel_version: linus-master
+          vm_artifact_upload_dir: blktests/results/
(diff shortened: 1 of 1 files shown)


.github/workflows/run-nightly-tests.yml
```diff
@@ -16,7 +16,56 @@ jobs:
...
+            liburing-devel nbd device-mapper ktls-utils dosfstools \
+            bc libnl3-cli cryptsetup sg3_utils pciutils unzip jq git wget \
+            meson json-c-devel openssl-devel keyutils-libs-devel dbus-devel \
+            swig xz-devel libcurl-devel curl libarchive libarchive-devel json-c
+
+            # mdadm has a buffer overflow in v4.3 and was fixed with this commit:
+            # https://git.kernel.org/pub/scm/utils/mdadm/mdadm.git/commit/?id=7f960c3bd050e76f8bf0a8a0c8fbdcbaa565fc78
+            # The Fedora packages ship v4.3.
+            # We are now compiling and installing mdadm from source with the most
+            # recent known good commit. The v4.4 release does not compile.
```
igaw, line 41: Just thinking aloud here: it would be possible to package up this version special version and just install it.
Another solution would be to prebuild the container as we currently do in the container-ci project. I'd just mentioning this because the install step failed very often due to $reason. Since we have the containers pre build things got way more stable.

PR #2920, opened 2025-09-26
nvme: add `--block-size` option to IO interface to bypass identify command

NateThornton:
Sometimes it is desired to only issue the underlying I/O command through submit_io (compare, read, write). Currently, at least one identify namespace is used to discover the logical block size.

This change adds a new `--block-size=` option to the submit_io interface. When provided, the `--block-size=` is used as the logical block size in bytes, and the `--metadata-size=` is also used as the metadata size in bytes. Storage and Reference tag checks are bypassed.

(I am less certain about the `--metadata-size=` acting per-block, and how best to support or document it. One idea would be to take the metadata-size as a the total size across all blocks, and setting `ms = cfg.metadata_size / (cfg.block_count + 1)`. I'm open to ideas or changes here)


igaw:
Thanks!

igaw:
Right after merging I saw the the nvme-write documentation had an typo for the short-hand. Fixed it.

PR #2923, opened 2025-09-30
plugins/amzn: Enable namespace metrics for EC2 local storage

awz1234:
This patch enables namespace metrics for EC2 local storage.


awz1234:
added `Signed-off-by`.

Issue #2868, opened 2025-07-16
wdc/sndk plug-in not functioning as expected for SanDisk SN861.


Sayoojntly:
Steps

1. Power on the NVME OF enclosure with SN861 drives connected to host.
2. Check the wdc/sndk plug in commands.

Observations:
WDC Plug-in not functioning as expected
```
nvme wdc get-drive-status /dev/nvme1
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: wdc_get_dev_mng_log_entry: Detected unaligned end of the data. Data Offset: 0x8 Entry Size: 0x0, Remaining Log Length: 0xba Entry Id: 0x0
ERROR: WDC: C2 log id 0x8 not found with uuid index 0
Not found, requesting log page with uuid_index 0
ERROR: WDC: wdc_get_dev_mng_log_entry: Detected unaligned end of the data. Data Offset: 0x8 Entry Size: 0x0, Remaining Log Length: 0xba Entry Id: 0x0
ERROR: WDC: C2 log id 0x8 not found with uuid index 0
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: wdc_get_dev_mng_log_entry: Detected unaligned end of the data. Data Offset: 0x8 Entry Size: 0x0, Remaining Log Length: 0xba Entry Id: 0x0
ERROR: WDC: C2 log id 0x8 not found with uuid index 0
Not found, requesting log page with uuid_index 0
ERROR: WDC: wdc_get_dev_mng_log_entry: Detected unaligned end of the data. Data Offset: 0x8 Entry Size: 0x0, Remaining Log Length: 0xba Entry Id: 0x0
ERROR: WDC: C2 log id 0x8 not found with uuid index 0
ERROR: WDC: 0xC2 Log Page entry ID 0x8 not found
ERROR: WDC: wdc_get_pci_ids : Open vendor file failed
ERROR: WDC: wdc_get_dev_mng_log_entry: Detected unaligned end of the data. Data Offset: 0x8 Entry Size: 0x0, Remaining Log Length: 0xba Entry Id: 0x0
ERROR: WDC: C2 log id 0x8 not found with uuid index 0
Not found, requesting log page with uuid_index 0
[...]

PR #2926, opened 2025-10-02
wdc: Fix SN861 capabilities in NVME-OF cases

jeff-lien-sndk:
In NVMe over Fabric cases, the pci device id is not available to use for identifying a drive.  Therefore, other fields from the Wd  C2 log page are used instead.


jeff-lien-sndk:
> I suppose this should address some of the problems reported in #2868. Good to close then this one, or is there more to come?

Yes, this should fix the wdc plugin commands.  I have another fix coming for the sandisk plugin commands.

PR #2928, opened 2025-10-08
sndk: Support fabric attached SN861

jeff-lien-sndk:
Since the pci device id is not available in the fabric attached case, other fields in the C2 log page are used to identify the drive.

The vs-fw-activate-history and clear-fw-activate-history wdc plugin commands were ported to the sandisk plugin.

Update the required libnvme commit to pick up new required function - nvme_get_uuid_list.


jeff-lien-sndk:
@igaw Could you take a look at this PR and merge if all looks good?  Thanks.

igaw:
Oh sorry, fell through the cracks. Alright I'll wait for your update :)

PR #2940, opened 2025-10-15
plugin/dera: update dera smart-log-add command output

Dequan-Dera:
plugin/dera: update dera smart-log-add command output. Delete some smart log items.


igaw:
Thanks!

PR #2949, opened 2025-10-28
ibm persist events log

wen2xiong:
Add two IBM specific persistent events log:
DE-01: Change Definition
DE-02: Reported Error


igaw:
couple of changes from me:
- renamed `pel_header` to `nvme_show_pel_header`
- add prefix to subject

PR #2953, opened 2025-11-04
sndk: Add plugin command support for SN-861

jeff-lien-sndk:
These commit will add support for the vs-smart-add-log sandisk plugin command to the SN-861 drive when in a nvme over fabric attached case.


igaw:
Thanks!

PR #2967, opened 2025-11-10
meson: build libnvme directly instead of as a subproject

martin-belanger:
Change the build flow so that libnvme is no longer defined as a Meson subproject of nvme-cli. Instead, include libnvme's meson.build directly from the top-level meson.build.

This allows sharing a common ccan directory between nvme-cli and libnvme, and generates a single configuration header, project-config.h (formerly config.h), used by all components. The old file name conflicted with plugins/solidigm/solidigm-telemetry/config.h, hence the rename.

Also reformat all meson.build files to fix indentation inconsistencies and other cosmetic discrepancies.

@igaw - I tested different build options (enabling all possible options including generating the docs). Everything is  working although I couldn't test cross-compilation and other exotic options (e.g. muon). 

One thing that I wasn't sure about is the licensing which is slightly different between nvme-cli and libnvme. I have kept both licenses as follows. I made sure these would get used in the `nvme.spec.in` and `libnvme.spec.in`.

```
project(
    'nvme-cli', ['c'],
    meson_version: '>= 0.62.0',
    license: [
        'GPL-2.0-only',         # nvme-cli
        'LGPL-2.1-or-later',    # libnvme
    ],
    version: '3.0-a.1',
    default_options: [
        'c_std=gnu99',
        'buildtype=debugoptimized',
        'prefix=/usr/local',
        'warning_level=1',
        'sysconfdir=etc',
        'wrap_mode=nofallback',
    ],
)
```


martin-belanger:
@igaw - Regarding your earlier comment:
> we need a way to configure/build the library independent of nvme-cli. This means also we should be able to build nvme-cli without library. In this case the dependency for the library (e.g. OpenSSL) should not be required. This is an important feature the distros ask for.

Building the library (`libnvme.so`) without nvme-cli (`nvme`) is not a problem. However, building nvme-cli w/o the library present will result in a whole bunch of missing symbols (see example below). Maybe there's a gcc option to allow for that, but I'm not aware of it.

```
/usr/bin/ld: fabrics.c:(.text+0x3239): undefined reference to `nvme_ctrl_find'
/usr/bin/ld: fabrics.c:(.text+0x3281): undefined reference to `nvme_subsystem_next_ctrl'
/usr/bin/ld: fabrics.c:(.text+0x3319): undefined reference to `nvme_ctrl_get_subsysnqn'
/usr/bin/ld: fabrics.c:(.text+0x3326): undefined reference to `nvme_ctrl_is_persistent'
/usr/bin/ld: fabrics.c:(.text+0x3454): undefined reference to `nvme_free_ctrl'
/usr/bin/ld: fabrics.c:(.text+0x34e2): undefined reference to `nvme_host_is_pdc_enabled'
/usr/bin/ld: fabrics.c:(.text+0x34f2): undefined reference to `nvme_disconnect_ctrl'
etc...
```

Please let me know how you want to proceed.

martin-belanger:
> How do you build nvme-cli here? When building the project separately, the library needs to be build AND installed first., basically what ./scripts/build.sh distro does.

Sorry, I misunderstood. When you said "_This means also we should be able to build nvme-cli without library_" I thought you meant that we should be able to build nvme-cli w/o libnvme present. 

We are on the same page. This patch will allow building nvme-cli by itself as long as libnvme has previously been built and installed.

igaw:
> Thanks for the info. I tried it and it works. The only problem is that all the NBFT tests are failing because the utility nbft-dump is complaining of a bad format.

This is likely caused by a missing hooks in the kernel for executing cross compiled binaries. That is the hook is missing in  `/proc/sys/fs/binfmt_misc/`

PR #2968, opened 2025-11-10
plugins/sandisk:  Use Correct PCI Vendor ID for SNESSD1 Drives

jeff-lien-sndk:
The current Sandisk plugin code is using the Western Digital PCI vendor id for the SNESSD1 drives.  It should be using the Sandisk Vendor ID (0x15B7) instead.  The wdc plugin code also needs to be updated to recognize this new Sandisk device.


jeff-lien-sndk:
@igaw Dan, could your review and merge this PR today if possible?  Thanks

PR #2974, opened 2025-11-18
nvme-print-stdout: Display id-ctrl ONCS fields according to NVMe spec…

lgdacunh:
… v.2.3

Adjusted ONCS field descriptions and variable names to match spec names.


igaw:
Thanks!

PR #2977, opened 2025-11-22
libnvme/util: fix libnvme/src/nvme/util.c endian error

ikegami-t:
Since converted to big endian so fix to little endian.


igaw:
Thanks for working on such things!

PR #2983, opened 2025-11-27
nvme: refactor PI tagging for uniform behavior

MaisenbacherD:
Introduce `init_pi_tags` helper function that is called for all IO commands, to ensure uniform behavior.

Renaming related variables to match the spec.


igaw:
Just ingore those checkpatch warnings. It's just a stupid reporting tool and we should not blindly follow it. Your changes look good to me. The only thing I would probably change is the overlong lines for the function definition.

PR #2984, opened 2025-11-28
ioctl: fix dms command

igaw:
The DSM ranges data struct needs to be initialized. The rework dropped this part. Also the number of ranges for CDW10 is a 0's value (sic).

Fixes: 6c991dce76d4 ("src: rework nvme_dsm command")
Fixes: #https://github.com/nvme-experiments/nvme-cli/issues/97

### DIFF
--- a/libnvme/src/nvme/ioctl.h
+++ b/libnvme/src/nvme/ioctl.h
@@ -4393,9 +4393,9 @@ static inline int nvme_flush(struct nvme_transport_handle *hdl, __u32 nsid)
  * @cmd:	Passthru command to use
  * @nsid:	Namespace identifier
  * @nr:		Number of block ranges in the data set management attributes
- * @idr:	DSM Deallocate attribute
- * @idw:	DSM Integral Dataset for Read attribute
- * @ad:		DSM Integral Dataset for Read attribute
+ * @idr:	DSM Integral Dataset for Read attribute
+ * @idw:	DSM Integral Dataset for Write attribute
+ * @ad:		DSM Deallocate attribute
  * @data:	User space destination address to transfer the data
  * @len:	Length of provided user buffer to hold the log data in bytes
  */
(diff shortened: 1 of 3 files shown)


nvme.c
```diff
@@ -7364,13 +7364,17 @@ static int dsm(int argc, char **argv, struct command *acmd, struct plugin *plugi
...
+		cfg.ad = NVME_GET(cfg.cdw11, DSM_CDW11_AD);
+		cfg.idw = NVME_GET(cfg.cdw11, DSM_CDW11_IDW);
+		cfg.idr = NVME_GET(cfg.cdw11, DSM_CDW11_IDR);
+	}
 
 	dsm = nvme_alloc(sizeof(*dsm) * 256);
 	if (!dsm)
 		return -ENOMEM;
 
+	nvme_init_dsm_range(dsm, ctx_attrs, nlbs, slbas, nr);
```
MaisenbacherD: ~~`nlbs` is not correctly wired up with the `struct config` above.
The `char *blocks;` member of `struct config` must be `char *nlbs;` .~~
Edit: Sorry, it is correctly wired up with the `argconfig_parse_comma_sep_array_u32` helper.

I think we should rename the `struct config` members according to the spec to avoid further confusion :)

libnvme/src/nvme/ioctl.h
```diff
@@ -4410,7 +4410,7 @@ nvme_init_dsm(struct nvme_passthru_cmd *cmd,
 	cmd->nsid	= nsid;
```
MaisenbacherD: The docs for the arguments got a bit shuffled around.
Should be something like:
```
...
 * @idr:	DSM Integral Dataset for Read attribute
 * @idw:	DSM Integral Dataset for Write attribute
 * @ad: 	DSM Deallocate attribute
...
```

PR #2989, opened 2025-11-29
nvme: add support for nvme top

ikegami-t:
This is to monitor the system changes.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -272,6 +272,21 @@ conf.set(
     ),
     description: 'Does struct tm have a tm_gmtoff field?'
 )
+conf.set(
+    'HAVE_STDOUT_TO_SET',
+    cc.compiles(
+        '''
+            #include <stdio.h>
+            int main(void) {
+                char *buf;
+                size_t len;
+                stdout = open_memstream(&buf, &len);
+            }
+        ''',
+        name: 'stdout_to_set'
+    ),
+    description: 'Is stdout to set a mstream available?'
+)
 
 if cc.has_function_attribute('fallthrough')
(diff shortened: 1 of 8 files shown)


ikegami-t:
Just refixed the patch again so please review again and if any comment or advice please let me know.
  Note: Also tested the delay option on locally by changing the show-topology command as simulated.

ikegami-t:
Sure will do wait the nvme-top command. For now will close the PR then later will create a new PR for show-topology command if needed. Thank you so much for your reviewing and comments!

ikegami-t:
Just reopend the PR since seems still no update for the nvme-top command and just rebased the changes with the latest master and also changed to add the delay option into the smart-log and sanitize-log commands since those commands are not necessary to consider the key-down and key-up inputs as the outpus are not long lines. I will do fix separately the comment https://github.com/linux-nvme/nvme-cli/pull/2989#issuecomment-3782875953 since it needs more time to investigate then try to add the dealy option again into the show-topology command. Also will do consider to merge the delay feature functionality with the nvme-top comand after the nvme-top command applied.

ikegami-t:
Thank you so much for your update and explanation. My suggestion is the option 4 as to add the delay feature at first then update it to use the dashboard functionality later after it applied if you okay.

igaw:
I've merged @shroffni version of `nvme top`. Still many thanks for your PR. I know you spend a bit of time on this one. It's not an easy feature. Anyway, let's make `nvme top` even better :)

ikegami-t:
Thank you.

PR #3009, opened 2025-12-10
ioctl: avoid segfault for list & list-subsys commands

martin-gpy:
Below segfault seen in nvme_submit_passthru64() while running list and list-subsys commands:

scan controller nvme0
lookup subsystem /sys/class/nvme-subsystem/nvme-subsys0/nvme0 scan controller nvme0 path nvme0c0n1
failed to scan ctrl nvme0: Success
scan subsystem nvme-subsys0
scan subsystem nvme-subsys0 namespace nvme0n1

This is due to an invalid dereferencing of the nvme_transport_handle structure content at hdl->ctx->dry_run. Fix the same.

### DIFF
--- a/libnvme/src/nvme/ioctl.c
+++ b/libnvme/src/nvme/ioctl.c
@@ -126,7 +126,7 @@ static int nvme_submit_passthru32(struct nvme_transport_handle *hdl,
 	int err = 0;
 
 	user_data = hdl->submit_entry(hdl, cmd);
-	if (hdl->ctx->dry_run)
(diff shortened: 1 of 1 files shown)


igaw:
On my system with a recent kernel there are no ioctl. That means the fallback code for older kernels is trying to issue the command.

igaw:
There is plenty of documentation and analysis through the decades of computer on this topic: https://en.wikipedia.org/wiki/Fail-fast_system

I wont accept random `if (!ptr) return` patches.

PR #3014, opened 2025-12-12
libnvme: add accessor code generation

martin-belanger:
This adds a code generator that parses struct definitions and generates accessor functions (setters/getters).

Note that the generated code is added to `libnvme.so`, but it is not currently being called by anything. Also, the generated code is not added to the `libnvme.map` file. So, basically it is dead code...

The next step will be to actually invoke the generated functions and add them to the `libnvme.map` file. I need to check if we are allowed to have more than one `*.map` files. It would be easier to generate the map file for the accessor functions separately from the `libnvme.map`.


martin-belanger:
@igaw — I fixed the compilation issue from my last commit. When building libnvme separately from nvme-cli, there are several cases to handle—for example, whether libnvme is installed in a standard location (`/usr`, `/usr/local`) or in a non-standard `prefix`.

For non-standard install locations, the compiler must be explicitly told where to find `libnvme.so` and the libnvme headers using absolute paths, rather than relying on Meson’s automatic discovery.

This took some experimentation to get right, but I learned a few new things about Meson along the way. Figuring this out was a lot of fun... :stuck_out_tongue_winking_eye:

PR #3017, opened 2025-12-16
refactor fabrics: move discovery and connect logic code to the library

igaw:
In order to be able to really split the logic from the parsing code move the discovery and connect code to the library. Since some interaction between nvme-cli and library is necessary introduce a `struct nvmf_context` object which holds the configuration state and also callbacks. Since it should be extensible in future, don't expose the data structure. Just introduce setters for it. If we need  to extend it, we can just add new setters and don't break existing code. 

This series doesn't not really cleanup the existing code, it just moving around and starts introducing this new API. It needs more work but I think it's better to built on top of it instead trying to do all in one go.

This survives simple testing.


johnmeneghini:
This pull request broke nvme-connect-all and nvme-discover with /etc/nvme/discover.conf.

See the issue   https://github.com/linux-nvme/nvme-cli/issues/3437

Issue #2993, opened 2025-12-02
tests: Namespace Granularity Controller Attribute support breaks all tests


NateThornton:
When running any of the nvme tests, a controller that supports Namespace Granularity Attribute will fail in `setUp()`, with the call chain create_and_validate_ns() -> create-ns(). The create-ns command is run with `--verbose` option, which will elevate the log level to INFO, and thus the ["enforce [nsze|ncap] alignment to %PRIx64 because of namespace granularity requirement"](https://github.com/linux-nvme/nvme-cli/blob/bcf72cf8eb3a51180b048cc7a34e4445bebcbced/nvme.c#L3364) message will print every time. This causes the json parsing to fail.

Here, I print the command and the stdout:
```
nvme create-ns /dev/nvme0 --nsze=3750748848 --ncap=3750748848 --flbas=0 --dps=0 --verbose --output-format=json
enforce nsze alignment to 200 because of namespace granularity requirements
enforce ncap alignment to 200 because of namespace granularity requirements
{
  "create-ns":"success",
  "nsid":"1"
}
```

I'd be happy to fix this, but am looking for direction. Couple of solutions:

1. Drop the `--verbose` option in nvme_test.py
2. Only print the "enforce..." strings for normal output in nvme.c

Any preferences?

NateThornton:
@igaw Would this modification to `print_info` and `print_debug` align with what you were envisioning?

```c
// logging.h
#define print_info(...)					\
	do {						\
		if (is_printable_at_level(LOG_INFO))	\
			printf(__VA_ARGS__);		\
	} while (false)
#define print_debug(...)				\
	do {						\
		if (is_printable_at_level(LOG_DEBUG))	\
			printf(__VA_ARGS__);		\
	} while (false)

bool is_printable_at_level(int level);
```
```c
// logging.c
bool is_printable_at_level(int level)
{
	return log_level >= level && strcmp(nvme_cfg.output_format, "normal") == 0;
}
```

igaw:
Yes, this makes sense to me. Please send PR for this. Thanks!

PR #3023, opened 2025-12-22
logging: suppress logging except during normal output

NateThornton:
Suppress the logging macros `print_info` and `print_debug` except during "normal" output. This ensures that "json" and "binary" output formats do not include unnecessary warnings.

This was originally causing issue with create-ns command printing warnings about namespace granularity alignment when using JSON option. See issue #2993.


igaw:
Thanks!

Issue #3034, opened 2026-01-04
concat connections fail during connect-all


martin-gpy:
Seems secure channel concat connections fail during nvme connect-all while using a JSON config file, as shown below:

```
# cat /usr/local/etc/nvme/config.json
[
  {
    "hostnqn":"nqn.2014-08.org.nvmexpress:uuid:e6550026-173e-4959-ba74-be367844bd8a",
    "hostid":"e6550026-173e-4959-ba74-be367844bd8a",
    "subsystems":[
      {
        "nqn":"nqn.1992-08.com.netapp:sn.48391d66c0a611ecaaa5d039ea165514:subsystem.subsys_CLIENT116_0",
        "ports":[
          {
            "transport":"tcp",
            "traddr":"192.168.1.116",
            "host_traddr":"192.168.1.6",
            "trsvcid":"4420",
            "dhchap_key":"DHHC-1:01:gTtZdM+d+SjkxHcCcJKuI0OQKEnIJiXFASAlNDQP94FD/lKe:",
            "concat":"true"
          }
        ]
      }
    ]
  }
]
```
Now attempt a connect-all on the above JSON config file:

```
# nvme connect-all -J /usr/local/etc/nvme/config.json -vv
kernel supports: instance cntlid transport traddr trsvcid nqn queue_size nr_io_queues reconnect_delay ctrl_loss_tmo keep_alive_tmo hostnqn host_traddr host_iface hostid duplicate_connect disable_sqflow hdr_digest data_digest nr_write_queues nr_poll_queues tos keyring tls_key fast_io_fail_tmo discovery dhchap_secret dhchap_ctrl_secret tls concat recovery_delay
connect ctrl, 'nqn=nqn.2014-08.org.nvmexpress.discovery,transport=tcp,traddr=192.168.1.116,host_traddr=192.168.1.6,trsvcid=4420,hostnqn=nqn.2014-08.org.nvmexpress:uuid:e6550026-173e-4959-ba74-be367844bd8a,hostid=e6550026-173e-4959-ba74-be367844bd8a,ctrl_loss_tmo=600'
connect ctrl, response 'instance=0,cntlid=192'
nvme0: nqn.2014-08.org.nvmexpress.discovery connected
lookup subsystem /sys/class/nvme-subsystem/nvme-subsys0/nvme0
nvme0: nqn.2014-08.org.nvmexpress.discovery disconnected
[...]

PR #3035, opened 2026-01-04
fabrics: fix concat during connect-all

martin-gpy:
During nvme connect-all, if a discovery log page record reports the sectype as anything other than NVMF_TCP_SECTYPE_NONE in nvmf_connect_disc_entry(), it then assumes that tls should be default set for the same. But this holds true only for configured PSK TLS connections alone and not generated PSK TLS (i.e. secure channel concat) connections since both concat and tls flags are meant to be mutually exclusive. Fix the same.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/3034

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -1193,8 +1193,12 @@ static int nvmf_connect_disc_entry(nvme_host_t h,
 		c->cfg.disable_sqflow = true;
 
 	if (e->trtype == NVMF_TRTYPE_TCP &&
-	    e->tsas.tcp.sectype != NVMF_TCP_SECTYPE_NONE)
-		c->cfg.tls = true;
+	    e->tsas.tcp.sectype != NVMF_TCP_SECTYPE_NONE) {
+		if (e->treq & NVMF_TREQ_REQUIRED)
+			c->cfg.tls = true;
+		else if (e->treq & NVMF_TREQ_NOT_REQUIRED)
+			c->cfg.concat = true;
+	}
 
 	ret = nvmf_add_ctrl(h, c, cfg);
 	if (!ret) {


martin-gpy:
> I think the `nvmf_connect_disc_entry` function should not blindly enable `tls` in this case.
> 

Yes, I agree. This code snippet at nvmf_connect_disc_entry() is really unwarranted IMO. But anyways...

> Something like this here:
> 
> ```c
> 	if (e->trtype == NVMF_TRTYPE_TCP &&
> 	    e->tsas.tcp.sectype != NVMF_TCP_SECTYPE_NONE &&
> 	    !cfg->tls)
> 		c->cfg.tls = true;
> ```
> 
> If we check the command line for `--concat` and `--tls` and ensure that the user doesn't provide both options, then this here would be a good place to log what is happening. It seems relevant to report to me. What do you think?

This won't help because ```c->cfg.tls``` (or even ```c->cfg.concat``` for that matter) is always false at this point while parsing the discovery log page entry at nvmf_connect_disc_entry(). IIRC, the user provided options for these flags are updated only in nvmf_add_ctrl() alone during the merge_config(). 

That's the reason I made this fix in nvmf_add_ctrl() itself in the above commit, and not in nvmf_connect_disc_entry().

martin-gpy:
> If I read it correctly, the TSC value `not supported or support but not required` is not automatically mandating concatenation. Altough when `TASC` is it seems asking for concat. Do you have a pointer where `not supported` means concatenation should be used?

This was the interpretation following the discussion with Hannes:

|  Configuration  |     TASC     |       TSC      | SECTYPE | 
|          :---:          |      :---:       |       :---:     |     :---:     | 
| Configured PSK TLS with inband disabled | 00b (Not Specified)                   | 01b (Required)         | 02 (TLS 1.3) |
| Configured PSK TLS with inband enabled  | 01b (Auth Required)                 | 01b (Required)         | 02 (TLS 1.3) |
| Generated PSK TLS (with inband)                 | 10b (Auth with secure concat)  | 10b (Not Required)  | 02 (TLS 1.3) |

martin-gpy:
Any updates here? 

Without this fix, one is unable to secure concat TLS connect across several subsystems at one go using the JSON config file. Instead, one has to manually connect to each subsystem here one at a time, which is a big pain.

igaw:
I've started to look into it but run out of time.

PR #3045, opened 2026-01-05
docs: Fix read-the-docs

martin-belanger:
Allow Read-the-Docs web site to be able to build the docs.

@igaw - I made a change to the `meson.build` files to allow building the documentation without building anything else. This will allow the Read-the-Docs website to only build the documentation w/o anything else.

Note that the Read-the-Docs `libnvme` project (https://app.readthedocs.org/projects/libnvme/) needs to be updated to point to `linux-nvme/nvme-cli` instead of `linux-nvme/libnvme`.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -16,8 +16,8 @@ build:
     post_install:
       - pip3 install lxml
     pre_build:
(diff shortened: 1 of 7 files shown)


igaw:
Thanks!

Ideally we get the libnvme read the docs configuration so it has the docs for the 3.x and the 1.x version at the same time.

PR #3047, opened 2026-01-08
PyPI: add meson options to build without nvme executable

martin-belanger:
Now that nvme-cli and libnvme live in the same GitHub repository, meson builds both by default. When building the PyPI package, however, we only want to build `libnvme` and the Python bindings, not the `nvme` executable.

@igaw - I think this will fix the PyPI issue. This adds "`-Dnvme=disabled -Dlibnvme=enabled -Dpython=enabled`" to the `meson setup` options. If there are other options you would like to set (maybe `--prefix`, `--sysconfdir`, `--buildtype`, etc.) please feel free to update accordingly.


igaw:
Let's see if does the trick :)

PR #3054, opened 2026-01-11
nvme: fix get_pi_info() handling for submit_io() error

ikegami-t:
If nvme_identify_csi_ns() failed so caused floating point exception.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -7287,12 +7287,18 @@ static int get_pi_info(struct nvme_transport_handle *hdl,
 	if (!nvm_ns)
 		return -ENOMEM;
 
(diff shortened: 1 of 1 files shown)


nvme.c
igaw: When a function fails, in this case `get_pi_info`, it should not touch the output variables. Only on success it should update it. So I don't think there is need to change the above code.

Though I agree we should not blindly use `logical_block_size` here. I suggest to add a check here, when `logical_block_size` is 0, issue an error message an abort the operation. In this case without a valid `logical_block_size`, thus it's not possible calculate the `nblocks` value.

ikegami-t: Just fixed the patch as mentioned. Thank you.

igaw:
Thanks a lot! Nice fix!

PR #3055, opened 2026-01-11
Nvme show err

ikegami-t:


nvme-print.c
```diff
@@ -520,6 +520,16 @@ void nvme_show_status(int status)
 		ops->show_status(status);
 }
 
+int nvme_show_err(const char *msg, int err)
```
igaw: I think this function should not return an error code, so `void nvme_show_err`.

PR #3060, opened 2026-01-14
PyPI: Only use major version for libnvme.so

martin-belanger:
When building a pip (PyPI) package, do not set the libnvme.so version to major.minor.patch (e.g. libnvme.so.3.0.0), but instead just set it to the major (e.g. libnvme.so.3). pip/wheel do not care about the full versioning, and since symbolic links cannot be installed via pip (i.e. libnvme.so.3 -> libnvme.so.3.0.0), it's better to just skip the major.minor.patch versioning for PyPI.

To support this we're introducing a new config flag "pypi" to let meson know we're building for pip/PyPI.

@igaw - After doing more testing I found that the previous fix to set the symbolic links did not work. This PR reverts the  previous change and provides a simpler fix more in line with PyPI packages.

### DIFF
--- a/libnvme/src/meson.build
+++ b/libnvme/src/meson.build
@@ -55,25 +55,6 @@ libnvme = library(
     install: true,
 )
 
-# Explicitly create the symlinks in the same directory. When building the
-# PyPI package (pipx run build --sdist), the symbolic link don't get installed
-# unless we explicitly set them as below.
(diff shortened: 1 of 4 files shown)


igaw:
Okay, sounds like asking for trouble. Let's go with your solution. If it works we can still clean it up if needed.

PR #3064, opened 2026-01-16
fabrics: fix fabrics_connect() error message

martin-gpy:
Seeing below error message displayed during nvme connect error scenarios:

failed to connected: Unknown error -1013

Fix this by ensuring the appropriate ret value is passed to nvme_strerror(). And while at it, correct the message text grammar as well.


igaw:
Thanks!

Issue #2913, opened 2025-09-23
Prevent NVMe-oF boot devices from being removed by disconnect-all


johnmeneghini:
On systems that boot from SAN over NVMe-TCP or NVMe-FC, it is quite easy to make the system inoperable by issuing a 'nvme disconnect-all". For instance, after issuing the disconnect on a system which boots over NVMe-FC, the console reports the following:

```
[  103.827323] block nvme1n1: no available path - failing I/O 
[  103.827337] nvme1n1p4: writeback error on inode 67250623, offset 90112, sector 59489176 
[  103.827342] block nvme1n1: no available path - failing I/O 
[  103.827346] nvme1n1p4: writeback error on inode 67250623, offset 94208, sector 59489184 
[  103.832144] XFS (nvme1n1p2): Block device removal (0x20) detected at fs_bdev_mark_dead+0x43/0x60 (fs/xfs/xfs_super.c:1179).  Shutting down filesystem. 
[  103.832152] XFS (nvme1n1p2): Please unmount the filesystem and rectify the problem(s) 
[  103.832200] block nvme1n1: no available path - failing I/O 
[  103.832204] XFS (nvme1n1p4): log I/O error -5 
[  103.832206] XFS (nvme1n1p4): Filesystem has been shut down due to log error (0x2). 
[  103.832207] XFS (nvme1n1p4): Please unmount the filesystem and rectify the problem(s). 
```

igaw:
The same problem exists with any FS mounted from an 'attached' storage. As far I am aware the general consensus here is, just don't do it.  `ip down`, `ifconfig` etc. don't prevent you to shutdown an networking interface which is serving a mounted FS.

I am not convinced that adding even more logic to the ill defined 'connect-all' and 'disconnect-all' command is not going to make things better. 

Anyway, I assume you are also attending the ALPSS conference; we can discuss this it there.

tbzatek:
There are numerous occassions when you can shoot yourselves in the foot. Preventing that is often a battle with a wind. This bugreport has been sitting downstream for a while yet it was treated as low priority and seemingly not really viable for spending much time on it.

One idea was to somehow mark the `nbftX` network interfaces as private in NetworkManager, but after some other related discussions and the general upstream tendency to avoid network interface renaming this idea didn't make much sense after all.

Another idea was to use nvme-cli application strings to tag nbft-initiated connections, so they're excluded from 'disconnect-all' and such. No attempt to test this in practice that was ever made.

[...]

PR #3066, opened 2026-01-17
Allow service to also disconnect at shutdown

dschwartz783:
Added to keep service running in background, so that nvme can be disconnected at shutdown.

Currently, at least on Ubuntu, a variety of services hang due to the nvme device files being tied to active connections, which no longer work once the network has been brought down. This fixes that. Connections will be terminated before passing network-online.target.


tbzatek:
See also #2913

tbzatek:
I believe we can close this in favour of #2913?

igaw:
yes, one issue for this topic is enough.

PR #3074, opened 2026-01-23
ocp: Fix core dump during telemetry log decode

jeff-lien-sndk:
When decoding ocp 2.5 telemetry log with json output the internal-log ocp plugin command was hitting an issue that caused a core dump.  This commit will resolve that error.

Here's the actual error that this PR resolves:
nvme ocp internal-log /dev/nvme0n1 -f fifo_da_1-2_cctemp-telemetry.bin -s fifo_da_1-2_cctemp-string.bin --output-format json -t host -d 2

Missing telemetry-log. Fetching from drive...

telemetry.bin generated. Proceeding with next steps.

Extracting Telemetry Host Dump (Data Area 2)...
nvme: ./json_object.c:1490: json_object_array_add: Assertion `json_object_get_type(jso) == json_type_array' failed.
Aborted (core dumped)

### DIFF
--- a/plugins/ocp/ocp-nvme.h
+++ b/plugins/ocp/ocp-nvme.h
@@ -11,7 +11,7 @@
 #if !defined(OCP_NVME) || defined(CMD_HEADER_MULTI_READ)
 #define OCP_NVME
 
-#define OCP_PLUGIN_VERSION   "2.16.0"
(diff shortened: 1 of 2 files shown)


jeff-lien-sndk:
@igaw Hey Dan, could you review and merge (if no issues)?  We have a few users looking for these fixes.  Same for Brandon's PR #3082.  Thanks

PR #3079, opened 2026-01-29
Fix build failure in generate-accessors with -Werror set.

jimmunn:
The generate-accessors native executable uses asprintf() which requires _GNU_SOURCE. While the main project defines this flag via add_project_arguments(), native executables don't inherit these settings and need it explicitly defined.

This fix adds -D_GNU_SOURCE to the c_args for the generate_accessors executable, resolving the implicit declaration warning that becomes a build error when -Werror is enabled.

### DIFF
--- a/libnvme/src/nvme/meson.build
+++ b/libnvme/src/nvme/meson.build
@@ -20,6 +20,7 @@ headers_to_scan = [
 generate_accessors = executable(
     'generate-accessors',
     'generate-accessors.c',
+    c_args: ['-D_GNU_SOURCE'],
     dependencies: [
         config_dep,
     ],


martin-belanger:
Hi @jimmunn - Just curious. You mention that _native executables don't inherit settings_ set by `add_project_arguments()`. I was just reading the documentation on this function, and it seems that the arguments should apply to everything in this project. If not, there will be issues with other components. Should we use `add_global_arguments()` instead? However, the documentation for `add_global_arguments()` says: _Usually you should use add_project_arguments() instead, because that works even when your project is used as a subproject._

Ref: https://mesonbuild.com/Reference-manual_functions.html#add_global_arguments

igaw:
Just reformatted the commit message, so it doesn't have overlong lines.

PR #3080, opened 2026-01-29
nvme: Remove duplicated variables

sc108-lee:
already defined as static const char variable

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -2710,7 +2710,6 @@ static int list_ns(int argc, char **argv, struct command *acmd, struct plugin *p
 	const char *desc = "For the specified controller handle, show the "
 		"namespace list in the associated NVMe subsystem, optionally starting with a given nsid.";
 	const char *namespace_id = "first nsid returned list should start from";
-	const char *csi = "I/O command set identifier";
 	const char *all = "show all namespaces in the subsystem, whether attached or inactive";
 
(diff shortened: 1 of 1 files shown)


igaw:
Hmm, it is a bit odd that we have some of the variables as globals and some not. Can't we come up with something more sane in the long run. Any ideas?

igaw:
So with the recent improvement of the LLMs I think this change can be done very simple. I'll try it later.

sc108-lee:
sorry for late reply
I think your suggestions are reasonable.

I look forward to seeing the LLM based results.

If we proceed with an LLM-based implementation,
it would be beneficial to manage the rules for modifying descriptions in an MD file that is easily understood by both users and the LLM.
This way, whenever a new option is added, it can be generated based on that file.

igaw:
Personally, I’m not opposed to using “smarter” tools to assist with programming. So far, though, the results have been a bit of a mixed bag. These tools work well for renaming functions or handling simple, repetitive tasks like this, but they’re less effective for more complex work. Of course, things are evolving quickly in this field, so who knows how good these tools will become.

In any case, I think it would be great if we started improving our documentation for both users and developers. Martin has already made some improvements recently, which is highly appreciated. If you’d like to contribute in this area, feel free to do so.

Anyway, let me share the result of the refactoring patch.

igaw:
Let's move the discussion the the draft PR.

PR #3090, opened 2026-02-10
libnvme: Add nvme_getifaddrs()

martin-belanger:
The POSIX `getifaddrs()` API returns the list of network interfaces on the system, but invoking it can be costly. On large-scale systems with hundreds of NVMe-over-TCP connections, this API may be called hundreds of times, resulting in increased latency.

This patch introduces `nvme_getifaddrs()`, a wrapper around `getifaddrs()` that caches the results on the first invocation and reuses the cached data on subsequent calls.

Fixes: https://github.com/linux-nvme/libnvme/issues/1098

### DIFF
--- a/libnvme/src/nvme/private.h
+++ b/libnvme/src/nvme/private.h
@@ -13,6 +13,8 @@
 #include <poll.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
+#include <sys/types.h>
+#include <ifaddrs.h>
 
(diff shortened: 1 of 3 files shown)


martin-belanger:
Ah! You're right. It's been so long since I've worked on libnvme that I forgot how to access the ctx from a ctrl or subsys object. Let me rework this.

igaw:
Yes, this looks way better! Thanks!

PR #3093, opened 2026-02-12
nvme: standardize nvme_strerror() usage

martin-gpy:
nvme_strerror() is a useful wrapper function for printing proper error messages corresponding to error codes including library specific error codes too. So it would be helpful to standardize the use of the same here across the codebase.


martin-gpy:
> @igaw - Are you fine with the responses I posted above? Let me know your thoughts.

I actually went ahead and added new commits to this same pull request to address the review comments raised by Copilot. So I believe we should be good now. Thanks.

PR #3097, opened 2026-02-13
udev: don't spread variables through all nodes

howl:
This has been discussed here https://github.com/systemd/systemd/issues/40670

NVME_HOST_IFACE=none is being set to every existing node device but this
shouldn't be done, we don't want this in /dev/peanutbutter.

Set it only for nvme subsystem devices.

While at it remove all the action change conditions that are always true
given the if it's any other kind of action a goto end in the beginning.

### DIFF
--- a/nvmf-autoconnect/udev-rules/70-nvmf-autoconnect.rules.in
+++ b/nvmf-autoconnect/udev-rules/70-nvmf-autoconnect.rules.in
@@ -7,24 +7,24 @@
 ACTION!="change", GOTO="autoconnect_end"
 
 # For backwards compatibility. Make sure HOST_IFACE is not an empty string.
-ENV{NVME_HOST_IFACE}=="", ENV{NVME_HOST_IFACE}="none"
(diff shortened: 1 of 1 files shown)


igaw:
Could you elaborate 'we don't want'? Who is we and why? Just trying to figure out what the problem is. Maybe you could share some logs or explain what's going wrong. Thanks!

howl:
@igaw you need to ask? Is anything but not normal to set variables unconditionally to absolutely all the nodes in dev.

Just reporting it into systemd @yuwata set this to v260 milestone and look the issue was here.

No need for logs just do yourself sudo udevadm trigger and then udevadm info yourself the device you want as is spreaded over absolutely all nodes, I suggest /dev/zero, /dev/null, /dev/full, /dev/kmesg you shouldn't do that by no mean.

I have set it to the just the only same subsystem this project checks that variable and removed unnecessary conditionals.

yuwata (approved):

PR #3110, opened 2026-02-23
micron: add smart-log and id-ctrl commands

swapnansam:
Add smart-log command to retrieve SMART/Health log with vendor-specific
  OLEC and IPM fields. Add id-ctrl command to display Identify Controller
  with vendor-specific PMS, IPMSR, and MSMT fields.

  Signed-off-by: Swapna Samoju <ssamoju@micron.com>


swapnansam:
Thank You @igaw

PR #3118, opened 2026-02-27
feat: fix bugs in arbitration and power-thresh commands

jimmunn:
arbitration_set() used temp_thresh_feat instead of arbitration_feat in the error path, and feat_arbitration() passed TEMP_THRESH_DESC instead of ARBITRATION_DESC to parse_and_open().

power_thresh_set() used __u8 instead of __u16 for the ptv parameter, truncating the power threshold value.


ikegami-t:
Looks good. Thank you for your fix. Sorry for the errors.

PR #3119, opened 2026-02-27
Sndk plugin fixes

jeff-lien-sndk:
This PR contains fixes or changes for the following issues:

- Add Support for SN655 UUID checking in sandisk plugin
- Add Clear Assert Capability for new drives
- vs-internal-log Sandisk plugin command does not set data area value correctly
- Updates to vs-internal-log plugin command


igaw:
No worries, happens :)

I squashed the fixup patch into the last one, which broke the build. The goal is that every commit builds. Helps in cased someone bissects the tree.

Issue #3105, opened 2026-02-20
blktests nvme/046 failure


kawasaki:
When I tested the latest libnvme and nvme-cli, I observed blktests test case nvme/046 failed.

```
nvme/046 => nvme0n1 (basic test for unprivileged passthrough on /dev/ngX) [failed]
    runtime  0.248s  ...  0.375s
    --- tests/nvme/046.out      2023-04-06 10:11:07.925670528 +0900
    +++ /home/shin/Blktests/blktests/results/nvme0n1/nvme/046.out.bad   2026-02-20 16:35:18.834545847 +0900
    @@ -1,2 +1,5 @@
     Running nvme/046
    +Error: io-passthru read failed
    +Error: id-ns failed
    +Error: id-ctrl failed
     Test complete
```

The failure was not observed with older libnvme/nvme-cli (1.15/2.15). I tried to bisect pinning libnvme version to the git hash d65b44cd. The failure was observed with nvme-cli git hash 4fb93ad2d, and was not observed with git hash 77b6d0c72. There are some commits between these commits, but they all look failing to build nvme-cli.

I guess and hope that libnvme/nvme-cli experts' knowledge to resolve this failure quicker than my effort.

ikegami-t:
Seems currently only expected the device: /dev/nvmeX as direct device but not for the device: /dev/ngX.
```c
static int __nvme_transport_handle_open_direct(struct nvme_transport_handle *hdl, const char *devname)
{
	struct nvme_passthru_cmd dummy = { 0 };
	_cleanup_free_ char *path = NULL;
	char *name = basename(devname);
	int ret, id, ns;
	bool c;

	hdl->type = NVME_TRANSPORT_HANDLE_TYPE_DIRECT;

	ret = sscanf(name, "nvme%dn%d", &id, &ns);
	if (ret != 1 && ret != 2)
		return -EINVAL;
```

ikegami-t:
Note: Still failed the id-ctrl and show-topology commands on the older version (commit: 77b6d0c) with the device: /dev/ngX. (So the issue is not only for the latest version as existed on the older version on some commands.)

igaw:
should work with the fix from #3107 

```shell
nvme/046 => nvme0n1 (basic test for unprivileged passthrough on /dev/ngX) [passed]
    runtime  0.518s  ...  0.509s
```

Thanks @ikegami-t for the fix!

kawasaki:
[...]

PR #3120, opened 2026-03-01
nvme: allow char dev filter for show-pology and list-subsys commands

ikegami-t:
Previously only the block device can be set for the commands filter. This allows the filer for both the nvme and ng character devices.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -3456,19 +3456,29 @@ parse_lba:
 	return err;
 }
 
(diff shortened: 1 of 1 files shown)


ikegami-t:
This is for the issue mentioend the issue comment: https://github.com/linux-nvme/nvme-cli/issues/3105#issuecomment-3940568885.

Issue #3128, opened 2026-03-03
Host Identifier is showing only 64bit in get feature command output for FID 81h


sukhi61166:
The drive supports 128-bit Host Id. when get-feature command is used without the -H option, it returns full 128-but string but with -H it trucates to 64-bit. 

```
[mlx5/mlx5/i40e/i40e/i40e/i40e]root@perf-256-2:/home/kazan/suki# nvme --version
nvme version 2.16 (git 2.16)
libnvme version 1.16.1 (git 1.16.1)
```

```
nvme get-feature --feature-id=0x81 -c 1 --sel=0 /dev/nvme1
get-feature:0x81 (Host Identifier), Current value:00000000
       0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
0000: a9 ff 96 00 3a c5 11 eb 80 00 3c ec ef 1f c3 22 "....:.....<....""
[mlx5/mlx5/i40e/i40e/i40e/i40e]root@perf-256-2:/home/kazan/suki# nvme get-feature --feature-id=0x81 -c 1 --sel=0 /dev/nvme1 -H
get-feature:0x81 (Host Identifier), Current value:00000000
        Host Identifier (HOSTID):  16938536526425817001
 

```

```
mlx5/mlx5/i40e/i40e/i40e/i40e]root@perf-256-2:/home/kazan/suki#  nvme id-ctrl /dev/nvme1 -H | grep -i 128-bit
  [0:0] : 0x1   128-bit Host Identifier Supported
[mlx5/mlx5/i40e/i40e/i40e/i40e]root@perf-256-2:/home/kazan/suki# nvme get-feature --feature-id=0x81 -c 1 --sel=0 /dev/nvme1 -vv -H
opcode       : 0a
flags        : 00
rsvd1        : 0000
nsid         : ffffffff
cdw2         : 00000000
cdw3         : 00000000
data_len     : 00000010
metadata_len : 00000000
addr         : 5e7c13a4a000
metadata     : 0
cdw10        : 00000081
cdw11        : 00000001
cdw12        : 00000000
cdw13        : 00000000
cdw14        : 00000000
cdw15        : 00000000
timeout_ms   : 00000000
result       : 00000000
err          : 0
latency      : 24357 us
get-feature:0x81 (Host Identifier), Current value:00000000
        Host Identifier (HOSTID):  16938536526425817001
[...]

PR #3136, opened 2026-03-08
nvme-print-json: fix get feature 128-bit host id print

ikegami-t:
Since implemented as always 64-bit host id print.


ikegami-t:
Changes for the json output version fix based on PR: #3134 for the issue: #3128.

PR #3138, opened 2026-03-09
Set/update SPDX and Copyright statements

igaw:
Update and/or set the SPDX tag and, where possible, also the copyright statement.

I've used [reuse](https://reuse.software/)  to identify files that are missing the SPDX tag and copyright statements and updated them where it was clear how to do so. I asked Hannes and Martin in advance whether they were okay with changing licenses. This involves either changing GPL to LGPL or replacing Apache with GPL/LGPL. Both gave me their approval.

This work is not complete. There are still many files missing copyright statements, and it is not clear what to do in those cases. I think I could ask the original authors if they could provide them (e.g., Jeremy, Margin B., ...), but I am not sure how to proceed with all those files from Keith when he was still at Intel.

Another question I have is what to do with all the documentation (Documentation and libnvme/doc). I am not sure what the original intended license is. If I understand correctly, it may not be wise to release them under GPL or LGPL. The GNU project provides specific licenses intended for documentation.

@hreinecke 
@jk-ozlabs 
@martin-belanger
@keithbusch

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,48 @@
+<!-- SPDX-License-Identifier: GPL-2.0-only -->
+<!-- Copyright (C) 2026 Daniel Wagner, SUSE LLC -->
+<!-- Author: Daniel Wagner <dwagner@suse.de> -->
+# Security Policy
+
+## Reporting a Vulnerability
+
+If you discover a security vulnerability in this project, please report it
+responsibly.
+
+Please send an email to:
+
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
> I think I could ask the original authors if they could provide them (e.g., Jeremy, Margin B., ...)

Yes, absolutely fine here. The mi*.[ch] files should have existing copyright notices (and I am fine to replace those with SPDX equivalents), but let me know whether you need additional headers/SPDX tags elsewhere too.

PR #3237, opened 2026-04-03
windows: added platform abstraction headers

brgoing:
Added Windows and Linux platform abstraction headers.  Updated files that include platform-specific files to include the new platform compatibility headers.

Platform abstraction headers get installed under the platform subfolder.

### DIFF
--- a/libnvme/src/meson.build
+++ b/libnvme/src/meson.build
@@ -160,3 +160,14 @@ install_headers(
     subdir: 'nvme',
     install_mode: mode,
 )
+
+# Install platform abstraction headers
+install_headers([
+        'platform/includes.h',
+        'platform/linux.h',
+        'platform/windows.h',
+        'platform/types.h'
+    ],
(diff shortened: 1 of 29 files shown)


igaw:
As discussed during our call, I am not really a big fan of this approach. It creates a header file which includes everything and everything is depended on this header. From experience this gets easily out of hand. Thus I decided to go through the existing code and split/refactor the code so that there is generic code which is supposed to work on all platforms and platform depended code. The build system needs to pull in the right files. In the end the only Windows specific changes are the the new headers types.h (`__u8`, `__u16`, ...) and endian.h (`htole16`, ...). Overall I really think this is a better approach thus I've merged my #3270. Hope this is okay.

Obviously, there is still more cleanups necessary (I am working on refactoring the fabrics code for the libnvme_fabrics_option removal from the API anyway) but I think with this in place the porting path is unlocked.

PR #3259, opened 2026-04-09
libnvmf: replace nvmf_get_discovery_log() with opaque-args API

martin-belanger:
This PR contains 4 commits: 3 preparatory bug-fix / clean-up commits followed by the actual refactor.

**Commit 1** – bug fix: nvmf: fix timeout not applied to discovery log commands
The `args->timeout` value was never forwarded to `cmd.timeout_ms` in `nvme_discovery_log()`, so all three Get Log Page commands always used the kernel default timeout.

**Commit 2** – bug fix: nvmf: fix generator emitting setters for read-only members in .ld files
The generator's `generate_ld()` was unconditionally emitting both getter and setter entries for every struct member, ignoring the `is_const` flag that correctly suppresses setters in the .h/.c output. Regenerate accessors.ld to drop the spurious setter entries.

**Commit 3** – clean-up: nvmf: clean up discovery args struct initialization
Remove redundant zero-initializers (`.result = 0`) and replace the magic literal 0 for lsp with the named constant `NVMF_LOG_DISC_LSP_NONE`.

**Commit 4** – refactor: nvmf: replace `nvmf_get_discovery_log()` with opaque-args API
Introduce `struct nvmf_discovery_args` as an ABI-stable opaque object with auto-generated setter/getter accessors. The mandatory controller argument moves to the function signature; optional parameters (`timeout`, `max_retries`, `lsp`) are set via accessors. Callers may pass `NULL` for args to get built-in defaults. Remove the old `nvmf_get_discovery_wargs()` API entirely.

Split the accessor generator output into a common NVMe triplet (`accessors.{h,c,ld}`) and an NVMe-oF-specific triplet (`nvmf-accessors.{h,c,ld}`) so that non-fabrics builds can exclude all fabrics code. The generator script is made generic; two meson targets (`update-common-accessors`, `update-fabrics-accessors`) are exposed under the `update-accessors` alias.

### DIFF
--- a/libnvme/examples/discover-loop.c
+++ b/libnvme/examples/discover-loop.c
@@ -51,12 +51,13 @@ static void print_discover_log(struct nvmf_discovery_log *log)
 
 int main()
 {
-	struct nvmf_discovery_log *log;
+	struct nvmf_discovery_log *log = NULL;
 	struct libnvme_global_ctx *ctx;
 	libnvme_host_t h;
 	libnvme_ctrl_t c;
 	int ret;
 	struct libnvme_fabrics_config cfg;
+	struct nvmf_discovery_args *args;
 
 	nvmf_default_config(&cfg);
 
@@ -86,7 +87,13 @@ int main()
 		return 1;
 	}
 
-	ret = nvmf_get_discovery_log(c, &log, 4);
+	ret = nvmf_discovery_args_create(&args);
+	if (!ret) {
+		nvmf_discovery_args_set_max_retries(args, 4);
+		ret = nvmf_get_discovery_log(c, args, &log);
+		nvmf_discovery_args_free(args);
+	}
+
 	libnvme_disconnect_ctrl(c);
 	libnvme_free_ctrl(c);
 
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -36,6 +36,7 @@
 
 #include "cleanup.h"
 #include "private.h"
+#include "private-fabrics.h"
 #include "compiler_attributes.h"
 
 const char *nvmf_dev = "/dev/nvme-fabrics";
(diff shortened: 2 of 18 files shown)


libnvme/examples/discover-loop.c
```diff
@@ -86,7 +86,16 @@ int main()
 		return 1;
 	}
 
-	ret = nvmf_get_discovery_log(c, &log, 4);
+	{
+		struct nvmf_discovery_args *dargs;
```
igaw: Just move the decleration at the beginnning of the function, this avoids this block then. It just looks a bit funky.

libnvme/src/nvme/fabrics.c
```diff
@@ -1421,37 +1425,57 @@ static void sanitize_discovery_log_entry(struct libnvme_global_ctx *ctx,
...
+
+	if (!argsp)
+		return -EINVAL;
+
+	args = calloc(1, sizeof(*args));
+	if (!args)
+		return -ENOMEM;
 
+	args->max_retries = 6;
+	args->timeout = NVME_DEFAULT_IOCTL_TIMEOUT;
```
igaw: That reminds me, I've ripped out the timeout argument from the init commands, because everything was just using the defaults. I think we could do this here as well. If we need to change the default we can add it due the new API design  without a problem.

igaw:
Looks good. Just the one nitpick and as I said, I would drop the timeout value for now. There is little point explicitly setting the default value '0' which means the kernel is using its defaults. If there are commands which need a special treatment we can add it without a problem for this specific case.

libnvme/examples/discover-loop.c
martin-belanger: Done.

libnvme/src/nvme/fabrics.c
martin-belanger: Done

igaw:
Thanks!

PR #3260, opened 2026-04-09
fabrics: issue a warning on invalid traddr format

igaw:
Report that the traddr is invalid instead silently fix it only.


igaw:
Fixes: #3173

PR #3264, opened 2026-04-10
nvme-print: fix traddr transformation for ave discovery log

igaw:
inet_ntop wants the network address structure as input not converted string.

Fixes: 1571543dcd9b ("nvme: add ave-discovery-log command")

@ikegami-t did you tests the original implementation? If so, can you try this one too?


ikegami-t:
Thanks for your fix. Just tested the changes with the dummy data as works correctly and the changes look good. Sorry for the error and previously I could not test the original changes correctly.

igaw:
Thanks! The copilot review has brought up some good points. I'll add them next week. running out of steam for today.

PR #3265, opened 2026-04-10
doc: consolidate contributor guidance into CONTRIBUTING.md

martin-belanger:
`README.md` was mixing user-facing content (build instructions, configuration, distro support) with developer-facing content (how to add commands, how to add plugins, API naming conventions, PR workflow). Readers looking to contribute had no obvious starting point.

Move the three developer-facing sections — "Developers", "API naming conventions", and "How to contribute" — into `CONTRIBUTING.md`, where they join and supersede the existing commit-conventions and bug-report stubs. Replace all three sections in`README.md` with a short "Contributing" paragraph that links to `CONTRIBUTING.md`.

While reorganizing `CONTRIBUTING.md`, document the new API naming convention (`libnvme_`/`libnvmf_` for libnvme, `nvme_`/`nvmf_` for nvme-cli) so contributors know which prefix to use when adding new functions.


CONTRIBUTING.md
```diff
@@ -14,13 +14,178 @@ under the [GPLv2-style License used for the NVMe CLI](https://github.com/linux-n
...
+|--------|-------|---------|
+| `nvme_` | Common NVMe (PCIe and NVMe-oF) | `nvme_show_id_ctrl()`, `nvme_get_log_smart()`, `nvme_identify_ctrl()`, `nvme_show_status()` |
+| `nvmf_` | NVMe-oF only | `nvmf_connect()`, `nvmf_disconnect()`, `nvmf_discover()` |
+
+### Summary
+
+| Component | Common NVMe prefix | NVMe-oF prefix |
+|-----------|--------------------|----------------|
+| libnvme   | `libnvme_`         | `libnvmf_`     |
+| nvme-cli  | `nvme_`            | `nvmf_`        |
```
igaw: The long time goal here is also to stop using the `nvme_` and `nvmf_` prefix for the CLI. Only the spec API should use `nvme_` and `nvmf_`. This makes clear what is CLI code and what is spec.

martin-belanger: Ah! I did not fully understand this distinction. Let me update the docs.

igaw:
Rest looks great!

CONTRIBUTING.md
martin-belanger: Done

CONTRIBUTING.md
```diff
@@ -1,8 +1,8 @@
 <!-- SPDX-License-Identifier: GPL-2.0-only -->
-# Contributing to the NVM-e CLI
+# Contributing to nvme-cli and libnvme
 
 Here you will find instructions on how to contribute to the NVM-Express command
-line interface.
+line interface and the libnvme library.
```
igaw, line 5: I think we need to update this section as well. The CLI wants the GPL-2.0-or-later license while libnvme LGPL-2.1-or-later.

igaw: Ah, that reminds me about the readme in libnvme. That needs to be merged into these two documents as well. Sorry :)

martin-belanger: All done. I actually folded the content of `./libnvme/README.md` into `./README.md` and `./CONTRINUTING.md`. Also, because the file `./README.md` was getting a bit big, I created a new `./TESTING.md` document and moved content related to Testing/CI to that document.

igaw:
Thanks a lot!

PR #3266, opened 2026-04-10
plugins/solidigm: Fixed parse Telemetry of multidimensional arrays

lgdacunh:
Array dimensions window were not sliding to correctly in recursive call.


igaw:
Thanks!

PR #3270, opened 2026-04-14
Header cleanups for Windows support

igaw:
Refactor and restructure the code base so the generic code is not platform dependent. The platform dependent part should be in separate files which are added via the build environment. 

The fabric and MI patch need some more love to avoid the ifdeffery but it's not too bad luckily.

~~The last two patches add some window specific code. Not sure yet if this is the best idea but it seems it's the only  platform depended code in the generic code base.~~

The TODO list from https://github.com/linux-nvme/nvme-cli/pull/3237


libnvme/tools/generator/generate-accessors.py
```diff
@@ -630,8 +630,8 @@ def main():
             f'#include <stdlib.h>\n'
             f'#include <string.h>\n'
             f'#include <stdbool.h>\n'
-            f'#include <stdint.h>\n'
-            f'#include <linux/types.h> /* __u32, __u64, etc. */\n'
+            f'#include <stdint.h>\n\n'
+            f'#include <nvme/types.h>\n'
```
martin-belanger, line 634: `nvme/nvme-types.h`

igaw:
The Windows build includes the `no-mi.c` file and this file has a `#include <libnvme.h>`. The build succeeds thus the library headers do pass the sanity check on Windows too.

I consider this a major milestone. 

Nevertheless there is still some more cleanups necessary, e.g. moving the fabrics code into fabrics.c. Currently, it's just behind some ugly ifdefs. This is all in the c files and not in the headers. The headers should be in good state. So when reviewing concentrate on the headers. They are the important step right now.

libnvme/tools/generator/generate-accessors.py
martin-belanger: I was wrong with my previous comment. I thought that you renamed `nvme/types.h` to `nvme/nvme-types.h`, but I see that the two files are needed. Here, we need to include the file that defines `__u32`, `__u64`, etc. So your original change was correct.

martin-belanger: I think that "Check accessors" failed for another reason. It checks that the `accessors.h` committed to the repo matches with the `generate-accessors.py` output. If not, then it would mean that `accessors.h` needs to be regenerated and committed to the repo.

martin-belanger:
@igaw - Everything looks good. Disregard my earlier comment. You made the right change to `generate-accessors.py`. All you need to do to clear the "Check accessors" error is to regenerate the accessors, `meson compile -C <build-dir> update-accessors` (or simply `make update-accessors`) and commit/push the updated accessor files.

igaw:
I missed to update the headers when updating the code generator. While at it also improved the commit message a bit. It should be a bit more standard English :)

PR #3272, opened 2026-04-15
min: Cleaned up usage of MIN in favor of safer CCAN min.

bgoing-micron-oss:
Libnvme's util.c uses a combination of CCAN's min from minmax.h and the MIN defined in param.h.  The CCAN implementation has typechecking and is more portable, whereas the param.h definition of MIN only exists on Linux.
- Updating util.c to use CCAN's min implementation consistently.
- Removing unused MIN definition in nbft.c.

### DIFF
--- a/libnvme/src/nvme/nbft.c
+++ b/libnvme/src/nvme/nbft.c
@@ -19,8 +19,6 @@
 #include "private.h"
 #include "compiler_attributes.h"
 
-#define MIN(a, b) (((a) < (b)) ? (a) : (b))
-
 static __u8 csum(const __u8 *buffer, ssize_t length)
 {
 	int n;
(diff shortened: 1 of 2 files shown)


igaw:
This looks good. Are you resubmitting it?

bgoing-micron-oss:
> This looks good. Are you resubmitting it?
After opening this I saw that you had a major refactoring pull request submitted, so I was waiting to review that today and see if this was still applicable.

igaw:
Yes, sorry busy moving stuff around. Whenever I pull on a string I end up a big refactoring excersice. Anyway, the code base gets better. I check what is missing after the libnvme_fabrics_config refactoring. What needs to go from the public API. Hopefully this was the last big piece. Obviously, there is still a bit to do in the c files. The main goal is to get the public API done first.

igaw:
Thanks!

PR #3273, opened 2026-04-15
Make struct libnvme_fabrics_config private

igaw:
Move `struct libnvme_fabrics_config` into the private section, so it's not part of the stable API.


libnvme/libnvme/nvme.i
```diff
@@ -696,27 +746,26 @@ struct libnvme_ns {
...
 
-	void connect(struct libnvme_host *h,
-		     struct libnvme_fabrics_config *cfg = NULL) {
+	void connect(struct libnvme_host *h) {
 		int ret;
 		const char *dev;
 
 		dev = libnvme_ctrl_get_name($self);
-		if (dev && !cfg->duplicate_connect) {
+		if (dev && self->cfg.duplicate_connect) {
```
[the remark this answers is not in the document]
martin-belanger: I agree with Copilot here.

libnvme/src/nvme/fabrics.c
```diff
@@ -714,22 +752,29 @@ static int inet_pton_with_scope(struct libnvme_global_ctx *ctx, int af,
...
-	if (!c->traddr)
+	char addrstr[NVMF_TRADDR_SIZE];
+
+	if (!traddr || !transport)
 		return false;
-	if (strcmp(c->transport, "tcp") && strcmp(c->transport, "rdma"))
+	if (!strcmp(traddr, "none"))
 		return false;
-	if (inet_pton_with_scope(ctx, AF_UNSPEC, c->traddr, c->trsvcid, &addr) == 0)
+
```
martin-belanger: `traddr_is_hostname()` drops scoped-IPv6 support.

The old implementation called `inet_pton_with_scope()` which handles IPv6 addresses with scope IDs (e.g. `fe80::1%eth0`). The new one calls `inet_pton(AF_INET6, ...)` directly, which will fail on scope IDs and misclassify them as hostnames. 

Unlikely to matter in practice (NVMe over scoped-link-local is exotic), but worth noting.

tbzatek: > NVMe over scoped-link-local is exotic

Actually that's what we've attempted for our NBFT testbed since we currently lack native ipv6/dualstack in our testbed. Worth noting there were more issues such exotic identified at various places and it never truly worked :-)

libnvme/libnvme/nvme.i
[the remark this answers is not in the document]
igaw: Interesting, I did build it locally and it worked. Anyway fixed...

libnvme/src/nvme/fabrics.c
```diff
@@ -409,23 +447,23 @@ static struct libnvme_fabrics_config *merge_config(libnvme_ctrl_t c,
 			  NVMF_DEF_CTRL_LOSS_TMO);
 	MERGE_CFG_OPTION(ctrl_cfg, cfg, fast_io_fail_tmo, 0);
 	MERGE_CFG_OPTION(ctrl_cfg, cfg, tos, -1);
-	MERGE_CFG_OPTION(ctrl_cfg, cfg, keyring, 0);
-	MERGE_CFG_OPTION(ctrl_cfg, cfg, tls_key, 0);
+	MERGE_CFG_OPTION(ctrl_cfg, cfg, keyring_id, 0);
+	MERGE_CFG_OPTION(ctrl_cfg, cfg, tls_key_id, 0);
 	MERGE_CFG_OPTION(ctrl_cfg, cfg, duplicate_connect, false);
```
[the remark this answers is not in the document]
igaw: added

libnvme/src/nvme/tree.c
```diff
@@ -1576,6 +1508,7 @@ libnvme_ctrl_t libnvme_lookup_ctrl(libnvme_subsystem_t s,
 	ctx = s->h ? s->h->ctx : NULL;
 	/* Set the NQN to the subsystem the controller should be created in */
 	fctx->subsysnqn = s->subsysnqn;
+	libnvmf_default_config(&fctx->cfg);
```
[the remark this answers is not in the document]
igaw: fixed

libnvme/src/nvme/fabrics.c
igaw: is this below not good enough?

```
+       if (inet_pton(AF_INET, traddr, addrstr) > 0 ||
+           inet_pton(AF_INET6, traddr, addrstr) > 0)
```

or do we need this and the dropped `inet_pton_with_scope` test?

igaw: two implementations and they don't agree. nice one...

libnvme/src/nvme/fabrics.c
```diff
@@ -196,6 +196,12 @@ __public const char *libnvmf_cms_str(__u8 cm)
 	return arg_str(cms, ARRAY_SIZE(cms), cm);
 }
 
+void libnvmf_default_config(struct libnvme_fabrics_config *cfg)
+{
```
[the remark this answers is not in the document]
igaw: nope, all callsites which do not use `libnvmf_context_create()` initialize the whole data struct.

libnvme/src/nvme/fabrics.c
igaw: or let me formulate my question differently, the version which was already in fabrics.c is the better one?

libnvme/src/nvme/json.c
```diff
@@ -25,7 +26,7 @@
 static void json_update_attributes(libnvme_ctrl_t c,
 				   struct json_object *ctrl_obj)
 {
-	struct libnvme_fabrics_config *cfg = libnvme_ctrl_get_config(c);
+	struct libnvme_fabrics_config *cfg = libnvmf_ctrl_get_fabrics_config(c);
```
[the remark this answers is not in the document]
igaw: I suppose we only need the json stuff for the config.json and this is a fabrics feature. I am going to json dependent on fabrics.

libnvme/src/nvme/fabrics.c
martin-belanger: In theory, `inet_pton(AF_INET6, traddr, addrstr)` should support scoped IPv6 addresses but let me write a quick test program to confirm. BTW, I'm the one that initially introduced `inet_pton_with_scope()` (which I copied from the kernel) because the original code would consider something like `fe80::1%eth0` to be hostname.

martin-belanger: I wrote a test program to compare the old implementation that uses `inet_pton_with_scope()` with the new implementation. The results are in the below table.

The new implementation gets 4 scoped IPv6 cases wrong — all misclassified as hostnames. Notice it also gets the `eth99` case wrong even though that interface doesn't exist; the address format is still valid.

There's also an unexpected finding: the old implementation has a bug for the literal string `"none"` — it falls through to the IP parsing, fails, and returns true (hostname). The new implementation correctly short-circuits it with an explicit `!strcmp(traddr, "none")` check. So, the new code actually fixes one edge case while introducing the scoped address regression.

```
transport   traddr                               expected  old       new       description
----------  -----------------------------------  --------  --------  --------  -----------
tcp         192.168.1.10                         IP        IP        IP        IPv4 address
rdma        192.168.1.10                         IP        IP        IP        IPv4 address (rdma)
tcp         fe80::1                              IP        IP        IP        IPv6 link-local (no scope)
tcp         2001:db8::1                          IP        IP        IP        IPv6 global unicast
tcp         ::1                                  IP        IP        IP        IPv6 loopback
tcp         fe80::1%lo                           IP        IP        hostname  IPv6 scoped (lo — exists locally)  <-- NEW_WRONG MISMATCH
tcp         fe80::1%enp0s31f6                    IP        IP        hostname  IPv6 scoped (enp0s31f6 — exists locally)  <-- NEW_WRONG MISMATCH

  [note] interface 'eth99' not found locally (if_nametoindex failed) — treating as valid IP

tcp         fe80::1%eth99                        IP        IP        hostname  IPv6 scoped (eth99 — does not exist)  <-- NEW_WRONG MISMATCH
rdma        fe80::1%lo                           IP        IP        hostname  IPv6 scoped rdma (lo)  <-- NEW_WRONG MISMATCH
tcp         storage.example.com                  hostname  hostname  hostname  hostname (FQDN)
tcp         mynvme-target                        hostname  hostname  hostname  hostname (short)
rdma        storage.example.com                  hostname  hostname  hostname  hostname rdma
pcie        192.168.1.10                         IP        IP        IP        IPv4 — pcie transport (check skipped)
fc          fe80::1%lo                           IP        IP        IP        scoped IPv6 — fc transport (check skipped)
loop        storage.example.com                  IP        IP        IP        hostname — loop transport (check skipped)
tcp         none                                 IP        hostname  IP        literal "none"  <-- OLD_WRONG MISMATCH
tcp         (null)                               IP        IP        IP        NULL traddr
(null)      192.168.1.1                          IP        IP        IP        NULL transport

Results: 13/18 passed  |  old wrong: 1  |  new wrong: 4  |  mismatches: 5
```

The ideal fix for this PR is simple: keep the `"none"` guard from the new implementation, but switch back to the scope-aware parsing for the IP detection:

```
bool traddr_is_hostname(const char *transport, const char *traddr)
{
    struct sockaddr_storage addr;

    if (!traddr || !transport)
        return false;
    if (!strcmp(traddr, "none"))               
        return false;
    if (strcmp(transport, "tcp") && strcmp(transport, "rdma"))
        return false;
    if (inet_pton_with_scope(traddr, &addr) == 0)  /* scope-aware */
        return false;
    return true;
}

```

martin-belanger: Just FYI: I added a new Wiki page describing `--traddr`, `--host-traddr`, and `--host-iface`. I describe how to use each parameter for each transport type: TCP, RDMA, FC. I also added a note regarding scoped IPv6 addresses saying to stay away from them due to lack of testing.

https://github.com/linux-nvme/nvme-cli/wiki/Using-%E2%80%90%E2%80%90host%E2%80%90traddr-and-%E2%80%90%E2%80%90host%E2%80%90iface

igaw: Thanks a lot digging into this. I'll update the function accordingly.

igaw: added the version you proposed.

libnvme/tools/generator/generate-accessors.md
```diff
@@ -265,8 +297,11 @@ __public const char *person_get_role(const struct person *p)
...
 1. **Dynamic strings** (`char *`) — setters store a `strdup()` copy; passing `NULL` clears the field.
 2. **Fixed char arrays** (`char foo[N]`) — setters use `snprintf`, always NUL-terminated.
-3. **`const` members** — only a getter is generated, no setter.
+3. **`const` members** — only a getter is generated, no setter (applies regardless of any annotation).
 4. **`//!accessors:readonly`** — same effect as `const`: getter only.
-5. **`//!accessors:none`** — member is completely ignored by the generator.
-6. **`--prefix`** — prepended to every function name (e.g. `--prefix nvme_` turns `ctrl_set_name` into `nvme_ctrl_set_name`).
-7. **Line length** — generated code is automatically wrapped to stay within the 80-column limit required by `checkpatch.pl`.
+5. **`//!accessors:writeonly`** — setter only; getter is suppressed.
+6. **`//!accessors:readwrite`** — both getter and setter; overrides a restrictive struct-level default.
```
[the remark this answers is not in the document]
igaw: it's documented in the notes.

PR #3274, opened 2026-04-15
Add nvme_fabrics module loading function

maurizio-lombardi:
As proposed in private by @igaw, I open a PR here for discussing this feature that we are considering for RHEL:

Introduce a new function to check and load the nvme_fabrics module when performing a connect/discover command, if it is not already loaded.

It uses the libkmod library.


igaw:
Very cool feature! Thanks a lot!

PR #3275, opened 2026-04-15
mi: make private-mi.h include conditional on MI support

brgoing:
MI support was made optional, but lib.c still included private-mi.h unconditionally.  Made include contitional based on MI support.

### DIFF
--- a/libnvme/src/nvme/lib.c
+++ b/libnvme/src/nvme/lib.c
@@ -24,7 +24,11 @@
 #include "cleanup.h"
 #include "cleanup-linux.h"
 #include "private.h"
+
+#ifdef CONFIG_MI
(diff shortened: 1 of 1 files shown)


bgoing-micron-oss:
This was an attempt to avoid including poll.h from private-mi.h when building on Windows, but it is failing the minimal build.  Additional testing on a Linux build configured with `-Dmi=disabled` has additional failures, even without this change.

* libnvme/test build fails - needs `if wants_mi` around mi tests
* libnvme_mi_status_to_string in no-mi.c needs to be __public

I'll put together a different PR with those changes for review.

igaw:
yep, sorry. there will be some more of this kind but hopefully it's not too bad.

PR #3276, opened 2026-04-15
mi: fix build with MI support disabled

brgoing:
MI support was made optional, but building with mi=disabled currently fails.  Fixed build issues and made inclusion of Linux-specific code in private-mi.h conditional on CONFIG_MI.


igaw:
Thanks!

PR #3277, opened 2026-04-16
test: add unit tests for static helpers in fabrics.c

martin-belanger:
@igaw - I thought it would be a good idea to add unit tests for the `static` functions in `fabrics.c` including `traddr_is_hostname()`. 

Adds `libnvme/test/test-fabrics.c`, which tests the static helper functions in `src/nvme/fabrics.c` without modifying any production source.

The technique is to define `static` to nothing and then `#include` the source file directly.  This exposes all translation-unit-local symbols in the test binary.  No `#ifdef UNIT_TEST` guards or visibility changes are needed — the production files are left completely untouched.

Functions covered:
  `strchomp, hostid_from_hostnqn, __add_bool_argument, __add_hex_argument, _add_int_argument,
  __add_int_or_minus_one_argument, __add_argument, inet4_pton, inet_pton_with_scope,  traddr_is_hostname, unescape_uri`

The test is gated on `want_fabrics` in `meson.build`, matching the conditional that guards the uriparser test above it.

The `traddr_is_hostname` tests include the key regression case: scoped IPv6 addresses (e.g. `"fe80::1%lo"`) must not be mis-classified as hostnames.  A plain `inet_pton()` call would fail on the `%` delimiter and wrongly return true; `inet_pton_with_scope()` handles it correctly.

### DIFF
--- a/libnvme/test/meson.build
+++ b/libnvme/test/meson.build
@@ -126,7 +126,26 @@ if want_fabrics
         ],
     )
 
-test('libnvme - uriparser', uriparser)
+    test('libnvme - uriparser', uriparser)
+
+    test_fabrics = executable(
+        'test-fabrics',
+        ['test-fabrics.c'],
+        # -fgnu89-inline: '#define static' (used to expose static helpers
+        # for unit testing) turns 'static inline' into plain 'inline'.
+        # Under C99, plain 'inline' has external linkage and may require an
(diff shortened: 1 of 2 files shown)


igaw:
Thanks. Let's see how this goes in the long run. I do think having proper unit tests also for such functions is not a bad thing. Though it has the change to get out of hand. But that's tomorrow's problem.

PR #3280, opened 2026-04-16
generate-accessors: fix stale examples and redundant code

martin-belanger:
This is just a bit of clean up.

The examples in generate-accessors.md were generated with an older version of the tool:
  - excluded members leave no trace; the phantom "no accessors" comment has been removed
  - car struct accessors were missing their KernelDoc comments
  - accessors.c used 4-space indentation instead of tabs
  - accessors.ld was not shown at all

In generate-accessors.py, collapse the redundant two-branch isdigit() check in emit_src_setter_chararray() — both branches produced identical output.


igaw:
Thanks!

PR #3281, opened 2026-04-16
lib: Gave blkdev and chardev checks platform agnostic names.

brgoing:
On Linux, controller devices are character devices, and namespace devices are block devices.  This is not true on platforms like Windows.
- Changed platform-specific libnvme_transport_handle_is_chardev and libnvme_transport_handle_is_blkdev to platform-agnostic libnvme_transport_handle_is_ctrl and libnvme_transport_handle_is_ns. This also helps to make the intent of the checks more clear.
- Changed error messages to use platform-agnostic controller/namespace wording as well.

### DIFF
--- a/libnvme/src/libnvme.ld
+++ b/libnvme/src/libnvme.ld
@@ -181,10 +181,10 @@ LIBNVME_3 {
 		libnvme_subsystem_release_fds;
 		libnvme_transport_handle_get_fd;
 		libnvme_transport_handle_get_name;
-		libnvme_transport_handle_is_blkdev;
-		libnvme_transport_handle_is_chardev;
+		libnvme_transport_handle_is_ctrl;
 		libnvme_transport_handle_is_direct;
 		libnvme_transport_handle_is_mi;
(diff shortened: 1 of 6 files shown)


bgoing-micron-oss:
@igaw This is the change we discussed last week to make checking the handle type platform agnostic and based on the intended use rather than whether it is block device or character device.

igaw:
Looks good to me. I wonder if we should just extend the help text  to something like

```
"ERROR : The NVMe namespace device (e.g. /dev/nvme0n1) must be specified\n");
```

FWIW, we can even make the example platform specific later.

bgoing-micron-oss:
@igaw I've updated the error string as suggested.  I left the error strings in nvme.c unchanged because they seemed consistent with the other error strings, but I updated the main help string which shows examples of controller and namespace device names.  If you would like further changes, let me know.  Thanks.

igaw:
I've changed the commit messages to present tense to keep it consist with the rest and applied it directly. Thanks!

PR #3282, opened 2026-04-16
Split nbft header

igaw:
Split the NVMe spec part from the library types and rename them. Follow the pattern, thus use `libnbft_` as prefix.


igaw:
@tbzatek   can you have a look at this?

libnvme/src/nvme/nbft-types.h
```diff
@@ -0,0 +1,989 @@
...
+#pragma once
+
+#include <nvme/lib-types.h>
+
+/**
+ * DOC: nbft.h
+ *
+ * NVM Express Boot Specification, Revision 1.0
+ *
+ * Note: this API is currently unstable, subject to further additions.
```
tbzatek: I believe it's rather stable now, at least this `-types` include.

tbzatek (approved):
Looks good, thanks!

tbzatek:
FYI: I've added TODO for myself to sync `nbft-types.h` with the last Boot Spec (if any)

libnvme/src/nvme/nbft-types.h
igaw: Right, I drop this then.

igaw:
FWIW, the current spec is at 1.3 but I don't know what has changed.

tbzatek:
> FWIW, the current spec is at 1.3 but I don't know what has changed.

I'm involved in the work group, but never took the time to sync the changes. Not much has changed really, mostly wording and perhaps some extra flags. The nvme-cli connection code doesn't respect many of the flags anyway as actual firmware implementations do not support everything either, having no reference firmware implementation to test the code against.

I'll have a look anyway.

PR #3283, opened 2026-04-16
generate-accessors: add lifecycle, defaults, and annotation redesign

martin-belanger:
Extend the accessor generator with three new capabilities.

1. Lifecycle (constructor + destructor): annotate a struct's opening brace with `//!generate-lifecycle` to generate `foo_new()` and `foo_free()`. `foo_new()` allocates a zeroed instance with `calloc()` and returns `-EINVAL` / `-ENOMEM` on error. `foo_free()` frees all owned `char*` and `char**` members then frees the struct. Passing `NULL` to `foo_free()` is safe: destructors that dereference members guard with `if (!p) return;` those with no members to dereference rely on `free(NULL)` being a no-op. `//!lifecycle:none` on a member excludes it from the destructor. `const char*` members are never freed (assumed externally owned).

2. Defaults (init function): annotate individual members with `//!default:VALUE` to generate `foo_init_defaults()`, which assigns each annotated field its compile-time default. Any valid C expression is accepted as the value. When combined with `//!generate-lifecycle`, `foo_new()` calls `foo_init_defaults()` after allocation. `foo_init_defaults()` is also useful standalone to re-initialise a struct without reallocating.

3. Annotation style: drop support for the `/*!annotation*/` block-comment style. Annotations now use only the `//` line-comment style. The parser was redesigned accordingly: after `//`, each `!keyword` token is treated as a command, so multiple annotations may share one comment:
```
  struct foo { //!generate-accessors !generate-lifecycle
```
`private.h` and `private-fabrics.h` are updated throughout to use the new style.

The generated `.c` files now include `<errno.h>` for `EINVAL`/`ENOMEM`.

Apply these features to `struct libnvmf_discovery_args`, replacing the manually-written `libnvmf_discovery_args_create()` and `libnvmf_discovery_args_free()` with generated equivalents. The constructor is renamed `_new` (consistent with the generated naming convention).

### DIFF
--- a/libnvme/examples/discover-loop.c
+++ b/libnvme/examples/discover-loop.c
@@ -94,7 +94,7 @@ int main()
 		goto free_fctx;
 	}
 
-	ret = libnvmf_discovery_args_create(&args);
+	ret = libnvmf_discovery_args_new(&args);
 	if (!ret) {
 		libnvmf_discovery_args_set_max_retries(args, 4);
 		ret = libnvmf_get_discovery_log(c, args, &log);
--- a/libnvme/src/libnvmf.ld
+++ b/libnvme/src/libnvmf.ld
@@ -24,8 +24,7 @@ LIBNVMF_3 {
 		libnvmf_ctrl_get_fabrics_config;
 		libnvmf_disconnect_ctrl;
 		libnvmf_discovery;
-		libnvmf_discovery_args_create;
-		libnvmf_discovery_args_free;
+
 		libnvmf_discovery_config_file;
 		libnvmf_discovery_config_json;
 		libnvmf_discovery_nbft;
--- a/libnvme/src/nvme/private.h
+++ b/libnvme/src/nvme/private.h
@@ -167,7 +167,7 @@ struct libnvme_transport_handle {
 	struct libnvme_log *log;
 };
 
-struct libnvme_path { /*!generate-accessors*/
+struct libnvme_path { //!generate-accessors
 	struct list_node entry;
 	struct list_node nentry;
 
@@ -189,7 +189,7 @@ struct libnvme_ns_head {
 	char *sysfs_dir;
 };
 
-struct libnvme_ns { /*!generate-accessors*/
+struct libnvme_ns { //!generate-accessors
 	struct list_node entry;
 
 	struct libnvme_subsystem *s;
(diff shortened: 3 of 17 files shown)


libnvme/src/nvme/private.h
```diff
@@ -167,7 +167,7 @@ struct libnvme_transport_handle {
 	struct libnvme_log *log;
 };
 
-struct libnvme_path { /*!generate-accessors*/
+struct libnvme_path { //!generate-accessors
```
igaw, line 170: Could you move these changes into a separate patch? Thanks!

And while at it an extra space wouldn't hurt `// !generate-accessors`

libnvme/src/libnvmf.ld
```diff
@@ -24,8 +24,7 @@ LIBNVMF_3 {
 		libnvmf_ctrl_get_fabrics_config;
 		libnvmf_disconnect_ctrl;
 		libnvmf_discovery;
-		libnvmf_discovery_args_create;
-		libnvmf_discovery_args_free;
+
```
igaw, line 27: Looks like an unnecessary empty line

libnvme/examples/discover-loop.c
```diff
@@ -94,7 +94,7 @@ int main()
 		goto free_fctx;
 	}
 
-	ret = libnvmf_discovery_args_create(&args);
+	ret = libnvmf_discovery_args_new(&args);
```
igaw, line 97: Should we look into making this consistent through out the code base? Having two types of wording is not good IMO.

igaw:
The general feature looks fine as far as I can tell. The Python code is slowly turning into a bit of a monster, but I don’t care too much about that. There may come a point where we need to start from scratch. Apparently, that’s just how things go. So the important part is getting the requirements right.

We’ve already added quite a few features beyond the initial simple design, and I think it really helps avoid writing annoying boilerplate code. In short, I’m fine with this, just please try to split the changes into smaller patches.

libnvme/src/nvme/private.h
martin-belanger: Yes. will do.

libnvme/src/libnvmf.ld
martin-belanger: Ok

libnvme/examples/discover-loop.c
martin-belanger: As per your text message. I will have a separate patch to change all `_create` to `_new`.

martin-belanger:
Closing this PR. Will submit a fresh one with incremental commits instead of everything in a single commit.

PR #3287, opened 2026-04-20
nvme: refactor NVME_ARGS help text constants

igaw:
Move all help text to a central place so it's easy to maintain a consistent documentation and wording.

TODO: 
- [ ] - all other commands in fabrics and the plugins...
- [x] - fix the whitespace damage in this patch
- [x] - fix build failures

For later:
- [ ] - consolidate the help texts (remove duplicates, grammar fix, etc.)

Fixes: #3080

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -181,74 +181,651 @@ static struct program nvme = {
 	.extensions = &builtin,
 };
 
-const char *uuid_index = "UUID index";
-const char *namespace_id_desired = "identifier of desired namespace";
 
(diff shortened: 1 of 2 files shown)


igaw:
I am not really sold on this approach. The functions get a bit simpler to skim over but honestly it never really bothered me much.

One thing though I do like is the style improvements style etc but we can do this with the functions. Also making the help text consistent isn't that difficult with copilot tbh. 

Idk.

PR #3288, opened 2026-04-20
generate-accessors: annotation style + lifecycle generation

martin-belanger:
This PR is a two-patch rework of the accessor generator, split from the earlier `generate-lifecycle` branch per reviewer's request for smaller incremental patches (see earlier PR: https://github.com/linux-nvme/nvme-cli/pull/3283).

**Patch 1 — Switch annotations to `//` comment style**

Replaces the `/*!annotation*/` block-comment style with `// !annotation` line comments. The canonical form is `// !token` (one space between `//` and `!`). The parser accepts any amount of whitespace between `//` and `!`, so `//!token`, `// !token`, and `//\t!token` are all equivalent — making annotations resilient to contributor variation.

Updates all annotated structs in `private.h` and `private-fabrics.h`, and updates the documentation and helper scripts. The generated `accessors.{h,c}` and `accessors-fabrics.{h,c}` are unchanged by this patch.

**Patch 2 — Add lifecycle and default-value generation**

Extends the generator with two new annotations:

- `// !generate-lifecycle` (on a `struct` brace) — generates `_new()`, `_free()`, and (when defaults exist) `_init_defaults()`
- `// !default:<value>` (on a member line) — sets the initial value of that member in `_init_defaults()`

Applies `// !generate-lifecycle` to `struct libnvmf_discovery_args`, replacing the hand-written `libnvmf_discovery_args_create()` / `_free()` pair:

- `libnvmf_discovery_args_new()` replaces `libnvmf_discovery_args_create()` — same `int` return / double-pointer signature; sets defaults via `libnvmf_discovery_args_init_defaults()`
- `libnvmf_discovery_args_free()` is now generated (same name, same semantics — `NULL`-safe via `free(NULL)`)
- `libnvmf_discovery_args_init_defaults()` is new; sets `max_retries = 6` and `lsp = NVMF_LOG_DISC_LSP_NONE`

Updates all callers (`nvme.i`, `discover-loop.c`, `test/ioctl/discovery.c`) to use the new `_new()` API. Moves the ABI entries for the removed functions from `libnvmf.ld` to `accessors-fabrics.ld`.

Note on `_create` → `_new` renaming

A patch 3 to rename remaining `_create` functions to `_new` was considered but dropped. The `_new` suffix generated here carries a specific contract: allocate a zeroed struct on the heap, set default field values, and return it via a double-pointer with no other parameters. The remaining `_create` functions in the public API (`libnvme_create_global_ctx`, `libnvme_create_raw_secret`, `libnvmf_context_create`, `libnvmf_create_ctrl`) are fundamentally different — they accept multiple parameters and perform substantial initialization logic beyond just setting defaults. Renaming them to `_new` would misrepresent what they do and break the semantic contract that `_new` is meant to convey.

### DIFF
--- a/libnvme/examples/discover-loop.c
+++ b/libnvme/examples/discover-loop.c
@@ -94,7 +94,7 @@ int main()
 		goto free_fctx;
 	}
 
-	ret = libnvmf_discovery_args_create(&args);
+	ret = libnvmf_discovery_args_new(&args);
 	if (!ret) {
 		libnvmf_discovery_args_set_max_retries(args, 4);
 		ret = libnvmf_get_discovery_log(c, args, &log);
--- a/libnvme/libnvme/nvme.i
+++ b/libnvme/libnvme/nvme.i
@@ -829,7 +829,7 @@ struct libnvmf_context {};
 			discover_err = 1;
 			return NULL;
 		}
-		discover_err = libnvmf_discovery_args_create(&args);
+		discover_err = libnvmf_discovery_args_new(&args);
 		if (discover_err)
 			return NULL;
 		libnvmf_discovery_args_set_lsp(args, lsp);
--- a/libnvme/src/accessors-fabrics.ld
+++ b/libnvme/src/accessors-fabrics.ld
@@ -10,8 +10,11 @@
 
 LIBNVMF_ACCESSORS_3 {
 	global:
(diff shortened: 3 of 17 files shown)


igaw:
Thanks!

PR #3290, opened 2026-04-20
python: add combat shim for older distros

igaw:
Newer SWIG started generating Py_NewRef() calls in its own runtime boilerplate, but Py_NewRef wasn't added until Python 3.10. The mismatch hits any distro with SWIG ≥ 4.1 but Python < 3.10.

Suggested-by: Martin Belanger <Martin.Belanger@dell.com>

### DIFF
--- a/libnvme/libnvme/meson.build
+++ b/libnvme/libnvme/meson.build
@@ -6,6 +6,10 @@
 # Authors: Martin Belanger <Martin.Belanger@dell.com>
 #
 if want_python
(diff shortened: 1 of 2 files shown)


martin-belanger:
@igaw - Hey Daniel, I've already added this change in a PR I'm working on. I was planning to submit later today. The PR will contain other Python bug-fixes/improvements as per my previous email.

igaw:
ah okay, just though I do it quickly. I'll pick your version then :)

PR #3292, opened 2026-04-20
fabrics: Made more includes conditional on fabrics support

brgoing:
Made more fabrics-specific and netdb-specific  includes conditional on fabrics and netdb support.  Also removed unneeded ioctl.h include from logging.c.


igaw:
Thanks!

FYI, @martin-belanger wants to split the util.c file into util.c and util-fabrics.c.

PR #3293, opened 2026-04-20
python, swig: fix bugs and restore Python ergonomics after libnvmf_context refactor

martin-belanger:
Three commits fixing regressions and compatibility issues introduced when the C API moved from `libnvme_fabrics_config` to the opaque `libnvmf_context`.

**Commit 1 — Bug fixes from the libnvmf_context refactor**
- `%pythonappend` for `connect()` had a stale `fctx` parameter — SWIG silently ignored it, so the GC guard (`self.__host = h`) never ran, leaving a latent use-after-free.
- `duplicate_connect` check was inverted in the SWIG binding. Moved to the C library (`libnvmf_add_ctrl()`) where it belongs, with the correct polarity.
- `fctx_set_fabrics_config` was missing its `%rename`, exposing it to Python with the ugly `fctx_` prefix intact.

**Commit 2 — Restore Python ergonomics: dict-based ctrl constructor**
The refactor forced Python callers to manually create, configure, and discard a `libnvmf_context` — a C implementation detail with no value at the Python level. This commit hides it entirely behind a `%typemap(in/freearg)` pair that converts a Python dict to an fctx transparently. The typemap is reusable for any future function that takes `struct libnvmf_context *`.

**Commit 3 — Py_NewRef compatibility (Python < 3.10)**
SWIG ≥ 4.1 generates calls to `Py_NewRef()`, which was added in Python 3.10. Distributions with SWIG ≥ 4.1 and Python < 3.10 (e.g. SLES 15.x) got an `undefined symbol` at import time. Fixed with a `%begin` compat shim and a meson warning when the condition is detected.

The dict covers all configurable fields: connection identity, the full `libnvme_fabrics_config` (set via the generated accessors), host identity, DH-HMAC-CHAP keys, TLS parameters, and persistence. `help(nvme.ctrl)` documents every supported key via an autodoc string.

### DIFF
--- a/libnvme/libnvme/meson.build
+++ b/libnvme/libnvme/meson.build
@@ -6,6 +6,10 @@
 # Authors: Martin Belanger <Martin.Belanger@dell.com>
 #
 if want_python
+    if not cc.has_header_symbol('Python.h', 'Py_NewRef', dependencies: py3_dep)
+        warning('Python < 3.10 detected: Py_NewRef compatibility shim will be used')
+    endif
+
     r = run_command(swig, ['-version'], check: true)  # Returns: "\nSWIG Version 4.1.1\n\nCompiled with ..."
     swig_version = r.stdout().split('\n')[1].split()[2].strip()
     if swig_version.version_compare('<4.1.0')
--- a/libnvme/libnvme/nvme.i
+++ b/libnvme/libnvme/nvme.i
@@ -10,6 +10,20 @@
                clashes with the same macro defined in Python.h.
  */
 #undef fallthrough
+
+#include <Python.h>
+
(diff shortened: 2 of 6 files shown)


igaw:
Looks great. Thanks

As indicated on the chat I'd to see the callbacks also be added to the Python binding:

```
`if (libnvmf_context_create(arg1, NULL, NULL, NULL, NULL, &temp)) {
```

This would make the Python binding offering the same features as the C binding and allows to write any discovery/connect logic in Python. The C library should just do the low level parsing/iterating over the log pages and the rest can be somewhere else. This would allow to have way simpler custom logic supported on default.

martin-belanger:
I agree with the callbacks and I do plan to revisit this soon.

PR #3294, opened 2026-04-20
mmap_registers: Made mmap functionality conditional on mmap support

brgoing:
Bypassed the mmap functionality in mmap_registers if the target platform doesn't support mmap.  Created munmap_registers for conditional unmap usage as well.

Used NVME_HAVE_MMAP to avoid conflict with python's HAVE_MMAP definition when building python wrappers.


igaw:
I've updated the commit message so it uses present tense and applied it directly. Thanks!

PR #3295, opened 2026-04-20
lib: Added libnvme wrapper for updating block size after format.

brgoing:
Moved the linux-specific ioctl calls used to update the block size after it is changed by a format command into a wrapper function in libnvme. Consolidates ioctl calls into ioctl.c, and allows for platform-specific implementations in the future as needed.


igaw:
Updated the commit message to use present tense, fixed up the merge conflict and applied directly to the master branch. Thanks!

PR #3298, opened 2026-04-21
libnvme: add check-public-headers test and fix missing prototypes

martin-belanger:
A few public function prototypes were missing from the installed public headers. 

A symbol that appears in a version script but has no prototype in any installed header is technically callable by external code, but callers have no declaration to include — they must write their own prototype or fall back to `dlsym()`. 

This series fixes three such symbols and adds a test to prevent regressions.

## Patch 1 — add missing prototypes for exported symbols

Three `__public` symbols were listed in version scripts (`*.ld`) but had no prototype in any installed header:

- `libnvme_mi_submit_entry()` — weak tracing hook in `mi.c`
- `libnvme_mi_submit_exit()` — weak tracing hook in `mi.c`
- `libnvmf_context_set_fabrics_config()` — setter in `fabrics.c` whose
  getter `libnvmf_context_get_fabrics_config()` was already declared

Prototypes added to `mi.h` and `fabrics.h` respectively.

## Patch 2 — add check-public-headers test

Adds `tools/check-public-headers.py`, a companion to the existing `check-public-symbols.py`. It reads all `*.ld` version scripts, extracts their `global:` symbols, then scans each installed header for a matching prototype. Conditional headers (MI, fabrics) are silently skipped when absent, matching the build's own conditionality.

The script is registered as a meson test alongside `check-public-symbols` so it runs automatically under `meson test`.

### DIFF
--- a/libnvme/src/nvme/fabrics.h
+++ b/libnvme/src/nvme/fabrics.h
@@ -405,6 +405,19 @@ struct libnvme_fabrics_config *libnvmf_context_get_fabrics_config(
 		struct libnvmf_context *fctx);
 
 /**
+ * libnvmf_context_set_fabrics_config() - Set fabrics configuration for a
+ * fabrics context
+ * @fctx: Fabrics context
+ * @cfg:  Fabrics configuration to apply
+ *
+ * Copies the fields of @cfg into the fabrics configuration of @fctx.
+ *
+ * Return: 0 on success, or a negative error code on failure.
+ */
+int libnvmf_context_set_fabrics_config(struct libnvmf_context *fctx,
+		struct libnvme_fabrics_config *cfg);
+
+/**
  * libnvmf_ctrl_get_fabrics_config() - Fabrics configuration of a controller
  * @c:	Controller instance
  *
--- a/libnvme/src/nvme/mi.h
+++ b/libnvme/src/nvme/mi.h
@@ -947,3 +947,37 @@ int libnvme_mi_aem_disable(libnvme_mi_ep_t ep);
  * Return: 0 is a success, nonzero is an error and errno may be read for further details
  */
 int libnvme_mi_aem_process(libnvme_mi_ep_t ep, void *userdata);
+
+/**
+ * libnvme_mi_submit_entry() - Weak hook called before an MI message is sent.
+ * @type:	MCTP message type
+ * @hdr:	Pointer to the MI message header
+ * @hdr_len:	Length of the message header in bytes
+ * @data:	Pointer to message payload data
+ * @data_len:	Length of payload data in bytes
+ *
+ * This is a weak symbol that can be overridden by an application to intercept
+ * outgoing MI messages for tracing or testing purposes.  The return value is
+ * passed back as @user_data to the matching libnvme_mi_submit_exit() call.
+ *
+ * Return: An opaque pointer passed to libnvme_mi_submit_exit(), or NULL.
+ */
+void *libnvme_mi_submit_entry(__u8 type, const struct nvme_mi_msg_hdr *hdr,
+			      size_t hdr_len, const void *data, size_t data_len);
+
+/**
+ * libnvme_mi_submit_exit() - Weak hook called after an MI message completes.
+ * @type:	MCTP message type
+ * @hdr:	Pointer to the MI response message header
+ * @hdr_len:	Length of the response message header in bytes
+ * @data:	Pointer to response payload data
+ * @data_len:	Length of response payload data in bytes
+ * @user_data:	Value returned by the matching libnvme_mi_submit_entry() call
+ *
+ * This is a weak symbol that can be overridden by an application to intercept
+ * completed MI transactions.  Called with the opaque pointer returned by the
+ * corresponding libnvme_mi_submit_entry() call.
+ */
+void libnvme_mi_submit_exit(__u8 type, const struct nvme_mi_msg_hdr *hdr,
+			    size_t hdr_len, const void *data, size_t data_len,
+			    void *user_data);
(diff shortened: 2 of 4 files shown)


libnvme/tools/check-public-headers.py
```diff
@@ -0,0 +1,109 @@
...
+    SRC_DIR / 'libnvme.ld',
+    SRC_DIR / 'libnvmf.ld',
+    SRC_DIR / 'accessors.ld',
+    SRC_DIR / 'accessors-fabrics.ld',
+]
+
+# Mirrors the install_headers() calls in src/meson.build.
+# Conditional headers (fabrics, MI) are included here and silently skipped
+# if they don't exist, matching the same conditionality as the build.
+INSTALLED_HEADERS = [
```
igaw: Would it possible to hand in the list of installed headers from meson? I fear having several places where we define the headers is likely to get out of sync. We could use the same list as input for the header include test which currently also is maintained separately (and which already got out of sync...)

martin-belanger: Yep! I thought you'd notice this 😄 
Let me see. It should be possible to have all the necessary parameters passed to the script from meson. I'll push an update.

libnvme/src/nvme/fabrics.h
```diff
@@ -404,6 +404,19 @@ int libnvmf_context_set_device(struct libnvmf_context *fctx, const char *device)
...
+ * libnvmf_context_set_fabrics_config() - Set fabrics configuration for a
+ * fabrics context
+ * @fctx: Fabrics context
+ * @cfg:  Fabrics configuration to apply
+ *
+ * Copies the fields of @cfg into the fabrics configuration of @fctx.
+ *
+ * Return: 0 on success, or a negative error code on failure.
+ */
+int libnvmf_context_set_fabrics_config(struct libnvmf_context *fctx,
```
igaw, line 417: I was unsure if we actually want to provide this interface at all. My initial plan was to embed all members of fabrics config directly into libnvmf_context. But we have the same members also for the controller. That's why fabrics config still exists. I think it would be good to get rid of fabrics config completely, at least from the API point of view. The python binding basically does this now and I do like it.

martin-belanger: No we don't really need to provide it. I just noticed that the function was define as `__public` but had no prototype in the installed headers.

libnvme/src/nvme/mi.h
```diff
@@ -947,3 +947,37 @@ int libnvme_mi_aem_disable(libnvme_mi_ep_t ep);
...
+ * @data:	Pointer to message payload data
+ * @data_len:	Length of payload data in bytes
+ *
+ * This is a weak symbol that can be overridden by an application to intercept
+ * outgoing MI messages for tracing or testing purposes.  The return value is
+ * passed back as @user_data to the matching libnvme_mi_submit_exit() call.
+ *
+ * Return: An opaque pointer passed to libnvme_mi_submit_exit(), or NULL.
+ */
+void *libnvme_mi_submit_entry(__u8 type, const struct nvme_mi_msg_hdr *hdr,
```
igaw, line 965: Ohh, left overs. For the  ioclt based interface we have callbacks instead these weak symbols. We should do the same thing here, switch over to callbacks, see b73c4a49f15e ("ioctl: add hooks into submission path")

IIRC, I was not sure how to deal with `struct nvme_mi_msg_hdr` here. With some fresh eyes, I'd say we just add a set of setters for the callbacks:

`nvme_transport_handle_set_mi_submit_entry` ?

libnvme/src/nvme/fabrics.h
igaw: Sure, glad you doing this stuff! Getting this right is a good thing :)

As I said, I think ideally we don't expose `struct libnvme_config_fabrics`. Maybe we could extend the generator to create setters/getters for fctx->cfg->members. I avoided this part because I didn't wanted to get stuck at this problem. But now with the series merged we could just make this happen.

libnvme/src/nvme/mi.h
martin-belanger: I'm not familiar with the MI stuff. Do you know someone that would have the "fresh eyes" 👀 required to take care of this?

martin-belanger:
@igaw - I changed the script to take the list of headers and *.ld as input parameters as per your suggestion.

If you don't mind, I would like to address the other review comments as separate PRs. In other words, would it be possible to merge this PR, and I will work on fixes and submit another PR for your other comments (although for the MI I would prefer if someone else could handle it).

libnvme/tools/check-public-headers.py
martin-belanger: A fix is in

libnvme/src/nvme/mi.h
igaw: I'll take it. In libnvme v1 we used the `weak` symbol approach to hookup into the submit path for being able to print debug logs. While this is likely to be more efficient, it's a bit of a strange debug interface. Thus in v3 we have callbacks which add a couple of `if`s into the `hotpath`. But I don't think anyone sane would use nvme-cli for performance tasks.

libnvme/src/nvme/fabrics.h
martin-belanger: I created this issue for extending the generator for `fctx->cfg->members`. You can assign the issue to me (I'm unable to assign issues - must be a permission thing).

https://github.com/linux-nvme/nvme-cli/issues/3303

libnvme/src/nvme/mi.h
martin-belanger: Great. Want me to write an issue for this?

igaw: Thanks, just done it myself. So let's go ahead with this PR.

PR #3299, opened 2026-04-21
CI: fix podman device permissions

MaisenbacherD:
The nvme_create_max_ns_test test case is failing to create more than one namespace due to the lack of access permissions in the /dev tree within the test container.

Instead of passing only the default namespace of the testing device to the container we now bind mount the entire /dev tree, giving the container access to all devices within the test VM.


ikegami-t:
Looks good. Thank you.

igaw:
Thanks!

PR #3300, opened 2026-04-21
swig: guard against silent attribute typos on SWIG objects

martin-belanger:
### Problem

SWIG-generated Python classes allow assigning arbitrary attributes on instances. A caller writing:

```python
ctrl.dhchap_key = x        # typo — correct name is dhchap_ctrl_key
```
silently creates a new Python instance attribute instead of calling the C-level property setter. The value is discarded and the bug is invisible at runtime. This has already caused a real bug in `nvme-stas` where `dhchap_key` was written instead of `dhchap_ctrl_key`, meaning the controller DHCHAP key was never actually set. What actually happened is that the property `dhchap_key` had been renamed to `dhchap_ctrl_key` in libnvme's python bindings, but had not been changed in `nvme-stas` causing a silent failure.

The same silent-failure applies to the ctrl constructor dict: a typo in a key such as `subsysnqn_typo` was previously ignored without error.

**Changes**

#### `libnvme/libnvme/nvme.i`

Add a `__setattr__` guard (via `%pythoncode`) to four classes:
`libnvme_global_ctx`, `libnvme_host`, `libnvme_subsystem`, and `libnvme_ctrl`.

The guard:
- Passes through this (SWIG's raw C pointer) and `_`-prefixed names (GC-anchoring refs set by `%pythonappend` blocks such as `self.__host`) via `object.__setattr__`, bypassing property lookup.
- Walks the MRO to find a `property` with an `fset` and calls it directly, equivalent to Python's default machinery but without allowing fallback to instance-dict creation.
- Raises `AttributeError` for everything else — covering both unknown names and `%immutable` properties (which have no `fset`).

Note: `thisown` has an `fset`, so it is handled correctly by the MRO walk without any special case.

In `set_fctx_from_dict()`, replace the silent fall-through at the end of the key-dispatch `while` loop with a `PyErr_Format(PyExc_KeyError) - return -1`. Every known key already hits `continue`; any key that falls through is therefore unknown. This avoids a second pass over the dict.

#### `libnvme/libnvme/tests/test-setattr.py` (new file)

Six test cases:
- Valid writable property assignment does not raise.
- Typo on a property name raises `AttributeError` immediately.
- Assignment to a `%immutable` (read-only) property raises `AttributeError`.
- Completely unknown attribute name raises `AttributeError`.
- Unknown key in the constructor dict raises `KeyError`.
- Missing required key (`subsysnqn` or `transport`) raises `KeyError`.

Testing
```bash
meson test -C .build
```

### DIFF
--- a/libnvme/libnvme/tests/test-setattr.py
+++ b/libnvme/libnvme/tests/test-setattr.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# This file is part of libnvme.
+# Copyright (c) 2026, Dell Technologies Inc. or its subsidiaries.
+# Authors: Martin Belanger <Martin.Belanger@dell.com>
+"""Tests that __setattr__ guards on SWIG-generated classes raise on bad names."""
+
+import unittest
+from libnvme import nvme
+
+
+class TestCtrlSetattr(unittest.TestCase):
+
+    def setUp(self):
+        self.ctx = nvme.global_ctx()
+        self.ctrl = nvme.ctrl(self.ctx, {
+            'subsysnqn': nvme.NVME_DISC_SUBSYS_NAME,
+            'transport': 'loop',
+        })
+
+    def test_valid_writable_property(self):
+        """Writing a valid writable property must not raise."""
+        self.ctrl.discovery_ctrl = True
+
+    def test_typo_raises(self):
+        """A typo in a property name must raise AttributeError immediately."""
+        with self.assertRaises(AttributeError):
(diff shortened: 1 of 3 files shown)


libnvme/libnvme/tests/test-setattr.py
```diff
@@ -0,0 +1,60 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
igaw, line 1: Could you also add your copyright statement to new files? thanks!

martin-belanger: Fixed

igaw:
Thanks!

PR #3301, opened 2026-04-21
libnvme: move fabrics-only helpers to util-fabrics.c

martin-belanger:
## Summary

- `libnvmf_exat_ptr_next()` and `libnvme_getifaddrs()` were compiled unconditionally even in PCIe-only/embedded builds that disable the fabrics layer. This patch moves them to a new `util-fabrics.c` which is only compiled when `want_fabrics` is set (in `meson.build`).
- `libnvme_getifaddrs()` is renamed to `libnvmf_getifaddrs()` to reflect that it is a fabrics-only function (consistent with the `libnvmf_` prefix convention).
- The `nvmf_exat_len()` and `nvmf_exat_size()` static inlines, along with the `libnvmf_getifaddrs()` declaration, are moved from `private.h` to `private-fabrics.h` for the same reason.
- The include guard in `util.c` is simplified from `#if defined(HAVE_NETDB) || defined(CONFIG_FABRICS)` to `#ifdef HAVE_NETDB` now that `CONFIG_FABRICS` is no longer needed there.

This is low-hanging fruit — a small, self-contained first step toward a cleaner fabrics/PCIe separation. More work is needed in `tree.c`, which still contains several `#ifdef CONFIG_FABRICS` blocks (TCP-specific match functions, `libnvme_iface_matching_addr()`, etc.) that belong in a future `tree-fabrics.c`. That follow-up is more invasive and will be handled separately.

## Test plan

- [x] Builds cleanly with `meson setup .build && meson compile -C .build`
- [x] Builds cleanly with fabrics disabled: `meson setup .build -Dfabrics=disabled`
- [x] Unit tests pass: `meson test -C .build`

### DIFF
--- a/libnvme/src/nvme/private-fabrics.h
+++ b/libnvme/src/nvme/private-fabrics.h
@@ -7,6 +7,10 @@
  */
 #pragma once
 
+#if defined(HAVE_NETDB) || defined(CONFIG_FABRICS)
+#include <ifaddrs.h>
+#endif
+
 #include <nvme/fabrics.h>
 #include <nvme/tree.h>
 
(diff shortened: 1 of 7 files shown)


libnvme/src/nvme/util-fabrics.c
```diff
@@ -0,0 +1,36 @@
...
+#include "util.h"
+
+__public struct nvmf_ext_attr *libnvmf_exat_ptr_next(struct nvmf_ext_attr *p)
+{
+	__u16 size = nvmf_exat_size(le16_to_cpu(p->exatlen));
+
+	return (struct nvmf_ext_attr *)((uintptr_t)p + (ptrdiff_t)size);
+}
+
+#if defined(HAVE_NETDB) || defined(CONFIG_FABRICS)
```
igaw: `CONFIG_FABRICS` shouldn't be necessary in this c file, as it should only be part of the compilation when fabrics is enabled.

libnvme/src/nvme/private-fabrics.h
```diff
@@ -106,6 +110,54 @@ struct libnvmf_uri { // !generate-accessors
...
+ * nvmf_exat_len() - Return length rounded up by 4
+ * @val_len: Value length
+ *
+ * Return the size in bytes, rounded to a multiple of 4 (e.g., size of
+ * __u32), of the buffer needed to hold the exat value of size
+ * @val_len.
+ *
+ * Return: Length rounded up by 4
+ */
+static inline __u16 nvmf_exat_len(size_t val_len)
```
igaw: To be consistent this should be `libnvmf_exat_len`

libnvme/src/nvme/private-fabrics.h
```diff
@@ -106,6 +110,54 @@ struct libnvmf_uri { // !generate-accessors
...
+ *           field of a "struct nvmf_ext_attr".
+ *
+ * Return the size of the "struct nvmf_ext_attr" needed to hold
+ * a value of size @val_len.
+ *
+ * Return: The size in bytes, rounded to a multiple of 4 (i.e. size of
+ * __u32), of the "struct nvmf_ext_attr" required to hold a string of
+ * length @val_len.
+ */
+static inline __u16 nvmf_exat_size(size_t val_len)
```
igaw: same here `libnvmf_` prefix

igaw: we can do the renaming as new patch on top of this one.

libnvme/src/nvme/util-fabrics.c
martin-belanger: I had a lot of problems with the musl and muon builds. This was the only way to fix it. I can try again to confirm.

libnvme/src/nvme/private-fabrics.h
martin-belanger: Ok

libnvme/src/nvme/util-fabrics.c
igaw: `want_fabrics` is only true when also `CONFIG_FABRICS` is set. Thus it should work. 

Only in the C file, the header still needs it.

martin-belanger: You are correct. I was thinking of something similar in another file. I'm working on an update that will address all your comments. I will push shortly.

martin-belanger: Fixed

libnvme/src/nvme/private-fabrics.h
martin-belanger: Fixed

libnvme/src/nvme/private-fabrics.h
martin-belanger: Fixed

igaw:
Thanks!

PR #3302, opened 2026-04-21
nvme: apply user-specified timeout to all admin passthru commands

darkfiberiru:
The --timeout / -t flag is parsed into nvme_args.timeout but was only applied to format-nvm and admin-passthru. All other commands that use nvme_init_*() helpers followed by libnvme_submit_admin_passthru() had their timeout_ms left at 0 (memset by the init functions), silently falling back to the kernel's admin_timeout (60s).

This meant commands like delete-ns, create-ns, attach-ns, detach-ns, sanitize, fw-download, fw-commit, device-self-test, security-send, security-receive, capacity-mgmt, and all identify passthru commands ignored the user-supplied timeout.

Set cmd.timeout_ms = nvme_args.timeout after every nvme_init_*() call and before submission. When -t is not given, nvme_args.timeout is NVME_DEFAULT_IOCTL_TIMEOUT (0), preserving the existing kernel-default behavior.

### DIFF
--- a/libnvme/src/libnvme.ld
+++ b/libnvme/src/libnvme.ld
@@ -188,6 +188,7 @@ LIBNVME_3 {
 		libnvme_transport_handle_set_decide_retry;
 		libnvme_transport_handle_set_submit_entry;
 		libnvme_transport_handle_set_submit_exit;
+		libnvme_transport_handle_set_timeout;
 		libnvme_unlink_ctrl;
 		libnvme_update_block_size;
 		libnvme_update_key;
--- a/libnvme/src/nvme/ioctl.c
+++ b/libnvme/src/nvme/ioctl.c
@@ -193,6 +193,9 @@ out:
 __public int libnvme_submit_io_passthru(struct libnvme_transport_handle *hdl,
 		struct libnvme_passthru_cmd *cmd)
 {
+	if (!cmd->timeout_ms && hdl->timeout)
+		cmd->timeout_ms = hdl->timeout;
+
 	if (hdl->ioctl_io64)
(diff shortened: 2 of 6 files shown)


igaw:
I suggest something simpler, set timeout on the handle and use this value when submitting in ioctl.c:

```
--- a/nvme.c
+++ b/nvme.c
@@ -381,6 +381,8 @@ int parse_and_open(struct libnvme_global_ctx **ctx,
        libnvme_transport_handle_set_submit_exit(hdl_new , nvme_submit_exit);
        libnvme_transport_handle_set_decide_retry(hdl_new, nvme_decide_retry);
        libnvme_set_dry_run(ctx_new, argconfig_parse_seen(opts, "dry-run"));
+       if (argconfig_parse_seen(opts, "timeout"))
+               libnvme_transport_handle_set_timeout(hdl, nvme_args.timeout);

        *ctx = ctx_new;
        *hdl = hdl_new;
```

In v1 we passed the timeout value around using always the default value with a few exception. For special cases we can still do the approach you suggest, but I would like to avoid this if possible.

darkfiberiru:
@igaw yes I see that's much cleaner. Sorry for brute force I will refactor

darkfiberiru:
@igaw Updated

igaw:
I've updated some formatting (nitpicking), moved linker symbol into the correct section and trimmed the commit message. Thanks!

ixhamza:
Unrelated to this commit, but in [`format_cmd()` the `nvme_args.timeout` check runs before`_parse_args()`](https://github.com/linux-nvme/nvme-cli/blob/master/nvme.c#L6640), so I believe it never fires.

PR #3304, opened 2026-04-21
swig: replace hand-written accessor wrappers with auto-generated #defines

martin-belanger:
# swig: auto-generate #define bridges for accessors

SWIG uses a naming convention (`libnvme_STRUCT_MEMBER_get/set`) that differs from the libnvme C API convention (`libnvme_STRUCT_get/set_MEMBER`).  Until now this mismatch was papered over by ~35 hand-written C wrapper functions inside `nvme.i` — one thin shim per accessor that did nothing but call the real function.

This commit replaces all those shims with automatically generated C preprocessor `#define` object-like macros.  Because the macros have no parameter list, the arguments flow through naturally, so no wrapper function body is needed at all.

## What changed

**New script: `libnvme/tools/generator/generate-swig-accessors.py`**

Post-processing step that reads `accessors.h` and `accessors-fabrics.h` (already kept up to date by `generate-accessors.py`) and emits `libnvme/libnvme/nvme-swig-accessors.i`.  It uses a non-greedy regex to split each accessor name at the first `_get_` / `_set_` occurrence so that multi-word member names (e.g. `dhchap_host_key`) and long struct prefixes (e.g. `libnvmf_discovery_args`) are handled correctly.  Output is only rewritten when the content changes, avoiding spurious mtime updates.

**Updated: `libnvme/tools/generator/meson.build`**

- Added `update-swig-accessors` run-target that invokes the new script.
- Folded it into the existing `update-accessors` alias so that `meson compile -C <build-dir> update-accessors` regenerates all three artefact sets in one shot.

**New generated file: `libnvme/libnvme/nvme-swig-accessors.i`**

Contains one `#define` per accessor for every struct annotated in `private.h` and `private-fabrics.h`, grouped by struct and with the third column vertically aligned within each group for readability.

**Updated: `libnvme/libnvme/nvme.i`**

- Added `%include "nvme-swig-accessors.i"` (alongside the existing `%include "exception.i"`) so SWIG picks up all the bridges automatically.
- Removed all four hand-written wrapper `%{ %}` blocks (host, subsystem, ctrl, ns) — 168 lines deleted.
- Kept four `#define` bridges that cannot be auto-generated (virtual Python property names or `!accessors:none` annotations), consolidated into the main `%{ %}` preamble with a comment explaining why each one exists:

  ```c
  /* Cannot be auto-generated — virtual names or !accessors:none fields */
  #define libnvme_subsystem_host_get libnvme_subsystem_get_host
  #define libnvme_ctrl_state_get     libnvme_ctrl_get_state
  #define libnvme_ctrl_subsystem_get libnvme_ctrl_get_subsystem
  #define libnvme_ctrl_address_get   libnvme_ctrl_get_traddr
  ```

## Result

New accessors added to libnvme (via `private.h` annotations) now automatically appear in the Python bindings after running `meson compile -C <build-dir> update-accessors` — zero manual maintenance of `nvme.i` required.


libnvme/libnvme/nvme.i
```diff
@@ -46,6 +47,33 @@ static inline PyObject *Py_NewRef(PyObject *obj)
...
+	 *   the C struct member is named 's' and the hand-written accessor is
+	 *   libnvme_ctrl_get_subsystem() in tree.c.
+	 *
+	 * - libnvme_ctrl_address_get: 'address' has !accessors:none in private.h
+	 *   and maps to libnvme_ctrl_get_traddr(), not libnvme_ctrl_get_address().
+	 */
+	#define libnvme_subsystem_host_get libnvme_subsystem_get_host
+	#define libnvme_ctrl_state_get     libnvme_ctrl_get_state
+	#define libnvme_ctrl_subsystem_get libnvme_ctrl_get_subsystem
+	#define libnvme_ctrl_address_get   libnvme_ctrl_get_traddr
```
igaw, line 75: I am trying to figure out if the `traddr` vs `address` name is going to bite us.

igaw:
Thanks!

libnvme/libnvme/nvme.i
martin-belanger: From what I can tell just looking at `struct libnvme_ctrl` ([private.h, line 264](https://github.com/linux-nvme/nvme-cli/blob/master/libnvme/src/nvme/private.h#L264)), `address` and `traddr` are not the same thing. Or are they?

igaw: `address` got renamed to `traddr` to match with the kernel argument names. I suppose we should remove `address` from the Python binding.

PR #3311, opened 2026-04-23
compat (errno): Adds cross-platform compatibility for errno values.

brgoing:
Some errno values used in the code are only available on Linux. Adds compatible handling of errno values not available on all platforms.
- Substitutes good equivalent errno value for unsupported values used by nvme/util.c when converting status to errno.
- Replaces ENAVAIL usage in nvme.c with a local error code. It is only used internally within nvme.c to signal that an "invalid field" error was returned and should be ignored.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -7466,7 +7466,7 @@ static int get_pi_info(struct libnvme_transport_handle *hdl,
 		 * Keep the I/O commands behavior same as before.
 		 * Since the error returned by drives unsupported.
 		 */
-		return -ENAVAIL;
+		return NVME_SC_INVALID_FIELD;
 
 	pi_size = (pif == NVME_NVM_PIF_16B_GUARD) ? 8 : 16;
 	if (NVME_FLBAS_META_EXT(ns->flbas)) {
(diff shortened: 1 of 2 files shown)


nvme.c
```diff
@@ -7412,7 +7418,7 @@ static int get_pi_info(struct libnvme_transport_handle *hdl,
 		 * Keep the I/O commands behavior same as before.
 		 * Since the error returned by drives unsupported.
 		 */
-		return -ENAVAIL;
+		return ERR_IGNORE_INVALID_FIELD;
```
igaw: Why not returning `NVME_SC_INVALID_FIELD`. This avoids introducing a new define and matches the existing return pattern. Negative values for internal errors (userland/kernel) and positive values for the protocol level status info.

igaw:
Could you also split these two changes into two patches? Thanks!

nvme.c
brgoing: Good idea.

brgoing:
> Could you also split these two changes into two patches? Thanks!

@igaw I have split the changes.  Would you like them in separate pull requests, or will this work?

igaw:
No need for separate PRs. Just a sec

igaw:
i've tweaked the commit message slightly. The subject prefix is used to 'point' to the file which is edited. I know this is different to the usage in different context but let's stick to what we have so far.  Since I can't push these changes back to this branch I merged them directly.

igaw:
Thanks!

PR #3314, opened 2026-04-23
compat (fd): Adds cross-platform-compatible file descriptor type

brgoing:
Creates and uses new libnvme_fd_t to handle differences in file descriptor types across platforms.  Also defines platform-specific values for invalid file descriptors and test file descriptors.

### DIFF
--- a/libnvme/src/nvme/ioctl.c
+++ b/libnvme/src/nvme/ioctl.c
@@ -101,7 +101,7 @@ __libnvme_public int libnvme_update_block_size(
 		struct libnvme_transport_handle *hdl, int block_size)
 {
 	int ret;
-	int fd = libnvme_transport_handle_get_fd(hdl);
+	libnvme_fd_t fd = libnvme_transport_handle_get_fd(hdl);
 
 	ret = ioctl(fd, BLKBSZSET, &block_size);
 	if (ret < 0)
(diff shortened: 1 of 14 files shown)


libnvme/src/nvme/lib-compat.h
```diff
@@ -0,0 +1,27 @@
...
+#define LIBNVME_INVALID_FD INVALID_HANDLE_VALUE
+#define LIBNVME_TEST_FD ((HANDLE)0xFD)
+
+#else
+
+typedef int libnvme_fd_t;
+#define LIBNVME_INVALID_FD -1
+#define LIBNVME_TEST_FD 0xFD
+
+#endif
```
igaw: Move this to `libnvme/src/nvme/lib-types.h`, where we define library types and libnvme_fd_t is a clear fit IMO.

igaw:
Could you rebase and move the typedefs to lib-types.h? Rest looks good, not too much noise and code churn which is good news :) 

Thanks!

libnvme/src/nvme/lib-compat.h
brgoing: Moved.

igaw:
Same here, I tweaked the commit message slightly to use the different style for the prefix and applied it directly. Thanks!

Issue #3112, opened 2025-11-26
Create NVMe PDC by default for each unique target port


martin-gpy:
It would help to create an NVMe PDC (Persistent Discovery Controller) by default for each unique target port listed in the discovery log page data, provided of course the 'Explicit discovery connections' flag is set in the the log page data.

This would help the NVMe host to automatically react to discovery AENs raised whenever there is any change to the discovery log page data (like when a subsystem is added/removed, target ports created/onlined/disabled, etc.).

igaw:
There are currently two PRs (#3536, #3472) which are still based on the json configuration. Let me check the situation with the INI configuration before dealing with the implementation.

Mr-Bossman:
There are currently *three* PRs (https://github.com/linux-nvme/nvme-cli/pull/3536, https://github.com/linux-nvme/nvme-cli/pull/3472, #3698 )

igaw:
I've merged all relevant PRs, along with my cleanup PR #3720. The last PR may have introduced some regressions, but I considered the cleanup important. I'd rather fix any regressions on top of this version than fix all bugs first and refactor afterward.

The good news is that we are now left with a single code path that performs discovery, and there are only two places where the decision is made whether a Discovery Controller (DC) is persistent:

* `discover_lookup_ctrl`: This function, along with its helper functions, tries to find existing controllers. Unfortunately, it also modifies the user input to steer `dc_should_connect`. I don't think this is ideal, but let's see whether it turns out to be a problem. Since this is all internal library code, we can improve it after the 3.0 release without affecting the public API.
* `dc_should_connect`: Contains the decision logic for when to connect to or disconnect from a DC, without any side effects.

PR #3315, opened 2026-04-23
swig, generator: unify accessor annotations and add custom-access groundwork

martin-belanger:
## Summary

This PR contains two small, incremental patches that prepare the codebase for upcoming work on Python/SWIG binding generation.

The goal is to make accessor semantics explicit and machine-readable so downstream tooling can reason about read/write behavior and custom implementations.

## Patch 1: Rename annotation

* Rename `!accessors:*` → `!access:*`
* Update generator, headers, and documentation
* Purely mechanical change (no behavior change)

## Patch 2: Explicit access model + custom support

* Introduce a unified annotation model:

  * `!access:read=<mode>,write=<mode>`
  * `!generate-accessors:read=<mode>,write=<mode>`
* Modes: `generated`, `custom`, `none`
* Struct-level annotations define defaults; member-level overrides per axis
* Add support in the generator
* Update documentation
* Annotate existing fields that already rely on custom accessors

## Rationale

The previous annotation scheme could not clearly express:

* Direction (read vs write)
* Implementation (generated vs custom)

This becomes important for:

* Python/SWIG binding generation
* Consistency checks between C and bindings

These changes establish a clear and extensible model without altering existing runtime behavior.

## Notes

* No functional changes intended
* Changes are kept incremental for easier review
* This is groundwork for follow-up patches adding binding generation

---

Feedback welcome before building on top of this.

### DIFF
--- a/libnvme/src/nvme/private.h
+++ b/libnvme/src/nvme/private.h
@@ -196,26 +196,27 @@ struct libnvme_stat {
 	double ts_ms;			/* timestamp when the stat is updated */
 };
 
-struct libnvme_path {			// !generate-accessors
+struct libnvme_path {			// !generate-accessors:read=custom,write=none
 	struct list_node entry;
 	struct list_node nentry;
 
 	struct libnvme_stat stat[2];	/* gendisk I/O stat */
-	unsigned int curr_idx;		/* current index into the stat[] */
-	bool diffstat;			// !access:none
+	// curr_idx: current index into the stat[]
+	unsigned int curr_idx;		// !access:read=generated,write=generated
+	bool diffstat;			// !access:read=none
 
 	struct libnvme_ctrl *c;
 	struct libnvme_ns *n;
 
-	char *name;
-	char *sysfs_dir;
-	char *ana_state;		// !access:none
-	char *numa_nodes;		// !access:none
-	int grpid;
-	int queue_depth;		// !access:none
-	long multipath_failover_count;	// !access:none
-	long command_retry_count;	// !access:none
-	long command_error_count;	// !access:none
+	char *name;			// !access:read=generated,write=generated
+	char *sysfs_dir;		// !access:read=generated,write=generated
+	char *ana_state;
+	char *numa_nodes;
+	int grpid;			// !access:read=generated,write=generated
+	int queue_depth;
+	long multipath_failover_count;
+	long command_retry_count;
+	long command_error_count;
 };
 
 struct libnvme_ns_head {
@@ -236,12 +237,12 @@ struct libnvme_ns {			// !generate-accessors
 
 	struct libnvme_stat stat[2];	/* gendisk I/O stat */
 	unsigned int curr_idx;		/* current index into the stat[] */
-	bool diffstat;			// !access:none
+	bool diffstat;			// !access:read=none,write=none
 
 	struct libnvme_transport_handle *hdl;
 	__u32 nsid;
 	char *name;
-	char *generic_name;		// !access:none
+	char *generic_name;		// !access:read=custom,write=none
 	char *sysfs_dir;
 
 	int lba_shift;
@@ -255,13 +256,13 @@ struct libnvme_ns {			// !generate-accessors
 	unsigned char uuid[NVME_UUID_LEN];
 	enum nvme_csi csi;
 
-	long command_retry_count;		// !access:none
-	long command_error_count;		// !access:none
-	long requeue_no_usable_path_count;	// !access:none
-	long fail_no_available_path_count;	// !access:none
+	long command_retry_count;	     // !access:read=custom,write=none
+	long command_error_count;	     // !access:read=custom,write=none
+	long requeue_no_usable_path_count;   // !access:read=custom,write=none
+	long fail_no_available_path_count;   // !access:read=custom,write=none
 };
 
-struct libnvme_ctrl {			// !generate-accessors
+struct libnvme_ctrl {			// !generate-accessors:read=generated,write=none
 	struct list_node entry;
 	struct list_head paths;
 	struct list_head namespaces;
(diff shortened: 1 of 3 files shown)


libnvme/src/nvme/private.h
```diff
@@ -202,20 +202,20 @@ struct libnvme_path {			// !generate-accessors
 
 	struct libnvme_stat stat[2];	/* gendisk I/O stat */
 	unsigned int curr_idx;		/* current index into the stat[] */
-	bool diffstat;			// !access:none
+	bool diffstat;			// !access:read=none,write=none
```
igaw: Is this change necessary? It looks like a no-op. The same comment on  diffstat below

libnvme/src/nvme/private.h
```diff
@@ -202,20 +202,20 @@ struct libnvme_path {			// !generate-accessors
...
+	char *numa_nodes;		// !access:read=custom,write=none
 	int grpid;
-	int queue_depth;		// !access:none
-	long multipath_failover_count;	// !access:none
-	long command_retry_count;	// !access:none
-	long command_error_count;	// !access:none
+	int queue_depth;		// !access:read=custom,write=none
+	long multipath_failover_count;	// !access:read=custom,write=none
+	long command_retry_count;	// !access:read=custom,write=none
+	long command_error_count;	// !access:read=custom,write=none
```
igaw: Just wondering if the default should always be `none` when not listed, e.g.

```// !access:read=custom```

or would this be confusing? I'd like to reduce the 'noise' level if possible

File: libnvme/src/nvme/private.h (hunk 1 above)
Comment (martin-belanger) [reply libnvme/src/nvme/private.h]: Good catch — it does look like a no-op at first glance.

The behavior isn’t defined only at the member level though — it comes from the struct-level default, and member annotations act as overrides.

In this case:
```
    struct libnvme_path { // !generate-accessors
```
implies the default:

    read=generated,write=generated

So:
```
    bool diffstat; // !access:read=none,write=none
```
is explicitly overriding that default to disable both directions.

That said, I agree this adds noise when we’re just “turning things off”. One thing we can improve is choosing struct-level defaults that better match the majority of members, which would reduce the need for these overrides.

File: libnvme/src/nvme/private.h (hunk 2 above)
Comment (martin-belanger) [reply libnvme/src/nvme/private.h]: The current behavior is that struct-level annotations define defaults, and member-level annotations override per axis.

So something like:
```
    // !access:read=custom
```
would inherit the struct-level default for `write`, not imply `write=none`.

That’s why we currently need to spell out:
```
    // !access:read=custom,write=none
```
I do agree with the goal of reducing noise though. The best way to do that is probably to pick struct-level defaults that match the common case, so we only annotate exceptions at the member level.

**Comment by martin-belanger** [comment]
If it helps, we could also consider making `write=none` the default in some structs where setters are rarely needed, which would allow shorter forms like `!access:read=custom` without ambiguity.

I can make that change if you'd like.

File: libnvme/src/nvme/private.h (hunk 2 above)
Comment (martin-belanger) [reply libnvme/src/nvme/private.h]: I changed the struct-level defaults to minimize the number of member-level override and reduce the noise.

**Comment by martin-belanger** [comment]
By setting the struct-level defaults I was able to reduce the number of member-level overrides.

BTW, we may need to revisit the member annotations. I'm not sure that we need accessors for all of them For example, we are generating accessors for `libnvme_path::curr_idx`, but this looks like an internal parameter. I'm not sure that needs to be exposed through the public API at all.

File: libnvme/src/nvme/private.h [hunk 3]
```diff
@@ -196,26 +196,27 @@ struct libnvme_stat {
...
-struct libnvme_path {			// !generate-accessors
+struct libnvme_path {			// !generate-accessors:read=custom,write=none
 	struct list_node entry;
 	struct list_node nentry;
 
 	struct libnvme_stat stat[2];	/* gendisk I/O stat */
-	unsigned int curr_idx;		/* current index into the stat[] */
-	bool diffstat;			// !access:none
+	// curr_idx: current index into the stat[]
+	unsigned int curr_idx;		// !access:read=generated,write=generated
```
igaw, line 205: As you already noted, this is an internal member and should not be exposed to the public API.

libnvme/src/nvme/private.h
```diff
@@ -255,82 +256,82 @@ struct libnvme_ns {			// !generate-accessors
...
 	struct libnvme_global_ctx *ctx;
 
-	char *hostnqn;			// !access:readonly
-	char *hostid;			// !access:readonly
+	char *hostnqn;			// !access:read=generated,write=none
+	char *hostid;			// !access:read=generated,write=none
 	char *dhchap_host_key;
 	char *hostsymname;
-	bool pdc_enabled;		// !access:none
+	bool pdc_enabled;		// !access:read=none,write=custom
```
igaw, line 334: This one looks a bit wierd, but we have to touch the PDC anyway (see #3112), so this will be likely go away.

igaw:
I've just trimmed the very verbose commit message. Rest looks good. Let's fix the the comments I had on top of them. Thanks!

libnvme/src/nvme/private.h
martin-belanger: I submitted PR https://github.com/linux-nvme/nvme-cli/pull/3324 that clearly indicates that those are internal and not to be exposed with setter/getter functions.

libnvme/src/nvme/private.h
martin-belanger: Yes, this is a bit awkward as currently implemented. In PR https://github.com/linux-nvme/nvme-cli/pull/3324, I added a note clarifying that `pdc_enabled` and `pdc_enabled_valid` are intended to operate as a pair.

Good to hear this may go away, since the current approach is indeed a bit odd.

PR #3318, opened 2026-04-24
json: Enables json printing and ocp plugin without fabrics

brgoing:
Enables non fabrics-specific json functionality like printing and the ocp plugin when json-c is available, even if fabrics is disabled. Only nvme/json.c functionality is excluded when fabrics is disabled.

### DIFF
--- a/libnvme/src/meson.build
+++ b/libnvme/src/meson.build
@@ -79,7 +79,7 @@ else
     sources += 'nvme/no-uring.c'
 endif
 
-if json_c_dep.found()
+if json_c_dep.found() and want_fabrics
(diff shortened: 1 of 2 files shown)


brgoing:
@igaw We have been testing the ocp plugin and json printing on Windows, which doesn't use fabrics.  A recent change disabled json-c when fabrics was disabled, which broke that functionality for us.  It looks like only nvme/json.c is fabrics-only, so I made the fabrics dependency only apply to that json support specifically.

igaw:
Sorry, didn't realize this dependency. But sure this fix looks good! Thanks!

PR #3319, opened 2026-04-25
tests: fix python TypeError errors

ikegami-t:
This is to fix the errors below.

    if self.host_behavior_data[4] & cross_namespace_copy:
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for &: 'str' and 'int'

    data = self.host_behavior_data[:4] + cross_namespace_copy.to_bytes...
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
TypeError: can only concatenate str (not "bytes") to str

### DIFF
--- a/tests/nvme_copy_test.py
+++ b/tests/nvme_copy_test.py
@@ -16,6 +16,8 @@ NVMe Copy Testcase:-
 
 """
 
+import base64
+
 from nvme_test import TestNVMe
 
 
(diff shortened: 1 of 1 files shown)


igaw:
Thanks for looking into this one!

PR #3320, opened 2026-04-26
nvme: handle get-feature output format in nvme-print layer

ikegami-t:
Also the --raw-binary just selects the binary output format.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -4838,19 +4838,19 @@ static int filter_out_flags(int status)
 static void get_feature_id_print(struct feat_cfg cfg, int err, __u64 result,
 		void *buf, nvme_print_flags_t flags)
 {
-	int status = filter_out_flags(err);
+	int status = err > 0 ? filter_out_flags(err) : err;
 	enum nvme_status_type type = NVME_STATUS_TYPE_NVME;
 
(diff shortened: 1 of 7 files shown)


nvme.c
```diff
@@ -4838,19 +4838,18 @@ static int filter_out_flags(int status)
...
-		if (!nvme_status_equals(status, type, NVME_SC_INVALID_FIELD) &&
-		    !nvme_status_equals(status, type, NVME_SC_INVALID_NS))
-			nvme_show_status(err);
-	} else {
-		nvme_show_error("get-feature: %s", libnvme_strerror(err));
+	if (err) {
+		if (nvme_status_equals(status, type, NVME_SC_INVALID_FIELD) ||
+		    nvme_status_equals(status, type, NVME_SC_INVALID_NS))
+			return;
+		nvme_show_err(err, "get-feature");
```
igaw, line 4848: should we not do a return here? I think this missing return is also causing the compile warning and the next patch is not needed.

ikegami-t: Yes right and just fixed the patch. Sorry for the error. Thank you.

igaw:
Thanks a lot for cleaning this up!

Issue #3322, opened 2026-04-27
tests: `nvme-cli - nvme_copy_test`: Command Size Limit Exceeded


igaw:
When running the tests against a Qemu nvme device, the `nvme_copy_test` fails. Figure out what's the problem.

```
root@localhost:/home/wagi/work/nvme-cli/.build-debian# /usr/bin/python3 /home/wagi/work/nvme-cli/tests/tap_runner.py --start-dir /home/wagi/work/nvme-cli/tests nvme_copy_test

TAP version 13
1..1
# Using nvme binary 'nvme'
# Running: find /sys/devices -name \*nvme0 | grep -i pci
# /sys/devices/pci0000:00/0000:00:05.0/nvme/nvme0
#
# Running: nvme id-ctrl /dev/nvme0 --output-format=json
# {
#   "vid":6966,
#   "ssvid":6900,
#   "sn":"nvme-0              ",
#   "mn":"QEMU NVMe Ctrl                          ",
#   "fr":"10.1.5  ",
#   "rab":6,
#   "ieee":5395456,
#   "cmic":0,
#   "mdts":7,
#   "cntlid":0,
#   "ver":66560,
#   "rtd3r":0,
#   "rtd3e":0,
#   "oaes":256,
#   "ctratt":32768,
#   "rrls":0,
#   "bpcap":0,
#   "nssl":0,
#   "plsi":0,
#   "cntrltype":1,
#   "fguid":"00000000-0000-0000-0000-000000000000",
#   "crdt1":0,
#   "crdt2":0,
#   "crdt3":0,
#   "crcap":0,
#   "nvmsr":0,
#   "vwci":0,
#   "mec":0,
#   "oacs":298,
#   "acl":3,
#   "aerl":3,
#   "frmw":3,
#   "lpa":7,
#   "elpe":0,
#   "npss":0,
#   "avscc":0,
#   "apsta":0,
#   "wctemp":343,
#   "cctemp":373,
#   "mtfa":0,
#   "hmpre":0,
#   "hmmin":0,
#   "tnvmcap":0,
#   "unvmcap":0,
#   "rpmbs":0,
#   "edstt":0,
#   "dsto":0,
#   "fwug":0,
#   "kas":0,
#   "hctma":0,
#   "mntmt":0,
#   "mxtmt":0,
#   "sanicap":0,
#   "hmminds":0,
#   "hmmaxd":0,
#   "nsetidmax":0,
#   "endgidmax":0,
[...]

PR #3323, opened 2026-04-27
nvme: fix copy command nr value as zero based

ikegami-t:
Since it was set the maximum value of argconfig_parse_comma_sep_array_u**. But those values are not zero based then fix to set with minus 1.


ikegami-t:
Let me close this PR as commented by the issue comment: https://github.com/linux-nvme/nvme-cli/issues/3322#issuecomment-4328175765.

PR #3325, opened 2026-04-27
nvme: Allow commit action = 0b011 (replace and activate immediate)

NateThornton:
Prior commit 63e7fcf refactored the supported commit-action types from hard coded values to enumerated types, but left out commit action 0b011 `NVME_FW_COMMIT_CA_REPLACE_AND_ACTIVATE_IMMEDIATE` as a support value.

I don't believe this was intentional, but maybe @ikegami-t can comment.


ikegami-t:
Looks good. Yes the error was not intended. Sorry for the error. Thank you.

igaw:
Thanks!

PR #3330, opened 2026-04-28
nbft: Moves public nbft methods from libnvme to libnvmf

brgoing:
Fixes check-public-headers test with fabrics disabed.  Since nbft is not included when fabrics is disabled, moves public nbft methods from libnvme.ld to libnvmf.ld and renames them with the libnvmf prefix.

Also fixes build warnings in private.h when building with fabrics disables but netdb available.  Includes ifaddrs in private.h if needed.


igaw:
Because I had to rebase this change, I decided to split the change into two patches. The doc is autogenerated thus I dropped this part. Applied to master, thanks!

PR #3334, opened 2026-04-30
python bindings: exception hierarchy + PUBLISHING.md

martin-belanger:
### Exception hierarchy

Replaces the `connect_err` / `discover_err` globals and `%exception` shims with a proper exception hierarchy in `libnvme/_exceptions.py`:

```
NvmeError              base class — carries .errno (int) and .message (str)
├── ConnectError       raised by ctrl.connect()
├── DisconnectError    raised by ctrl.disconnect()
├── DiscoverError      raised by ctrl.discover()
└── NotConnectedError  raised when a connected controller is required (.errno is always 0)
```

Exceptions are imported into the `nvme` module namespace at init time so callers can use `nvme.ConnectError` etc. directly.

On the C side, `raise_nvme(cls, err)` sets the exception; `raise_not_connected()` handles the no-errno case. One subtlety worth noting: `SWIG_fail` (`goto fail;`) cannot be used inside `%extend` function bodies because SWIG extracts those into `SWIGINTERN` helper functions where the `fail:` label is absent. Minimal `%exception` blocks are used instead to propagate any pending exception from the helper back into the wrapper where `fail:` is defined.

`supported_log_pages()` is renamed to `get_supported_log_pages()` and now raises `NvmeError` on failure instead of returning `None`.

`discover-loop.py` and `README.md` are updated to match the new API.

### PUBLISHING.md

Documents the PyPI publishing workflow: source distribution only (no wheels), GitHub Actions handles both TestPyPI (on every master push) and PyPI (on release tags) using OIDC authentication.

### DIFF
--- a/libnvme/libnvme/_exceptions.py
+++ b/libnvme/libnvme/_exceptions.py
@@ -1,0 +1,35 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# This file is part of libnvme.
+# Copyright (c) 2026, Dell Technologies Inc. or its subsidiaries.
+# Authors: Martin Belanger <Martin.Belanger@dell.com>
+
+
+class NvmeError(Exception):
+    """Base class for all libnvme errors.
+
+    Attributes:
+        errno:   OS error number (negative values are stored as-is).
+        message: Human-readable description from libnvme_errno_to_string().
+    """
+    def __init__(self, errno, message):
+        self.errno = errno
+        self.message = message
+        super().__init__(f"[Errno {errno}] {message}")
+
+
+class ConnectError(NvmeError):
+    """Raised when a controller connection attempt fails."""
+
+
+class DisconnectError(NvmeError):
+    """Raised when a controller disconnect attempt fails."""
+
+
+class DiscoverError(NvmeError):
+    """Raised when a discovery log retrieval fails."""
+
+
+class NotConnectedError(NvmeError):
+    """Raised when an operation requires a connected controller but none exists."""
+    def __init__(self, message="Not connected"):
+        super().__init__(0, message)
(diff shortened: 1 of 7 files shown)


libnvme/libnvme/_exceptions.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
igaw, line 1: Could you add your Copyright statement here as well?

igaw:
Rest looks good.

libnvme/libnvme/_exceptions.py
martin-belanger: Done.

martin-belanger:
Quick note: This PR still fails the checkpatch (even with the new checkpatch) because it does not have quite the right Assisted-by signature. Future commits will have the right one.

igaw:
Christoph doesn't like the `Assisted-by` tag. It doesn't really tell anything except a tool was involved. The SoB from you is what counts. Sure if a LLM or so finds a bug it might be worth to mention how it was found etc, but with general code contributions it is not really necessary. I agree with him, I don't see a good reason for it. I'll merge this one now, but I suggest we stop using it in future.

martin-belanger:
Ok I will stop using it. BTW, this was suggested by Linus Torvalds:
https://www.zdnet.com/article/linus-torvalds-and-maintainers-finalize-ai-policy-for-linux-kernel-developers/

igaw:
Ahh, okay, so we have conflicting rules from Linus and Christoph :) So not sure what is best then. I think it makes sense to follow what Linus says?

martin-belanger:
> Ahh, okay, so we have conflicting rules from Linus and Christoph :) So not sure what is best then. I think it makes sense to follow what Linus says?

Yep, I was just trying to follow what the "big boss" said. Maybe Christoph  is not aware. I will do what you tell me. You're the maintainer after all. 😉

igaw:
yeah, confusing times. I'd say we do what Linus says here...

PR #3335, opened 2026-04-30
ci: replace webispy/checkpatch-action with upstream checkpatch.pl

martin-belanger:
## Summary

- Drops the `webispy/checkpatch-action` third-party action (frozen at 2022) in favour of fetching `checkpatch.pl` directly from `torvalds/linux` at workflow runtime
- Eliminates supply chain risk from a pinned, unmaintained third-party action
- Picks up current kernel checkpatch improvements, including proper support for the `Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL]` signature format
- All project-specific settings remain in `.checkpatch.conf` and are auto-read by `checkpatch.pl` — no behaviour change there

## Notes

The old action had an unused `Calculate PR commits + 1` step left over from an earlier `fetch-depth` strategy; removed as part of this cleanup since `fetch-depth: 0` is hardcoded in the checkout step.

Tested locally by running the new pipeline against this branch and against a branch with known `Assisted-by` commits — 0 errors, 0 warnings on both.

### DIFF
--- a/.github/workflows/checkpatch.yml
+++ b/.github/workflows/checkpatch.yml
@@ -13,11 +13,17 @@ jobs:
     name: checkpatch review
     runs-on: ubuntu-latest
     steps:
-    - name: 'Calculate PR commits + 1'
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3336, opened 2026-05-01
python bindings: remove deprecated files

martin-belanger:
These are leftover files no longer needed following the redesing of of the python bindings.


igaw:
Thanks!

PR #3337, opened 2026-05-01
Makefile: add checkpatch targets

martin-belanger:
Add two targets for running checkpatch locally before committing, mirroring the CI workflow.

`make checkpatch` checks all commits on the current branch not yet in the base. The base defaults to `master` but can be overridden via `BASE=`:

```
make checkpatch                       # compare vs local master
make checkpatch BASE=upstream/master  # for fork workflows where origin= personal fork
make checkpatch BASE=HEAD~3           # check only the last 3 commits
```

`make checkpatch-diff` checks uncommitted work — staged and unstaged changes to tracked files, plus any untracked files:

```
make checkpatch-diff
```

`checkpatch.pl` is fetched from `torvalds/linux` on first use and cached in `/tmp`, matching the CI fetch introduced in #3335.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,3 +91,21 @@ static:
 		-Dexamples=false
 
 	meson compile -C ${BUILD-DIR}
+
+CHECKPATCH     := /tmp/checkpatch.pl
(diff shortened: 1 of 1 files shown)


igaw:
Thanks a lot. One more down of the 3rd party actions :)

PR #3338, opened 2026-05-02
ci: extend check-accessors to cover all generated files

martin-belanger:
### Problem

The `check-accessors` CI workflow only verified two of the eight files produced by `generate-accessors.py`:
|File|CI coverage (before)|
|:---|:---|
| libnvme/src/nvme/accessors.h | ✓ checked |
| libnvme/src/nvme/accessors.c | ✓ checked |
| libnvme/src/accessors.ld              | ✗ not checked |
| libnvme/src/nvme/accessors-fabrics.h  | ✗ not checked |
| libnvme/src/nvme/accessors-fabrics.c  | ✗ not checked |
| libnvme/src/accessors-fabrics.ld      | ✗ not checked |
| libnvme/libnvme/accessors.i           | ✗ not checked |
| libnvme/libnvme/accessors-fabrics.i   | ✗ not checked |

A developer could modify `private.h` or `private-fabrics.h`, regenerate, and push with any of those six files stale — and CI would pass silently.

The `.ld` files have an additional wrinkle: they cannot be checked with a simple `git diff` because their version-section labels (`LIBNVME_ACCESSORS_3`, etc.) are assigned manually by the maintainer. The right check is whether the **symbol set** has drifted, not whether the file is byte-identical.

### Solution

Add a `-Dcheck-accessors=true` meson option. When set, `update-accessors.sh` runs in read-only CI mode (`--check` flag):

- `.h`, `.c`, `.i` — byte-compared against the source tree; any mismatch is reported as STALE and causes a non-zero exit.
- `.ld` — symbol sets extracted and compared via `comm`; added or removed symbols are reported as drift and cause a non-zero exit. The file itself is not compared (version labels are intentionally different).

The existing `update-common-accessors` and `update-fabrics-accessors` targets receive `--check` when the option is enabled — no new targets needed. The CI workflow passes `-Dcheck-accessors=true` to `meson setup` and replaces the old two-step (regenerate + `git diff`) with a single `meson compile update-accessors`.

The developer workflow (`meson compile update-accessors` without the option) is unchanged: `.h`/`.c`/`.i` are updated in place, `.ld` drift is reported as an advisory warning, and the script always exits 0.

### Files changed

| File | Change |
|---|---|
| `meson_options.txt` | Add `check-accessors` boolean option (default `false`) |
| `libnvme/tools/generator/meson.build` | Splice `--check` into both targets when option is set |
| `libnvme/tools/generator/update-accessors.sh` | Add `--check` mode: read-only, exits non-zero on any drift |
| `.github/workflows/check-accessors.yml` | Pass `-Dcheck-accessors=true`; drop separate `git diff` step |

### DIFF
--- a/.github/workflows/check-accessors.yml
+++ b/.github/workflows/check-accessors.yml
@@ -6,12 +6,18 @@
 #
 # Authors: Martin Belanger <Martin.Belanger@dell.com>
 #
-# Verify that the committed accessor files (accessors.h, accessors.c) are
-# up to date with the generate-accessors tool and the structs in private.h.
+# Verify that all committed generated accessor files (.h, .c, .i, .ld) are
+# in sync with the generate-accessors tool and the annotated private*.h files.
 #
-# If this check fails, run the following command and commit the result:
+# .h, .c, and .i files must be byte-identical to what the generator produces.
+# .ld files are checked at the symbol level: the committed file may carry
+# different version-section labels but must export the same set of symbols.
 #
-#   meson compile -C <build-dir> update-accessors
+# If this check fails:
+#   - For .h/.c/.i: run 'meson compile -C <build-dir> update-accessors'
+#                   and commit the result.
+#   - For .ld:      manually add/remove the symbols listed in the WARNING and
+#                   commit the result.
 
 name: Check accessors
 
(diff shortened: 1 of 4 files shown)


igaw:
Thanks!

PR #3340, opened 2026-05-05
CI: Add host_devices list to kubevirt-action

MaisenbacherD:
We need to specify the devices that are passed though to the VM when running the tests. The kubevirt-action should not be responsible for picking the testing devices.


igaw:
Thanks!

PR #3342, opened 2026-05-08
python bindings: table-driven fabrics config dict parsing

martin-belanger:
## python bindings: table-driven fabrics config dict parsing

Replace the hand-written if-chain in set_fctx_from_dict() with a table-driven approach. The new design splits the function into two focused helpers:

- set_fctx_fabrics_config() — iterates auto-generated sentinel tables to write int/long/bool scalar fields directly by offset; adding a new scalar field to libnvme_fabrics_config requires no change here.
- set_fctx_host_params() — handles fields that need libnvmf_context_set_*() helpers (hostnqn, hostid, TLS keys, persistent flag).

The sentinel tables are generated into fctx_field_tables.h by a new `!generate-dict-table` struct annotation processed by generate-accessors.py. Members annotated `!dict-table:none` are excluded from the tables.

Changes:
| File | Change |
| :--- | :--- |
| `libnvme/libnvme/fctx_field_tables.h`           | new generated header |
| `libnvme/libnvme/nvme.i`                        | table-driven `set_fctx_from_dict()` |
| `libnvme/libnvme/meson.build`                   | depend_files + include_directories |
| `libnvme/src/nvme/private.h`                    | `!generate-dict-table` annotation |
| `libnvme/tools/generator/generate-accessors.py` | `!generate-dict-table` support |
| `libnvme/tools/generator/generate-accessors.md` | updated docs |
| `libnvme/tools/generator/update-accessors.sh`   | `--dict-table-out` in both branches |
| `libnvme/tools/generator/meson.build`           | `--dict-table-out` wired in |

### DIFF
--- a/libnvme/libnvme/nvme.i
+++ b/libnvme/libnvme/nvme.i
@@ -97,18 +97,21 @@ static PyObject *NvmeDiscoverError     = NULL;
 static PyObject *NvmeNotConnectedError = NULL;
 static PyObject *fctx_known_keys       = NULL;
 
-static void raise_nvme(PyObject *cls, int err) {
+static void raise_nvme(PyObject *cls, int err)
+{
 	const char *s = libnvme_errno_to_string(err < 0 ? -err : err);
 	PyObject *args = Py_BuildValue("(is)", err, s ? s : "unknown");
 	PyErr_SetObject(cls, args);
 	Py_DECREF(args);
 }
 
-static void raise_not_connected(void) {
+static void raise_not_connected(void)
+{
 	PyErr_SetString(NvmeNotConnectedError, "Not connected");
 }
 
-static void PyDict_SetItemStringDecRef(PyObject * p, const char *key, PyObject *val) {
+static void PyDict_SetItemStringDecRef(PyObject *p, const char *key, PyObject *val)
+{
 	PyDict_SetItemString(p, key, val); /* Does NOT steal reference to val .. */
 	Py_XDECREF(val);                   /* .. therefore decrement ref. count. */
 }
--- a/libnvme/libnvme/nvme.i
+++ b/libnvme/libnvme/nvme.i
@@ -86,6 +86,7 @@ PyObject *read_hostid();
 #include <libnvme.h>
 #include "nvme/private.h"
 #include "nvme/private-fabrics.h"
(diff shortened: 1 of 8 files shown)


libnvme/libnvme/nvme.i
```diff
@@ -94,34 +95,26 @@ static PyObject *NvmeConnectError      = NULL;
 static PyObject *NvmeDisconnectError   = NULL;
 static PyObject *NvmeDiscoverError     = NULL;
 static PyObject *NvmeNotConnectedError = NULL;
+static PyObject *fctx_known_keys       = NULL;
 
-static void raise_nvme(PyObject *cls, int err) {
+static void raise_nvme(PyObject *cls, int err)
+{
```
igaw, line 101: doesn't seem necessary this whitespace change. Also I'd like to keep the style consistent in nvme.i, that is the bracket should stay on the same line (yes, it is inconsistent with any block start but that's a different discussion :))

igaw: btw, I don't mind to change it for nvme.i if you say it makes more sense to follow a different coding style for SWIG. Just saying that I'd like to keep the formatting within a file consistent.

igaw:
Apart my nitpicking of the coding style, looks good. The table parser approach is a good cleanup.

martin-belanger:
Thanks for the review — glad the table-driven approach reads well!

On the brace style: the change was intentional. The C helper functions in nvme.i were the last holdouts using K&R-style opening braces, inconsistent with the rest of libnvme's C code which follows Linux kernel style (opening brace on its own line for function definitions). Since I was already touching those functions, I took the opportunity to align them.

SWIG-specific constructs (`%extend`, `%typemap`, etc.) still use braces on the same line — that's SWIG syntax, not C, so it's a different category. Within the C function definitions in the file, everything now follows the same convention.

Happy to discuss further if you'd prefer a different approach, but from my side the intent was to reduce inconsistency rather than introduce it.

igaw:
Okay, but then these changes should really go into a separate patch.

martin-belanger:
Ok. Got it. Let me rework it.

martin-belanger:
Done. Two separate commits in one PR.

igaw:
Thanks!

PR #3344, opened 2026-05-09
doc: add error-log command description for fields and ordering

ikegami-t:
As all fields including the ordering described in the spec.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -521,16 +521,21 @@ void json_nvme_id_ctrl(struct nvme_id_ctrl *ctrl,
 }
 
 static void json_error_log(struct nvme_error_log_page *err_log, int entries,
-			   const char *devname)
+			   const char *devname,
+			   struct nvme_error_log_filter *flt)
 {
 	struct json_object *r = json_create_object();
 	struct json_object *errors = json_create_array();
+	struct json_object *error;
 	int i;
 
 	obj_add_array(r, "errors", errors);
 
 	for (i = 0; i < entries; i++) {
-		struct json_object *error = json_create_object();
+		if (nvme_is_error_log_filter(&err_log[i], flt))
+			continue;
+
+		error = json_create_object();
 
 		obj_add_uint64(error, "error_count", le64_to_cpu(err_log[i].error_count));
 		obj_add_int(error, "sqid", le16_to_cpu(err_log[i].sqid));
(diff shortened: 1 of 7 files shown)


ikegami-t:
Currently added 9 commits for the error-log command changes then if needed will do combine the commits or separate the PR so please let me know.

nvme.c
```diff
@@ -1206,8 +1206,15 @@ static int get_error_log(int argc, char **argv, struct command *acmd, struct plu
 		"error log entries from a given device "
 		"in either decoded format (default) or binary.";
 	const char *log_entries = "number of entries to retrieve";
-	const char *raw = "dump in binary format";
+	const char *status = "output specified STATUS entry only";
+	const char *nsid = "output specified NSID entry only";
+	const char *trtype = "output specified TRTYPE entry only";
+	const char *opcode = "output specified OPC entry only";
+	const char *sqid = "output specified SQID entry only";
 	const char *valid_entry = "output valid entry only";
```
igaw, line 1214: as far I can tell, `valid_entry` is not used anymore. Maybe you could merge this patches together so it's only one patch? I mean the patches which introduce this filter logic.

nvme-print-json.c
```diff
@@ -562,6 +565,12 @@ static void json_error_log(struct nvme_error_log_page *err_log, int entries,
 		array_add_obj(errors, error);
 	}
 
+	if (entries == filtered) {
+		error = json_create_object();
+		obj_add_str(error, "no error", "all entries filtered");
+		array_add_obj(errors, error);
+	}
+
```
igaw: Not sure if we should add this at all. The users asked for filtering and if there are no entries, we just return an empty array, no?

igaw:
Looks good, I would just merge the filter logic patches into one. 

Could also update the documentation and the shell completion? But this can go also afterwards. Thanks!

nvme.c
ikegami-t: No still the valid-entry option is used so needed to remain. Okay I will do merge the patches to introduce the filter logic.

nvme-print-json.c
ikegami-t: Just followed as same with the stdout output but yes will do delete the object for the json output.

ikegami-t:
Just fixed the patches as metioned. Sorry let me update the documentation and the shell completion separately. I will do that later by additional a patch or PR. Thank you.

igaw:
Thanks!

PR #3345, opened 2026-05-11
libnvme: add __libnvme_ prefix to all compiler attributes

martin-belanger:
### Problem

The macros defined in `compiler-attributes.h` — `__public`, `__weak`, and `__unused` — use short, generic names that sit in the C implementation-reserved namespace (`__` prefix). This caused a real build failure: musl libc uses `__unused` as a struct field name in `bits/stat.h`:

```c
long __unused[3];
```

When `compiler-attributes.h` is included before a musl system header, the preprocessor rewrites that field declaration to:

```c
long __attribute__((__unused__))[3];
```

which is a syntax error.

### Fix

Rename all three macros with a project-specific prefix:

| Old name | New name |
|---|---|
| `__public` | `__libnvme_public` |
| `__weak` | `__libnvme_weak` |
| `__unused` | `__libnvme_unused` |

The `__libnvme_` prefix is specific enough that no system header will ever use it as an identifier, while keeping the double-underscore convention already used throughout the codebase.

### Scope

- All source files under `libnvme/src/nvme/` updated
- Accessor generator (`generate-accessors.py`) updated; `accessors.c` and `accessors-fabrics.c` regenerated
- `check-public-symbols.py` and `check-public-headers.py` updated

### checkpatch

The longer prefix pushes some function signatures over the 80-column limit. Long signatures are wrapped so arguments begin on the continuation line. Three signatures remain marginally over 80 columns — `__libnvme_public` plus the return type and function name alone fills the line before any arguments appear. Separating the attribute from the function name is not valid, so these three warnings are unavoidable.

### DIFF
--- a/libnvme/src/nvme/accessors-fabrics.c
+++ b/libnvme/src/nvme/accessors-fabrics.c
@@ -29,7 +29,8 @@
  * Accessors for: struct libnvmf_discovery_args
  ****************************************************************************/
 
-__public int libnvmf_discovery_args_new(struct libnvmf_discovery_args **pp)
+__libnvme_public int libnvmf_discovery_args_new(
+		struct libnvmf_discovery_args **pp)
 {
 	if (!pp)
 		return -EINVAL;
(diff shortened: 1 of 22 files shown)


igaw:
Thanks!

PR #3349, opened 2026-05-12
libnvme: probe 64-bit ioctl support without submitting a command

Mateusz-Nowicki-Embedded:
Replace the dummy admin SQE submission used to probe `NVME_IOCTL_ADMIN64_CMD` availability with a NULL-argp ioctl,
which distinguishes "ioctl unknown" (ENOTTY) from "ioctl known, arg invalid" (EFAULT) without sending any
NVMe command to the controller.

Side benefits:
- eliminates the uninitialized-`ns` info leak through dummy.nsid 
- removes a malformed `Delete I/O SQ` SQE per controller open, which would otherwise be observable in admin SQ ring memory
   and as an `Invalid Queue Identifier` CQE in traces

Compatible with all kernels supported by nvme-cli (>= v4.15): 
- 4.15..5.6: `NVME_IOCTL_ADMIN64_CMD` doesn't exist → ENOTTY, same outcome as today
- 5.7+: `copy_from_user(NULL)` fails with EFAULT before the command leaves nvme_user_cmd64()

### DIFF
--- a/libnvme/src/nvme/lib.c
+++ b/libnvme/src/nvme/lib.c
@@ -146,7 +146,6 @@ __public void libnvme_transport_handle_set_timeout(
 static int __nvme_transport_handle_open_direct(
 		struct libnvme_transport_handle *hdl, const char *devname)
 {
-	struct libnvme_passthru_cmd dummy = { 0 };
 	__cleanup_free char *path = NULL;
 	char *name;
 	int ret, id, ns;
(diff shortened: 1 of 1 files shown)


igaw:
Thanks for the fix. Ideally we don't probe at all anymore, instead we just try the CMD64 variant when we issue the first command and if this fails we use the CMD32 version. Though this is slightly difficult to implement, it would make the overhead of probing going away for modern system, which is a good thing IMO.

igaw:
I've merged the defer/lazy ioctl approach. Hopefully, this should address the issue and make it actually work also for IO. Let me know if this doesn't work as expected. Thanks!

Mateusz-Nowicki-Embedded:
> I've merged the defer/lazy ioctl approach. Hopefully, this should address the issue and make it actually work also for IO. Let me know if this doesn't work as expected. Thanks!

yep, I verified it and dummy cmd is not generated anymore, thanks!

PR #3351, opened 2026-05-12
libnvme: abstract memory (de)allocation API

igaw:
This is based on https://github.com/linux-nvme/nvme-cli/pull/3310 which does too many things in one patch. 
Split the patch into baby steps and see how things go. 

- [x] replace dprintf
- [x] platform support for mem allocation helpers
- [ ]


libnvme/src/nvme/mem-win.c
```diff
@@ -0,0 +1,152 @@
...
+
+__libnvme_public void *libnvme_realloc(void *p, size_t len)
+{
+	size_t old_len;
+	void *result;
+
+	if (!p)
+		return libnvme_alloc(len);
+
+	old_len = _msize(p);
```
brgoing, line 57: I found today during some testing that _msize was incorrect for working with aligned allocations.  There is an _aligned_msize that needs to be used instead.
```suggestion
	old_len = _aligned_msize(p, getpagesize(), 0);
```

igaw: alright, I'll pick up this version then :)

igaw:
merged these patches separately.

PR #3354, opened 2026-05-12
Refactore memory API to make them portable

igaw:
Refactore the memory APIs so that it's possible to provide versions for Windows. This reduces the code duplication between between libnvme and nvme-cli but means there are a handful new exported functions. But given this is tricky code it might makes sense to provide this for all users anyway.


igaw:
@bgoing-micron-oss would you mind to review this here? I splitted the memory API changes into this single PR. I think I got most of it covered. 

I've prepare another PR based on your #3310. So far I covered dprintf and getline (#3355). And then I want to tackle the signal API.

brgoing (approved):

igaw:
okay let's go with those changes and see what I missed :)

PR #3360, opened 2026-05-13
mem-win: Remove unneeded includes and definitions.

brgoing:
Removes fsync definition and includes that are not needed for the functionality contained in mem-win.c.


igaw:
Thanks!

PR #3361, opened 2026-05-14
nvme: Use libnvme_free for all memory allocated with libnvme_alloc

brgoing:
Fixes remaining __cleanup_free uses for buffers allocated using libnvme_alloc.  Updates to use __cleanup_libnvme_free.


igaw:
Thanks!

PR #3362, opened 2026-05-14
Minor OCP Fixes

brgoing:
Fixes a no-uuid parameter detection issue and an error printing issue in the ocp plugin.


igaw:
Thanks!

PR #3364, opened 2026-05-15
micron-nvme cleanup

brgoing:
Clean-up and fixes for the Micron plugin.

- Updates all log buffers to use libnvme_alloc and libnvme_free.
- Fixes telemetry log reading.
- Fixes vs-drive-info parameter validation logic that was always failing.
- Cleans up vs-nand-stats error reporting and duplicate id-ctrl.
- Fixes error reporting for positive error numbers in multiple commands.


igaw:
Thanks!

PR #3365, opened 2026-05-15
libnvme: extract fabrics-specific logic into tree-fabrics.c

martin-belanger:
## Summary

`tree.c` currently mixes PCIe/generic tree management with NVMe-oF-specific logic, including TCP address matching, TLS/DHChap sysfs readers, and controller lookup helpers that are only meaningful in a fabrics context. This series begins extracting that fabrics-specific code into a new `tree-fabrics.c`, compiled only when `want_fabrics=true`, following the existing `*-fabrics.c` convention already established by `util-fabrics.c` and `accessors-fabrics.c`.

This is a **preparation PR**. Phases 1–3 are included here. Phase 4 (moving the full ctrl-find cluster: `libnvme_lookup_ctrl`, `libnvme_candidate_init`, `libnvme_ctrl_find`, `libnvme_tree_ctrl_match`) will follow in a separate series once the `libnvmf_context` coupling in `libnvme_ctrl_alloc()` has been resolved.

### Phase 1 — Create `tree-fabrics.c`, eliminate `#ifdef CONFIG_FABRICS` from `tree.c`

- Creates `src/nvme/tree-fabrics.c` and wires it into `src/meson.build`.
- Moves the five `_tcp_*` static matching functions out of `tree.c`.
- Extracts the TCP/RDMA transport-selection block from `_candidate_init()` into a new `libnvmf_candidate_init()` hook called from the generic path.
- Adds a `libnvmf_candidate_init()` stub to `no-fabrics.c` (returns `NULL`).
- Moves `struct candidate_args` and the `ctrl_match_t` typedef to `private-fabrics.h`.
- After this phase, `tree.c` has **zero** `#ifdef CONFIG_FABRICS` guards.

### Phase 2 — Move TLS/DHChap sysfs readers

- Moves `libnvmf_read_sysfs_dhchap()`, `libnvmf_read_sysfs_tls()`, and `libnvmf_read_sysfs_tls_mode()` from `tree.c` to `tree-fabrics.c` as `static` helpers, behind a single non-static wrapper `libnvmf_read_sysfs_fabrics_attrs()`.
- `libnvme_reconfigure_ctrl()` in `tree.c` is updated to call the wrapper.
- A no-op stub is added to `no-fabrics.c`.
- These attributes only exist on NVMe-oF controllers; in PCIe-only builds the sysfs reads were compiled in but never returned data — now they are excluded entirely.

### Phase 3 — Move `libnvmf_ctrl_match_config` and `libnvmf_ctrl_find`; rename cleanup

- `libnvmf_ctrl_match_config()` (previously `_libnvme_ctrl_match_config`) and `libnvmf_ctrl_find()` (previously the 2-param `libnvme_ctrl_find`) are moved to `tree-fabrics.c`. Both are purely fabrics-facing (all callers are in `fabrics.c`).
- `libnvme_ctrl_match_config()` is **dropped** from the public API entirely. It was exported in `libnvme.ld` but had no external callers — nvme-cli does not call it, and nvme-stas goes through the Python bindings. It was dead API.
- All underscore-prefixed functions that became non-static (crossing file boundaries) are renamed to use proper `libnvme_` / `libnvmf_` prefixes: `__nvme_ctrl_find` → `libnvme_ctrl_find` (3-param), `_candidate_init` → `libnvme_candidate_init`, `_tree_ctrl_match` → `libnvme_tree_ctrl_match`, `_libnvme_create_ctrl` → `libnvme_create_ctrl`.
- `_discovery_config_json()` in `fabrics.c` is made `static` (it was accidentally non-static despite being file-local).
- No new stubs are needed in `no-fabrics.c` — the moved functions have no callers outside the fabrics layer.

## What remains for the follow-up series (Phase 4)

The following functions still live in `tree.c` because `libnvme_lookup_ctrl()` is called by both `fabrics.c` and `libnvme_ctrl_alloc()` (which constructs a `libnvmf_context` as a parameter bag for all transport types, including PCIe). This coupling prevents `libnvme_lookup_ctrl()` and its dependencies from moving cleanly:

- `libnvme_lookup_ctrl()`
- `libnvme_candidate_init()`
- `libnvme_ctrl_find()` (3-param)
- `libnvme_tree_ctrl_match()`

Phase 4 will introduce a PCIe-specific path in `libnvme_ctrl_alloc()` that does not go through `libnvmf_context`, after which the full ctrl-find cluster can move to `tree-fabrics.c` and `tree.c` will no longer need to include `private-fabrics.h`.


martin-belanger:
@igaw - **Note:** One class of checkpatch warning could not be resolved without introducing a different warning. The test file uses `printf("test_foo:\n")` to label test sections. Checkpatch flags this with "Prefer using `__func__`". Switching to `printf("%s:\n", __func__)` eliminates that warning but immediately triggers "`ftrace`-like logging detected". There is no form that satisfies both checks simultaneously, so the hardcoded function-name strings were kept.

libnvme/src/nvme/private-fabrics.h
```diff
@@ -155,9 +154,43 @@ static inline __u16 libnvmf_exat_size(size_t val_len)
  *
  * Return: Pointer to I/F data, NULL on error (with errno set).
  */
+#if defined(HAVE_NETDB) || defined(CONFIG_FABRICS)
```
igaw: I would keep the #if before the documentation, I think this is what we already have in other places.

igaw: or we drop the kernel doc completely. I haven't checked if this will be in the main documetation. that would be wrong as this is in internal function.

igaw:
I like it :) 

We should check if the kdoc is actually showing up in the API doc, if not than this is good to go, otherwise I would say it's good time to address it when the code is moved around. I guess it is then '/* */' style doc.

libnvme/src/nvme/private-fabrics.h
martin-belanger: I moves the `#if defined(HAVE_NETDB) || defined(CONFIG_FABRICS)` before the documentation.

martin-belanger:
> We should check if the kdoc is actually showing up in the API doc, if not than this is good to go, otherwise I would say it's good time to address it when the code is moved around. I guess it is then '/* */' style doc.

The kdoc in `private-fabrics.h` does not show up in the API docs. The doc build (`doc/meson.build`) feeds only the public headers to kernel-doc, and `private-fabrics.h` is not among them. So we're good to go.

igaw:
Great cleanup. I like it! Thanks!

PR #3366, opened 2026-05-15
sigaction compatibility for windows

brgoing:
Addresses the lack of sigaction on Windows.
- Provides platform-specific implementations for sighdl (sighdl-linux.c/sighdl-win.c).
- Adds a check for sigaction support, and updates uses of sigaction to fall back to using signal if sigaction is not supported.


igaw:
Nice! Thanks this looks way nicer instead of platform headers,  IMO :)

PR #3367, opened 2026-05-15
common: Add compatibility wrappers for common functions.

brgoing:
Adds windows compatibility wrappers for common functions used by nvme-cli and plugins.


igaw:
Okay, for the time being this looks like good compromise. At least the min/max thing should be addressed by using the ccan versions eventually. But one impossible thing at the time. Thanks!

PR #3368, opened 2026-05-16
util: Move get_entity_name and get_entity_version into util-fabrics.

brgoing:
The get_entity_name and get_entity_version utility functions were only used for fabrics.  Moves their declarations into private-fabrics.h and their implementations into util-fabrics.c, and adds libnvmf prefix for consistency with other fabrics-only utilities.


igaw:
Sorry for the long delay.  I was on two conferences last week and didn't have time to work on upstream. Operations resumed to normal :)

Thanks!

PR #3369, opened 2026-05-18
tree: remove fabrics-only code from tree.c

martin-belanger:
# Summary

The ultimate goal of this series is to completely remove fabrics-only code from `tree.c` so that PCIe-only builds carry zero NVMe-oF overhead. The previous series (already merged) created `tree-fabrics.c` and moved the TCP matching functions and sysfs readers there. This series goes further by decoupling the controller allocation and lookup machinery from `libnvmf_context`, then moving the ctrl-find cluster itself into `tree-fabrics.c`, leaving `tree.c` with no `#include "private-fabrics.h"` dependency.

## fabrics: rename callbacks to hooks (327e17cc2)

Renames `libnvmf_callbacks` → `libnvmf_hooks` and the corresponding `libnvmf_context_set_*_cbs()` accessors to `libnvmf_context_set_*_hooks()`. The term "hooks" is more accurate — these are injection points for application behaviour, not observer callbacks. Preparatory rename before the structural changes that follow.

## fabrics: group hooks and user_data into struct libnvmf_hooks (7efbed4b7)

Embeds all hook function pointers and `user_data` into a nested `struct libnvmf_hooks hooks` inside `libnvmf_context`. This makes the relationship between the hooks and their shared `user_data` explicit in the type system. Call sites change from `fctx->hook_name(...)` to `fctx->hooks.hook_name(...)`.

## tree/fabrics: introduce libnvme\_ctrl\_params and embed in libnvmf\_context (01d61a178)

Introduces `struct libnvme_ctrl_params` to group the connection-target fields (`transport`, `traddr`, `trsvcid`, `subsysnqn`, `host_traddr`, `host_iface`, and the embedded `struct libnvme_fabrics_config`). The struct is embedded as `ctrl_params` in `libnvmf_context`, replacing the individual fields that were scattered across the context. Also introduces `libnvme_fabrics_config_copy()` as a safe alternative to raw struct assignment for `struct libnvme_fabrics_config`, guarding against future additions of owned pointer members.

## fabrics: fix libnvmf\_connect() not propagating resolved TLS key IDs (3a1667ef1)

Pre-existing bug uncovered during refactoring: `nvme_parse_tls_args()` resolves symbolic TLS key identifiers into numeric IDs and writes them into `fctx->ctrl_params.cfg`, but this happens *after* `libnvme_create_ctrl()` has already copied `cfg` to `c->cfg`, so the resolved IDs never reached the controller. The fix mirrors the pattern already present in `libnvmf_config_modify()`: call `update_config(c, &fctx->ctrl_params.cfg)` after `nvme_parse_tls_args()`.

## tree: decouple ctrl-find/alloc from libnvmf\_context (d3ed1ed98)

The ctrl-find/alloc cluster (`libnvme_candidate_init`, `libnvme_ctrl_find`, `libnvme_lookup_ctrl`, `libnvme_ctrl_alloc`) only ever accessed `fctx->ctrl_params.*`, so passing the full `libnvmf_context` was unnecessary coupling. All four functions are changed to accept `const struct libnvme_ctrl_params *` instead. `libnvme_ctrl_alloc` becomes static. `json.c` is simplified to use `struct libnvme_ctrl_params` directly and drops its `private-fabrics.h` include.

## tree: move ctrl-find cluster to tree-fabrics.c (8648654)

Moves the ctrl-find/match functions from `tree.c` into `tree-fabrics.c`: `_libnvmf_tree_ctrl_match` and `_libnvmf_candidate_init` become static; `libnvme_ctrl_find` is non-static because `tree.c::libnvme_lookup_ctrl` still calls it, so a simplified no-fabrics stub (transport + traddr matching) is added to `no-fabrics.c`. The declarations of `traddr_is_hostname`, `libnvmf_default_config`, and `libnvme_ctrl_find` are promoted from `private-fabrics.h` to `private.h`. The `#include "private-fabrics.h"` is removed from `tree.c`.

### DIFF
--- a/libnvme/src/nvme/no-fabrics.c
+++ b/libnvme/src/nvme/no-fabrics.c
@@ -20,7 +20,7 @@ void libnvmf_default_config(struct libnvme_fabrics_config *cfg)
 
 ctrl_match_t libnvmf_candidate_init(struct libnvme_global_ctx *ctx,
 		struct candidate_args *candidate,
-		const struct libnvmf_context *fctx)
+		const struct libnvme_ctrl_params *params)
 {
 	return NULL;
 }
--- a/libnvme/src/nvme/no-fabrics.c
+++ b/libnvme/src/nvme/no-fabrics.c
@@ -6,7 +6,7 @@
  * Authors: Daniel Wagner <dwagner@suse.de>
  */
 
-#include "private-fabrics.h"
+#include "private.h"
 
 bool traddr_is_hostname(struct libnvme_global_ctx *ctx,
 		const char *transport, const char *traddr)
@@ -18,14 +18,27 @@ void libnvmf_default_config(struct libnvme_fabrics_config *cfg)
 {
 }
 
-ctrl_match_t libnvmf_candidate_init(struct libnvme_global_ctx *ctx,
-		struct candidate_args *candidate,
-		const struct libnvme_ctrl_params *params)
-{
-	return NULL;
-}
-
 void libnvmf_read_sysfs_fabrics_attrs(struct libnvme_global_ctx *ctx,
 		libnvme_ctrl_t c)
 {
+}
+
+libnvme_ctrl_t libnvme_ctrl_find(libnvme_subsystem_t s,
+		const struct libnvme_ctrl_params *params, libnvme_ctrl_t p)
+{
+#ifndef _WIN32
+	struct libnvme_ctrl *c;
+
+	c = p ? libnvme_subsystem_next_ctrl(s, p) :
+		libnvme_subsystem_first_ctrl(s);
+	for (; c; c = libnvme_subsystem_next_ctrl(s, c)) {
+		if (!streq0(c->transport, params->transport))
+			continue;
+		if (params->traddr && c->traddr &&
+		    !streqcase0(c->traddr, params->traddr))
+			continue;
+		return c;
+	}
+#endif
+	return NULL;
 }
--- a/fabrics.c
+++ b/fabrics.c
@@ -196,7 +196,7 @@ static void save_discovery_log(char *raw, struct nvmf_discovery_log *log)
 static int setup_common_context(struct libnvmf_context *fctx,
 		struct nvmf_args *fa);
 
-struct cb_fabrics_data {
+struct hook_fabrics_data {
 	struct nvmf_args *fa;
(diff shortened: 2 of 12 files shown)


libnvme/src/nvme/no-fabrics.c
```diff
@@ -6,26 +6,43 @@
...
+void libnvmf_read_sysfs_fabrics_attrs(struct libnvme_global_ctx *ctx,
+		libnvme_ctrl_t c)
 {
-	return NULL;
 }
 
-void libnvmf_read_sysfs_fabrics_attrs(struct libnvme_global_ctx *ctx,
-		libnvme_ctrl_t c)
+/* No-fabrics stub; real implementation in tree-fabrics.c */
+libnvme_ctrl_t libnvme_ctrl_find(libnvme_subsystem_t s,
```
igaw, line 26: It's a bit strange that we have to implement a libnvme_ctrl_find here. But let's sort this out later.

igaw:
trimmed the commit message slighly (removed the how part). dropped the doc changes (autogenerated) and 
removed some unnecessary comments in no-fabrics.c. 

Thanks!

libnvme/src/nvme/no-fabrics.c
martin-belanger: The implementation is needed for the sysfs scan path:
`libnvme_scan_ctrl()` → `libnvme_ctrl_alloc()` → `libnvme_lookup_ctrl()` → `libnvme_ctrl_find()`.

When (re-)scanning existing PCIe controllers from sysfs, `libnvme_ctrl_alloc()` uses `libnvme_ctrl_find()` to check if the controller is already in the in-memory tree before adding it. It's not about creating PCIe devices — it's deduplication during tree scan.

That said, you're right that it's strange. I was in a somewhat mechanical state where I moved all functions with no-fabrics counterparts into `no-fabrics.c`, but this file is really meant for trivial stubs. The basic find loop is pure tree traversal with no fabrics dependency — it belongs in `tree.c`. The fix would be to move it there and let the `tree-fabrics.c` version augment it with TCP-specific matching, leaving the no-fabrics stub truly trivial. I can make that fix in a separate PR since this one already has 6 commits.

martin-belanger:
Thanks! I know this was a lot of changes to review. I appreciate it.

PR #3370, opened 2026-05-18
util: Fix buffer overflow in process_field_size_default

### LINKED ISSUE #2954: Temporary buffer size hardcoded in process_field_size_default is too small in some cases and can cause a seg fault
`void process_field_size_default(int offset, char *sfield, __u8 *buf, int size, char *datastr)
{
        __u8  cval;
        char description_str[256] = "0x";           <----------------------------------
        char temp_buffer[3] = { 0 };

        for (unsigned char i = 0; i < (unsigned char)size; i++) {
                cval = (buf[offset + i]);

                sprintf(temp_buffer, "%02X", cval);
                strcat(description_str, temp_buffer);
        }
        sprintf(datastr, "%s", description_str);`

Micron's plugin uses this method to parse it's log data.   On some drives the description_str is between 256 and 512 bytes.   This function should not allow anything larger than 256 bytes if that is the limit it supports.  It should check for the size and fail out instead of segfaulting.  If 256 bytes is not an uppler limit that needs enforced then change the defintion of description_str from 256 bytes in size to 512 bytes resolved the issue for us.  Requesting this method get revisited to enforce its limits and possibly allow up to 512 bytes intead of 256 or allow the description_str to get allocated dynamically based on the size that is needed.

There are other methods in there with similar hardcoded temporary arrays that could also be impacted by a value that exceeds the size and will also seg fault.

jimmunn:
The process_field_size_default function used a fixed 256-byte buffer for formatting hex output, which overflowed when processing fields larger than 127 bytes (such as Reserved fields up to 302 bytes in Micron SMART logs). This caused segmentation faults.

Additionally, the loop counter was an unsigned char, limiting processing to 255 bytes maximum.

Fix by:
- Converting all process_field_size_* functions to return dynamically allocated strings instead of writing to caller-provided buffers
- Using asprintf() for consistent dynamic allocation
- Using __cleanup_free in the caller for automatic memory management
- Also fixes a pre-existing memory leak where hex_to_ascii() return value was never freed

Fixes #2954

### DIFF
--- a/util/utils.c
+++ b/util/utils.c
@@ -10,6 +10,7 @@
 #include "utils.h"
 #include "types.h"
 #include "json.h"
+#include "cleanup.h"
 
 int hex_to_int(char c)
 {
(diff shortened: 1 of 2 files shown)


igaw:
Nice fix, Thanks!

PR #3371, opened 2026-05-18
mem-win: Disable Windows min/max macros in favor of project solutions

brgoing:
Disables definitions of min/max macros coming from Windows includes. Favors local solutions for min/max definitions isntead.
- Removes #ifndef guards around min/max in common.h that were added due to Windows default min/max definitions.
- Updates mem-win.c to use ccan/minmax definitions.

### DIFF
--- a/common.h
+++ b/common.h
@@ -15,12 +15,8 @@
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
-#ifndef min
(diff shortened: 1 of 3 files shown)


brgoing:
@igaw I found a way to disable the Windows min/max macros that were conflicting with the min/max macros already being defined and used in the project.  I've backed out the workaround in common.h, and updated the existing Windows code to use ccan/minmax instead where needed.

igaw:
Nice cleanup, that is way better :)

Thanks!

PR #3372, opened 2026-05-19
Minor cleanup for Windows support

brgoing:
- Replaces single use of uint with __u32 for better consistency and compatibility. 
- Cleans up Windows check in endian.h.


igaw:
Thanks!

PR #3373, opened 2026-05-19
log: Add Windows compatibility for __libnvme_msg formatting

brgoing:
MinGW on Windows requires gnu_printf formatting to match the C99 formats used by printf on Linux by default.  Updates __libnvme_msg to use gnu_printf formatting when building on Windows with MinGW for compatibility.


igaw:
Thanks!

PR #3374, opened 2026-05-19
CI: add no-fabrics build workflow with unit tests

martin-belanger:
Add a CI workflow that configures and builds nvme-cli with all optional dependencies disabled (fabrics, mi, python, json-c, libkmod, openssl, keyutils) and runs the unit test suite against that build.

The existing build-musl-minimal-static job also builds without fabrics but explicitly disables tests (-Dtests=false), leaving the no-fabrics code path unvalidated by CI. This workflow closes that gap.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -200,6 +200,26 @@ jobs:
         run: |
           scripts/build.sh minimal_static
 
+  build-nofabrics:
+    name: no-fabrics build with unit tests
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/linux-nvme/debian:latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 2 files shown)


.github/workflows/build-nofabrics.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          meson setup .build-nofabrics   \
+            --werror                     \
+            -Dfabrics=disabled           \
+            -Dmi=disabled                \
+            -Dpython=disabled            \
+            -Djson-c=disabled            \
+            -Dlibkmod=disabled           \
+            -Dopenssl=disabled           \
+            -Dkeyutils=disabled
```
igaw: Makes sense. Though, let's follow the existing pattern: add this configuration to `scripts/build.sh` and just add a new job to the exising `build.yml`? Thanks!

martin-belanger: Done

igaw:
Thanks!

PR #3376, opened 2026-05-19
fabrics: avoid segfault at sanitize_discovery_log_entry()

martin-gpy:
Avoid a potential segfault at sanitize_discovery_log_entry() while dereferencing the nvmf_disc_log_entry pointer.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -1490,6 +1490,9 @@ out_free_log:
 static void sanitize_discovery_log_entry(struct libnvme_global_ctx *ctx,
 		struct nvmf_disc_log_entry *e)
 {
+	if (!e)
+		return;
+
 	strchomp(e->trsvcid, sizeof(e->trsvcid));
 	strchomp(e->traddr, sizeof(e->traddr));
 


libnvme/src/nvme/fabrics.c
```diff
@@ -1490,6 +1490,9 @@ static int nvme_discovery_log(libnvme_ctrl_t ctrl,
 static void sanitize_discovery_log_entry(struct libnvme_global_ctx *ctx,
 		struct nvmf_disc_log_entry *e)
 {
+	if (!e)
+		return;
```
igaw, line 1494: Did you observe a crash? I don't see how this is potential possible. All call side pass in `&log->entries[i]`. So this will be always a pointer value which is `!= NULL`. 

I don't mind starting to sanity check the log page input (all input is evil) but this seems at the wrong place.

martin-gpy: I thought I saw a core for this in the past, but not able to repro this now.  Given your reasoning above, I guess it's best we ignore this patch then.

igaw: There was a crash in `nvmf_update_tls_concat` via the nbft code. 

0f4fe4592005 ("fabrics: fix NULL dereference in nvmf_update_tls_concat")

igaw:
Let's close it for now, see above reasoning.

PR #3377, opened 2026-05-19
ocp: Fix ocp internal-log output-format handling

brgoing:
If output-format is specified, copy from nvme_args to the local options. Was always reporting missing output format and defaulting to json.


igaw:
Thanks!

PR #3378, opened 2026-05-19
sed: fixing SID and MSID reporting

bbain:
The bit that reports SID and MSID equality is zero when they are equal and one when they are not.

### DIFF
--- a/plugins/sed/sedopal_cmd.c
+++ b/plugins/sed/sedopal_cmd.c
@@ -822,7 +822,7 @@ void sedopal_print_sid_auth(void *data)
 
 	printf("\nSED Block SID Authentication:\n");
 	printf("\tSID value equal MSID            : %s\n",
-		(sid_auth_d->states & BLOCK_SID_VALUE_STATE) ? "yes" : "no");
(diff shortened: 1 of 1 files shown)


bbain:
From TCG Storage Feature Set: Block SID Authentication: 

> 4.1.1.2 SID Value State 
>
> This field specifies whether the C_PIN_SID object’s PIN column value is equal to the C_PIN_MSID object’s PIN
> column value.
> 
> This bit SHALL be cleared to zero if the C_PIN_SID object’s PIN column value is equal to the C_PIN_MSID object’s
> PIN column value.
> 
> This bit SHALL be set to one if the C_PIN_SID object’s PIN column value is not equal to the C_PIN_MSID object’s
> PIN column value.

This was displaying incorrectly for me. Confused my when setting up the drive. Only really clicked once I enabled Opal and changed the SID value and it was still reported as being the same.

igaw:
Looks good to me. Let me add Greg who maintains the sed plugin.

@gjoyce-ibm

gjoyce-ibm:
looks good

PR #3379, opened 2026-05-20
linux: Add no-linux.c with shims for methods implemented in linux.c

brgoing:
Adds compatibility shims for linux.c methods when building on Windows.


igaw:
I suggest we split these function out of linux.c into crypto.c and have a matching no-crypto.c. This way we keep the existing feature grouping. linux.c is not a good name anyway. I think after factoring the crypto code out, there isn't much left. sysfs and hostid code which we could then move to new files eventually.

bcapener:
along with the two functions libnvme_generate_hostnqn, libnvme_read_hostnqn from no-linux.c, I also moved these related functions from linux.c to crypto.c/no-crypto.c, libnvme_generate_hostid, libnvme_generate_hostnqn_from_hostid, libnvme_read_hostid, to keep things consistent.

igaw:
Thanks!

I think it's possible to make the crypto.c file also a bit simpler in the future. E.g. there is code for the !HAVE_OPENSSL case. In should be possible to tie this to this feature detection I suppose. Anyway, this is something we can look into this later.

PR #3380, opened 2026-05-21
ioctl: Add Windows ioctl support

brgoing:
Adds Windows ioctl support in ioctl-win.c, and pulls Linux-specific methods into ioctl-linux.c.


brgoing:
The change in #3371 fixes the warnings that are causing the Windows build to fail for this PR.  Once those changes have been applied, the warnings will clear up and the Windows build will pass.

igaw:
Nice! Thanks!

PR #3381, opened 2026-05-21
nvme-top: avoid NULL dereference on non-multipath host

shroffni:
When the kernel is built without CONFIG_NVME_MULTIPATH, the iopolicy sysfs attribute is not exposed under the NVMe subsystem hierarchy.

In this case, libnvme_subsystem_get_iopolicy() returns NULL. However, nvme-top dereferences the returned pointer without validating it first, leading to a NULL pointer dereference and a resulting SIGSEGV on non- multipath systems.

Fix this by validating the return value of nvme_subsystem_get_iopolicy() before dereferencing it. If the helper returns NULL, display "NA" and thus avoid NULL pointer dereference.


igaw:
Thanks!

PR #3382, opened 2026-05-21
libnvme: fix memleak in libnvme_subsystem_set_ns_path

shroffni:
While building the NVMe topology on multipath systems, libnvme constructs the namespace path list by scanning the  /sys/block/<nshead-dev>/multipath directory to discover all paths associated with a namespace head node.

The scandir() helper allocates an array of struct dirent entries using malloc(), and it is the caller's responsibility to free the allocated array once it is no longer needed.

libnvme_subsystem_set_ns_path() invokes scandir() to discover namespace paths and populate head->paths, but it never frees the allocated dirent array, resulting in a memory leak.

Fix this by annotating the struct dirents instance with __cleanup_dirents, ensuring that the allocated dirent array is automatically freed when leaving scope.


igaw:
Thanks!

PR #3383, opened 2026-05-21
build, CI: add memory testing infrastructure (valgrind, ASan+UBSan)

martin-belanger:
## Summary

This PR adds memory testing support to the build system and fixes one leak found in the process.

The existing CI workflow (`build.yml`) already passes `-x` to `scripts/build.sh` on every run. Before this PR, `-x` used `--wrapper valgrind` without `--error-exitcode`, so valgrind ran but its findings were silently ignored. This PR wires `-x` up properly: it now uses a named meson test setup with `--leak-check=full` and `--error-exitcode=1`, so valgrind failures are caught and fail the build.

### What's added

**Valgrind** (`meson test --setup valgrind`)

A named meson test setup wraps all unit tests under valgrind with `--leak-check=full`. A suppression file (`valgrind.supp`) is bundled to silence known false positives in three categories:
- Bash interpreter internals (shell-script-wrapped tests)
- CPython interpreter internals (Python test scripts)
- SWIG module teardown (libnvme Python bindings cleanup)

The system `/usr/lib/valgrind/python3.supp` is also loaded automatically when present.

**AddressSanitizer + UndefinedBehaviourSanitizer** (`--setup asanubsan`)

A named meson test setup for ASan+UBSan instrumented builds. Requires a dedicated build directory compiled with `-Db_sanitize=address,undefined`:

```bash
meson setup .build-asanubsan -Db_sanitize=address,undefined
meson compile -C .build-asanubsan
meson test -C .build-asanubsan --setup asanubsan
```

**`scripts/build.sh` alignment**

- `-x` now uses `--setup valgrind` instead of the bare `--wrapper valgrind`
- `-s` (new flag) enables ASan+UBSan: sets `-Db_sanitize=address,undefined` at configure time and runs `--setup asanubsan`

**`TESTING.md`**

New "Memory Testing" section documents both setups, the suppression file, and known limitations.

## Memory issue found and fixed

Running these tools against the existing test suite revealed one confirmed memory leak:

- `libnvme_create_host()` in `tree.c`: `nvme_hostid_from_hostnqn()` returns a `strdup`'d string that is immediately copied again into `h->hostid` via a second `strdup`, leaking the first allocation. Fixed in this PR by transferring ownership directly instead of strdup'ing twice.

## Test plan

- [x] `meson setup .build && meson compile -C .build && meson test -C .build --setup valgrind` — all 65 tests pass, 2 expected failures, 0 valgrind errors
- [x] `meson setup .build-asanubsan -Db_sanitize=address,undefined && meson compile -C .build-asanubsan && meson test -C .build-asanubsan --setup asanubsan` — all nvme-cli unit tests pass
- [x] `./scripts/build.sh -x` — valgrind run via named setup
- [x] `./scripts/build.sh -s` — ASan+UBSan run via named setup

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -682,6 +682,24 @@ if want_nvme
     subdir('Documentation')
 endif
 
+if host_system != 'windows'
+    valgrind_cmd = [
+        'valgrind',
+        '--leak-check=full',
+        '--error-exitcode=1',
+        '--suppressions=' + meson.current_source_dir() / 'valgrind.supp',
+    ]
+    python3_supp = '/usr/lib/valgrind/python3.supp'
+    if fs.is_file(python3_supp)
+        valgrind_cmd += ['--suppressions=' + python3_supp]
+    endif
+
+    add_test_setup('valgrind',
+        exe_wrapper: valgrind_cmd,
+        timeout_multiplier: 5,
+    )
+endif
+
 if doc_tgts.length() == 0
     doc_tgts += custom_target(
         'docs-dummy',
--- a/meson.build
+++ b/meson.build
@@ -698,6 +698,10 @@ if host_system != 'windows'
         exe_wrapper: valgrind_cmd,
         timeout_multiplier: 5,
     )
+
+    add_test_setup('asanubsan',
+        timeout_multiplier: 2,
+    )
 endif
 
 if doc_tgts.length() == 0
--- a/TESTING.md
+++ b/TESTING.md
@@ -131,3 +131,38 @@ Example: Cross-build for `s390x`:
 ```
 
 The exact supported targets depend on the toolchains installed in the container.
+
+## Memory Testing
+
+### Valgrind
+
+A named Valgrind test setup is registered in `meson.build`. It runs all unit
+tests under `valgrind --leak-check=full` and treats any leak as a test failure.
+
+Prerequisites: Valgrind must be installed (`apt install valgrind` or equivalent).
+
+Build and run:
+
+```bash
+meson setup .build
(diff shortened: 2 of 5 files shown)


meson.build
```diff
@@ -673,6 +673,28 @@ if want_nvme
...
+    if fs.is_file(python3_supp)
+        valgrind_cmd += ['--suppressions=' + python3_supp]
+    endif
+
+    add_test_setup('valgrind',
+        exe_wrapper: valgrind_cmd,
+        timeout_multiplier: 5,
+    )
+
+    add_test_setup('asanubsan',
```
igaw, line 702: should this be in the second patch?

meson.build
```diff
@@ -576,6 +576,7 @@ if want_nvme
     if host_system != 'windows'
         subdir('unit')
     endif
+
```
igaw: whitespace change and nothing touches near this place. I'd leave it away.

igaw:
I like it, the only nitpick is that the second patch seems a bit empty and the missing part is in the first patch. Either merge both patches into one or move the missing line to the second patch.

martin-belanger:
> I like it, the only nitpick is that the second patch seems a bit empty and the missing part is in the first patch. Either merge both patches into one or move the missing line to the second patch.

Will do. Thanks

meson.build
martin-belanger: You're absolutely right. Done.

meson.build
martin-belanger: Done

igaw:
Thanks!

Issue #2904, opened 2025-09-03
Add support for `nvme top` (monitor feature)


igaw:
There are use cases where the user wants to monitor the system changes. nvme-cli currently shows a snapshot. A simple way to achieve this in shell is to do: 

```shell
watch -n1 nvme show-topology
```

Though sometimes shell scripting is adding an additional layer of problems, e.g. when creating logs for other tools to consume https://github.com/linux-nvme/nvme-cli/issues/2189

@hreinecke pointed out in https://lore.kernel.org/linux-nvme/ee0d329a-334b-429c-a916-f03c64bbb323@suse.de/

```
Well, that might be tricky. The current 'tree' structure is build
once when the program starts up. Any changes to that structure after
that are not tracked, and there (currently) are no hooks for updating
it. So having a 'monitor' function will get tricky.

To do that we would either need a udev event receiver (using uevents
to update the tree structure) or looking into fanotify()/inotify().
Then the tree structure would always be up-to-date and things like
'monitor' will be possible.
Will be hell for the python bindings, mind 🙁 
```

shroffni:
For the reference, here I have listed down community feedback received for nvme-top 
presentation during LPC-2025:

1. Grafana Export
   - Having a Grafana export was viewed as a very useful and valuable addition to the tool.

2. Polling vs. Event-Based Updates
   - Relying on inotify/fsnotify or uevents for regular parameter updates may introduce
     complexity and risk missing kernel updates. A poll-based approach is recommended for
     commonly changing parameters such as queue depth, statistics, controller state, and
     I/O policy.

   - Exception: topology changes (e.g., adding or removing namespaces or controllers).
     For topology updates, listening to kernel events is necessary so that nvme-top can
     rebuild the topology using nvme-cli and update the dashboard in real time.

3. UI Framework
   - ncurses is not required; a simple termios-based approach is sufficient to build an 
     effective and responsive dashboard.

4. Handling Large Topologies
   - In environments with many namespaces per subsystem, fitting all subsystems, namespaces, 
     controllers, and nsheads on a single page may not be practical.
     One suggestion is to add an interactive menu to toggle between different views or tables.
   - Refer  the htop implementation if simplifying or scaling the dashboard becomes challenging.

5. Controller Summary Enhancements
   In addition to displaying the controller state, the controller summary should include:
   - Number of controller resets,
   - Number of reconnect attempts (especially for NVMe-oF),
   - Potentially other reliability-related counters.

[...]

PR #3384, opened 2026-05-21
nvme-top: preserve selected subsystem across topology refresh

shroffni:
The nvme top UI consists of two dashboards. The first dashboard displays the subsystem list and allows the user to scroll through available subsystems and select one for further inspection. Pressing ENTER on a selected subsystem opens the second dashboard, which displays detailed topology and subsystem information.

When the subsystem list dashboard is initially displayed, the first subsystem entry (row #0) is highlighted by default.

However, while the user is navigating through the subsystem list, the highlighted selection may unexpectedly jump back to the first subsystem entry. This happens when nvme top receives a uevent (for block or NVMe), triggering a topology rescan and dashboard refresh.

During the redraw operation, nvme-top currently resets the selected entry to the first subsystem in the list, even if the previously selected subsystem still exists after the topology refresh. This results in an inconsistent and disruptive user experience.

Fix this by preserving the currently selected subsystem across topology rescans. If the previously highlighted subsystem is still present after rebuilding the topology, continue highlighting the same subsystem instead of resetting the selection to the first entry in the list.

### DIFF
--- a/nvme-print-stdout-top.c
+++ b/nvme-print-stdout-top.c
@@ -1144,6 +1144,22 @@ static libnvme_subsystem_t *stdout_top_build_subsys_arr(
 	return subsys_arr;
 }
 
+static int stdout_top_find_subsys_by_name(libnvme_subsystem_t *subsys_arr,
+		int num_subsys, const char *subsys_name)
+{
+	int idx;
+	libnvme_subsystem_t s;
+
+	for (idx = 0; idx < num_subsys; idx++) {
+		s = subsys_arr[idx];
+
+		if (!strcmp(libnvme_subsystem_get_name(s), subsys_name))
+			return idx;
+	}
+
+	return -1;
+}
+
 /*
  * Draws subsys topology screen of susbystem @s
  * Returns: 0 if ESC key is pressed or needs to draw subsystem selection screen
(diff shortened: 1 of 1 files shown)


shroffni:
@igaw @ikegami-t 
Does this help resolve the issue reported here: https://github.com/linux-nvme/nvme-cli/issues/2904#issuecomment-4443077063 (look at second issue) ?

nvme-print-stdout-top.c
```diff
@@ -1460,21 +1476,40 @@ void stdout_top(int refresh_interval)
 				break;
 			fallthrough;
 		case EVENT_TYPE_NVME_UEVENT:
-			libnvme_free_global_ctx(ctx);
+		{
```
ikegami-t: Is this brace coding style correct? Since some other code implemenmted as below.
```c
  switch (aaa) {
    case bbb: {
       ...
    }
    case ccc:
       ...
  }
```
(By the way I do not like to use the brace for the swicth case as move the variables to the funcion head part or split the code in the case statement as a function.)

nvme-print-stdout-top.c
```diff
@@ -1460,21 +1476,40 @@ void stdout_top(int refresh_interval)
...
+			libnvme_subsystem_t s;
+			__cleanup_free char *subsys_name = NULL;
```
ikegami-t, line 1479: FYI: By changing the variables ordering as an inverse Christmas tree style as below then we can avoid the checkpatch.pl warning below. (But the warning seems the checkpatch.pl issue.)
1. Variable ordering
```c
			__cleanup_free char *subsys_name = NULL;
			libnvme_subsystem_t s;
```
2. checkpatch.pl warning
```sh
WARNING: Missing a blank line after declarations
#71: FILE: nvme-print-stdout-top.c:1481:
+                       libnvme_subsystem_t s;
+                       __cleanup_free char *subsys_name = NULL;

total: 0 errors, 1 warnings, 65 lines checked
```

ikegami-t (commented):
Looks good as no issue caused with the changes. But just commented as minor comments. Thank you.

nvme-print-stdout-top.c
shroffni: Yeah I will put the opening brace on the same line as the case label.

nvme-print-stdout-top.c
shroffni: Yes that seems to be issue with checkpatch.pl. But I can move ordering of variables to avoid complain from the checkpatch.pl.

ikegami-t: Thank you for this also.

nvme-print-stdout-top.c
ikegami-t: Thank you!

shroffni:
Force pushed PR with following changes:
- Reorder the declaration of variables to avoid checkpatch complain
- Moved open brace inline to the case statement

ikegami-t (approved):
Looks good. Thank you.

igaw:
Thanks!

PR #3386, opened 2026-05-21
Small changes for cross-platform compatibility

brgoing:
- Updates check-public-symbols to work across platforms.
- Changes use of non-standard bzero to memset.
- Changes unsigned long casts that need to be 64-bits to use fixed width __u64 type.


igaw:
Thanks!

PR #3387, opened 2026-05-21
ocp-nvme: Replace fstat checking of S_ISBLK with built-in check

brgoing:
The ocp code was using fstat to check S_ISBLK to determine if the device handle was a namespace device before calling libnvme_get_nsid.  Update to use the built-in libnvme_transport_handle_is_ns check for cross-platform compatibility.


igaw:
Thanks!

PR #3388, opened 2026-05-22
support fixed width table column

shroffni:
Currently table API only support automatic resize of the table columns. This series adds support for fixed width columns. This would then allow us to use auto size columns for one shot nvme-cli commands and use fixed width columns for real-time dashboard display such as nvme-top.

This pacthset adds three commits:
The first commit extends the table API to support fixed width column. With this change now auto size table columns requires the column width to be inited using AUTO_WIDTH (a special macro defined as INT_MAX) and fixed width columns shall be inited using constant width integer value. 
The second commit then update the show topology tabular output  (which is one shot nvme-cli command) using AUTO_WIDTH for the column.
The third patch updates the nvme-top tables using fixed width columns.

----------------------------------------------------------------------------------------------------------- 

Force pushed PR with the following change:
Split the first patch into two patches and created the new commit to fix a potential memlelak in table API

The following patch is splitted into two patches: 
util/table: extend table APIs to support fixed width column

first patch: util/table: fix memory leak in table_add_columns_filter()
second patch: util/table: extend table APIs to support fixed width column

The remaining patches remain as is.

### DIFF
--- a/util/table.c
+++ b/util/table.c
@@ -256,6 +256,9 @@ void table_add_row(struct table *t, int row_id)
 
 	/* Adjust the column width based on the row value. */
 	for (col = 0; col < t->num_columns; col++) {
+		if (!t->columns[col].auto_adjust)
+			continue;
+
 		max_width = t->columns[col].width;
 		width = table_get_value_width(&row->val[col]);
 		if (width > max_width)
@@ -299,7 +302,17 @@ static int table_add_column(struct table *t, struct table_column *c)
 	if (!t->columns[col].name)
 		return -ENOMEM;
 	t->columns[col].align = c->align;
-	t->columns[col].width = strlen(c->name);
+
+	if (c->width == AUTO_WIDTH) {
+		t->columns[col].width = strlen(c->name);
+		t->columns[col].auto_adjust = true;
+	} else {
+		if (c->width < strlen(c->name))
+			return -EINVAL;
+
+		t->columns[col].width = c->width;
+		t->columns[col].auto_adjust = false;
+	}
 
 	return 0;
 }
@@ -335,7 +348,7 @@ free_col:
 
 int table_add_columns(struct table *t, struct table_column *c, int num_columns)
 {
-	int col;
+	int ret = 0, col;
 
 	t->columns = calloc(num_columns, sizeof(struct table_column));
 	if (!t->columns)
@@ -343,24 +356,36 @@ int table_add_columns(struct table *t, struct table_column *c, int num_columns)
 
 	for (col = 0; col < num_columns; col++) {
 		t->columns[col].name = strdup(c[col].name);
-		if (!t->columns[col].name)
+		if (!t->columns[col].name) {
+			ret = -ENOMEM;
 			goto free_col;
+		}
 
 		t->columns[col].align = c[col].align;
-		if (c[col].width > strlen(t->columns[col].name))
-			t->columns[col].width = c[col].width;
-		else
+
+		if (c[col].width == AUTO_WIDTH) {
 			t->columns[col].width = strlen(t->columns[col].name);
+			t->columns[col].auto_adjust = true;
+		} else {
+			if (c[col].width < strlen(t->columns[col].name)) {
+				ret = -EINVAL;
+				goto free_col;
+			}
+			t->columns[col].width = c[col].width;
+			t->columns[col].auto_adjust = false;
+		}
 	}
 	t->num_columns = num_columns;
 
-	return 0;
+	return ret;
 free_col:
-	while (--col >= 0)
+	while (col >= 0) {
 		free(t->columns[col].name);
+		col--;
+	}
 	free(t->columns);
 	t->columns = NULL;
-	return -ENOMEM;
+	return ret;
 }
 
 void table_free(struct table *t)
--- a/util/table.c
+++ b/util/table.c
@@ -293,13 +293,13 @@ static int table_add_column(struct table *t, struct table_column *c)
 	if (!new_columns)
 		return -ENOMEM;
 
+	t->num_columns++;
 	t->columns = new_columns;
 	t->columns[col].name = strdup(c->name);
 	if (!t->columns[col].name)
 		return -ENOMEM;
 	t->columns[col].align = c->align;
 	t->columns[col].width = strlen(c->name);
-	t->num_columns++;
 
 	return 0;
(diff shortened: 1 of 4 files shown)


shroffni:
@igaw @ikegami-t 
Does this help solve the issue with nvme-top dashboard which you both reported earlier here: https://github.com/linux-nvme/nvme-cli/issues/2904#issuecomment-4443077063 (look at first issue)

util/table.c
```diff
@@ -293,13 +296,23 @@ static int table_add_column(struct table *t, struct table_column *c)
...
 	t->columns[col].align = c->align;
-	t->columns[col].width = strlen(c->name);
-	t->num_columns++;
+
+	if (c->width == AUTO_WIDTH) {
+		t->columns[col].width = strlen(c->name);
+		t->columns[col].auto_adjust = true;
+	} else {
+		if (c->width < strlen(c->name))
+			return -EINVAL;
```
ikegami-t: Is this error returning also expected? BUt the t->num_columns value already incremented.

util/table.c
```diff
@@ -293,13 +296,23 @@ static int table_add_column(struct table *t, struct table_column *c)
 	if (!new_columns)
 		return -ENOMEM;
 
+	t->num_columns++;
 	t->columns = new_columns;
```
ikegami-t, line 300: The change is not for to support the fixed width column so better to be done as a separated patch.

util/table.c
```diff
@@ -293,13 +296,23 @@ static int table_add_column(struct table *t, struct table_column *c)
...
 		return -ENOMEM;
 	t->columns[col].align = c->align;
-	t->columns[col].width = strlen(c->name);
-	t->num_columns++;
+
+	if (c->width == AUTO_WIDTH) {
+		t->columns[col].width = strlen(c->name);
+		t->columns[col].auto_adjust = true;
+	} else {
+		if (c->width < strlen(c->name))
```
ikegami-t, line 310: Is this error checking better than to limit the c->name length to the c->width value?

ikegami-t (commented):
Seems good but some minor comments so please do confirm them. By the way how about the change to the stats counters align to right from left? Since seems better to monitor the counter values with the right align.

util/table.c
shroffni: Okay I will split this change into another patch.

util/table.c
shroffni: Yes, when AUTO_WIDTH is not specified, the caller is expected to explicitly provide the column width. The specified width must be at least as large as the length of the column name.

The reason for this requirement is that the table renderer also draws a separator line made up of dashes (-) underneath each column header. If the specified width is smaller than strlen(column_name), the separator line and the column contents will become misaligned, resulting in improperly formatted table output.

ikegami-t: Noted then just locally tested to limit the length instead as below then seems as it works correctly.
```diff
tokunori@tokunori-X570-Taichi:~/nvme-cli$ git diff
diff --git a/nvme-print-stdout-top.c b/nvme-print-stdout-top.c
index 292ced4c6..b8e362cad 100644
--- a/nvme-print-stdout-top.c
+++ b/nvme-print-stdout-top.c
@@ -1267,7 +1267,7 @@ static int stdout_top_draw_subsys_screen(struct dashboard_ctx *db_ctx,
                {"Paths",      LEFT, AUTO_WIDTH},
                {"Ctrls",      LEFT, AUTO_WIDTH},
                {"IOPolicy",   LEFT, AUTO_WIDTH},
-               {"r_IOPS",     LEFT, 9},
+               {"r_IOPS789a", LEFT, 9},
                {"w_IOPS",     LEFT, 9},
                {"r_clat",     LEFT, 8},
                {"w_clat",     LEFT, 8},
diff --git a/util/table.c b/util/table.c
index 1c37b5128..7dc004ed9 100644
--- a/util/table.c
+++ b/util/table.c
@@ -308,7 +308,7 @@ static int table_add_column(struct table *t, struct table_column *c)
                t->columns[col].auto_adjust = true;
        } else {
                if (c->width < strlen(c->name))
-                       return -EINVAL;
+                       t->columns[col].name[c->width] = 0;
 
                t->columns[col].width = c->width;
                t->columns[col].auto_adjust = false;
@@ -367,10 +367,8 @@ int table_add_columns(struct table *t, struct table_column *c, int num_columns)
                        t->columns[col].width = strlen(t->columns[col].name);
                        t->columns[col].auto_adjust = true;
                } else {
-                       if (c[col].width < strlen(t->columns[col].name)) {
-                               ret = -EINVAL;
-                               goto free_col;
-                       }
+                       if (c[col].width < strlen(t->columns[col].name))
+                               t->columns[col].name[c[col].width] = 0;
                        t->columns[col].width = c[col].width;
                        t->columns[col].auto_adjust = false;
                }
```
image

util/table.c
ikegami-t: Thank you.

util/table.c
shroffni: But that doesn't look correct. I think we shouldn't truncate the column name and instead we should ideally return error (-EINVAL : which this patch implements) when user specified column length is less that the strlen of column name.

shroffni:
> Seems good but some minor comments so please do confirm them. By the way how about the change to the stats counters align to right from left? Since seems better to monitor the counter values with the right align.

Yes agreed, but then the way we draw the table it looks pretty ugly. For instance, I tried this out for subsystem list screen by aligning all numbers on right and the this is how it look:
image

So as you could see above, we have all column name aligned to the left but the numbers aligned to the right then it looks pretty bad.

Then, I aligned column names as well numbers printed under each column to the right (but kept columns which print string as-is) and this is how the output appears:
image

This also looks pretty convoluted to me. That's because the columns which display the string is aligned to left and columns which display the numbers are aligned to right. 

Than I aligned everything (columns and row values) on the right and this is how it looks:
image

So based on above experiment, I'd suggest either align everything on right or left. The current nvme top implementation aligns everything to the left. To me aligning everything on the left looks good (but yes I am open to modify the alignment if others also feel the same).

ikegami-t (approved):
Understood about the aligning everything on left then looks good. Thank you.

util/table.c
ikegami-t: Yes agreed as mentioend. Thank you.

igaw:
Thanks!

PR #3389, opened 2026-05-22
util: Add compatibility for libnvme_random_uuid

brgoing:
Adds random_bytes helper that uses Windows' bcrypt support if available and otherwise defaults to using /dev/urandom.
Updates libnvme_random_uuid to use the new helper.


igaw:
Thanks!

PR #3390, opened 2026-05-22
fdp: Remove unused linux/fs.h include

brgoing:
Remove include of linux/fs.h from fdp.c.
Nothing in fdp.c relies on linux/fs.h, and removing it allows compilation on Windows.


igaw:
Thanks!

PR #3395, opened 2026-05-28
libnvme: flat accessors for nested structs + grouped setters

martin-belanger:
## Background

This is the follow-up to the text message I sent you yesterday about the `libnvmf_context` accessor design. PR #3369 has since merged, so here is the full implementation of the plan we discussed.

---

## What this PR does

`libnvmf_context` is an opaque type containing nested structs (`libnvme_ctrl_params` → `libnvme_fabrics_config`). Before this series, the only way to set fabrics config fields was through bulk APIs like `libnvmf_context_set_fabrics_config()` that accepted a `struct libnvme_fabrics_config *`, forcing callers to construct and manage an internal struct type directly and leaking the private layout through the public API.

The fix: extend the accessor generator to flatten nested struct fields into the parent's accessor set, generate the full suite of flat `libnvmf_context_get/set_*()` accessors, then delete the bulk APIs. On top of that, two hand-written grouped setters cover parameter sets with internal coupling that individual setters cannot express.

---

### 1. generator: add nested struct accessor support

Adds two new annotations:

- `// !nested-accessors` on a struct definition — marks it as available for expansion into a parent's accessor set.
- `// !access:nested` on a struct member — tells the generator to flatten that member's fields rather than treating the member itself as an opaque field. Accepts an optional per-usage metadata override (e.g. `!access:nested:write=none`).

The generator uses a two-pass approach: Pass 1 scans all input files and builds a `nested_map` of every `!nested-accessors` struct; Pass 2 walks each `!generate-accessors` struct and recursively expands any `!access:nested` members using the map. Arbitrary nesting depth is supported; cycle detection raises an error. Unreferenced `!nested-accessors` structs produce a warning.

A `--nested-source` flag allows a header that defines `!nested-accessors` structs to be provided as lookup-only input when it also carries `!generate-accessors` structs belonging to a separate generator run.

Nested members are excluded from SWIG fragments — dotted field paths cannot be used as direct struct members in SWIG-generated wrappers.

### 2. libnvme: expose libnvmf_context fields through flat accessors

Annotates `libnvmf_context`, `libnvme_ctrl_params`, and `libnvme_fabrics_config` in `private-fabrics.h` / `private.h` and regenerates `accessors-fabrics.{h,c,ld}`. The result is a complete set of flat `libnvmf_context_get/set_*()` accessors covering all three levels of nesting, with callers needing no knowledge of the internal struct hierarchy.

### 3. fabrics: replace bulk config APIs with flat accessors

With flat accessors in place, `libnvmf_context_set_fabrics_config()` and `libnvmf_context_get_fabrics_config()` are removed from the public API and their call sites in `fabrics.c` migrated to individual accessors. The standalone `libnvme_fabrics_config_set/get_*()` accessors are also removed since all access now goes through the parent `libnvmf_context` accessors. Hand-written trivial setters (`set_persistent`, `set_discovery_defaults`) are replaced by their generated equivalents.

Net: ~700 lines deleted, ~40 added.

### 4. libnvme: replace libnvmf_ctrl_get_fabrics_config() with flat accessors

`libnvmf_ctrl_get_fabrics_config()` returned a pointer to an internal struct. All its callers were inside libnvme itself (`fabrics.c`, `json.c`), so it provided no value as a public API. Adding `!access:nested:write=none` to `libnvme_ctrl::cfg` generates read-only flat `libnvme_ctrl_get_*()` accessors for every fabrics config field. Internal callers are migrated to direct `c->cfg` access and the public function is removed.

### 5. libnvmf: add grouped setters for coupled fabrics context parameters

Two sets of `libnvme_fabrics_config` fields have internal coupling that individual flat setters cannot express:

**`libnvmf_context_set_io_queues(fctx, nr_io_queues, nr_write_queues, nr_poll_queues, queue_size, disable_sqflow)`** — all five parameters feed into a single Connect command at queue creation time and describe one coherent decision: the I/O queue topology of the connection. Total I/O queue count is `nr_io_queues + nr_write_queues + nr_poll_queues`.

**`libnvmf_context_set_reconnect_policy(fctx, ctrl_loss_tmo, reconnect_delay, fast_io_fail_tmo)`** — the kernel derives `max_reconnects` from `ceil(ctrl_loss_tmo / reconnect_delay)`, and `fast_io_fail_tmo` must not exceed `ctrl_loss_tmo`. A grouped setter makes this coupling visible at the call site.

Also fixes `queue_size` being silently dropped in nvme-cli's `set_fabrics_options()` despite `--queue-size` being a valid CLI option.

Individual flat accessors remain available for callers that only need to change a single field.


igaw:
I do like the result.

The only nitpick I have is that the commit messages are really hard to understand. For example, the commit message for the first patch is borderline understandable to me. I had to read it several times. The documentation that was added is actually much simpler to read and understand.

martin-belanger:
Let me rework the commit messages

martin-belanger:
@igaw - I simplified/clarified all the commit messages.

igaw:
Thanks a lot!

PR #3400, opened 2026-05-29
libnvme: NVMe-oF controller ownership registry

martin-belanger:
This PR addresses: https://github.com/linux-nvme/nvme-cli/issues/2913

Multiple independent orchestrators (nvme-stas, NBFT, nvme-discoverd, manual nvme-cli) can establish NVMe-oF controller connections on the same host with no coordination. `nvme disconnect-all` is indiscriminate and dangerous in such environments.

This series introduces a cooperative userspace ownership registry so orchestrators can declare and respect ownership boundaries. Full design rationale and decisions are documented in: [nvme-registry-plan.md](https://github.com/user-attachments/files/28163658/nvme-registry-plan.md).

## What's new

- **libnvme**: registry API (`libnvmf_registry_retrieve/update/delete/for_each`), automatic registration on connect when `libnvmf_context_set_owner()` is set, Python bindings, unit tests
- **udev rule**: `70-nvmf-registry.rules` removes stale entries on controller removal
- **nvme disconnect-all**: ownership-aware by default; new `--owner` and `--force` options
- **nvme connect-all --nbft**: automatically registers controllers as owned by `"nbft"`
- **nvme registry-list/retrieve/update/delete**: new subcommands for manual registry management
- **nvme list -v**: new Orchestrator column

## Commits

1. `libnvme: add NVMe controller ownership registry` (78c9055)
2. `libnvmf: register controller ownership on successful connect` (6892470)
3. `nvme: add registry CLI commands and ownership-aware disconnect-all` (1c4dbdf)
4. `libnvme: expose registry API and owner in Python bindings` (43818fb)
5. `nvme: show orchestrator ownership in verbose list output` (dc0823e)
6. `libnvme: install udev rule to clean up registry on controller removal` (fa9e34b)
7. `nvme: add man pages for registry commands and update disconnect-all` (e5f500c)

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -376,6 +376,15 @@ __libnvme_public int libnvmf_context_set_reconnect_policy(
 	return 0;
 }
 
+__libnvme_public void libnvmf_context_set_owner(struct libnvme_global_ctx *ctx,
+					    const char *owner)
+{
+	if (!ctx)
+		return;
+	free(ctx->owner);
+	ctx->owner = owner ? strdup(owner) : NULL;
+}
+
 /*
  * Derived from Linux's supported options (the opt_tokens table)
  * when the mechanism to report supported options was added (f18ee3d988157).
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -984,6 +984,7 @@ static int __nvmf_add_ctrl(struct libnvme_global_ctx *ctx, const char *argstr)
 {
 	__cleanup_fd int fd = -1;
 	int ret, len = strlen(argstr);
+	int instance;
 	char buf[0x1000], *options, *p;
 
 	fd = open(nvmf_dev, O_RDWR);
@@ -1035,8 +1036,11 @@ static int __nvmf_add_ctrl(struct libnvme_global_ctx *ctx, const char *argstr)
 	while ((p = strsep(&options, ",\n")) != NULL) {
 		if (!*p)
 			continue;
-		if (sscanf(p, "instance=%d", &ret) == 1)
-			return ret;
+		if (sscanf(p, "instance=%d", &instance) == 1) {
+			if (ctx->owner)
+				libnvmf_registry_create(instance, ctx->owner);
+			return instance;
+		}
 	}
 
 	libnvme_msg(ctx, LIBNVME_LOG_ERR, "Failed to parse ctrl info for \"%s\"\n", argstr);
--- a/libnvme/src/nvme/registry.c
+++ b/libnvme/src/nvme/registry.c
@@ -1,0 +1,258 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ * Copyright (c) 2026 Dell Technologies Inc. or its subsidiaries.
+ *
+ * Authors: Martin Belanger <martin.belanger@dell.com>
+ */
+
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <libgen.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <json.h>
+
+#include "cleanup.h"
+#include "compiler-attributes.h"
+#include "private.h"
+#include "registry.h"
+
+static DEFINE_CLEANUP_FUNC(cleanup_json_object, struct json_object *,
+			    json_object_put)
+#define __cleanup_json	__cleanup(cleanup_json_object)
+
+/*
+ * __libnvme_weak allows test binaries to override this function with a strong
+ * definition that returns a temporary directory, redirecting all registry I/O
+ * away from /run/nvme/registry without any runtime conditionals in production
+ * code.
+ */
+__libnvme_weak const char *registry_dir(void)
+{
+	return "/run/nvme/registry";
+}
+
+static int ensure_registry_dir(void)
+{
+	__cleanup_free char *dir_copy = NULL;
+
+	if (mkdir(registry_dir(), 0755) == 0 || errno == EEXIST)
+		return 0;
+	if (errno != ENOENT)
+		return -errno;
+
+	/* Parent directory is missing; create it then retry. */
+	dir_copy = strdup(registry_dir());
+	if (!dir_copy)
+		return -ENOMEM;
+	if (mkdir(dirname(dir_copy), 0755) < 0 && errno != EEXIST)
+		return -errno;
+	if (mkdir(registry_dir(), 0755) < 0 && errno != EEXIST)
+		return -errno;
+	return 0;
+}
+
+static int write_json_atomic(struct json_object *root, const char *path)
+{
+	__cleanup_free char *tmp_path = NULL;
+	int fd, dir_fd, ret;
+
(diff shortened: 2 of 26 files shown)


libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,257 @@
...
+	if (ret < 0) {
+		close(fd);
+		unlink(tmp_path);
+		return -EIO;
+	}
+
+	if (write(fd, "\n", 1) < 0) {
+		close(fd);
+		unlink(tmp_path);
+		return -errno;
```
igaw: to be extra correct, you should cache errno before calling close or unlink.

igaw:
just a quick thoughs as discussed on the matrix channel:

- 'context' and 'owner' are the same concept. ideally we just use one.
- add a test which covers parallel access to the registry (ideally it should also cover the python binding)
- move the registry commands to a plugin, no need to be top level.

libnvme/src/nvme/registry.c
martin-belanger: Addressed

martin-belanger:
@igaw - Thanks for the review!

On the context/owner naming: agreed, they are the same concept. I already have a follow-up planned that addresses this, so I'd like to handle it there rather than here to avoid a mid-flight rename.

On the parallel access test and plugin conversion: both are good ideas. I'd like to keep this PR focused on the core feature introduction and handle those in a follow-up as well, if that's acceptable to you. The concurrent registry access should be safe, but I agree that a test that proves it is worth having.

I see that there are conflicts with the PR that I need to resolve. Once I'm done fixing the conflicts, would you be OK merging this as-is with those items tracked as follow-up work?

igaw:
yes, I don't have an issue to go ahead as soon the core concept and API are sound. Adding tests cases later shouldn't really be a problem.

libnvme/src/nvme/fabrics.c
```diff
@@ -376,6 +376,15 @@ __libnvme_public int libnvmf_context_set_reconnect_policy(
 	return 0;
 }
 
+__libnvme_public void libnvmf_context_set_owner(struct libnvme_global_ctx *ctx,
+					    const char *owner)
+{
+	if (!ctx)
+		return;
+	free(ctx->owner);
+	ctx->owner = owner ? strdup(owner) : NULL;
```
MichaelRabek, line 385: `strdup(3)` can fail and the `libnvmf_context_set_owner()` function is going to be public, so perhaps we should not return `void` and let the user know it failed?

MichaelRabek (commented):
Other than that, I like it!

libnvme/src/nvme/fabrics.c
martin-belanger: @MichaelRabek - Thanks for the review.

You are correct that `strdup()` can fail. However, `libnvmf_context_set_owner()` follows the same void setter pattern used by hundreds of existing libnvme APIs (`libnvme_set_application()`, `libnvme_ctrl_set_dhchap_host_key()`, `libnvme_ctrl_set_tls_key_identity()`, etc.). Changing just this one function would be inconsistent.

That said, since we are in the middle of a major version change where API breaks are permitted, fixing all these setters to return an error in one sweep would be a reasonable follow-up PR. Would you be willing to take that on, or should I add it to my backlog?

MichaelRabek: > Changing just this one function would be inconsistent.

Yes, that is true.

> That said, since we are in the middle of a major version change where API breaks are permitted, fixing all these setters to return an error in one sweep would be a reasonable follow-up PR.

It does sound like a good "excuse" for me to touch the codebase :)

However, when looking at how these similar functions are being used currently, many of them are being called from functions that return `void` as well. So, the return value would just end up being ignored, which is not particularly nice either.

In my personal opinion, mentioning the fact that the member variable that is being manipulated may get set to `NULL` due to an allocation failure in the docstring should be enough. But I think it _should_ be mentioned in there. I like seeing the full side effect behavior of C library functions as transparently documented as possible. It saved me a lot of hassle when stuff like this **was** mentioned and caused frustration when it wasn't.

I can write a follow-up PR that puts a mention of the possible failure in the docstrings and instructs the user to check for `errno == ENOMEM` or the target value for `NULL` if they care about it.

How does that sound to you?

igaw: Don't worry strdup failing for small allocation, pretty sure if nvme-cli gets executed when running out of memory things will fail spectacularly and it wont be strdup. If you really care, wecould replace strdup with our own version and if we run out of memory it calls abort. Don't even try to error handle this situation. OOM for the win :)

Let's not go down the errno path again. This error handling style is just wrong.

martin-belanger: Leaving it as-is for consistency with other APIs.

martin-belanger:
Hi @igaw — all review comments are now addressed and the strdup thread is resolved. Anything else remaining before merging?

As discussed yesterday, all remaining items will be addressed in a follow up PR.

Thanks.

igaw:
One thing I think we should also add from the beginning is a schema for the new json file and also a verifier for it.

martin-belanger:
> One thing I think we should also add from the beginning is a schema for the new json file and also a verifier for it.

@igaw — before you ask for a JSON schema and verifier, I want to be upfront: I'm planning to replace the JSON backend with a sysfs-inspired layout in the follow-up PR. The schema would end up as throwaway code.

The directory layout looks like this:
```
/run/nvme/registry/
    nvme3/
        owner
```
One directory per controller, one plain text file per attribute — same convention the kernel uses for sysfs. The advantages over JSON:

- No `json-c` dependency — the registry works in all build configurations, including musl/no-json builds that currently get `-ENOTSUP` stubs
- No parser — reading an attribute is a plain `read()`; writing is `write()` + `rename()`
- Trivially inspectable: `cat /run/nvme/registry/nvme3/owner`
- No read-modify-write — attributes are independent files, so updating one doesn't require rewriting the whole record

The public API is unchanged by this; it's a purely internal storage format change. I'd rather not invest in a JSON schema that will be deleted in the next PR. Would you be OK merging the current PR as-is and letting me land the format switch in the follow-up?

igaw:
Sounds reasonable to me. The path lookup and then read and write is not trivial, e.g  TOCTOU. Though it's doable and if this is done properly in the library it should be fine.

martin-belanger:
@igaw - Agreed, the implementation needs to be careful. 

The main mitigations: 
* per-process unique tmp filenames (e.g. `owner.tmp.<pid>`) to prevent concurrent writers from clobbering each other; 
* `mkdir()` with `EEXIST` treated as success for the device directory.

The read path has no TOCTOU since it holds an open fd throughout.

martin-belanger:
Hey @igaw, just checking if there's anything I can do to get this PR merged. I can't start the follow-up PR until this gets merged.

Thanks

igaw:
Hmm, I am waiting for the update of this PR addressing the points we discussed above. I want to cut down a bit on the code churn.

martin-belanger:
Ah, yes — a bit of a miscommunication on my end. I thought the more pressing issues had already been addressed (I amended a few commits and force-pushed), and I understood you to be OK with a follow-up PR for the remaining ones. Clearly I got that wrong. Sorry about that.

Two options going forward — let me know which you prefer:
1. Add commits on top of the existing PR rather than amending. It's getting complex to keep amending without breaking something.
2. Close this PR and resubmit a clean one with all issues addressed. Given the number of fixes needed, that might actually be the cleaner path.

Happy either way.

martin-belanger:
Closing. I will resubmit a fresh PR with all concerns addressed.

PR #3401, opened 2026-05-29
build: include more source files in windows build

bcapener:
add files to source list that will compile on windows with minimal changes.


igaw:
Thanks!

PR #3403, opened 2026-05-30
nvme: check MUD support before firmware commit

### LINKED ISSUE #3396: fw-commit prints "identify-ctrl: No such device or address" after successful Firmware Commit
Hi, we observed an extra user-visible error after a successful `nvme fw-commit` when using immediate firmware activation.

The output was:
Image

Environment:
- device: SOLIDIGM SB5PH27X076T Gen5 NVMe
- firmware update path: G70YG472 -> G70YG473
- kernel: Linux 6.12
- OS: Debian 13.2
- nvme-cli: 2.13

So far this reproduced 100% in our production environment for this device and firmware update path. I suspect this is because the firmware activation takes long enough for the post-command Identify Controller command to hit the controller reset window.

From the current code path, `fw_commit_print_mud()` calls `fw_commit_support_mud()` after `libnvme_exec_admin_passthru()` succeeds. That sends an Identify Controller command only to decide whether the Firmware Commit completion MUD value should be printed.

With immediate activation, the controller may already be entering the firmware activation/reset window, so this post-success Identify can fail even though Firmware Commit itself has already completed successfully.

I sent a small patch to the linux-nvme mailing list to cache `FRMW.SMUD` before submitting Firmware Commit, and then use that cached value when printing the completion MUD value:
https://lore.kernel.org/linux-nvme/20260519134542.1841435-1-guzebing1612@gmail.com/.

It also makes the MUD handling use the pre-activation firmware capability. Since the MUD value belongs to the Firmware Commit command completion, whether it should be decoded should be determined before activation, not from a later Identify result that may reflect the newly activated firmware.

Could you please advise whether this is an acceptable fix for nvme-cli? If yes, should I resend the patch, open a PR, or adjust the approach?

I am raising this here because the mailing-list patch may have been missed, possibly because I used the wrong maintainer email address. I hope this does not split the discussion; my goal is to clarify the preferred next step for this fix.

Thanks
Guzebing

guzebing1612-dev:
This PR follows up on #3396 and the mailing-list patch:

https://lore.kernel.org/linux-nvme/20260519134542.1841435-1-guzebing1612@gmail.com/

The current code checks `FRMW.SMUD` after Firmware Commit succeeds, which may issue an Identify Controller command after firmware activation has started.

This patch caches `FRMW.SMUD` before submitting Firmware Commit and uses the cached value when printing the completion MUD value. This avoids issuing Identify Controller after activation has started, and makes MUD handling use the pre-activation controller capability.

Closes #3396

Tested:
- make checkpatch-diff
- meson test -C .build

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -5350,9 +5350,9 @@ static bool fw_commit_support_mud(struct libnvme_transport_handle *hdl)
 	return false;
 }
 
-static void fw_commit_print_mud(struct libnvme_transport_handle *hdl, __u64 result)
+static void fw_commit_print_mud(bool mud_supported, __u64 result)
 {
-	if (!fw_commit_support_mud(hdl))
+	if (!mud_supported)
 		return;
 
 	printf("Multiple Update Detected (MUD) Value: %#" PRIx64 "\n",
@@ -5412,6 +5412,7 @@ static int fw_commit(int argc, char **argv, struct command *acmd, struct plugin 
 	struct libnvme_passthru_cmd cmd;
 	int err;
 	nvme_print_flags_t flags;
+	bool mud_supported;
 
 	struct config {
 		bool	ish;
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3404, opened 2026-05-30
Build nvme have

ikegami-t:


meson.build
```diff
@@ -408,33 +408,33 @@ if not is_static
...
 dl_dep = dependency('dl', required: false)
 conf.set(
-    'HAVE_LIBC_DLSYM',
+    'NVME_HAVE_LIBC_DLSYM',
     cc.has_function('dlsym', dependencies: dl_dep),
     description: 'Is dlsym function present',
 )
 
-conf.set('NVME_HAVE_SENDFILE', cc.has_function('sendfile'))
+conf.set('NVME_NVME_HAVE_SENDFILE', cc.has_function('sendfile'))
```
igaw: This already has the correct prefix, so no need to add another prefix.

meson.build
```diff
@@ -408,33 +408,33 @@ if not is_static
...
 
-conf.set('NVME_HAVE_MMAP', cc.has_function('mmap'))
+conf.set('NVME_NVME_HAVE_MMAP', cc.has_function('mmap'))
```
igaw: Same here, it alreayd has the `NVME` prefix.

igaw:
looks good, just the two double prefixes need to be dropped.

ikegami-t:
Just fixed the patch as mentioned. Sorry for the error. Thank you.

igaw:
No worries, that's why we have reviews. Thanks for your contributions!

PR #3405, opened 2026-05-30
nvme-types-base: rename nvme_cmd_get_log_telemetry_host_lsp

ikegami-t:
Rename to nvme_log_telemetry_host_lsp to drop the get.


igaw:
Thanks!

PR #3406, opened 2026-05-30
nvme: use #pragma once to reduce header file #ifndef code

ikegami-t:
This just follows libnvme header files implementation.


igaw:
Very nice cleanup! Thanks!

PR #3408, opened 2026-06-01
nvme-print: fix LBA range type 0h as general purpose

ikegami-t:
Since changed from reserved by NVMe specification revision 1.3.


ikegami-t:
1. NVMe specification revision 1.2.1
image

2. NVMe specification 1.3d
image

igaw:
Thanks!

PR #3411, opened 2026-06-01
plugins/solidigm: smart attribute C7 named dynamically, add desc for …

SolidigmTools:
…E7, E8

Smart Attributes - Added names for E7, E8 and set C7 name based on E7, E8


lgdacunh:
This change looks good to me.

igaw:
Thanks!

PR #3412, opened 2026-06-02
sed: clarifying SUM sed policy reporting

bbain:
The SUM policy bit indicates if the linked SUM user is in control of the the range (and common name) of its SUM locking range.  A policy of 0 indicates the SUM user is in control.  A policy of 1 indicates that any admin is in control. The value is set to 1 by default when SUM is disabled. An admin will set the policy when SUM is enabled. See section 4.2.1.4 of TCG Storage Opal SSC Feature Set: Single User Mode v1.00.

### DIFF
--- a/plugins/sed/sedopal_cmd.c
+++ b/plugins/sed/sedopal_cmd.c
@@ -791,8 +791,8 @@ void sedopal_print_sum(void *data)
 		(sumd->flags & SUM_FEATURE_ANY) ? "yes" : "no");
 	printf("\tAll Locking Objects in SUM?     : %s\n",
 		(sumd->flags & SUM_FEATURE_ALL) ? "yes" : "no");
-	printf("\tUser Authority of Objects       : %s\n",
-		(sumd->flags & SUM_FEATURE_POLICY) ? "yes" : "no");
+	printf("\tUser Controls Locking Range     : %s\n",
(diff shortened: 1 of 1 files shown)


bbain:
I tried to clarify this reporting from the SED plugin. SUM lets users take control of the locking regions. SUM lets the region's user unlock, secure erase and set the password for the region (i.e. but setting their own password). The admin is never able to see the data as it does not know the password for the region. The policy indicates whether or not the admin also wants to let the user configure the location of the region (i.e. its start and length) and its common name. 

From § 4.2.1.4 of TCG Storage Opal SSC Feature Set: Single User Mode v1.0 

> 4.2.1.4 Policy
> 
> This bit is 1 if Admins authority maintains ownership of the RangeStart and RangeLength of Locking objects in
> Single User Mode, or if the Locking SP is in any life cycle state other than Manufactured or Issued.
>
> This bit is 0 if User authorities of Locking objects in Single User Mode have ownership of their associated
> RangeStart and RangeLength columns.

igaw:
Looks good to me. The only thing I wondering if we should keep it backwards comparability? We could argue this will be released under new a major version update :)

@gjoyce-ibm any objections?

gjoyce-ibm:
Clearly I'd looked at too many bits at that point and got that one inverted. The fix is good.

igaw:
Thanks!

PR #3413, opened 2026-06-02
nvme-print-json: split fabrics Address into fields

igaw:
The Address string combines several fields into a string, which forces the user to parse the string. Add AddressDetails which splits out fields into separate key/value pairs.

Fixes: #https://github.com/linux-nvme/nvme-cli/issues/3256


igaw:
@mwilck please have a look

PR #3419, opened 2026-06-02
nvme: call nvme_show_finish by closing transport handle

ikegami-t:
Since previously done by closing device for the debug log level. But currently only calls nvme_show_init by opening tranceport handle.


igaw:
Thanks!

PR #3422, opened 2026-06-03
logging: add data and metadata dump to show command

ikegami-t:
This is for debugging the command data and metadata.

### DIFF
--- a/libnvme/src/nvme/lib.h
+++ b/libnvme/src/nvme/lib.h
@@ -14,10 +14,11 @@
 #include <nvme/lib-types.h>
 
 enum libnvme_log_level {
-	LIBNVME_LOG_ERR	  = 0,
-	LIBNVME_LOG_WARN  = 1,
-	LIBNVME_LOG_INFO  = 2,
-	LIBNVME_LOG_DEBUG = 3,
+	LIBNVME_LOG_ERR           = 0,
+	LIBNVME_LOG_WARN          = 1,
+	LIBNVME_LOG_INFO          = 2,
+	LIBNVME_LOG_DEBUG         = 3,
+	LIBNVME_LOG_DEBUG_VERBOSE = 4,
 };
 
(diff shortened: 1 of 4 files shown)


igaw:
If we have to introduce another log level it should be called DEBUG_VERBOSE, I think.

Though my question is, do we really need it? Are the data/metadata that big? Or can we make  the data/metadata output a bit smarter, e.g. skipping NULLs etc, or even parse it instead just dumping the raw buffers?

ikegami-t:
The command dword data output by the log level debug. And the data dump function `d()` doe s not skip NULLs by the same log level debug. Then alway NULLs output by the function. To skip NULLs change the function as only dumps NULLs by the log level verbose. So just changed the log level to DEBUG_VERBOSE.

libnvme/src/nvme/lib.h
```diff
@@ -14,10 +14,11 @@
...
 enum libnvme_log_level {
-	LIBNVME_LOG_ERR	  = 0,
-	LIBNVME_LOG_WARN  = 1,
-	LIBNVME_LOG_INFO  = 2,
-	LIBNVME_LOG_DEBUG = 3,
+	LIBNVME_LOG_ERR           = 0,
+	LIBNVME_LOG_WARN          = 1,
+	LIBNVME_LOG_INFO          = 2,
+	LIBNVME_LOG_DEBUG         = 3,
+	LIBNVME_LOG_DEBUG_VERBOSE = 4,
```
igaw, line 21: You need also to update `libnvme/libnvme/nvme.i` accordingly.

igaw:
Okay, then let's have the DEBUG_VERBOSE, but it's the last log level! (famous last words)

libnvme/src/nvme/lib.h
ikegami-t: Just updated `libnvme/libnvme/nvme.i`. Thank you.

igaw:
Thanks!

PR #3423, opened 2026-06-03
nvme: ensure log_level is properly updated

martin-gpy:
The log_level for certain commands like show-topology and tls-key does not get updated even when the verbose option is passed to them. Fix this by ensuring parse_args() is invoked in these commands instead of argconfig_parse(), so that the log_level gets properly updated here.


igaw:
Thanks!

PR #3425, opened 2026-06-03
libnvmf: NVMe Controller Ownership Registry (V2)

martin-belanger:
This addresses issue https://github.com/linux-nvme/nvme-cli/issues/2913. It is a clean replacement for PR #3400, which was closed in favor of this redesigned version.

---

## Problem

Multiple independent orchestrators can establish NVMe-oF controller connections on the same Linux host simultaneously:

- Manual `nvme connect` / `nvme connect-all` (stateless, one-shot)
- `nvme-stas` (`stafd` / `stacd`) — production-grade daemon with CDC/fabric-zoning support
- NBFT firmware boot connections
- `nvme-discoverd` — ships with nvme-cli, runs by default (future project)

All connected controllers appear in a single flat namespace (`/dev/nvmeX`) with no indication of which orchestrator created or manages each one. Commands like `nvme disconnect-all` are therefore indiscriminate and dangerous.

The registry gives well-behaved orchestrators the information they need to avoid accidentally disconnecting controllers managed by another tool. It is a **cooperative coordination mechanism**, not an enforcement boundary — all participants are assumed to cooperate.

A guiding principle throughout: **libnvme provides the mechanism (primitives), the orchestrator layer provides the policy.** `connect-all` / `disconnect-all` are themselves human-driven orchestrators, so coordination *policy* (ownership rules, the confirmation prompt, which transports to skip) lives there, while libnvme exposes only the building blocks.

---

## What changed from PR #3400

| Area | PR #3400 | This PR |
|------|----------|---------|
| Storage format | One JSON file per controller (`nvme3.json`) | One directory per controller, one plain-text file per attribute (`nvme3/owner`) |
| json-c dependency | Required in the registry path | Eliminated — registry works in all build configurations |
| CLI commands | Top-level built-ins (`nvme registry-list`) | Optional plugin (`nvme registry list`) |
| API terminology | `key` | `attr` |
| Iteration API | `for_each(device, owner, user_data)` | `device_for_each(device, user_data)` + `attr_for_each(attr, value, user_data)` |
| `application` field | Left in place | Removed from `libnvme_global_ctx` and `libnvme_subsystem` |
| Parallel tests | None | 10-process concurrent write test (C and Python) |

---

### Directory-based storage

```
/run/nvme/registry/
    nvme1/
        owner          ← contains "stas\n"
    nvme3/
        owner          ← contains "nbft\n"
```

Mirrors the kernel's sysfs convention. Trivially inspectable with `cat`. Extensible — new attributes are new files, no schema change needed. No json-c dependency in the registry path. `/run` is the FHS-correct home for ephemeral runtime state (cleared at boot), and the location is a fixed coordination point so that all orchestrators on the host agree on it.

Each attribute is written atomically (temp file + `rename`) so a concurrent reader never observes a half-written value:

```
mkostemp ->  <attr>.tmp.XXXXXX   (O_CLOEXEC; mkstemp + fcntl(FD_CLOEXEC) fallback where _GNU_SOURCE is undefined)
fchmod   ->  0644
write    ->  value + newline
rename   ->  <attr>              (atomic replace)
fsync    ->  directory
```

`rename()` provides the only guarantee we need (atomic visibility); there is no per-file `fsync`, since `/run` is ephemeral and durability across reboot is explicitly not wanted.

### Owner identity

A process declares its orchestrator identity once, after creating the context:

```c
struct libnvme_global_ctx *ctx = libnvme_create_global_ctx(fp, log_level);
libnvme_set_owner(ctx, "stas");   /* or "nbft", "discoverd", ... */
```

`libnvme_create_global_ctx()` keeps its original two-argument signature; a process that does not participate in the registry simply never calls `libnvme_set_owner()`. The library treats `owner` as plain data and does **not** enforce immutability — that is a policy decision, and policy belongs one layer up (the orchestrator). The CLI does apply such a policy: `nvme registry update` refuses to write the reserved `owner` attribute, since ownership is recorded at connect time rather than hand-edited.

### Transport classification

`libnvme_ctrl_is_transport_fabric()` reports whether a controller is reached over a fabrics transport, replacing open-coded `strcmp()` against the local transports (`pcie` / `apple-nvme`). It is defined by exclusion, so a newly added transport is treated as fabrics by default. This is the mechanism the orchestrator (`disconnect-all`) uses to skip local controllers; the policy of what to do with the answer stays in the orchestrator.

### Removal of `application`

The `application` field was a previous attempt to solve the same multi-orchestrator coordination problem. It assumed all orchestrators share a hand-written JSON config file, which is false for every real orchestrator in the ecosystem (NBFT reads firmware tables; nvme-stas uses DNS-SD; nvme-discoverd reacts to udev events). `owner` solves the same problem through a shared runtime path (`/run/nvme/registry/`) without any shared-config assumption.

Both `libnvme_global_ctx.application` and `libnvme_subsystem.application` are removed, along with `libnvme_get/set_application()` and the connect-time filtering logic that compared them. This is an intentional v3.0 API break. (The `nvme connect` `--context` option, which only existed to set that field, is removed too, and the autoconnect systemd units are updated accordingly.)

### Plugin placement

Registry commands live in `plugins/registry/` (`nvme registry list`, `nvme registry retrieve`, etc.) rather than as top-level built-ins, and the plugin is optional (selectable via `-Dplugins`). This keeps the core binary lean and signals that these are orchestration tools rather than general NVMe commands. The registry *library* is always present; only the CLI frontend is optional.

---

## Commit series

1. **`libnvme: add libnvme_ctrl_is_transport_fabric()`** — transport-classification primitive used by the ownership-aware `disconnect-all`.

2. **`libnvme: add NVMe controller ownership registry`** — `registry.c`, `registry.h`, the registry C API (`retrieve` / `update` / `delete` / `attr_equal` / `device_for_each` / `attr_for_each`), Python bindings, C unit tests with a 10-process parallel write test, Python unit tests, udev cleanup rule.

3. **`nvme: add registry plugin and orchestrator column`** — `plugins/registry/` plugin with four subcommands, the `nvme list -v` Orchestrator column (via `libnvme_ctrl_owner()`, with a no-fabrics fallback), CLI integration test (`tests/nvme_registry_test.py`, gated by `-Dnvme-tests`).

4. **`libnvmf: wire registry into connect and disconnect paths`** — connect hook in `__nvmf_add_ctrl()` (registers ownership; failures are logged but never fail the connection), `libnvme_set_owner()`, `connect-all --nbft` registers boot controllers as `owner=nbft`, ownership-aware `disconnect-all` with `--owner` / `--force` (confirmation prompt on a TTY), removal of `application` (and the `--context` option) from all associated code.

5. **`nvme: add man pages for registry commands and update disconnect-all`** — man pages for all four registry subcommands, updated `nvme-disconnect-all(1)`.

---

### Added

```c
/* Public, always present (libnvme.ld) */
bool  libnvme_ctrl_is_transport_fabric(libnvme_ctrl_t c);
char *libnvme_ctrl_owner(libnvme_ctrl_t c);                 /* registered owner, or NULL; NULL in no-fabrics builds */
int   libnvme_set_owner(struct libnvme_global_ctx *ctx, const char *owner);

/* Public, fabrics only (libnvmf.ld) */
int libnvmf_registry_retrieve(const char *device, const char *attr, char **value);
int libnvmf_registry_attr_equal(const char *device, const char *attr, const char *value);   /* 0 equal / >0 differ / <0 -errno; NULL value == "absent" */
int libnvmf_registry_update(const char *device, const char *attr, const char *value);       /* value == NULL removes the attribute */
int libnvmf_registry_delete(const char *device);
int libnvmf_registry_device_for_each(void (*callback)(const char *device, void *user_data), void *user_data);
int libnvmf_registry_attr_for_each(const char *device, void (*callback)(const char *attr, const char *value, void *user_data), void *user_data);
```

Python bindings expose `registry_retrieve`, `registry_update`, `registry_delete`, `registry_devices`, `registry_device_attrs`, and a `registry_entries()` generator — all backed by the C functions above (no file access is reimplemented in Python).

### Removed (intentional v3.0 break)

- `libnvme_get_application()` / `libnvme_set_application()`
- `libnvme_subsystem_get_application()` / `libnvme_subsystem_set_application()`
- `libnvme_global_ctx.application` and `libnvme_subsystem.application` fields
- `nvme connect`/`connect-all` `--context` option

`libnvme_create_global_ctx()` is **unchanged** (`fp`, `log_level`); registry participation is opt-in via `libnvme_set_owner()`.

---

## Testing

- C unit tests: CRUD operations, stale-entry skipping, iteration, NULL-arg validation, and a 10-process concurrent write test (`libnvme/test/registry.c`)
- Python unit tests: same coverage via the bindings, including a 10-process parallel write test (`libnvme/libnvme/tests/test-registry.py`)
- CLI integration tests: all four subcommands, argument validation, error cases (`tests/nvme_registry_test.py`, gated by `-Dnvme-tests`)
- No-fabrics build: registry code is correctly excluded and the `nvme list` Orchestrator column falls back via a stub
- Musl-style build (`-U_GNU_SOURCE`): clean build (the `mkostemp` / GNU-extension paths are guarded)
- Every commit in the series builds standalone (bisectable)


igaw:
Quickly glanced over it. Generally looks good. I don't like the `libnvme_create_global_ctx` change, most of the callsite pass in NULL, which is the default then. 

I need fresh eyes to look into a bit deeper :)

martin-belanger:
Thanks for the quick glance! On the `libnvme_create_global_ctx()` change — I made `owner` a constructor parameter rather than adding a separate setter specifically to keep it immutable for the lifetime of the context. If there were a setter, any code holding a pointer to the context could change ownership identity at any point during the process's life, which is exactly the kind of mutability I wanted to rule out for something meant to identify "who connected this" for the lifetime of a registry entry.

I agree the NULL-at-most-callsites pattern doesn't look great today, but that's mostly because most existing callers (largely unit tests) don't have an owner concept yet and were mechanically updated to pass NULL just to keep them compiling — not a sign the API shape itself is wrong, just that the call sites haven't caught up.

igaw:
If the user code want's to change the owner ship, it just creates another global context. It's not that this is really difficult to workaround. I'd really don't like to make the same mistake by adding arguments to these central objects, e.g. I even tried to get rid of the file descriptor and log level setting and I am still pondering if we should just do it. `libnvme_create_global_context` has fallback code when the user doesn't pass in a file descriptor anyway.

martin-belanger:
> If the user code want's to change the owner ship, it just creates another global context. It's not that this is really difficult to workaround. I'd really don't like to make the same mistake by adding arguments to these central objects, e.g. I even tried to get rid of the file descriptor and log level setting and I am still pondering if we should just do it. `libnvme_create_global_context` has fallback code when the user doesn't pass in a file descriptor anyway.

Fair enough — I won't fight the direction you're taking the constructor, especially if `fp` and `log_level` are on their way out too. I'll drop `owner` from `libnvme_create_global_ctx()`.

To keep the immutability I was after, I'll add a one-shot setter instead:

```c
int libnvme_set_owner(struct libnvme_global_ctx *ctx, const char *owner);
```

It can only be set once: a second call returns an error and leaves the existing value untouched. That keeps "who connected this" stable for the lifetime of the context without piling another argument onto the central object. Callers that don't participate in the registry simply never call it, so the NULL-at-every-call-site churn goes away.

And your point still holds on top of this — if a caller really does want a different owner, it creates another global context; the one-shot setter only rules out mutating the identity of an existing one.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,432 @@
...
+			return -errno;
+		fsync(dir_fd);
+		return 0;
+	}
+
+	fd = mkstemp(tmp_path);
+	if (fd < 0)
+		return -errno;
+	fcntl(fd, F_SETFD, FD_CLOEXEC);
+
```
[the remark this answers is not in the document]
martin-belanger: I originally used `mkostemp()`, but it's a GNU extension whose glibc declaration is gated behind `_GNU_SOURCE`. Our musl-style CI build (`config_meson_musl` in `scripts/build.sh`) compiles with `-U_GNU_SOURCE` to catch GNU-isms, which made `mkostemp()` fail with an implicit-declaration error — so I'd resorted to `mkstemp + fcntl`.

What I'll do instead is use `mkostemp` when it's available:
```
#ifdef _GNU_SOURCE
	fd = mkostemp(tmp_path, O_CLOEXEC);
	if (fd < 0)
		return -errno;
#else
	fd = mkstemp(tmp_path);
	if (fd < 0)
		return -errno;
	fcntl(fd, F_SETFD, FD_CLOEXEC);
#endif
```

igaw:
Yes, that sounds reasonable and we could also error out or warn if the owner is not set when doing an operation which could cause havoc?

libnvme/src/nvme/registry.c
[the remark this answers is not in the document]
martin-belanger: Done. I now use `mkostemp(tmp_path, O_CLOEXEC)` when `_GNU_SOURCE` is defined, and fall back to `mkstemp()` + `fcntl(FD_CLOEXEC)` otherwise. `mkostemp()` is a glibc extension whose declaration is gated behind `_GNU_SOURCE`, and the musl-style CI build (`config_meson_musl` in `scripts/build.sh`, which passes `-U_GNU_SOURCE`) rejects it with an implicit-declaration error — the `#ifdef` keeps both paths correct.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,432 @@
...
+	ret = write(fd, value, strlen(value));
+	if (ret < 0) {
+		ret = -errno;
+		goto err;
+	}
+	ret = write(fd, "\n", 1);
+	if (ret < 0) {
+		ret = -errno;
+		goto err;
+	}
```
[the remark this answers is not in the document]
martin-belanger: Done. Added a `write_all()` helper that loops until the whole buffer is written (and retries on `EINTR`), used for both the value and the trailing newline.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,432 @@
...
+	if (fd < 0)
+		return -errno;
+
+	n = read(fd, buf, sizeof(buf) - 1);
+	close(fd);
+	if (n < 0)
+		return -errno;
+	if (n == 0)
+		return -ENOENT;
+
```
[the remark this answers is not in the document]
martin-belanger: Done. `libnvmf_registry_retrieve()` now returns `-E2BIG` when the read fills the buffer completely, instead of returning a truncated value.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,432 @@
...
+	if (errno != ENOENT)
+		return -errno;
+
+	/* Parent /run/nvme missing; create it then retry. */
+	if (mkdir("/run/nvme", 0755) < 0 && errno != EEXIST)
+		return -errno;
+	if (mkdir(registry_dir(), 0755) < 0 && errno != EEXIST)
+		return -errno;
+	return 0;
+}
```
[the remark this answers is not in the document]
martin-belanger: Done. Replaced the hardcoded parent creation with a generic recursive `mkdir_p()` that creates `registry_dir()` and any missing parents. (`NVME_REGISTRY_DIR` is also now confined to `/tmp` so it can't redirect the registry's writes/deletes to an arbitrary location in a privileged process.)

fabrics.c
```diff
@@ -952,17 +974,39 @@ int fabrics_disconnect_all(const char *desc, int argc, char **argv)
...
+					libnvmf_registry_retrieve(name,
+						"owner", &reg_owner);
+					if (cfg.owner)
+						do_disconnect = reg_owner &&
+							!strcmp(reg_owner,
+								cfg.owner);
+					else
+						do_disconnect = !reg_owner;
+					free(reg_owner);
+				}
```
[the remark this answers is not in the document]
martin-belanger: Done. I now capture the return code and treat a controller as unowned only when it returns `-ENOENT`. On any other error (e.g. `EACCES`/`EIO`) the controller is skipped rather than disconnected, so a temporarily unreadable registry can't tear down an owned controller.

Documentation/nvme-registry-update.txt
```diff
@@ -0,0 +1,70 @@
...
+-----------
+Write an attribute to the ownership registry entry for an NVMe-oF
+controller.  The registry is stored under /run/nvme/registry/ as one
+directory per live controller, with one plain-text file per attribute.
+The entry directory is created if it does not already exist.
+
+This command is primarily used by orchestrators to claim or transfer
+ownership of an existing connection.  An orchestrator with greater
+management authority (e.g. nvme-stas with CDC/fabric-zoning knowledge)
+may call this command to take ownership from one that connected first.
```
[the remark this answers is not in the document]
martin-belanger: Fixed. The `owner` attribute is immutable (recorded at connect time), so I rewrote the DESCRIPTION to describe writing non-owner attributes and to state that updating `owner` is rejected.

Documentation/nvme-registry-update.txt
```diff
@@ -0,0 +1,70 @@
...
+OPTIONS
+-------
+-d <DEV>::
+--device=<DEV>::
+	NVMe device name (e.g. nvme3).  The /dev/ prefix is optional.
+
+-a <ATTR>::
+--attr=<ATTR>::
+	Attribute name to write (e.g. owner).
+
```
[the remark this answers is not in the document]
martin-belanger: Fixed. Changed the example to `note` and noted that the reserved `owner` attribute cannot be written with this command.

Documentation/nvme-registry-update.txt
```diff
@@ -0,0 +1,70 @@
...
++
+------------
+# nvme registry update --device nvme3 --attr owner --value stas
+------------
+
+* Transfer ownership to a different orchestrator:
++
+------------
+# nvme registry update --device nvme3 --attr owner --value nbft
+------------
```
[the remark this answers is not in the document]
martin-belanger: Fixed. Replaced both examples with a mutable attribute (`note`) so copy/paste usage works.

libnvme/libnvme/nvme.i
```diff
@@ -545,13 +573,71 @@ PyObject *nbft_get(struct libnvme_global_ctx *ctx, const char *filename)
...
+	Returns:
+		The attribute value as a string, or None if the device is not
+		registered or the attribute is not present.
+	"""
+	path = os.path.join(_registry_dir, device, attr)
+	try:
+		with open(path) as f:
+			return f.read().strip()
+	except OSError:
+		return None
```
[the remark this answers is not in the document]
martin-belanger: Fixed. Changed to `rstrip("\n\r")`, so only the trailing newline added by the writer is removed; leading/trailing spaces in a value are preserved, matching the C reader.

libnvme/libnvme/nvme.i
```diff
@@ -545,13 +573,71 @@ PyObject *nbft_get(struct libnvme_global_ctx *ctx, const char *filename)
...
+		attrs = {}
+		try:
+			for attr in os.listdir(dev_path):
+				if attr.startswith('.') or '.tmp.' in attr:
+					continue
+				try:
+					with open(os.path.join(dev_path, attr)) as f:
+						attrs[attr] = f.read().strip()
+				except OSError:
+					continue
```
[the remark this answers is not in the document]
martin-belanger: Fixed. Same change: `rstrip("\n\r")` to match the C behavior so values round-trip accurately.

libnvme/libnvme/nvme.i
```diff
@@ -545,13 +573,71 @@ PyObject *nbft_get(struct libnvme_global_ctx *ctx, const char *filename)
...
+	Returns:
+		The attribute value as a string, or None if the device is not
+		registered or the attribute is not present.
+	"""
+	path = os.path.join(_registry_dir, device, attr)
+	try:
+		with open(path) as f:
+			return f.read().rstrip("\n\r")
+	except OSError:
+		return None
```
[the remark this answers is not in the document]
martin-belanger: Fixed. The Python bindings now validate the arguments before building the path, mirroring the C API: device must match `nvme[0-9]+` and attr `[A-Za-z0-9_-]+`, otherwise `registry_retrieve()` raises `ValueError` (consistent with `registry_update`/`registry_delete`). I also confined the `NVME_REGISTRY_DIR` override on the Python side the same way the C `registry_dir()` does — honored only if it is under `/tmp` and free of `..`, otherwise the production directory — via a small `_init_registry_dir()` helper.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,492 @@
...
+		 * ".." traversal; anything else falls back to the production
+		 * directory.
+		 */
+		if (env && strncmp(env, "/tmp/", 5) == 0 && !strstr(env, ".."))
+			dir = env;
+		else
+			dir = "/run/nvme/registry";
+	}
+	return dir;
+}
```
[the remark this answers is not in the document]
martin-belanger: Fixed. `registry_dir()` now copies the value into a static `char buf[PATH_MAX]` with `snprintf()` instead of caching `getenv()`'s pointer, so a later `setenv()`/`putenv()` cannot invalidate it. (The Python side has no equivalent issue: `os.environ.get()` returns an immutable string copy.)

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,492 @@
...
+#ifdef _GNU_SOURCE
+	fd = mkostemp(tmp_path, O_CLOEXEC);
+	if (fd < 0)
+		return -errno;
+#else
+	fd = mkstemp(tmp_path);
+	if (fd < 0)
+		return -errno;
+	fcntl(fd, F_SETFD, FD_CLOEXEC);
+#endif
```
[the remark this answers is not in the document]
martin-belanger: Fixed. The fallback path now checks `fcntl()` and, on failure, cleans up the temp file and returns the error, matching the `mkostemp(O_CLOEXEC)` path.

fabrics.c
```diff
@@ -914,17 +909,45 @@ int fabrics_disconnect_all(const char *desc, int argc, char **argv)
...
+		else
+			fprintf(stderr,
+				"WARNING: --owner disconnects all NVMeoF controllers\n"
+				"owned by '%s'. Type 'yes' to confirm: ",
+				cfg.owner);
+		if (!fgets(ans, sizeof(ans), stdin) ||
+		    strncmp(ans, "yes", 3) != 0) {
+			fprintf(stderr, "Aborted.\n");
+			return -EINVAL;
+		}
```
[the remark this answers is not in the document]
martin-belanger: Fixed. The prompt now trims the trailing newline and requires an exact `strcmp(ans, "yes") == 0`, so only "yes" confirms.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,492 @@
...
+
+		fd = openat(dir_fd, de->d_name, O_RDONLY | O_CLOEXEC);
+		if (fd < 0)
+			continue; /* device may have been removed concurrently */
+
+		n = read(fd, buf, sizeof(buf) - 1);
+		close(fd);
+		if (n <= 0)
+			continue;
+
```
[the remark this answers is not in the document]
martin-belanger: Fixed. The iterator now detects a full-buffer read (`n == sizeof(buf) - 1`) and skips that attribute rather than passing truncated data to the callback, matching `libnvmf_registry_retrieve()` (which returns `-E2BIG`).

libnvme/src/nvme/lib.h
```diff
@@ -39,6 +39,20 @@ struct libnvme_global_ctx *libnvme_create_global_ctx(FILE *fp, int log_level);
...
+ * @ctx:	&struct libnvme_global_ctx object
+ * @owner:	Orchestrator identity string (e.g. "stas", "nbft").
+ *
+ * Records the orchestrator identity used when claiming registry ownership of
+ * connections made through @ctx.  May be set only once: a second call returns
+ * -EEXIST and leaves the existing identity untouched.  A process that does not
+ * participate in the registry simply never calls this.
+ *
+ * Return: 0 on success, -EEXIST if already set, -EINVAL or -ENOMEM on error.
+ */
```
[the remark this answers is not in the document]
martin-belanger: Fixed. The kdoc for `libnvme_set_owner()` now states that it is the supported way to record the registry owner and that `libnvme_create_global_ctx()` deliberately takes no owner parameter. (The PR description predates the API change; the series uses `create_global_ctx(fp, level)` + `set_owner(ctx, owner)`.)

libnvme/libnvme/tests/test-registry.py
```diff
@@ -0,0 +1,143 @@
...
+"""
+import multiprocessing
+import os
+import tempfile
+import unittest
+
+# meson sets VALGRIND_OPTS when running under valgrind.  Forked child
+# processes under valgrind can behave unexpectedly; the parallel write
+# test is already covered by the C test suite (libnvme/test/registry.c).
+_under_valgrind = 'VALGRIND_OPTS' in os.environ
```
igaw, line 17: interesting idea. I think this could also be used for some of the existing tests as well. have to check...

igaw:
> To keep the immutability I was after, I'll add a one-shot setter instead:
> 
> ```c
> int libnvme_set_owner(struct libnvme_global_ctx *ctx, const char *owner);
> ```
> 
> It can only be set once: a second call returns an error and leaves the existing value untouched. That keeps "who connected this" stable for the lifetime of the context without piling another argument onto the central object. Callers that don't participate in the registry simply never call it, so the NULL-at-every-call-site churn goes away.

In this case you should also have some check in `libnvmf_registry_update` which would prevent the user to update the `owner`, or do I miss something?

Anyway, I don't have a strong opinion if we really need this 'safety net', the design is based on trusting all parties to behave correctly. So maybe we just leave it with the check in `libnvme_set_owner`.

libnvme/libnvme/nvme.i
```diff
@@ -545,13 +573,101 @@ PyObject *nbft_get(struct libnvme_global_ctx *ctx, const char *filename)
...
+		and len(device) > 4 and all(c in '0123456789' for c in device[4:]))
+
+
+def _valid_registry_attr(attr):
+	# Mirror libnvme valid_attr(): non-empty [A-Za-z0-9_-]+.
+	return (isinstance(attr, str) and len(attr) > 0
+		and all(c in _REGISTRY_ATTR_CHARS for c in attr))
+
+
+def registry_retrieve(device, attr):
```
igaw: I don't understand why there is a `registry_retrieve` and `registry_entries` Python implementation, shouldn't there be just binding code? I'd like to see all the logic code in the library, especially this code where we really need to be careful accessing the registry files.

libnvme/src/meson.build
```diff
@@ -144,7 +141,7 @@ if want_fabrics
         '-Wl,--version-script=@0@'.format(nvmf_accessors_ld),
     ]
     libconf.set('FABRICS_INCLUDE',
-                '#include <nvme/fabrics.h>\n#include <nvme/nbft.h>\n#include <nvme/accessors-fabrics.h>')
+                '#include <nvme/fabrics.h>\n#include <nvme/nbft.h>\n#include <nvme/accessors-fabrics.h>\n#include <nvme/registry.h>')
```
igaw, line 149: Maybe we should start doing something like

`fabrics-includes.h.in`:
```
#include <nvme/accessors-fabrics.h>
#include <nvme/fabrics.h>
#include <nvme/nbft.h>
#include <nvme/registry.h>
```

```meson
fabrics_include = files('fabrics-includes.h.in')[0].full_path()
libconf.set('FABRICS_INCLUDE', fs.read(fabrics_include))
```

igaw: ideally as a preparation patch :)

libnvme/libnvme/tests/test-registry.py
```diff
@@ -0,0 +1,143 @@
...
+    os.rmdir(_tmpdir)
+
+
+class TestRegistryUpdate(unittest.TestCase):
+
+    def tearDown(self):
+        nvme.registry_delete('nvme5')
+
+    def test_update_creates_entry(self):
+        nvme.registry_update('nvme5', 'owner', 'stas')
```
igaw, line 43: When registry_update fails for some reason, do we get proper exceptions here?

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+		 * override to /tmp and reject ".." traversal; anything else
+		 * falls back to the production directory.
+		 *
+		 * Copy into a static buffer: getenv()'s pointer can be
+		 * invalidated by a later setenv()/putenv().
+		 */
+		if (env && strncmp(env, "/tmp/", 5) == 0 &&
+		    !strstr(env, "..")) {
+			snprintf(buf, sizeof(buf), "%s", env);
+			dir = buf;
```
igaw, line 56: I don't think this make sense at all. Either it runs with root rights or with user rights. I would follow what the  exising `make_sysfs_dir` version does.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+		} else {
+			dir = "/run/nvme/registry";
```
igaw, line 58: Do we need to make this configurable? I know we don't have it for the sysfs, because that's the kernel API, but this could be something distros have different takes. Or do has everyone settle on this file system layout?

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+		}
+	}
+	return dir;
+}
+
+/* mkdir -p: create @path and any missing parents. */
+static int mkdir_p(const char *path, mode_t mode)
+{
+	char tmp[256];
```
igaw: I would suggest to make this dynamic, this is an arbitrary limit.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	}
+	if (mkdir(tmp, mode) < 0 && errno != EEXIST)
+		return -errno;
+	return 0;
+}
+
+static int ensure_registry_dir(void)
+{
+	return mkdir_p(registry_dir(), 0755);
+}
```
igaw: just embed this little helper.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+static int ensure_device_dir(const char *device)
+{
+	char path[256];
+	int ret;
+
+	ret = ensure_registry_dir();
+	if (ret)
+		return ret;
+
+	if (snprintf(path, sizeof(path), "%s/%s", registry_dir(), device) >=
```
igaw: use asprintf here to eliminate the arbitrary path length limit.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	if (!device || strncmp(device, "nvme", 4) != 0)
+		return false;
+	p = device + 4;
+	if (!*p)
+		return false;
+	for (; *p; p++) {
+		if (*p < '0' || *p > '9')
+			return false;
+	}
+	return true;
```
igaw, line 139: a bunch of newlines between the code would look better.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+
+	if (!attr || !*attr)
+		return false;
+	for (p = attr; *p; p++) {
+		if ((*p >= 'a' && *p <= 'z') || (*p >= 'A' && *p <= 'Z') ||
+		    (*p >= '0' && *p <= '9') || *p == '_' || *p == '-')
+			continue;
+		return false;
+	}
+	return true;
```
igaw, line 154: same here, it is not really consistent coding style

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+}
+
+/* write() may return a short count; loop until the whole buffer is written. */
```
igaw: unnecessary comment, it's clear from the code.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+ *
+ * mkostemp() atomically creates the tmp file with a random suffix, preventing
+ * both name prediction and TOCTOU races on the tmp file itself.  Builds
+ * without _GNU_SOURCE fall back to mkstemp() + fcntl(FD_CLOEXEC) (see below).
+ */
+static int write_attr_atomic(int dir_fd, const char *dir_path,
+			     const char *attr, const char *value)
+{
+	char tmp_path[512];
+	char final_path[512];
```
igaw: again, use dynamic buffers with asprintf to avoid the arbitrary path length limit.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+		goto err;
+	ret = write_all(fd, "\n", 1);
+	if (ret)
+		goto err;
+	if (fsync(fd) < 0) {
+		ret = -errno;
+		goto err;
+	}
+	close(fd);
+	fd = -1;
```
igaw: unnecessary to clear the fd.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	close(fd);
+	unlink(tmp_path);
+	return ret;
+}
+
+/*
+ * libnvmf_registry_create - Write a registry entry for a freshly connected
+ * controller.  Internal; called from the connect path in fabrics.c once the
+ * kernel returns instance=N.  Always overwrites any pre-existing entry:
+ * instance recycling means an old nvmeN/ directory is stale by definition.
```
igaw: I think it would be good to have a warn log message in this case. For debugging purpose this would be very helpful I think.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+ */
+int libnvmf_registry_create(int instance, const char *owner)
+{
+	char device[32];
+	char dir_path[256];
```
igaw: same as above with the paths lenght limitation. if there are more of those, please address them as well. I stop commenting on this one.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+
+	if (snprintf(path, sizeof(path), "%s/%s/%s",
+		     registry_dir(), device, attr) >= (int)sizeof(path))
+		return -ENAMETOOLONG;
+
+	fd = open(path, O_RDONLY | O_CLOEXEC);
+	if (fd < 0)
+		return -errno;
+
+	n = read(fd, buf, sizeof(buf) - 1);
```
igaw: `read` can fail with EINTR, like `write`.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	close(fd);
+	if (n < 0)
+		return -errno;
+	if (n == 0)
+		return -ENOENT;
+	/*
+	 * A full buffer means the value is larger than we support; refuse it
+	 * rather than silently return a truncated value.
+	 */
```
igaw: we could also allocate the right size buffer, as later there is a strdup anyway.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	if (ret)
+		return ret;
+
+	if (snprintf(dir_path, sizeof(dir_path), "%s/%s",
+		     registry_dir(), device) >= (int)sizeof(dir_path))
+		return -ENAMETOOLONG;
+
+	dir_fd = open(dir_path, O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+	if (dir_fd < 0)
+		return -errno;
```
igaw: I would push this into write_attr_atomic, because there are two identically code snippets.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	if (ret)
+		return ret;
+
+	if (rmdir(path) < 0 && errno != ENOENT)
+		return -errno;
+	return 0;
+}
+
+__libnvme_public int libnvmf_registry_device_for_each(
+		void (*cback)(const char *device, void *user_data),
```
igaw: A better than `cback` would be great.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+			continue;
+		/* Skip in-flight tmp files from concurrent writers. */
+		if (strstr(de->d_name, ".tmp."))
+			continue;
+
+		fd = openat(dir_fd, de->d_name, O_RDONLY | O_CLOEXEC);
+		if (fd < 0)
+			continue; /* device may have been removed concurrently */
+
+		n = read(fd, buf, sizeof(buf) - 1);
```
igaw: same as above, EINTR and a dynamic allocated buffer would be great instead arbitrary limiting it.

libnvme/sysfiles/udev/rules.d/70-nvmf-registry.rules
```diff
@@ -0,0 +1,13 @@
...
+# %k expands to the kernel device name (e.g. nvme3).
+#
+# The kernel allocates controller instance numbers with ida_alloc(), which
+# returns the lowest available ID.  A freed instance number can therefore be
+# reassigned to a new controller before this rule runs for the old one.  The
+# [ -e /dev/%k ] guard detects that case: devtmpfs removes the old node
+# before KOBJ_REMOVE fires, so if the node exists again a new controller has
+# already claimed the name and its registry entry must not be deleted.
+ACTION=="remove", SUBSYSTEM=="nvme", KERNEL=="nvme[0-9]*", \
+    RUN+="/bin/sh -c '[ -e /dev/%k ] || rm -rf /run/nvme/registry/%k'"
```
igaw, line 13: Why don't we use nvme-cli to remove the entry? It has additional checks and it would make any logging simpler.

libnvme/test/registry.c
```diff
@@ -0,0 +1,427 @@
...
+	 */
+	struct dirent *de;
+	DIR *d;
+
+	d = opendir(tmpdir);
+	if (!d)
+		return;
+	while ((de = readdir(d)) != NULL) {
+		if (de->d_name[0] != '.')
+			libnvmf_registry_delete(de->d_name);
```
igaw, line 55: checking the return code?

igaw:
The 2nd commit message says:

```
nvme: add registry plugin and orchestrator column
The registry commands belong in a plugin rather than as top-level
built-ins to keep the core nvme binary lean and to make it clear
these are orchestration tools, not general NVMe commands.
```

which is strange: we discussed this but the new code was never on top level. doesn't really make sense IMO.

The third commit message seems to have same problem:

```
libnvme_create_global_ctx() gains an owner parameter so the identity
is set at construction time and cannot change during the context
lifetime.  All callers that do not participate in the registry pass
NULL.  connect-all --nbft passes "nbft", registering boot-volume
controllers so they are protected from accidental disconnection.
```

plugins/registry/registry-nvme.c
```diff
@@ -0,0 +1,171 @@
...
+#include "common.h"
+#include "nvme.h"
+#include "util/cleanup.h"
+
+#define CREATE_CMD
+#include "registry-nvme.h"
+
+static void strip_dev_prefix(char **device)
+{
+	if (!strncmp(*device, "/dev/", 5))
```
igaw, line 25: This should also handle spdk devices, which have the prefix `/dev/spdk`

tests/nvme_registry_test.py
```diff
@@ -0,0 +1,158 @@
...
+        self._populate('nvme2', 'owner', 'nbft')
+        result = self._run('registry', 'list')
+        self.assertEqual(result.returncode, 0)
+
+
+if __name__ == '__main__':
+    # Remove the binary path from argv so unittest.main() doesn't see it.
+    if len(sys.argv) >= 2 and not sys.argv[1].startswith('-'):
+        del sys.argv[1]
+    unittest.main()
```
igaw, line 158: Nice end to end test!

meson.build
```diff
@@ -596,6 +596,16 @@ if want_nvme
...
+
+    if want_fabrics and 'registry' in selected_plugins
+        _py3 = find_program('python3', required: false)
+        if _py3.found()
+            test('nvme-cli - registry-cli',
+                 _py3,
+                 args: [files('tests/nvme_registry_test.py'), nvme_exe],
+                 depends: nvme_exe)
+        endif
+    endif
```
igaw: this should be in the tests directory, where it is gated by `nvme-tests` option, see the next line.

nvme-print-stdout.c
```diff
@@ -5854,16 +5854,38 @@ static bool stdout_detailed_ctrl(const char *name, void *arg)
...
-	       libnvme_ctrl_get_phy_slot(c),
-	       libnvme_subsystem_get_name(libnvme_ctrl_get_subsystem(c)));
+	{
+		const char *tr = libnvme_ctrl_get_transport(c);
+		const char *owner_str = "-";
+#ifdef CONFIG_FABRICS
+		char *reg_owner = NULL;
+
+		if (!strcmp(tr, "pcie") || !strcmp(tr, "apple-nvme")) {
+			owner_str = "kernel";
```
igaw, line 5864: This logic should be in `libnvme_registry_retrieve` (or where it belongs) not here.

nvme-print-stdout.c
```diff
@@ -5854,16 +5854,38 @@ static bool stdout_detailed_ctrl(const char *name, void *arg)
...
+		} else {
+			libnvmf_registry_retrieve(libnvme_ctrl_get_name(c),
+						  "owner", &reg_owner);
+			if (reg_owner)
+				owner_str = reg_owner;
+		}
+#else
+		owner_str = "kernel";
```
igaw: given that this depends on the support of FABRICS enabled, libnvme should provide a fallback implementation for the registry when not available which returns tthis.

nvme-print-stdout.c
```diff
@@ -5854,16 +5854,38 @@ static bool stdout_detailed_ctrl(const char *name, void *arg)
...
+		       libnvme_ctrl_get_cntlid(c),
+		       libnvme_ctrl_get_serial(c),
+		       libnvme_ctrl_get_model(c),
+		       libnvme_ctrl_get_firmware(c),
+		       tr,
+		       libnvme_ctrl_get_traddr(c),
+		       libnvme_ctrl_get_phy_slot(c),
+		       libnvme_subsystem_get_name(libnvme_ctrl_get_subsystem(c)));
+#ifdef CONFIG_FABRICS
+		free(reg_owner);
```
igaw: and then we just use the cleanup helpers.

nvme-print-stdout.c
```diff
@@ -5935,11 +5957,11 @@ static void stdout_detailed_list(struct libnvme_global_ctx *ctx)
...
-	printf("%-16s %-6s %-20s %-40s %-8s %-6s %-14s %-6s %-12s %-16s\n", "Device",
-		"Cntlid", "SN", "MN", "FR", "TxPort", "Address", "Slot", "Subsystem",
-		"Namespaces");
-	printf("%-.16s %-.6s %-.20s %-.40s %-.8s %-.6s %-.14s %-.6s %-.12s %-.16s\n",
-			dash, dash, dash, dash, dash, dash, dash, dash, dash, dash);
+	printf("%-16s %-12s %-6s %-20s %-40s %-8s %-6s %-14s %-6s %-12s %-16s\n",
+		"Device", "Orchestrator", "Cntlid", "SN", "MN", "FR", "TxPort",
+		"Address", "Slot", "Subsystem", "Namespaces");
+	printf("%-.16s %-.12s %-.6s %-.20s %-.40s %-.8s %-.6s %-.14s %-.6s %-.12s %-.16s\n",
+		dash, dash, dash, dash, dash, dash, dash, dash, dash, dash, dash);
```
igaw, line 5954: I think we should also update the documentation accordingly

libnvme/src/nvme/json.c
```diff
@@ -140,9 +140,6 @@ static void json_parse_subsys(libnvme_host_t h, struct json_object *subsys_obj)
 	s = libnvme_lookup_subsystem(h, NULL, nqn);
 	if (!s)
 		return;
-	app_obj = json_object_object_get(subsys_obj, "application");
-	if (app_obj)
-		libnvme_subsystem_set_application(s, json_object_get_string(app_obj));
```
igaw, line 144: the schema should be updated accordingly.

libnvme/src/nvme/registry.c
```diff
@@ -250,19 +250,28 @@ static int write_attr_atomic(int dir_fd, const char *dir_path,
...
 	snprintf(device, sizeof(device), "nvme%d", instance);
 
+	/*
+	 * Delete any stale entry unconditionally before creating the new one.
+	 * Instance recycling means an old nvmeN/ directory is stale by
+	 * definition — any attributes left over from the previous owner (e.g.
+	 * a private attribute written via libnvmf_registry_update()) must not
+	 * leak into the new entry.  Ignore errors: ENOENT is the common case.
+	 */
+	libnvmf_registry_delete(device);
```
igaw, line 365: maybe log a warning?

fabrics.c
```diff
@@ -612,8 +609,8 @@ int fabrics_discovery(const char *desc, int argc, char **argv, bool connect)
 			libnvme_strerror(errno));
 		return -ENOMEM;
 	}
-	if (context)
-		libnvme_set_application(ctx, context);
+	if (nbft)
+		libnvme_set_owner(ctx, "nbft");
```
igaw, line 617: But isn't a user executing this here?

fabrics.c
```diff
@@ -952,17 +979,46 @@ int fabrics_disconnect_all(const char *desc, int argc, char **argv)
...
+						 * (e.g. EACCES/EIO) skip rather
+						 * than disconnect a possibly
+						 * owned controller.
+						 */
+						do_disconnect = (rc == -ENOENT);
+					free(reg_owner);
+				}
+
+				if (do_disconnect &&
+				    libnvmf_disconnect_ctrl(c))
```
igaw: This need to go into the library. The frontend is not supposed to have any logic anymore.

igaw:
the third patch removes the 'context' command line option, the documentation needs to be updated accordingly.

Documentation/nvme-disconnect-all.txt
```diff
@@ -9,30 +9,72 @@ SYNOPSIS
...
+orchestrator (e.g. nvme-stas, nbft) are left untouched.
+
+Use --owner to target a specific orchestrator's controllers, or --force
+to disconnect all controllers regardless of ownership.  --force and
+--owner are mutually exclusive.  When either flag is used interactively
+a confirmation prompt is shown; non-interactive invocations proceed
+without prompting.
+
+PCIe and apple-nvme controllers are never disconnected regardless of
+the options used.
```
igaw: This is a fabrics command as stated in the first paragraph, thus this doesn't need to be in.

Documentation/nvme-registry-delete.txt
```diff
@@ -0,0 +1,50 @@
...
+
+NAME
+----
+nvme-registry-delete - Remove a controller's ownership registry entry
+
+SYNOPSIS
+--------
+[verse]
+'nvme' [<global-options>] 'registry delete'
+			--device=<DEV> | -d <DEV>
```
igaw, line 12: `[--deivce ...]` ?

Documentation/nvme-registry-delete.txt
```diff
@@ -0,0 +1,50 @@
...
+DESCRIPTION
+-----------
+Remove the ownership registry entry for an NVMe-oF controller.  The
+registry entry is the directory /run/nvme/registry/<device>/ and all
+attribute files it contains.
+
+Under normal operation registry entries are removed automatically: a
+udev rule fires on the kernel REMOVE event and deletes the entry.  This
+command is provided for manual cleanup or for orchestrators that want to
+explicitly release ownership before disconnecting.
```
igaw, line 23: When udev is used. Don't know if this should be documented in the command documentation. It might be good to have some explanation in a README to explain how this all works.

Documentation/nvme-registry-list.txt
```diff
@@ -0,0 +1,52 @@
...
+DESCRIPTION
+-----------
+List all live entries in the NVMe-oF controller ownership registry.  The
+registry records which orchestrator (e.g. nvme-stas, nbft) owns each
+connected NVMe-oF controller.  It is stored under /run/nvme/registry/ as
+one directory per live controller containing one plain-text file per
+attribute (e.g. /run/nvme/registry/nvme3/owner).
+
+One entry is printed per live registered controller.  Controllers with no
+registry entry (unowned) and PCIe/apple-nvme controllers (which are never
```
igaw: maybe just say Non fabrics controllers or just PCIe transport based instead mentioning `apple-nvme`.

Documentation/nvme-registry-retrieve.txt
```diff
@@ -0,0 +1,60 @@
...
+
+NAME
+----
+nvme-registry-retrieve - Read an attribute from a controller's registry entry
+
+SYNOPSIS
+--------
+[verse]
+'nvme' [<global-options>] 'registry retrieve'
+			--device=<DEV> | -d <DEV>
```
igaw, line 12: `[--device ....]`

Documentation/nvme-registry-update.txt
```diff
@@ -0,0 +1,71 @@
...
+----
+nvme-registry-update - Write an attribute to a controller's registry entry
+
+SYNOPSIS
+--------
+[verse]
+'nvme' [<global-options>] 'registry update'
+			--device=<DEV> | -d <DEV>
+			--attr=<ATTR> | -a <ATTR>
+			--value=<VAL> | -V <VAL>
```
igaw, line 14: `[]`

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+		tmp[len - 1] = '\0';
+
+	for (p = tmp + 1; *p; p++) {
+		if (*p != '/')
+			continue;
+		*p = '\0';
+		if (mkdir(tmp, mode) < 0 && errno != EEXIST)
+			return -errno;
+		*p = '/';
+	}
```
hreinecke, line 101: I don't like this loop. Parsing strings is always a pain, and we should avoid it if possible.
I'd very much prefer to just do a 'stat' on the pathname, and check if it's a directory. If not, bail out.
The current method of recursively creating directories is a pain to clean up (how do you figure out where to stop)?
I'd prefer to have the caller creating the directory, which consequently also makes him responsible for cleaning up.

libnvme/src/nvme/registry.c
hreinecke: And use 'sscanf', as we _know_ how the string needs to look like.

libnvme/src/nvme/registry.c
hreinecke: isalnum() ?

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	return true;
+}
+
+/* write() may return a short count; loop until the whole buffer is written. */
+static int write_all(int fd, const char *buf, size_t len)
+{
+	while (len) {
+		ssize_t w = write(fd, buf, len);
+
```
hreinecke, line 161: Nit: write() can return '0', at which point we should terminate the loop, too.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+#endif
+
+	/* the temp file is created with 0600; open it to the registry world. */
+	fchmod(fd, 0644);
+
+	ret = write_all(fd, value, strlen(value));
+	if (ret)
+		goto err;
+	ret = write_all(fd, "\n", 1);
+	if (ret)
```
hreinecke, line 298: I really would like to get the rationale here.
Why two calls to 'write_all()'? Expand the race window?
Please reformat 'value' to have a trailing newline (which, incidnetally, will be very awkward if 'value' already _has_ a trailing newline) and use a single call to 'write_all'.
(And I wonder if 'write_all' is needed here, or whether a single call to 'write()' won't be sufficient. The whole point here is that we want to write atomically, _but_ might fail. If things fail the caller has to do a recovery action. So why not consider a short write a failure, and let the caller sort things out?)

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+		goto err;
+	if (fsync(fd) < 0) {
+		ret = -errno;
+		goto err;
+	}
+	close(fd);
```
hreinecke, line 300: Why not use 'O_SYNC' and drop the fsync()?

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	snprintf(device, sizeof(device), "nvme%d", instance);
+
+	ret = ensure_device_dir(device);
+	if (ret)
+		return ret;
+
+	if (snprintf(dir_path, sizeof(dir_path), "%s/%s",
+		     registry_dir(), device) >= (int)sizeof(dir_path))
+		return -ENAMETOOLONG;
+
```
hreinecke, line 335: ???
Why formatting the device name first, then checking the device directory (which cannot be part of the device name, rendering the whole formatting a bit pointless), and then reformat with the _actual_ pathname?
Arbitrary increasing any race windows with pathname attacks?
Don't.
Format the entire pathname, and do an 'open' on that.
No races, done.

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,498 @@
...
+	dir_fd = open(dir_path, O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+	if (dir_fd < 0)
+		return -errno;
+
+	ret = write_attr_atomic(dir_fd, dir_path, "owner", owner);
+	close(dir_fd);
```
hreinecke, line 341: And this whole 'dir_fd' swizzle feels a bit pointless.
The filesystem _really_ needs to guarantee that there are no race conditions when calling 'rename()', and all instances of 'nvme-cli' run on the _same_ filesystem.
So really, under which circumstances can there be a race condition with 'rename()'? 'fsync' etc only makes a difference if you have to worry about sudden power failures and the filesystem has to recover after that.
But if that's the case we'll be rebuilding the nvme structure, too, so I am not sure if we need to care...

hreinecke (changes requested):

fabrics.c
dwsuse: FTR, Martin and I having a discussion on this topic on the matrix channel.

Martin argues to keep this logic here as it is client specific (nvme-cli vs nvme-stas) and without having read through the RFC yet, I tend to agree with this point. 

So if we keep this here, please factor this into a function, so the it's a better readable. 

Also I would suggest to introduce a `libnvme_ctrl_is_transport_fabrice` (or better name) and get rid of the strcmp for PCI based transports.

martin-belanger: Thanks — and per our Matrix discussion (which you kindly summarized on the PR), I kept the fleet-level policy here because `disconnect-all` is itself a human-driven orchestrator, not a thin frontend. The mechanism is in libnvme: the per-controller decision is factored into a small `disconnect_all_match()` predicate, the transport test is the new `libnvme_ctrl_is_transport_fabric()` primitive (no more open-coded `strcmp`), and the disconnect itself is the one-line `libnvmf_disconnect_ctrl()`. So mechanism lives in libnvme, policy (ownership/`--force`/skip-rules) lives in the orchestrator.

nvme-print-stdout.c
martin-belanger: The lookup moved into `libnvme_ctrl_owner()` (with a no-fabrics stub returning NULL, so the `#ifdef CONFIG_FABRICS` is gone); the "kernel"/owner/"-" presentation stays in nvme-print; the retrieved string uses `__cleanup_free`.

nvme-print-stdout.c
martin-belanger: Same as previous. Moved to libnvme, with fallback implementation.

nvme-print-stdout.c
martin-belanger: Same as previous. Cleanup helpers added.

nvme-print-stdout.c
martin-belanger: Done — added the `Orchestrator` column to the `nvme list -v` example in `nvme-list.txt`.

libnvme/src/nvme/json.c
martin-belanger: Done — removed the `application` property from `config-schema.json[.in]` (the field is removed in this PR). The generated API docs (`*.rst`, `man/*.2`) still mention `set/get_application`, but those are regenerated from the headers at release, so I left them for that pass.

libnvme/src/nvme/registry.c
martin-belanger: I'd prefer to keep this. The `/tmp` confinement + `..` rejection + no-getenv-UAF caching were added at **Copilot's** request, and unlike read-only sysfs the registry path drives writes *and* a recursive delete, so confining the test-only `NVME_REGISTRY_DIR` override is deliberate. I did simplify it (fixed `buf[256]` instead of `PATH_MAX`, since it's a test-only override under `/tmp` and truncation is harmless). I also kept a fixed `static` buffer rather than `asprintf` here on purpose: the value is cached for the lifetime of the process, so an `asprintf`'d cache (which is what `make_sysfs_dir` does when its env var is set) is a permanent, never-freed allocation that valgrind reports as still-reachable — the fixed buffer avoids that.

libnvme/src/nvme/registry.c
martin-belanger: The registry is a coordination point between independent processes (discoverd, nvme-stas, `disconnect-all`, udev); they only see each other if they agree on one location, so a runtime knob invites silent divergence. `/run` is the FHS-correct home. If a distro ever needs to relocate it, a compile-time meson option is the safe lever (every component builds against the same value).

libnvme/src/nvme/registry.c
martin-belanger: Done — the transient path buffers now use `asprintf()` + `__cleanup_free`. One exception: `registry_dir()` keeps a fixed `static` buffer, because it caches for the process lifetime and an `asprintf` there can never be freed (valgrind would flag it); I shrank it from `PATH_MAX` to 256.

libnvme/src/nvme/registry.c
martin-belanger: Done — `retrieve` (and `attr_for_each`) now `fstat` + `malloc` the exact size via a shared helper; the fixed 4 KB cap (`-E2BIG`) is gone.

libnvme/src/nvme/registry.c
martin-belanger: Done — added an EINTR-safe `read_full()`.

libnvme/src/nvme/registry.c
martin-belanger: Done — factored the shared open-dir/write dance into `write_device_attr()`.

libnvme/src/nvme/registry.c
martin-belanger: Renamed to `callback`.

libnvme/src/nvme/registry.c
martin-belanger: Done, but at the caller: the registry functions are ctx-less primitives with no log handle, so the WARN is emitted on the connect path (`__nvmf_add_ctrl`), which has the ctx, when `create_instance`/`delete_instance` fails (the connection still succeeds).

libnvme/sysfiles/udev/rules.d/70-nvmf-registry.rules
martin-belanger: The rule ships in the **libnvme** package, which must never depend on nvme-cli (nvme-stas, for example, pulls in libnvme alone) — and `nvme registry` is an optional plugin, so the binary may be absent even when nvme-cli is installed. So the rule cleans up directly (guarded `[ -e /dev/%k ] || rm -rf …`).

libnvme/libnvme/tests/test-registry.py
martin-belanger: Yes — `registry_update` raises `ValueError` on `-EINVAL` and `OSError` otherwise (it's a C binding now).

libnvme/test/registry.c
martin-belanger: That line is in `cleanup_tmpdir()`, which is intentionally best-effort (documented). I did add return-code checks to the test *setup* mutations, where a silent failure would make a test pass vacuously.

meson.build
martin-belanger: Done — moved the `registry-cli` test into `tests/meson.build` so it's gated by `-Dnvme-tests`.

plugins/registry/registry-nvme.c
martin-belanger: SPDK is a userspace driver that doesn't go through libnvme, so spdk devices never get a registry entry and no orchestrator sees them. Supporting them is a device-model expansion beyond this PR; noting it as a future enhancement.

Documentation/nvme-disconnect-all.txt
martin-belanger: Removed the redundant "PCIe and apple-nvme are never disconnected" sentence.

Documentation/nvme-registry-list.txt
martin-belanger: Removed the redundant "PCIe and apple-nvme are never disconnected" sentence.

Documentation/nvme-registry-retrieve.txt
martin-belanger: `--device` is required (you have to say which controller to read), so it's intentionally not bracketed; `--attr` is optional — it defaults to `owner` — and is bracketed. That matches the convention used in the other man pages (optional arguments in `[]`).

Documentation/nvme-registry-delete.txt
martin-belanger: The current text already reads `--device` (no typo).

Documentation/nvme-registry-update.txt
martin-belanger: `--value` is required: the CLI rejects an update without it, and removing an attribute is a separate `registry delete`. So it's correctly shown without brackets, as are the equally-required `--device` and `--attr`.

Documentation/nvme-registry-delete.txt
martin-belanger: Reasonable — I'd add a short README covering the registry + udev cleanup as a follow-up rather than expand the per-command man pages.

libnvme/src/meson.build
martin-belanger: Agreed it's a nice cleanup; I'll do it as its own small PR since it's orthogonal to the registry.

martin-belanger:
> if (nbft)
> 		libnvme_set_owner(ctx, "nbft");

Done

fabrics.c
martin-belanger: Yes, a user (or the `nvmf-connect-nbft` unit) runs `connect-all --nbft` — but the owner string records the *source* of the connections, not who invoked the command. `--nbft` connects exactly the controllers described by the NBFT firmware table, i.e. firmware-defined boot volumes. Tagging them `owner=nbft` is a deliberate lifetime invariant that protects them: an orchestrator's `disconnect-all --owner <orch>` (and its registry-check skip logic) will never tear down a controller owned by `nbft`. nvme-discoverd preserves it too — it reconnects NBFT-sourced controllers with `--owner nbft`, not `--owner discoverd`, for the same reason. So `nbft` is a meaningful owner category ("firmware boot volume — hands off"), independent of the human or unit that happened to run the command.

libnvme/src/nvme/registry.c
martin-belanger: Good catch — fixed: `write() == 0` now returns `-EIO` (it would otherwise spin).

libnvme/src/nvme/registry.c
martin-belanger: I kept lazy creation (the registry dir is created on first write; there's no natural caller to own its lifecycle), but simplified the helper to a `strdup`-based walk with no fixed buffer.

libnvme/src/nvme/registry.c
martin-belanger: Kept the two writes: the value plus a deliberate trailing newline. The newline is load-bearing — it's what lets an empty value (`owner=""`) round-trip as `""`, distinct from "absent". `write_all` is the EINTR-safe primitive. (I considered `dprintf` for a single call but it adds format-parsing overhead and is gated behind `_GNU_SOURCE`, which our musl CI build undefines.)

libnvme/src/nvme/registry.c
martin-belanger: Went further and dropped the file `fsync` entirely — `/run` is ephemeral (cleared at boot), so durability isn't wanted; `rename()` gives the only guarantee we need (atomic visibility for readers). Kept the rename.

libnvme/src/nvme/registry.c
martin-belanger: Done — paths are now built once with `asprintf` and opened directly.

libnvme/src/nvme/registry.c
martin-belanger: You're right that fsync isn't about power-loss here; I dropped it. The `rename` stays, but its purpose is atomic visibility — a concurrent orchestrator reading the entry never sees a half-written value — not durability.

libnvme/src/nvme/registry.c
martin-belanger: I'd keep the explicit `nvme[0-9]+` check: `sscanf("nvme%d")` silently accepts leading whitespace and a sign (`nvme -3`), which we don't want.

libnvme/src/nvme/registry.c
martin-belanger: `isalnum()` doesn't cover `_`/`-` (our attr charset) and is locale-dependent; the explicit check is stricter and locale-independent.

libnvme/src/nvme/fabrics.c
```diff
@@ -1029,33 +1030,19 @@ static int __nvmf_add_ctrl(struct libnvme_global_ctx *ctx, const char *argstr)
...
 		if (!*p)
 			continue;
-		if (sscanf(p, "instance=%d", &ret) == 1)
-			return ret;
+		if (sscanf(p, "instance=%d", &instance) == 1) {
+			if (ctx->owner)
+				libnvmf_registry_create_instance(instance, ctx->owner);
+			else
+				libnvmf_registry_delete_instance(instance);
+			return instance;
```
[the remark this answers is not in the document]
martin-belanger: Now logged at WARN on the connect path (the connection still succeeds).

libnvme/src/nvme/registry.c
```diff
@@ -0,0 +1,517 @@
...
+		return -errno;
+	if (fcntl(fd, F_SETFD, FD_CLOEXEC) < 0) {
+		ret = -errno;
+		goto err;
+	}
+#endif
+
+	/* the temp file is created with 0600; open it to the registry world. */
+	fchmod(fd, 0644);
+
```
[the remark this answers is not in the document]
martin-belanger: Now checked (cleanup temp file + return the error).

libnvme/libnvme/nvme.i
```diff
@@ -545,13 +573,101 @@ PyObject *nbft_get(struct libnvme_global_ctx *ctx, const char *filename)
...
+	if not _valid_registry_device(device):
+		raise ValueError("invalid device name: %r" % (device,))
+	if not _valid_registry_attr(attr):
+		raise ValueError("invalid attribute name: %r" % (attr,))
+	path = os.path.join(_registry_dir, device, attr)
+	try:
+		with open(path) as f:
+			return f.read().rstrip("\n\r")
+	except OSError:
+		return None
```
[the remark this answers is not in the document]
martin-belanger: Moot — the Python reimplementation was deleted; the C `registry_retrieve` returns `None` only for `-ENOENT` and raises on real errors.

libnvme/src/nvme/lib.h
```diff
@@ -39,6 +39,23 @@ struct libnvme_global_ctx *libnvme_create_global_ctx(FILE *fp, int log_level);
...
+ * @owner:	Orchestrator identity string (e.g. "stas", "nbft").
+ *
+ * Records the orchestrator identity used when claiming registry ownership of
+ * connections made through @ctx.  May be set only once: a second call returns
+ * -EEXIST and leaves the existing identity untouched.  A process that does not
+ * participate in the registry simply never calls this.
+ *
+ * This is the supported way to record the registry owner;
+ * libnvme_create_global_ctx() deliberately takes no owner parameter.
+ *
```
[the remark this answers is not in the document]
martin-belanger: Fixed in the PR description: the final API keeps `libnvme_create_global_ctx(fp, level)` and adds `libnvme_set_owner(ctx, owner)`.

libnvme/libnvme/nvme.i
martin-belanger: Agreed — fixed. I exposed the C registry API via SWIG (`registry_retrieve`, `registry_devices`, `registry_device_attrs`) and deleted the Python-side directory walk / file reads (and the `_valid_*` / `_init_registry_dir` helpers). `registry_entries()` is now a 2-line generator over the C binders, so all file access is in libnvme. This also moots the Copilot `OSError`-vs-missing finding below.

libnvme/src/nvme/registry.c
martin-belanger: inlined ensure_registry_dir()

libnvme/src/nvme/registry.c
martin-belanger: asprintf() used with _cleanup helper

libnvme/src/nvme/registry.c
martin-belanger: Dropped

libnvme/src/nvme/registry.c
martin-belanger: Done

libnvme/src/nvme/registry.c
martin-belanger: removed fd = -1

libnvme/src/nvme/registry.c
martin-belanger: Done

libnvme/src/nvme/registry.c
martin-belanger: Done

libnvme/src/nvme/registry.c
martin-belanger: Done, but at the caller: the registry functions are ctx-less primitives with no log handle, so the WARN is emitted on the connect path (`__nvmf_add_ctrl`), which has the ctx, when `create_instance`/`delete_instance` fails (the connection still succeeds).

martin-belanger:
> > To keep the immutability I was after, I'll add a one-shot setter instead:
> > ```c
> > int libnvme_set_owner(struct libnvme_global_ctx *ctx, const char *owner);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It can only be set once: a second call returns an error and leaves the existing value untouched. That keeps "who connected this" stable for the lifetime of the context without piling another argument onto the central object. Callers that don't participate in the registry simply never call it, so the NULL-at-every-call-site churn goes away.
> 
> In this case you should also have some check in `libnvmf_registry_update` which would prevent the user to update the `owner`, or do I miss something?
> 
> Anyway, I don't have a strong opinion if we really need this 'safety net', the design is based on trusting all parties to behave correctly. So maybe we just leave it with the check in `libnvme_set_owner`.

I decided that the owner immutability is a policy decision that belongs outside libnvme.

martin-belanger:
> The 2nd commit message says:
> 
> ```
> nvme: add registry plugin and orchestrator column
> The registry commands belong in a plugin rather than as top-level
> built-ins to keep the core nvme binary lean and to make it clear
> these are orchestration tools, not general NVMe commands.
> ```
> 
> which is strange: we discussed this but the new code was never on top level. doesn't really make sense IMO.
> 
> The third commit message seems to have same problem:
> 
> ```
> libnvme_create_global_ctx() gains an owner parameter so the identity
> is set at construction time and cannot change during the context
> lifetime.  All callers that do not participate in the registry pass
> NULL.  connect-all --nbft passes "nbft", registering boot-volume
> controllers so they are protected from accidental disconnection.
> ```

I updated the commit messages.

martin-belanger:
Thanks @igaw and @hreinecke for the thorough review — it's a noticeably better PR for it.

I've gone through every comment: each is either fixed, or I've replied inline with the rationale where I felt a change wasn't warranted. The highlights of what changed:

- **All registry logic now lives in libnvme.** `registry_retrieve`/`registry_entries` are no longer reimplemented in Python — the bindings call the C API (`registry_devices`/`registry_device_attrs`), so the careful file access is entirely in `registry.c`.
- **`disconnect-all`** keeps its fleet-level *policy* (it's a human-driven orchestrator), but the *mechanism* moved into libnvme: a `disconnect_all_match()` predicate plus a new `libnvme_ctrl_is_transport_fabric()` primitive in place of open-coded `strcmp()`.
- **The `nvme list -v` Orchestrator column** lookup moved into `libnvme_ctrl_owner()` with a no-fabrics fallback, so the `#ifdef CONFIG_FABRICS` is gone.
- **Owner API:** reverted the `libnvme_create_global_ctx()` signature change — it keeps its original two-argument form, and ownership is set via `libnvme_set_owner()` after creation.
- **`application` removal** is cleaned up end-to-end, including the now-dead `connect --context` option and the autoconnect systemd units that still passed it.
- Plus the smaller items: EINTR-safe reads, `asprintf` paths, `fchmod` / `write()==0` handling, WARN logs on registry write failures, `fstat`-sized reads, the JSON config schema cleanup, the docs, and the CLI test moved under `-Dnvme-tests`.

A few things I deliberately kept, each with the reasoning inline on its thread: the `NVME_REGISTRY_DIR` `/tmp` confinement (it guards a write + recursive-delete path, unlike read-only sysfs), the hardcoded `/run/nvme/registry` location (a coordination point all orchestrators must agree on), the udev rule cleaning up directly (it ships in the libnvme package, which can't depend on nvme-cli), and the explicit device/attr validators.

The branch has been rebased onto current master (resolving the `tree.c` conflict from the `libnvme_ctrl_alloc` refactor that landed meanwhile), every commit builds standalone, and CI is green.

A few orthogonal items I've noted as separate follow-ups rather than growing this PR: a shared `write_all` helper, the `fabrics-includes.h.in` build cleanup, SPDK device support, and a registry/udev README.

I believe this is ready for final review or merge. Thanks again!

igaw:
Okay, let's go with this version.

A couple remarks/thoughts:
- the udev rule is not setting the owner. I understand we are going to address this with the next step, nvme-discoverd.
- should we add the global context to registry API?

igaw:
Thanks a lot!

PR #3426, opened 2026-06-04
libnvme: add CIU and CIRN to Identify Controller data structure

MaisenbacherD:
@igaw As we discussed at Lund Linux Conference, I took the spec and cross-referenced it with the code base to an LLM. It found some discrepancies. Here is the first one. Please let me know if this is useful, I will then continue sending the next ones in batches after manually checking the mismatches. :)

Commit description :

The I/O Command Set Independent Identify Controller data structure (Figure 328 in NVM Express Base Specification 2.3) defines two fields that were previously hidden inside the rsvd135[118] reserved block:

  - byte 135    Controller Instance Uniquifier (CIU)
  - bytes 143:136 Controller Instance Random Number (CIRN)

Split rsvd135[118] into ciu, cirn[8], and the two remaining reserved ranges.

Assisted-by: Claude:claude-opus-4-8 [devin]

### DIFF
--- a/libnvme/src/nvme/nvme-types-base.h
+++ b/libnvme/src/nvme/nvme-types-base.h
@@ -1298,7 +1298,10 @@ struct nvme_id_psd {
  *	       field is 3
  * @crcap:     Controller Reachability Capabilities (CRCAP), see
  *	       &enum nvme_id_ctrl_crcap
- * @rsvd135:   Reserved
+ * @ciu:       Controller Instance Uniquifier (CIU)
+ * @cirn:      Controller Instance Random Number (CIRN)
+ * @rsvd144:   Reserved
+ * @rsvd240:   Reserved for the NVMe Management Interface specification
  * @nvmsr:     NVM Subsystem Report, see &enum nvme_id_ctrl_nvmsr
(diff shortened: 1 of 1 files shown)


libnvme/doc/man/nvme_id_ctrl.2
```diff
@@ -58,7 +58,13 @@ struct nvme_id_ctrl {
...
 .BI "    __u8 crcap;"
 .br
-.BI "    __u8 rsvd135[118];"
+.BI "    __u8 ciu;"
+.br
+.BI "    __u8 cirn[8];"
+.br
+.BI "    __u8 rsvd144[96];"
+.br
+.BI "    __u8 rsvd240[13];"
```
igaw: No need to update the generated documentation. This will be updated on release.

igaw:
Yes, this is good stuff. Keep it coming!

And I think we are not going to do the `Assisted-by:` after all. I don't think it adds anything. You need to own your code, which tool you use it's up to you.

libnvme/doc/man/nvme_id_ctrl.2
MaisenbacherD: Right, forgot about that. Removed this change :)

igaw:
Thanks!

PR #3428, opened 2026-06-04
nvme-print: Fix nvme_bar_cap cps field alignment

brgoing:
Fixes field alignments of the nvme_bar_cap struct by making cps part of the __u16 that was missing those two bits. This change fixes warnings regarding the struct size encountered when compiling on Windows.


igaw:
Thanks!

PR #3429, opened 2026-06-05
nvme: delete deprecated human-readable option

ikegami-t:
Since the option is duplicated with the verbose option.

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -74,7 +74,6 @@ struct feat_cfg {
 	__u8 uuid_index;
 	__u32 data_len;
 	bool raw_binary;
-	bool human_readable;
 	bool changed;
 };
 
@@ -105,7 +104,6 @@ struct passthru_config {
 
 struct get_reg_config {
 	int offset;
-	bool human_readable;
 	bool cap;
 	bool vs;
 	bool intms;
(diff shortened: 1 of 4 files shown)


ikegami-t:
If really needed to remain the human-readable option we can add the option as a global option I think.

igaw:
Getting rid of this makes sense but we should have a fallback in place. From the past experience, any command line options change causes some regressions. 

Though we drop all information about it but nvme-cli still accepts it. So what about introducing an alias for `verbose` on global level? Obviously not mentioning it :)

ikegami-t:
Noted. I will do consider more to add an alias for verbose on global level. Thank you.

ikegami-t:
As suggested just added an alias for verbose on global level. Thank you.

igaw:
This should work even for command which use `-H` already. 

Maybe we should introduce somethin like
```
#define OPT_FLAG_HIDDEN(l, s, v, d, ...) \
	{l, s, NULL, CFG_FLAG_HIDDEN, v, no_argument, d, false, __VA_ARGS__}
```

or modife the argconfig so it doesn't print the option anymore. 

BTW, we could also remove the `seen` from `struct argconfig_commandline_options`

ikegami-t:
Added the argconfig `hidden` parameter instead of the `CFG_FLAG_HIDDEN`. About the argconfig `seen` it is still used by the function `argconfig_parse_seen()` to check if the command option is input or not. So we can not remove it I think.

(**Added**)

> This should work even for command which use `-H` already.

The hidden option added as the last entry then the existing `-H` short option should be worked correctly as same as current.

igaw:
Thanks!

igaw:
We should also remove it from the documentation as next.

ikegami-t:
> Since the option is duplicated with the verbose option.

Yes right then just created the PR: #3461. Thank you.

PR #3430, opened 2026-06-05
File and output stream line ending compatibility changes

brgoing:
Makes Windows output behavior match other platforms by using O_BINARY to prevent text-mode line ending translation (LF to CRLF) that can corrupt binary output.

* Introduces O_NVME_WRONLY, which includes O_BINARY when O_BINARY is defined, and updates file-open calls to use it.
* Configures stdout and stderr to use O_BINARY on Windows

### DIFF
--- a/common.h
+++ b/common.h
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: GPL-2.0-or-later */
 #pragma once
 
+#include <fcntl.h>
 #include <string.h>
 #include <stdbool.h>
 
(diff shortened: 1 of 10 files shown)


igaw:
I don't like this approach with overloading of the flags flags. The resulting code doesn't read well.

What about introducing something like `nvme_open_binary/nvme_open_rawfile/nvme_open_datafile`  (or any better name) which handles this internally.

brgoing:
> I don't like this approach with overloading of the flags flags. The resulting code doesn't read well.
> 
> What about introducing something like `nvme_open_binary/nvme_open_rawfile/nvme_open_datafile` (or any better name) which handles this internally.

Okay.  Thank you for the feedback.  We discussed multiple approaches to this, including defining O_BINARY as 0 on Linux and including it in the flags used, introducing a new nvme_open method like you mentioned, or the modified O_WRONLY approach we went with in this PR.  Each had its own drawbacks, so we figured we needed to pick one and go with it to start the conversation.

igaw:
Indeed, all have their drawbacks :) 

The new version looks more sane to me. Anyway, this isn't carved into stone, so if we find a better way, it's not a problem to change this later.

igaw:
Thanks!

igaw:
And just remember we did the same thing in libnvme with `libnvme_open`. So this kind of matches.

Issue #3433, opened 2026-06-08
libnvme/nvme: add NVMe controller ownership registry


martin-belanger:
## Summary

This issue tracks the design and implementation of an NVMe controller ownership registry — a cooperative coordination mechanism that allows NVMe-oF orchestrators to declare ownership of controllers and respect each other's boundaries.

## Motivation

Triggered by #2913 (Prevent NVMe-oF boot devices from being removed by disconnect-all). The root cause identified in that discussion is broader than `disconnect-all` alone: there is no mechanism for tools and daemons operating on the same host to know which controllers are managed by a running orchestrator. This leads to accidental disconnection of controllers that `nvme-stas`, NBFT, or other orchestrators depend on.

## Approach

A lightweight filesystem-based registry at `/run/nvme/registry/`. Each connected controller gets a directory named after its kernel device (`nvmeX/`) containing plain-text attribute files — primarily `owner`. Registry entries are written automatically by libnvme on connect when the caller provides an owner identity. Cleanup is handled by a udev rule on `KOBJ_REMOVE`.

`disconnect-all` uses the registry to skip owned controllers by default, with `--force` to override.

## Tracking

- Design document: `nvme-registry-guide.md` (shared separately as a DO-NOT-MERGE PR so that review comments can be attached directly to specific sections)
- Implementation: PR #3425 (branch `registry-v2`)

---

@igaw @mwilck @tbzatek — creating this issue to better track the ownership registry as a feature. I will submit a DO-NOT-MERGE PR with the design document shortly. The document is in Markdown format; it reads best in Typora or any Markdown-aware editor.

martin-belanger:
@mwilck @tbzatek — an updated version of the design document incorporating your review comments has been posted as an RFC PR for continued discussion: https://github.com/linux-nvme/nvme-cli/pull/3434

martin-belanger:
Closing in favor of a new GitHub Discussion. The scope has expanded beyond the ownership registry to cover nvme-discoverd and orchestrator coexistence; a Discussion is a better home for that broader conversation.

PR #3434, opened 2026-06-08
DO NOT MERGE: RFC: add ownership registry design documents

martin-belanger:
> **RFC — posted for design review and community feedback. Not intended for merge.**

One document is proposed for review, and a second is included for context:

- **nvme-registry-guide.md**: design guide for the NVMe controller ownership registry — a cooperative, filesystem-based mechanism that allows NVMe-oF orchestrators to declare ownership of controllers and avoid accidental disconnection of controllers managed by another orchestrator.
- **nvme-orchestrator-coexistence-summary.md**: companion design notes covering discovery policy coordination between nvme-discoverd and nvme-stas for environments where both daemons run simultaneously. Included for context — this is early-stage thinking, not a finished proposal.

Please use inline comments to discuss specific sections.

Related issue: https://github.com/linux-nvme/nvme-cli/issues/3433
Implementation PR: #3425

These documents are in Markdown format; it reads best in Typora or any Markdown-aware editor.

### DIFF
--- a/RFCs/nvme-orchestrator-coexistence-summary.md
+++ b/RFCs/nvme-orchestrator-coexistence-summary.md
@@ -1,0 +1,142 @@
+# NVMe Orchestrator Coexistence – Discussion Summary
+
+## Key Observation
+
+The NVMe Ownership Registry and orchestrator coexistence address two different problems:
+
+1. **Ownership coordination** – Who currently manages an existing controller?
+2. **Discovery policy coordination** – Which controllers should be connected in the first place?
+
+The ownership registry solves the first problem. Fabric zoning conflicts primarily arise from the second.
+
+---
+
+## Ownership Registry Remains Valuable
+
+The ownership registry provides a lightweight, cooperative mechanism for orchestrators to:
+
+- Declare ownership of connected controllers.
+- Avoid accidental disconnects of controllers managed by another orchestrator.
+- Support ownership transfer when a higher-authority orchestrator (e.g. STAS) determines it should manage a controller.
+- Enable ownership-aware `disconnect-all` behavior.
+
+The registry is intentionally advisory rather than enforceable, which is appropriate because all orchestrators run with root privileges.
+
+---
+
+## Systemd-Based Protection
+
+If running `nvme-discoverd` and `nvme-stas` simultaneously proves harmful, a useful safeguard is:
+
+```ini
+[Unit]
+Conflicts=nvme-discoverd.service
+```
+
+This provides runtime mutual exclusion as a safety net.
+
+However, `Conflicts=` alone does not establish priority. It only prevents both services from running at the same time.
+
+---
+
+## Discovery Policy Requires a Separate Mechanism
+
+The ownership registry cannot prevent reconnect loops because ownership is attached to a kernel controller instance (`nvmeX`), which disappears when the controller is disconnected.
+
+A reconnect creates a new controller instance and a new registry entry.
+
+Therefore, discovery-policy decisions must be expressed using stable identifiers such as:
+
+- Subsystem NQN
+- Transport type
+- Target address
+- Host interface
+- Other discovery-log attributes
+
+rather than kernel device names.
+
+---
+
+## Dynamic Exclusion Mechanism for discoverd
+
+> **Note:** This mechanism is not planned for immediate implementation. It becomes relevant when mDNS discovery support is added to nvme-discoverd — that is the point at which nvme-discoverd and nvme-stas can discover the same controllers independently, creating the conditions for serious connect/disconnect conflicts. The design is documented here to inform that future work.
+
+A promising approach is for discoverd to support runtime exclusion rules.
+
+Example:
+
+```
+/run/nvme/discoverd-exclusions/
+    stas.conf
+    admin.conf
+```
+
+discoverd would:
+
+1. Monitor the directory using inotify.
+2. Rebuild its effective exclusion set whenever files change.
+3. Apply exclusions to future discovery events only.
+
+**Design decision: exclusions are forward-only.** When an exclusion file is written or updated, discoverd stops connecting to matching controllers in the future but does not disconnect controllers it has already connected. Disconnecting existing controllers is the owner's responsibility — discoverd's role is connecting, not disconnecting things it did not connect. This keeps responsibility boundaries clean and avoids the same connect/disconnect bounce loop in reverse.
+
+### Exclusion File Format
+
+**Design decision: use the nvme-stas key=value format.**
+
+Each exclusion entry uses the same semicolon-separated key=value syntax already defined by nvme-stas:
+
+```
+exclude = transport=tcp;traddr=fe80::2c6e:dee7:857:26bb
+exclude = host-iface=enp0s8
+exclude = nqn=nqn.2014-08.org.nvmexpress:uuid:some-subsystem
+```
+
+A **minimal-match** approach is used: only the fields present in the entry are checked. Specifying `host-iface=enp0s8` alone excludes all controllers reachable on that interface, regardless of transport or address.
+
+Rationale for this choice:
+- nvme-stas is the primary producer of exclusion files — it can write entries in its native format with no transformation
+- libsystemd's conf-parser is internal to systemd and not part of the public API, so discoverd cannot link against it directly; however, it could serve as the basis for discoverd's own parser — provided the systemd source code is lifted with its copyright statements intact (systemd is LGPL-2.1-or-later, which is compatible with libnvme's license)
+- `sd-json` was only made public in systemd 257 (2025-09-17) and is too new to depend on for broad distro compatibility
+- The format is simple enough that discoverd needs only a small custom parser
+
+---
+
+## Advantages of Runtime Exclusions
+
+### Ephemeral State
+
+Using `/run` ensures exclusions automatically disappear on reboot and do not become stale.
+
+### Clear Ownership Boundaries
+
+- STAS owns its exclusion files.
+- discoverd owns its configuration.
+- No component edits another component's configuration files.
+
+### Extensible Design
+
+Multiple policy producers can coexist:
+
+- STAS
+- Administrative tools
+- Future orchestration components
+
+discoverd simply consumes the resulting exclusion set.
+
+### Fabric-Zoning Friendly
+
+STAS can translate CDC/fabric-zoning decisions into exclusion rules that discoverd respects, preventing connect/disconnect loops before they occur.
+
+---
+
+## Recommended Architecture
+
+The mechanisms are complementary:
+
+| Mechanism | Purpose |
+|------------|---------|
+| Ownership Registry | Coordinate ownership of existing controllers |
+| Runtime Exclusions | Prevent unwanted future connections |
+| systemd Conflicts | Provide operational safety and mutual exclusion |
+
+Together these provide a clean separation between ownership coordination, discovery policy coordination, and service management.
--- a/RFCs/nvme-registry-guide.md
+++ b/RFCs/nvme-registry-guide.md
@@ -1,0 +1,315 @@
+# NVMe Controller Ownership Registry
+
+## 1. Problem Statement
+
+Multiple independent NVMe-oF orchestrators can establish controller connections on the same Linux host simultaneously:
+
+- Manual `nvme connect` invocations (stateless, one-shot)
+- `nvme-stas` (stateful daemon — `stafd` / `stacd`)
+- NBFT firmware boot connections
+- `nvme-discoverd` and similar tools
+- UDisks (stateful daemon), libblockdev (stateless library)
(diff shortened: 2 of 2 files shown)


RFCs/nvme-orchestrator-coexistence-summary.md
```diff
@@ -0,0 +1,142 @@
...
+
+```
+/run/nvme/discoverd-exclusions/
+    stas.conf
+    admin.conf
+```
+
+discoverd would:
+
+1. Monitor the directory using inotify.
```
tbzatek, line 76: inotify is slow (but still useful!), I wonder if it makes sense to add another simple mechanism on top of it to briefly check if anything has changed in the monitored directory, before deciding on making new connection - something like timestamp comparison of the directory itself (granted that it works on tmpfs)? Trying to minimize race condition surface here...

RFCs/nvme-orchestrator-coexistence-summary.md
```diff
@@ -0,0 +1,142 @@
...
+exclude = host-iface=enp0s8
+exclude = nqn=nqn.2014-08.org.nvmexpress:uuid:some-subsystem
+```
+
+A **minimal-match** approach is used: only the fields present in the entry are checked. Specifying `host-iface=enp0s8` alone excludes all controllers reachable on that interface, regardless of transport or address.
+
+Rationale for this choice:
+- nvme-stas is the primary producer of exclusion files — it can write entries in its native format with no transformation
+- libsystemd's conf-parser is internal to systemd and not part of the public API, so discoverd cannot link against it directly; however, it could serve as the basis for discoverd's own parser — provided the systemd source code is lifted with its copyright statements intact (systemd is LGPL-2.1-or-later, which is compatible with libnvme's license)
+- `sd-json` was only made public in systemd 257 (2025-09-17) and is too new to depend on for broad distro compatibility
```
tbzatek, line 99: No need to make excuses, besides general design documents like this should be init system agnostic.

martin-belanger: Actually, @igaw specifically mandated libsystemd as a core dependency — sd_event loop, libsystemd APIs throughout. The reason is concrete: nvme-discoverd creates one systemd transient unit per NVMe controller, and that per-unit model is what makes correct shutdown ordering possible.

Because each unit knows its transport type at creation time, nvme-discoverd can add After=network.target to TCP and RDMA units and omit it for FC. systemd reverses After= at shutdown, so TCP/RDMA controllers are torn down before the network goes away — which is exactly the problem documented in issue #3309. That fix is not achievable with a single nvme connect-all invocation because it mixes all transport types in one shot.

So the libsystemd dependency isn't incidental — it's load-bearing. The parser note was a separate tangential point; happy to drop it from the design doc if it reads as noise.

RFCs/nvme-registry-guide.md
```diff
@@ -0,0 +1,315 @@
...
+- **nvme-stas** — zeroconf/ZTP via DNS-SD; controllers are discovered dynamically
+- **nvme-discoverd** — will read the JSON config file to set up initial bootup connections, so the `application` filtering would apply here — but only for the JSON-driven part of its workflow and not any potential mDNS discovered controllers.
+
+The `owner` registry works across all orchestrators through a shared runtime path (`/run/nvme/registry/`) without any shared config file. `owner` supersedes the `application` field, which is being removed in the v3.0 major version cleanup.
+
+---
+
+## 6. API Reference
+
+All registry functions are available when libnvme is built with fabrics support and are declared in `<libnvme.h>`. Registry entry creation is handled automatically by the connect path when `ctx->owner` is non-NULL. The internal function `libnvmf_registry_create()` is not part of the public API and should not be called directly (and therefore left out of the following list).
```
tbzatek, line 157: One important implementation TODO: sanitize `attr` and exclude path delimiters + reserved characters (`/`, perhaps `~`) as these will appear in the created filenames and we don't want the udev remove rule to delete anything outside.

RFCs/nvme-registry-guide.md
```diff
@@ -0,0 +1,315 @@
...
+
+**Kernel reconnect behavior.** During a reconnect within `ctrl-loss-tmo`, the kernel reuses the same `struct nvme_ctrl` — same instance number, same device name, no `KOBJ_REMOVE` event. The registry entry is untouched throughout. Only when `ctrl-loss-tmo` expires and the kernel calls `nvme_delete_ctrl()` does the `KOBJ_REMOVE` event fire and the entry get deleted. If an orchestrator later establishes a new connection, it creates a fresh registry entry under the new device name.
+
+**Boot device protection.** Controllers connected during the initramfs stage are automatically protected. The dracut nvmf module invokes `nvme connect-all --nbft`, which goes through libnvme and writes registry entries with `owner=nbft`. `switch_root(8)` explicitly moves `/run` to the new root — *"switch_root moves already mounted /proc, /dev, /sys and /run to newroot"* ([switch_root(8), util-linux](https://man7.org/linux/man-pages/man8/switch_root.8.html)) — so these entries are present in the runtime system with no special handling. No dracut changes are needed beyond using the updated nvme-cli in the initramfs. Boot devices are a first-class case, not a special one.
```
tbzatek, line 120: The dracut nvmf module actually handles various transports, not just NBFT - see the top of https://github.com/dracut-ng/dracut/blob/main/modules.d/74nvmf/parse-nvmf-boot-connections.sh and also notice the `fc,auto` discovery.

We need to update the module to explicitly provide context to any `nvme connect-all` calls there, incl. the `nvme connect-all` triggered from the custom udev rule: https://github.com/dracut-ng/dracut/blob/main/modules.d/74nvmf/95-nvmf-initqueue.rules

(that said, since this will be a new argument, we'll likely need to check `nvme` command version beforehand to be backwards-compatible...)

RFCs/nvme-orchestrator-coexistence-summary.md
martin-belanger: The slowness of inotify delivery is actually desirable here. When the first inotify event fires, we start a soak timer (e.g. 1–2 s). If additional inotify events arrive before the timer expires — the administrator is mid-edit, or `nvme exclusion add` is called several times in a script — the timer resets. Only once changes have stopped arriving for the full soak interval do we perform the reconciliation pass. This means we always act on a settled, consistent view of the exclusion set rather than reacting to each intermediate write.

A few additional properties that keep the race window tight:

**Atomic writes.** Every `nvme exclusion` write uses a temp-file + `rename()` protocol. inotify fires on the `rename()`, which is the commit point — we never observe a partial file. So the inotify event and the complete file are always in sync.

**Check at connect time, not at inotify time.** The exclusion check happens at the moment a connection is about to be made, not when inotify fires. The sequence is: `inotify → soak → rebuild in-memory exclusion set → reconcile (connect/skip)`. By the time we act, the in-memory set reflects the fully settled state of the directory.

**IN_Q_OVERFLOW safety net.** Under extreme load the kernel can overflow the inotify event queue and deliver `IN_Q_OVERFLOW`. We handle this the same way as startup: trigger a full directory re-scan and rebuild the exclusion set from scratch. This covers the missed-event scenario without needing a separate timestamp poll on every connection attempt.

Regarding the directory mtime approach: the exclusion list lives at `/etc/nvme/exclusions/` on persistent storage (not tmpfs), so mtime would be reliable there. That said, we believe the race surface is already small enough with atomic writes + soak timer + overflow handling, without adding a per-connection mtime check. Happy to reconsider if there is a specific scenario you have in mind that the above does not cover.

RFCs/nvme-registry-guide.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Single `nvme disconnect`
+
+`nvme disconnect <device>` targets a specific controller by name. Because the caller has made an explicit, deliberate choice, no guardrails are needed — the command always disconnects regardless of ownership, with no `--force` flag required or provided.
+
+This asymmetry is intentional: `disconnect-all` is a convenience command with broad scope and therefore carries ownership-aware defaults; `disconnect` is a precise, targeted operation where the caller's intent is unambiguous.
+
+### NBFT connect
+
+`nvme connect-all --nbft` automatically registers all controllers it connects as `owner=nbft`, protecting boot volumes from accidental disconnection.
```
tbzatek, line 251: Well, for the dracut context noted above, we'll need to add something like `--owner` for `nvme connect-all`. Since the dracut nvmf module only calls `nvme connect-all` as it creates `/etc/nvme/discovery.conf` from kernel cmdline arguments, there isn't currently any call to `nvme connect` (for now).

RFCs/nvme-registry-guide.md
martin-belanger: Good catch. The concern is about the `attr` parameter being used as a filename component without sanitization. If `attr` contains `/`, a caller could inadvertently write outside the device's registry subdirectory:

```c
libnvmf_registry_update("nvme3", "../../tmp/evil", "val");
// resolves to: /run/nvme/registry/nvme3/../../tmp/evil = /run/nvme/tmp/evil
```

The `~` case is subtler — a file literally named `~` in the registry directory could be misinterpreted by shell tooling that reads registry entries.

Two fixes:

1. **Sanitize `attr` in the API** — validate to `[a-zA-Z0-9_-]+` and reject anything else (in particular `/`, `~`, `.`, NUL). This is the primary fix.

2. **Tighten the udev rule** — add `KERNEL=="nvme[0-9]*"` as an explicit match condition. In practice the udev rule is already safe: `%k` is the kernel-assigned device name and for `SUBSYSTEM=="nvme"` the kernel always produces `nvme[0-9]+`, so path traversal via `%k` is not possible. But making the constraint explicit in the rule self-documents the intent and provides defense-in-depth:

```
ACTION=="remove", SUBSYSTEM=="nvme", KERNEL=="nvme[0-9]*", \
    RUN+="/bin/rm -rf /run/nvme/registry/%k"
```

I will add both to the implementation TODO list in the registry design doc.

RFCs/nvme-registry-guide.md
martin-belanger: You're right that the 74nvmf dracut module handles more than just NBFT — it also triggers FC discovery via `nvme_discovery` and connects controllers through the udev rule in `95-nvmf-initqueue.rules`. Those connections are made via `nvme connect-all` with no `--owner`, so they land in the registry as unowned.

I don't think this needs to change. We're already addressing the issue with the new nvme-discoverd daemon that starts after the switch_root takes place.

**There is no race window between dracut and nvme-discoverd.** The unowned period spans exactly the time between dracut's `connect-all` and nvme-discoverd's startup scan after `switch_root`. During `switch_root` itself, no user-space orchestrator is running. Nothing can disconnect unowned controllers in that window.

**After `switch_root`, nvme-discoverd wins the startup race.** nvme-discoverd is written in C and starts early in the boot sequence. At startup it scans sysfs for existing controllers and immediately claims ownership via `libnvmf_registry_update` for any controller it recognises. By the time nvme-stas (Python, slower to start) has initialised, nvme-discoverd already holds the registry entries. nvme-stas sees owned controllers and leaves them alone.

**nvme-stas does not manage FC connections by default.** Unless FC controllers are explicitly listed in `stafd.conf` or `stacd.conf`, nvme-stas never attempts to connect to or disconnect them. There is no default FC policy in stas that could accidentally disconnect the dracut-created FC connections.

**Adding `--owner` to dracut's `connect-all` would be counterproductive.** The design contract for plain `connect-all` is that connections are unowned so `disconnect-all` can always undo them. Making dracut's connections permanently owned (e.g. `owner=dracut`) would prevent `disconnect-all` from cleaning them up, and would require version-checking in dracut for backwards compatibility — complexity with no real benefit given the arguments above.

**The NBFT case is already handled correctly.** `nvme connect-all --nbft` registers `owner=nbft`, protecting firmware boot volumes from `disconnect-all`. That path is unchanged.

In summary: dracut creates unowned connections, nvme-discoverd claims them immediately after `switch_root` before any other orchestrator can act, and nvme-stas never touches FC connections by default. No dracut changes are needed.

RFCs/nvme-registry-guide.md
martin-belanger: Please refer to my previous answer on this topic. To summarise: the unowned window between dracut's `connect-all` and nvme-discoverd's startup scan is safe in practice — no orchestrator runs during `switch_root`, nvme-discoverd (C) claims ownership before nvme-stas (Python) is up, and nvme-stas has no default policy for FC connections anyway.

Beyond that, adding `--owner` to dracut's `nvme connect-all` would introduce a versioned dependency between dracut and nvme-cli that I'd like to avoid. The `--owner` option would only be available from nvme-cli 3.0 onwards, so dracut would need to version-check the `nvme` binary before using it — adding complexity to dracut for a problem that doesn't need solving.

martin-belanger:
Closing since the review is now complete and a new version of the doc is ready. 

The new doc will be part of a larger discussion about the ownership registry, nvme-discoverd, and orchestration coordination.

PR #3438, opened 2026-06-09
nvme: add support for sanitize wait option

ikegami-t:
The option is for waiting sanitize completion.


ikegami-t:
Hi, Just pushed the SANICAP verifiation changes also **and** some fixes for the ~~repaat~~ **repeat** option. Thank you.

igaw:
Alright, looks good. Let's merge it and see what explodes. Thanks!

nvme.c
```diff
@@ -5729,11 +5902,17 @@ static int sanitize_cmd(int argc, char **argv, struct command *acmd, struct plug
...
+	do {
+		err = libnvme_exec_admin_passthru(hdl, &cmd);
+		if (err) {
+			nvme_show_err(err, "sanitize");
+			return err;
+		}
+
+		if (cfg.wait)
+			err = wait_sanitize(hdl);
+	} while (--cfg.repeat && !err && is_sanitized(hdl));
```
maurizio-lombardi, line 5915: I am not sure I understand this point here. Supposing cfg.wait is zero, wouldn't this function return success even if is_sanitized() returns an error?

ikegami-t: Just fixed by the PR #3667. Thank you.

igaw: Thanks for reviewing and fixing it. A maintainers' dream :)

PR #3439, opened 2026-06-09
top: Make top a conditional feature and disable by default on Windows

brgoing:
Adds a 'top' configuration option that is enabled by default, and makes compilation of nvme-top code conditional on the feature enablement. Disables the top feature by default on Windows due to compatibility issues.


igaw:
Thanks!

PR #3440, opened 2026-06-09
tree: add windows tree support

bcapener:
Add tree.c and lib.c support. Adds ctrl-map.c for mapping controller name, nvmeX, to controller path, since windows controller paths might not fit within the d_name field of the dirent struct. The ctrl-map also creates a subsystem name to replicate linux's nvme-subsysX naming.

### DIFF
--- a/libnvme/src/nvme/ctrl-map.c
+++ b/libnvme/src/nvme/ctrl-map.c
@@ -1,0 +1,832 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ * Copyright (c) 2026 Micron Technology, Inc.
+ *
+ * Authors: Brandon Capener <bcapener@micron.com>
+ */
+#include <ctype.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wchar.h>
+
+#include <windows.h>
+#include <cfgmgr32.h>
+#include <setupapi.h>
+#include <winioctl.h>
+#include <ntddscsi.h>
+
+#include <libnvme.h>
+
+#include "private.h"
+#include "private-ctrl-map.h"
+#include "compiler-attributes.h"
+
+#include <initguid.h>
+#include <devpkey.h>
+
+struct ctrl_map_entry {
+	char *ctrl_name;   /* format "nvmeX" */
+	WCHAR *ctrl_path;  /* format "\\\\?\\pci..." */
+	struct nvme_id_ctrl id_ctrl;
+	int subsys_index;
+	char *subsys_name;
+};
+
+static struct {
+	struct ctrl_map_entry *entries;
+	size_t count;
+	size_t capacity;
+	bool initialized;
+} ctrl_map;
+
+size_t libnvme_ctrl_map_get_count(void)
+{
+	return ctrl_map.count;
+}
+
+const char *libnvme_ctrl_map_get_name(size_t index)
+{
+	if (index >= ctrl_map.count)
+		return NULL;
+	return ctrl_map.entries[index].ctrl_name;
+}
+
+void libnvme_ctrl_map_clear(void)
+{
+	size_t i;
+
+	for (i = 0; i < ctrl_map.count; i++) {
+		free(ctrl_map.entries[i].ctrl_name);
+		free(ctrl_map.entries[i].ctrl_path);
+		free(ctrl_map.entries[i].subsys_name);
+	}
+
+	free(ctrl_map.entries);
+	ctrl_map.entries = NULL;
+	ctrl_map.count = 0;
+	ctrl_map.capacity = 0;
+	ctrl_map.initialized = false;
+}
+
+static int find_or_create_subsys_index(const struct nvme_id_ctrl *id_ctrl)
+{
+	int max_subsys_index = -1;
+	size_t i;
+
+	for (i = 0; i < ctrl_map.count; i++) {
+		if (!memcmp(ctrl_map.entries[i].id_ctrl.subnqn,
+			    id_ctrl->subnqn, NVME_NQN_LENGTH))
+			return ctrl_map.entries[i].subsys_index;
+		if (ctrl_map.entries[i].subsys_index > max_subsys_index)
+			max_subsys_index = ctrl_map.entries[i].subsys_index;
+	}
+
+	return max_subsys_index + 1;
+}
+
+static int libnvme_ctrl_map_add(DWORD ctrl_index,
+				const WCHAR *ctrl_path,
+				const struct nvme_id_ctrl *id_ctrl)
+{
+	char *ctrl_name_copy;
+	WCHAR *ctrl_path_copy;
+	int subsys_index;
+	char *subsys_name;
+
+	if (!ctrl_path)
+		return -EINVAL;
+
+	if (ctrl_map.count == ctrl_map.capacity) {
+		size_t new_capacity = ctrl_map.capacity ?
+			ctrl_map.capacity * 2 : 8;
+		struct ctrl_map_entry *new_map;
+
+		new_map = realloc(ctrl_map.entries,
+				  new_capacity * sizeof(*ctrl_map.entries));
+		if (!new_map)
+			return -ENOMEM;
+
+		ctrl_map.entries = new_map;
+		ctrl_map.capacity = new_capacity;
+	}
+
+	if (asprintf(&ctrl_name_copy, "nvme%lu", ctrl_index) < 0)
+		return -ENOMEM;
+
+	ctrl_path_copy = _wcsdup(ctrl_path);
+	if (!ctrl_path_copy) {
+		free(ctrl_name_copy);
+		return -ENOMEM;
+	}
+
+	subsys_index = find_or_create_subsys_index(id_ctrl);
+
+	if (asprintf(&subsys_name, "nvme-subsys%d", subsys_index) < 0) {
+		free(ctrl_name_copy);
+		free(ctrl_path_copy);
+		return -ENOMEM;
+	}
+
+	ctrl_map.entries[ctrl_map.count].ctrl_name = ctrl_name_copy;
+	ctrl_map.entries[ctrl_map.count].ctrl_path = ctrl_path_copy;
+	ctrl_map.entries[ctrl_map.count].id_ctrl = *id_ctrl;
+	ctrl_map.entries[ctrl_map.count].subsys_index = subsys_index;
+	ctrl_map.entries[ctrl_map.count].subsys_name = subsys_name;
+	ctrl_map.count++;
+
+	return 0;
+}
+
+static int ctrl_map_find_index(const char *ctrl_name)
+{
+	size_t i;
+
+	for (i = 0; i < ctrl_map.count; i++) {
+		size_t sp_len = strlen(ctrl_map.entries[i].ctrl_name);
+		const char *suffix;
+
+		/* Skip entry if name does not start with nvmeX format */
+		if (strncmp(ctrl_map.entries[i].ctrl_name,
+			    ctrl_name, sp_len))
+			continue;
+
+		suffix = ctrl_name + sp_len;
+
+		/* Exact match: nvmeX */
+		if (*suffix == '\0')
+			return (int)i;
+
+		/* Namespace match: nvmeXnY (Y is a positive integer) */
+		if (*suffix != 'n' || suffix[1] < '1' || suffix[1] > '9')
+			continue;
+
+		suffix += 2;
+		while (*suffix >= '0' && *suffix <= '9')
+			suffix++;
+		if (*suffix == '\0')
+			return (int)i;
+	}
+
+	return -1;
+}
+
+/*
+ * Issue an Identify Controller command on the given handle.
+ * Creates a temporary transport context and handle internally.
+ * Returns 0 on success with id_ctrl filled in, negative errno on failure.
+ */
+static int identify_ctrl_from_handle(libnvme_fd_t h, struct nvme_id_ctrl *id_ctrl)
+{
+	struct libnvme_global_ctx *ctx;
+	struct libnvme_transport_handle *hdl;
+	struct libnvme_passthru_cmd cmd;
+	int ret;
+
+	ctx = libnvme_create_global_ctx();
+	if (!ctx)
+		return -ENOMEM;
+
+	hdl = __libnvme_create_transport_handle(ctx);
+	if (!hdl) {
+		libnvme_free_global_ctx(ctx);
+		return -ENOMEM;
+	}
+
+	hdl->fd = h;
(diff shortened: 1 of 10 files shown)


igaw:
The patch is too big to be properly reviewed. Can you at least split into two parts, the first one refactoring the existing code before adding new code?

bcapener:
Good idea. Sorry, I could not think of a better way to split this up. The refactoring of the existing code can be found here #3450 , I'll use the PR for adding the new code.

igaw:
I've merged #3450 so you can rebase this PR :)

bcapener:
Thanks Daniel, I have rebased this PR with upstream changes.

libnvme/src/nvme/ctrl-map.c
```diff
@@ -0,0 +1,838 @@
...
+ * Copyright (c) 2026 Micron Technology, Inc.
+ *
+ * Authors: Brandon Capener <bcapener@micron.com>
+ */
+#include <ctype.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wchar.h>
+#include "private-ctrl-map.h"
```
igaw, line 20: this should be grouped with the other local headers below

libnvme/src/nvme/ctrl-map.c
```diff
@@ -0,0 +1,838 @@
...
+	int subsys_index;
+	char *subsys_name;
+};
+
+static struct {
+	struct ctrl_map_entry *entries;
+	size_t count;
+	size_t capacity;
+	bool initialized;
+} ctrl_map;
```
igaw: Why not using the ccan hash table here? It should work for the use case here as far I can tell.

libnvme/src/nvme/ctrl-map.c
```diff
@@ -0,0 +1,838 @@
...
+
+	return -1;
+}
+
+/*
+ * Issue an Identify Controller command on the given handle.
+ * Creates a temporary transport context and handle internally.
+ * Returns 0 on success with id_ctrl filled in, negative errno on failure.
+ */
+static int identify_ctrl_from_handle(HANDLE h, struct nvme_id_ctrl *id_ctrl)
```
igaw: I would suggest to use the libnvme typedefs here, because it looks a bit strange below when assiging it to `hdl->fd = h`.

libnvme/src/nvme/ctrl-map.c
```diff
@@ -0,0 +1,838 @@
...
+{
+	struct libnvme_global_ctx *ctx;
+	struct libnvme_transport_handle *hdl;
+	struct libnvme_passthru_cmd cmd;
+	int ret;
+
+	ctx = libnvme_create_global_ctx(NULL, LIBNVME_DEFAULT_LOGLEVEL);
```
igaw: A don't think this should be here. Creating inside the library a global context is not correct.

libnvme/src/nvme/scan-win.c
```diff
@@ -29,7 +30,48 @@ __libnvme_public int libnvme_scan_subsystem_namespaces(
...
+	struct dirent **entries;
+	size_t i, count;
+	int ret;
+
+	if (!ctrls)
+		return -EINVAL;
+
+	*ctrls = NULL;
+
+	libnvme_ctrl_map_clear();
```
igaw, line 47: I see why you had to create global context hidden inside the ctrl map code. I suppose we should look how to extend the scan API in a sane way.

libnvme/src/nvme/lib-win.c
```diff
@@ -11,16 +11,198 @@
...
+{
+	/*
+	 * Controller device paths point to the PCI device, and begin with
+	 * "\\?\pci". Namespace device paths point to the disk device and begin
+	 * "\\.\PhysicalDrive".
+	 */
+	return strncasecmp(device_path, "\\\\?\\pci", 7) == 0;
+}
+
+static int __libnvme_transport_handle_open_direct(
```
igaw: could you split out the code for the handle libnvme_transport_handle implementation from the tree code? I'd rather review it separately.

libnvme/src/nvme/tree-win.c
```diff
@@ -16,29 +16,141 @@
...
-	return -ENOTSUP;
+	__cleanup_free char *hnqn = NULL;
+	__cleanup_free char *hid = NULL;
+	struct libnvme_host *h;
+
+	/* Use provided values or generate defaults */
+	if (hostnqn)
+		hnqn = strdup(hostnqn);
+	else
+		hnqn = strdup("nqn.2014-08.org.nvmexpress:uuid:00000000-0000-0000-0000-000000000000");
```
igaw, line 129: Doesn't Microsoft have a default UUID? Should be we just use one instead providing an NULL one?

libnvme/src/nvme/tree.c
```diff
@@ -588,13 +588,15 @@ static int nvme_subsystem_scan_namespaces(struct libnvme_global_ctx *ctx,
...
 
-	if (asprintf(&path, "%s/%s", libnvme_subsys_sysfs_dir(), name) < 0)
-		return -ENOMEM;
+	if (sysfs_dir)
+		if (asprintf(&path, "%s/%s", sysfs_dir, name) < 0)
+			return -ENOMEM;
 
 	s->model = libnvme_get_attr(path, "model");
-	if (!s->model)
+	if (path && !s->model)
```
igaw: This whole hunk doesn't look right. What does it try to do?

libnvme/src/nvme/tree.c
```diff
@@ -619,17 +621,20 @@ static int libnvme_scan_subsystem(struct libnvme_global_ctx *ctx,
...
+		if (ret < 0)
+			return -ENOMEM;
 
-	subsysnqn = libnvme_get_attr(path, "subsysnqn");
-	if (!subsysnqn)
-		return -ENODEV;
+		subsysnqn = libnvme_get_attr(path, "subsysnqn");
+		if (!subsysnqn)
+			return -ENODEV;
+	}
```
igaw: same here, what does it try to fix and if so, it should definitely be a separate patch with a proper commit message.

libnvme/src/nvme/lib-win.c
bcapener: sounds good, i created this PR https://github.com/linux-nvme/nvme-cli/pull/3456

libnvme/src/nvme/tree.c
bcapener: All of these changes in tree.c (including the other comment below) were our attempt to avoid using `#ifndef _WIN32`, instead we check if the `sysfs_dir` is not `NULL`, so that code will only run on linux and not windows.

```c
	if (path && !s->model)
		s->model = strdup("undefined");
```
here we don't want s->model set to "undefined" on windows, since it will be set somewhere else, that is why we added the path check.

libnvme/src/nvme/tree.c
bcapener: see other comment https://github.com/linux-nvme/nvme-cli/pull/3440#discussion_r3438904700

libnvme/src/nvme/lib-win.c
igaw: Thanks merged.

igaw:
BTW, I'll try to look into the scan API issue next week.

libnvme/src/nvme/scan-win.c
igaw: If I got this right, we should add the global context to the filter APIs. This would allow to pass it down into the library. Let me try this quickly.

igaw: It looks like we could even ditch the whole filter API on closer inspection...

igaw: I was able to get rid of all the `libnvme_filter_` public API and added the all `libnvme_scan_` functions a global context when it didn't have anything else like `libnvme_ctrl_t`. That means all the scan functions should then have access to the global context either directly or indirectly through an other object.

libnvme/src/nvme/ctrl-map.c
bcapener: done.

libnvme/src/nvme/ctrl-map.c
bcapener: done.

libnvme/src/nvme/tree.c
igaw: I am still a bit confused by the change. I look at it tomorrow with fresh eyes. But I think it should be in a separate patch and not in the middle of this large feature.

igaw:
fwiw, I making some good progress with the API review which should help this PR. It's still WIP but I hope to sort out some of the quirks soonish to unblock you here: https://github.com/igaw/nvme-cli/tree/api-review

libnvme/src/nvme/tree.c
bcapener: let me re-evaluate this code, I think the code has changed so much since we first wrote this that it no longer makes sense. I might just create a windows specific version of the libnvme_init_subsystem function and remove the libnvme_scan_subsystem changes, since this function is not called on windows. I will probably create a separate PR with the changes I propose.

bcapener: here is the PR for the changes I propose https://github.com/linux-nvme/nvme-cli/pull/3480, it covers this comment and the other comment below in tree.c

libnvme/src/nvme/tree.c
bcapener: ok, i plan on reverting these changes, for this function.

libnvme/src/nvme/tree.c
```diff
@@ -619,17 +621,20 @@ static int libnvme_scan_subsystem(struct libnvme_global_ctx *ctx,
...
 	int ret;
 
 	libnvme_msg(ctx, LIBNVME_LOG_DEBUG, "scan subsystem %s\n", name);
-	ret = asprintf(&path, "%s/%s", libnvme_subsys_sysfs_dir(), name);
-	if (ret < 0)
-		return -ENOMEM;
+	if (sysfs_dir) {
+		ret = asprintf(&path, "%s/%s", sysfs_dir, name);
+		if (ret < 0)
+			return -ENOMEM;
```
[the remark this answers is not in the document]
bcapener: reverted code. no longer applicable

libnvme/src/nvme/tree-win.c
```diff
@@ -16,29 +16,141 @@
...
 }
 
 __libnvme_public const char *libnvme_ctrl_get_state(libnvme_ctrl_t c)
 {
-	return NULL;
+	char *state = c->state;
+
+	c->state = strdup("");
+	free(state);
+	return c->state;
```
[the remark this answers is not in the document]
bcapener: I ignored this suggestion, since the code follows the pattern of the linux version. I am happy to reconsider.

libnvme/src/nvme/tree-win.c
bcapener: Is it worth this trouble for a value that will not be used on windows? from what I understand Host NQN and Host ID are for fabrics. Or are there other multi-host scenarios that I am not aware of?

Microsoft does have UUIDs but I am not sure they are equivalent.
Options are:
1. **Windows MachineGuid**: found in registry, HKLM\SOFTWARE\Microsoft\Cryptography\MachineGuid, used as a Windows-installation identifier. It is generated during Windows installation. **Caveat**: MachineGuid is tied to the Windows installation, not necessarily the physical host. It can be duplicated if an image/VM was cloned without proper generalization, and it can change after reinstall/sysprep. 
2. **SMBIOS System UUID**: From microsoft's docs "Universally unique identifier (UUID) for this product. A UUID is a 128-bit identifier that is guaranteed to be different from other generated UUIDs. If a UUID is not available, a UUID of all zeros is used." [link](https://learn.microsoft.com/en-us/windows/win32/cimwin32prov/win32-computersystemproduct). **Caveat**: In real systems, especially VMs, lab platforms, broken BIOS implementations, and cloned images, SMBIOS UUID can be all zeroes, all FFs, duplicated, or changed by virtualization settings.
3. **nvme-cli manages**: An LLM suggested the the only way to get the equivalent of linux's /etc/nvme/hostid would be to generate the UUID ourselves and save the value somewhere, like in the windows registry. This solution does not excite me, seems problematic.

Note: the information was sourced from an LLM, though I did try to verify it personally as best as I could. the caveats are a direct copy of LLM output.

bcapener: If we wanted to implement this, I was thinking about using the SMBIOS System UUID with maybe the MachineGuid as a fallback.

igaw: I checked what we do in Linux implementation in this scenario:

```c
__libnvme_public char *libnvmf_generate_hostid(void)
{
	int ret;
	char uuid_str[NVME_UUID_LEN_STRING];
	unsigned char uuid[NVME_UUID_LEN];

	ret = uuid_from_dmi(uuid_str);
	if (ret < 0)
		ret = uuid_from_device_tree(uuid_str);
	if (ret < 0) {
		if (libnvme_random_uuid(uuid) < 0)
			memset(uuid, 0, NVME_UUID_LEN);
		libnvme_uuid_to_string(uuid, uuid_str);
	}

	return strdup(uuid_str);
}


int libnvmf_host_get_ids(struct libnvme_global_ctx *ctx,
		      const char *hostnqn_arg, const char *hostid_arg,
		      char **hostnqn, char **hostid)
{
[...]

	if (!hid) {
		hid = libnvmf_generate_hostid();
		if (!hid)
			return -ENOMEM;

		libnvme_msg(ctx, LIBNVME_LOG_DEBUG,
			 "warning: using auto generated hostid and hostnqn\n");
	}

	/* incomplete configuration, thus derive hostnqn from hostid */
	if (!hnqn) {
		hnqn = libnvmf_generate_hostnqn_from_hostid(hid);
		if (!hnqn)
			return -ENOMEM;
	}
[...]
}
```

That means if all fails we just generate a UUID and use this one. I just realized I moved this part under CONFIG_FABRICS feature flag.  Looks like the generate hostid/hostnqn should always be available. Let me fix this.

Anyway, I think it would be good to have something similar on Windows, so the ports behave similar.

libnvme/src/nvme/ctrl-map.c
bcapener: ok, i have removed this and passed it as an argument. thanks!

libnvme/src/nvme/scan-win.c
bcapener: looks good. thanks

libnvme/src/nvme/ctrl-map.c
bcapener: Ok, I reworked it to use ccan hash tables

igaw:
From my first pass through the commits the final result looks good. I'll give it another look tomorrow, tired from reviewing the whole day, so not good to through such a big change too quickly.

igaw:
I've merged #3537 which is basically the final code from this PR. It looks in reasonable shape and let's improve the code in tree. No need to block this.

Thanks a lot!

PR #3444, opened 2026-06-12
ScaleFlux Plugin Cleanup AER Values

benreese0:
Fixing logic in status command to not immediately exit if unable to parse AER values in /sys/

Fixed a couple of whitespace errors highlighted by checkpatch


benreese0:
HI all, the `checkpatch` failures are only line lengths that are going to be kinda annoying to resolve.  We're working on a refactor of the ScaleFlux plugin for later this year which should make the line lengths not an issue anymore.
Can we let this one go through as is?

igaw:
Sure, glad to hear the plugin is still maintained :)

PR #3445, opened 2026-06-15
build: add Clang Static Analyzer support

shroffni:
Add support for building nvme-cli with the Clang Static Analyzer.

The static analyzer complements existing runtime verification tools such as ASAN/UBSan and Valgrind by performing compile-time analysis of code paths and reporting potential defects without requiring execution.

Introduce a new '-a' option to scripts/build.sh which configures the build to use the Clang Static Analyzer via scan-build. When selected, the build automatically uses the Clang compiler and generates analyzer reports as part of the build process. The generated reports can be viewed using a web browser.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -16,6 +16,7 @@ usage() {
     echo ""
     echo "CI build script."
     echo ""
+    echo " -a                   run static analysis with clang analyzer"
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3446, opened 2026-06-16
Introduce Qualified Protection Information Format (QPIF) 

MaisenbacherD:


MaisenbacherD:
I had this patch laying around for issue https://github.com/nvme-experiments/libnvme/issues/19. Finally got arround to finish it up, sorry for the long delay :)

igaw:
Looks reasonable and I do like the test case. Thanks!

PR #3447, opened 2026-06-16
Simplify code flow in fabrics.c

Mr-Bossman:
hook_parser_next_line used gotos to implement a do-while loop.
Replace the gotos with a do-while loop to improve readability.
Simplify create_common_context by calling setup_common_context.
Add to set_fabrics_options to setup_common_context as both callers  call it after setup_common_context.

Signed-off-by: Jesse Taube <jtaubepe@redhat.com>

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -334,25 +334,23 @@ static int hook_parser_next_line(struct libnvmf_context *fctx, void *user_data)
 		  OPT_FLAG("force",          0, &force,      "Force persistent discovery controller creation"));
 
 	memcpy(&fa, hfd->fa, sizeof(fa));
-next:
-	if (fgets(line, sizeof(line), hfd->f) == NULL)
-		return -EOF;
+	do {
+		if (fgets(line, sizeof(line), hfd->f) == NULL)
+			return -EOF;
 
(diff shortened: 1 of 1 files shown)


fabrics.c
igaw, line 358: let's use then the opportunity to do `return setup_comm_context(...)` here.

fabrics.c
```diff
@@ -391,7 +387,7 @@ static int setup_common_context(struct libnvmf_context *fctx,
 	if (err)
 		return err;
 
-	return 0;
+	return set_fabrics_options(fctx, fa);
 }
 
 static int create_common_context(struct libnvme_global_ctx *ctx,
```
igaw, line 395: I suppose it would make sense to drop the `common` in the function names. I introduced it because I thought there will be a need to differentiate between the 'base' info and the rest. But looks like it's not really the case.

igaw:
Looks good, just my nitpicks and a rebase is necessary.

fabrics.c
igaw: got confused by the web ui... stupid suggestion. sorry about it.

igaw:
Thanks!

PR #3448, opened 2026-06-16
Allow config.json to act like discovery.conf 

Mr-Bossman:
In  discovery.conf hostnqn and nqn are not required. to allow a seamless transition between the two formats make them also not required in config.json. This also allows a default config.json to be shipped without conflicting hostnqns.

Signed-off-by: Jesse Taube <jtaubepe@redhat.com>


igaw:
Looks good. BTW, thanks a lot for going through this part of the code base. It was in a terrible state and it starts to get better :)

PR #3450, opened 2026-06-16
tree: refactor tree.c, filters.c, and lib.c

bcapener:
Move existing linux specific code into separate files, tree-linux.c, filters-linux.c, and lib-linux.c. Create new files with windows stub function for tree-win.c, filters-win.c and lib-win.c. The windows functionality will be implemented in an upcoming change.


igaw:
Looks good. I am sure we need to reshuffle/restructure more stuff later. 

I've updated the commit message and pushed the change directly.

PR #3452, opened 2026-06-17
ioctl-win: Update sanitize implementation to use new ioctl

brgoing:
Sanitize is documented to work on Windows 11+ with IOCTL_STORAGE_PROTOCOL_COMMAND, but it doesn't.  Change to use IOCTL_STORAGE_REINITIALIZE_MEDIA on the namespace handle if possible.


igaw:
Thanks!

PR #3453, opened 2026-06-17
ioctl-win: Assert ctrl handle for Windows format nvm

brgoing:
Windows only supports format operations on namespace handles. Adds  handle type check before executing disk-based format operations. Fails with -ENOTSUP if not a namespace handle.


igaw:
Thanks!

PR #3454, opened 2026-06-17
completions: add power-measurement-log command

jimmunn:
Add shell completion support for the power-measurement-log command to both bash and zsh completions.


igaw:
Thanks!

PR #3455, opened 2026-06-17
Completions: remove -j and --json from top level sub-commands

jimmunn:
None of the top-level sub-commands support the -j or --json option. However, the bash completions included -j and --json. Removing those options from the completions code to be consistent with actual code behavior.

### DIFF
--- a/completions/bash-nvme-completion.sh
+++ b/completions/bash-nvme-completion.sh
@@ -542,7 +542,7 @@ nvme_list_opts () {
 			;;
 	esac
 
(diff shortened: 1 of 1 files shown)


igaw:
Ah, forgot to update the bindings. There a few new global options if you want to add them :)

Thanks!

jimmunn:
> Ah, forgot to update the bindings. There a few new global options if you want to add them :)

@igaw , I am exploring some options for generating the completions. I'll probably start a discussion in Matrix when I am further along.

PR #3456, opened 2026-06-18
lib: add windows direct open transport handle support

bcapener:
* rename __nvme_transport_handle_open_direct to __libnvme_transport_handle_open_direct. 
* Implement functionality for the windows __libnvme_transport_handle_open_direct function.


igaw:
Thanks!

PR #3460, opened 2026-06-19
libnvme, python: pass global context to registry public APIs and fix python test issue

martin-belanger:
@igaw - Per our discussion. I decided to bundle two commits into this PR.

- Add global context to all registry public APIs.
- Fix the python registry test. The problem was in the test fixture itself, not
  the python bindings (it relied on fork semantics, which broke under the
  spawn/forkserver default in Python 3.14).

I will wait on your decision for the `fedora:latest` CI workflow to catch python regressions early.

### DIFF
--- a/plugins/registry/registry-nvme.c
+++ b/plugins/registry/registry-nvme.c
@@ -33,13 +33,15 @@ static void print_attr(const char *attr, const char *value, void *user_data)
 
 static void print_device(const char *device, void *user_data)
 {
+	struct libnvme_global_ctx *ctx = user_data;
 	printf("%s\n", device);
-	libnvmf_registry_attr_for_each(device, print_attr, NULL);
+	libnvmf_registry_attr_for_each(ctx, device, print_attr, NULL);
 }
 
 static int registry_list(int argc, char **argv, struct command *acmd,
 			  struct plugin *plugin)
 {
+	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
 	const char *desc = "List all live NVMeoF controller ownership registry entries.";
 
 	NVME_ARGS(opts);
@@ -47,12 +49,14 @@ static int registry_list(int argc, char **argv, struct command *acmd,
 	if (argconfig_parse(argc, argv, desc, opts))
 		return -EINVAL;
 
-	return libnvmf_registry_device_for_each(print_device, NULL);
+	ctx = libnvme_create_global_ctx(stdout, LIBNVME_DEFAULT_LOGLEVEL);
+	return libnvmf_registry_device_for_each(ctx, print_device, ctx);
 }
 
 static int registry_retrieve(int argc, char **argv, struct command *acmd,
 			      struct plugin *plugin)
 {
+	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
 	const char *desc = "Read an attribute from a controller's registry entry.";
 	const char *device_help = "NVMe device name (e.g. nvme3)";
 	const char *attr_help = "attribute name (default: owner)";
@@ -80,7 +84,8 @@ static int registry_retrieve(int argc, char **argv, struct command *acmd,
 	}
 	strip_dev_prefix(&cfg.device);
 
-	ret = libnvmf_registry_retrieve(cfg.device, cfg.attr, &value);
+	ctx = libnvme_create_global_ctx(stdout, LIBNVME_DEFAULT_LOGLEVEL);
+	ret = libnvmf_registry_retrieve(ctx, cfg.device, cfg.attr, &value);
 	if (ret == -ENOENT) {
 		fprintf(stderr, "%s: not registered or '%s' not found\n",
 			cfg.device, cfg.attr);
@@ -97,6 +102,7 @@ static int registry_retrieve(int argc, char **argv, struct command *acmd,
 static int registry_update(int argc, char **argv, struct command *acmd,
 			    struct plugin *plugin)
 {
+	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
 	const char *desc = "Write an attribute to a controller's registry entry. The 'owner' attribute is immutable and cannot be changed.";
 	const char *device_help = "NVMe device name (e.g. nvme3)";
 	const char *attr_help = "attribute name (e.g. note); 'owner' is not allowed";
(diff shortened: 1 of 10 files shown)


.github/workflows/fedora-latest.yml
```diff
@@ -0,0 +1,75 @@
...
+    container:
+      image: fedora:latest
+    steps:
+      - name: Install build dependencies
+        run: |
+          dnf -y install \
+            gcc meson ninja-build pkgconf-pkg-config git tar gawk util-linux \
+            python3 python3-devel swig \
+            json-c-devel openssl-devel keyutils-libs-devel dbus-devel \
+            libnl3-devel libcurl-devel libarchive-devel xz-devel
```
igaw: This should use the pre-build CI containers as the install step fails way to often.

.github/workflows/fedora-latest.yml
```diff
@@ -0,0 +1,75 @@
...
+    branches: [master]
+  pull_request:
+    branches: [master]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  fedora-latest-no-valgrind:
```
igaw: I think this should be  in the build.yaml file, with the rest of the builds. If we want to split the build.yaml file up, we should first figure out what's a good strategy.

plugins/registry/registry-nvme.c
```diff
@@ -141,6 +154,7 @@ static int registry_update(int argc, char **argv, struct command *acmd,
 static int registry_delete(int argc, char **argv, struct command *acmd,
 			    struct plugin *plugin)
 {
+	struct libnvme_global_ctx *ctx;
```
igaw: __cleanup helpers?

plugins/registry/registry-nvme.c
```diff
@@ -33,26 +33,34 @@ static void print_attr(const char *attr, const char *value, void *user_data)
...
 	printf("%s\n", device);
-	libnvmf_registry_attr_for_each(device, print_attr, NULL);
+	libnvmf_registry_attr_for_each(ctx, device, print_attr, NULL);
 }
 
 static int registry_list(int argc, char **argv, struct command *acmd,
 			  struct plugin *plugin)
 {
+	int ret;
+	struct libnvme_global_ctx *ctx;
```
igaw: the ret could go as last, so we keep it a bit more inverved x-tree. though we have quiet a bit of offenders...

igaw:
btw, about the `Assisted-by` tag. I pondered on it a for a while and I think it doesn't make sense. The author has to know the code she/he submits and also 'own' it. We always use tools to write/modify code. Often we don't mention them, so I don't really see why we need to advertise those tools (it's more like an endorsement thing in my eyes). Since I didn't get any lawyers advice I don't need them. Would you mind to drop them in the future. I know it's not what the LF wants but this project is not part of LF in anyway.

.github/workflows/fedora-latest.yml
martin-belanger: Fair point on the flakiness. The one thing to keep in mind is that pointing this job at the pre-built `ghcr.io/linux-nvme/fedora:latest` image would mostly undo what it's for: that image trails the real Fedora release and the build.yml matrix already exercises it, so we'd be testing the same (older) Python the matrix already covers. The reason I used the upstream `fedora:latest` is to get the *newest* Python interpreter — that's where things like the 3.14 "forkserver" default surface first, which is what broke the fork-based test in the first place.

If you'd rather not carry this new CI workflow, I'm happy to drop it. If we do keep it, I can make the install more robust (retry the `dnf install`, maybe pin the package set) so it stops failing, and add a short note in the file explaining why it uses the upstream image rather than the pre-built one. Whatever you prefer.

.github/workflows/fedora-latest.yml
martin-belanger: This one is really tied to the first comment. The fedora:latest job *is* the no-valgrind run — that's what lets the fork-based tests actually execute (they're skipped under valgrind, e.g. the python-registry parallel-writes test is `@unittest.skipIf(_under_valgrind, ...)`). So if we keep it, there's no need for a second no-valgrind job in build.yml; that would just duplicate the coverage on a lagging image. The no-valgrind-in-build.yml route only really makes sense if you'd rather drop the fedora:latest workflow altogether — in which case I'm happy to add a no-valgrind job to build.yml instead. So it mostly comes down to your call on the first comment.

.github/workflows/fedora-latest.yml
igaw: Why would the pre-build undo it? We can add fedora-latest container and build it on a regular basis. The install step failed way too often in the past. I just don't want to waste time finding out it was a network issue again, why a build failed.

martin-belanger: Ah, I didn't have the full background on the pre-built images — I'd assumed they always trail the real Fedora release, which is the only thing I was trying to avoid (the whole point of this job is to run the Python bindings against the *newest* interpreter, since that's where things like the 3.14 "forkserver" default surface first). A regularly-rebuilt pre-built container is clearly the right fix for the flaky `dnf install` — that moves the install to image-build time and gets it out of the per-run path. Happy to go that route; I'll switch the job to the pre-built image.

One nuance so we keep the early-Python signal: the current `ci-containers` fedora image is pinned to `FROM fedora:43`, and the build workflow only triggers on push/tags (no schedule). A regular rebuild of a `fedora:43` image just refreshes packages within 43 — it won't ever surface a new Python *major* until someone manually bumps the Dockerfile to 44. So for the fedora image specifically, catching the next "forkserver"-style change early needs two things: `FROM fedora:latest` (or rawhide) instead of the pinned release, and a `schedule:` trigger so it actually rebuilds on a regular basis. With those, the pre-built image gives us both the stability you want and the newest-interpreter coverage this job was for. If you'd rather keep the shared fedora image pinned, I can point this one job at a dedicated `fedora-latest` variant instead — whichever you prefer.

martin-belanger:
@igaw - I've pulled the CI workflow out of this PR so it doesn't hold up the rest. I'll bring it back as its own PR once we've settled the `ci-containers` direction (using the pre-built image, and whether the fedora image moves to `fedora:latest` + a scheduled rebuild so it still catches new-Python regressions early).

.github/workflows/fedora-latest.yml
martin-belanger: I will do the CI workflow in a separate PR depending on yout decision.

plugins/registry/registry-nvme.c
martin-belanger: Done

.github/workflows/fedora-latest.yml
martin-belanger: Same as above (CI workflow changes will be in a separate PR)

.github/workflows/fedora-latest.yml
igaw: The ci-container project needs also some more improvements.  I haven't really look into the details regarding the upstream containers. I'd assume there is a tag for latest release container and latest container, just need to use it. Anyway, thanks for looking into making the ci-containers better. It's a simple project :)

igaw:
Thanks a lot!

PR #3462, opened 2026-06-21
libnvme: move fabrics include list into its own file

martin-belanger:
The fabrics-only headers appended to the generated `libnvme.h` were carried as a single `\n`-escaped string passed to `libconf.set()` in `libnvme/src/meson.build`. That form is hard to read and easy to get wrong when an include is added or removed.

This moves the list into `libnvme/src/fabrics-includes.h.in` (one `#include` per line) and pulls it in with `fs.read().strip()`. The generated `libnvme.h` is byte-identical.

Suggested by @igaw during the registry PR (#3425) review as a small, standalone preparation patch.

### DIFF
--- a/libnvme/src/fabrics-includes.h.in
+++ b/libnvme/src/fabrics-includes.h.in
@@ -1,0 +1,4 @@
+#include <nvme/fabrics.h>
+#include <nvme/nbft.h>
+#include <nvme/accessors-fabrics.h>
+#include <nvme/registry.h>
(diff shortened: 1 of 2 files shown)


igaw:
I claim the assisted tag here! :)

martin-belanger:
Ha! Fair enough — the idea was yours, so I swapped the Assisted-by for a Suggested-by: Daniel Wagner <wagi@kernel.org>. Credit where it's due. 😄

igaw:
Thanks!

PR #3463, opened 2026-06-21
feat: change nvme print flags to verbose

ikegami-t:
Since always set as normal but suitable for the human readable.

### DIFF
--- a/plugins/feat/feat-nvme.c
+++ b/plugins/feat/feat-nvme.c
@@ -79,6 +79,7 @@ static int feat_get_nsid(struct libnvme_transport_handle *hdl, __u32 nsid,
 	__u64 result;
 	int err;
 	__u32 len = 0;
+	nvme_print_flags_t flags = NORMAL;
 
 	__cleanup_libnvme_free void *buf = NULL;
 
@@ -98,9 +99,18 @@ static int feat_get_nsid(struct libnvme_transport_handle *hdl, __u32 nsid,
 		return err;
 	}
 
+	err = validate_output_format(nvme_args.output_format, &flags);
+	if (err < 0) {
+		nvme_show_error("Invalid output format");
+		return err;
+	}
+
+	if (nvme_args.verbose)
+		flags |= VERBOSE;
+
 	nvme_show_init();
 
-	nvme_show_feature(fid, sel, result, buf, len, NORMAL);
+	nvme_show_feature(fid, sel, result, buf, len, flags);
 
 	nvme_show_finish();
 


plugins/feat/feat-nvme.c
```diff
@@ -100,7 +100,7 @@ static int feat_get_nsid(struct libnvme_transport_handle *hdl, __u32 nsid,
 
 	nvme_show_init();
 
-	nvme_show_feature(fid, sel, result, buf, len, NORMAL);
+	nvme_show_feature(fid, sel, result, buf, len, VERBOSE);
```
igaw: Should the command line args verbose setting be use here?

ikegami-t: Currently the feat plugin command output is below for example as just output the value itself.
```sh
tokunori@tokunori-X570-Taichi:~/nvme-cli$ nvme-build feat arbitration /dev/nvme0
get-feature:0x01 (Arbitration), Current value:0x03030302
```
Then changed output by the PR verbose print flags is below as the value fields output.
```sh
tokunori@tokunori-X570-Taichi:~/nvme-cli$ nvme-build feat arbitration /dev/nvme0
get-feature:0x01 (Arbitration), Current value:0x03030302
        High Priority Weight   (HPW): 4
        Medium Priority Weight (MPW): 4
        Low Priority Weight    (LPW): 4
        Arbitration Burst       (AB): 4
```
For the reference the following outpus are get-feature comamnd normal and verbose outputs.
```sh
tokunori@tokunori-X570-Taichi:~/nvme-cli$ nvme-build get-feature /dev/nvme0 -f 1
get-feature:0x01 (Arbitration), Current value:0x03030302
tokunori@tokunori-X570-Taichi:~/nvme-cli$ nvme-build get-feature /dev/nvme0 -f 1 -v
get-feature:0x01 (Arbitration), Current value:0x03030302
        High Priority Weight   (HPW): 4
        Medium Priority Weight (MPW): 4
        Low Priority Weight    (LPW): 4
        Arbitration Burst       (AB): 4
```
Seems for the feat plugin command stdout print the vebose output as default is suitable to me. Do you think still the normal output should be keeped by the default for the feat plugin command?

igaw: I think it would be more consistent with the rest of the commands. If you don't provide the verbose flag you just the the raw value. I am not insisting on this but I think it's just a bit odd.

ikegami-t: Understood then just fixed the patch as mentioned. Thank you.

igaw:
Thanks!

PR #3464, opened 2026-06-21
build: add Clang analyzer workflow

shroffni:
Add a nightly workflow job which runs the Clang static analyzer, archives the generated reports and uploads them to the artifact store and SFTP server for later inspection.


.github/workflows/run-nightly-tests.yml
```diff
@@ -191,3 +191,65 @@ jobs:
...
+        with:
+          fetch-depth: 0
+
+      - name: Mark repo as safe for git
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y clang-tools meson
```
igaw: We should use the pre-build [ci-containers](https://github.com/linux-nvme/ci-containers) which avoids the install step which fails way to often in this environment. 

I suggest we grap the 'coverity' feature and rename it to 'analize' or something like this and add all missing dependencies to it: https://github.com/linux-nvme/ci-containers/blob/main/ci-containers.yaml

shroffni: > We should use the pre-build [ci-containers](https://github.com/linux-nvme/ci-containers) which avoids the install step which fails way to often in this environment.
 
Yeah makes sense. Frankly, I was not aware about this ci-container repo. Anyways, first I will send a PR for this repo which would add clang-tools in the container. 

> I suggest we grap the 'coverity' feature and rename it to 'analize' or something like this and add all missing dependencies to it: https://github.com/linux-nvme/ci-containers/blob/main/ci-containers.yaml

Grepping coverity in the current nvme-cli code base yields the following output:
.github/workflows/coverity.yml:23:  coverity-scan:
.github/workflows/coverity.yml:25:    name: coverity scan
.github/workflows/coverity.yml:48:          curl -o coverity_tool.tgz \
.github/workflows/coverity.yml:51:            https://scan.coverity.com/download/linux64
.github/workflows/coverity.yml:52:          mkdir coverity-tools
.github/workflows/coverity.yml:53:          tar xzf coverity_tool.tgz --strip 1 -C coverity-tools
.github/workflows/coverity.yml:61:          export PATH="$PWD/coverity-tools/bin:$PATH"
.github/workflows/coverity.yml:66:          tar czvf nvme-cli-coverity.tgz cov-int
.github/workflows/coverity.yml:72:            --form file=@nvme-cli-coverity.tgz \
.github/workflows/coverity.yml:75:            https://scan.coverity.com/builds?project=linux-nvme%2Fnvme-cli
.github/workflows/coverity.yml:78:        name: upload coverity artifacts
.github/workflows/coverity.yml:81:          name: coverity-results
.github/workflows/coverity.yml:84:            nvme-cli-coverity.tgz
README.md:12:[Coverity Scan Build Status](https://scan.coverity.com/projects/linux-nvme-nvme-cli)

So do you suggest instead of adding clang-analyzer support in the nightly build, I'd instead replace/rename the coverity.yml file with analyzer.yml and update it as necessary? And if yes, then it means we'd run clang-analyzer "weekly on Mondays at 00:00 UTC" and not everyday. 

Please suggest.

igaw: yes, I would rename/replace the existing coverity.yaml file with the new build steps and update the build trigger to nightly (or even for each merge ?) for the foreseeable future. The once per week was just because there wasn't much attention on the report and I didn't want to burn too much resources.

shroffni: Ok I will update the PR with suggested changes. I will rename/replace all coverity feature with the analyzer. For now, I will add support for triggering analyzer workflow nightly or manually.

igaw:
Thanks!

igaw:
BTW, I noticed when the build fails early (clang-tools missing) there is no error propagated.

shroffni:
> BTW, I noticed when the build fails early (clang-tools missing) there is no error propagated.

Yes correct and that's because we have continue-on-error is set to true. This was added so that even when build with clang analyzer report issues, the build step (scripts/build.sh -a) would exit with non-zero status. And this is not an issue as we want to continue here so in the next step we could upload artifacts (scan report) to github and then later to SFTP. 
So just to allow workflow continue even after build step fails, I added continue on error. Without this, workflow would abort after compilation and skip upload step.  This has consequences or side effect that in case workflow fails early (even before compilation starts) then the workflow is still marked as success.

igaw:
ah yes. so it was just a bad luck, clang-tools was still missing in the container. anyway seems to work. hopefully the next run gets uploaded and correctly untared.

PR #3465, opened 2026-06-22
nvme: rework the registry command-line interface

martin-belanger:
Take the controller as the positional <device> argument on retrieve, update and delete, matching every other nvme command and dropping the odd-one-out -d/--device option. As a side effect the generic "<device> may be a namespace or mctp address" help text now applies, since registry takes a positional device like everything else.

Drop the hard -EPERM block on writing the 'owner' attribute: it protected nothing (root can edit /run/nvme/registry/ directly) yet inconsistently let 'delete' clear ownership anyway. update and delete now confirm before an operation that changes or removes ownership; scripts proceed without prompting.

While here, give delete an optional --attr to remove a single attribute (clearer than "update with no value"), and require an explicit --attr on retrieve rather than silently defaulting to 'owner'.

### DIFF
--- a/Documentation/nvme-registry-delete.txt
+++ b/Documentation/nvme-registry-delete.txt
@@ -3,30 +3,42 @@ nvme-registry-delete(1)
 
 NAME
 ----
-nvme-registry-delete - Remove a controller's ownership registry entry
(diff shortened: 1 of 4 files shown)


igaw:
Thanks!

PR #3466, opened 2026-06-22
nvme: disconnect-all --transport short option + add --owner option

martin-belanger:
Two small, independent fabrics CLI changes building on the ownership registry (#3425).

### 1. disconnect-all: `-r` → `-t` for `--transport`

Every other fabrics command spells `--transport` as `-t` (the shared `NVMF_ARGS` macro), but `disconnect-all` defined its own option set and used `-r`. `-t` is the canonical transport short form and is now free on `disconnect-all` since `--timeout` no longer claims a short form. This is a deliberate break with no `-r` alias. Also drops a stray `(char *)` cast on the option's value argument.

### 2. Add `--owner NAME` to connect, connect-all and discover

These commands establish controller connections through libnvme but had no way to record the connecting orchestrator's identity, so the controllers they created always looked unowned and `disconnect-all` would tear them down. A new long-only `--owner NAME` option records ownership in the registry (via `libnvme_set_owner()`) before connecting.

- `--owner` has no short form on any command. To keep that consistent, `disconnect-all`'s `--owner` also loses its `-O` short form.
- For `connect-all`/`discover`, `--nbft` already records `owner=nbft` so legacy boot scripts get ownership for free. An explicit `--owner` now overrides that default; a plain `--nbft` (no `--owner`) still records `nbft`, so the boot path is unchanged.

Man pages and shell completions are updated to match.

### DIFF
--- a/Documentation/nvme-connect-all.txt
+++ b/Documentation/nvme-connect-all.txt
@@ -18,6 +18,7 @@ SYNOPSIS
 			[--nbft]
 			[--no-nbft]
 			[--nbft-path=<STR>]
+			[--owner=<NAME>]
 			[<fabrics-options>]
 			[<global-options>]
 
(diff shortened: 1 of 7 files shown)


igaw:
Thanks for checking all the command options and make it consistent!

PR #3467, opened 2026-06-22
doc: add libnvme registry overview (REGISTRY.md)

martin-belanger:
Add a concise overview of the NVMe controller ownership registry: why it exists (orchestrator coexistence and the disconnect-all coordination problem from issue #2913), the /run/nvme/registry layout, how an owner is set, the udev-driven cleanup and its instance-recycling guard, and the disconnect-all guardrails. The header kdoc and man pages remain the authoritative reference.

Suggested-by: Daniel Wagner <wagi@kernel.org>

### DIFF
--- a/libnvme/REGISTRY.md
+++ b/libnvme/REGISTRY.md
@@ -1,0 +1,139 @@
+# NVMe Controller Ownership Registry
+
+The registry records **which orchestrator owns each connected NVMe-oF controller**. It is a small, cooperative coordination layer that lets independent tools share a host without stepping on each other's controllers — most importantly, so that a sweeping command like `nvme disconnect-all` does not tear down a connection some other component depends on.
+
+> The code is the source of truth. This document summarizes behavior and intent; for exact signatures see the header kdoc in `src/nvme/registry.h` and the `nvme-registry-*` man pages.
+
+## Why it exists
+
+NVMe-oF connections on a host are rarely managed by a single actor. Independent **orchestrators** — agents that decide on their own which controllers to connect and disconnect — coexist on the same machine and share one flat device namespace (`/dev/nvmeX`), with no record of who created or manages each controller.
+
+This holds even on a plain system with no daemons installed. There are already at least two orchestrators: the **initramfs** (NBFT and FC-kickstart connections made during early boot) and a **human** running `nvme connect-all` / `nvme disconnect-all`. Installing `nvme-discoverd` or `nvme-stas` only adds more.
+
+Not every command is an orchestrator:
+
+- `nvme connect` / `nvme disconnect` — single, targeted actions; here the *human* is the orchestrator, deciding what to connect or disconnect.
+- `nvme connect-all` / `nvme disconnect-all` — orchestrating commands that, once invoked, decide on their own. `connect-all` reads a discovery controller's discovery log page (DLP), connects every DLP entry, and follows referrals into further discovery controllers, recursing through several layers; `disconnect-all` tears down across the whole host.
+- `nvme discover` — in between: it reads a discovery controller's DLP and prints it, but does not connect DLP entries or follow referrals. It is read-mostly; its only state-changing option is `--persistent`, which keeps the discovery connection itself.
+
+The trigger was [issue #2913](https://github.com/linux-nvme/nvme-cli/issues/2913): `disconnect-all` has host-wide scope but no way to tell a boot-critical NBFT connection — or one a daemon depends on — from a throwaway manual connect, so it would disconnect them all. To coordinate, an orchestrator must know who owns what.
+
+**This is why the registry is a new kind of state for `nvme`.** `nvme` already reads configuration (host identity, saved connections) and kernel state from sysfs, but it kept no memory of what it had *done* — each invocation was independent. The registry is that memory: runtime state `nvme`/libnvme writes itself, automatically, as a controller is connected (under `/run` — per-boot, not across reboot), recording which orchestrator owns each controller and read back by a later invocation or any other tool on the host.
+
+The registry is a **cooperative tool, not an enforcement mechanism**. Every orchestrator runs as root and could disconnect anything regardless. The registry simply lets well-behaved tools avoid doing so by accident.
+
+## What it is
+
+The registry lives under `/run/nvme/registry/` — runtime state that is tied to controller lifecycle and does not survive a reboot. It mirrors the sysfs convention: one directory per live controller, one plain-text file per attribute.
+
+```
+/run/nvme/registry/
+    nvme1/
+        owner
+        seqnum
+    nvme3/
+        owner
+        seqnum
+```
+
+```sh
+$ cat /run/nvme/registry/nvme3/owner
+nbft
+```
+
+- **Absence means unowned.** A controller with no directory (or no `owner` file) is managed by nobody. There is no explicit "unowned" marker.
+- The well-known attribute is **`owner`** — the orchestrator identity string (e.g. `stas`, `nbft`, `discoverd`). Orchestrators may write additional private attributes; unknown attributes are ignored by everyone else.
+- libnvme also writes a reserved **`seqnum`** attribute next to `owner` — the value of `/sys/kernel/uevent_seqnum` captured at connect. It is internal bookkeeping for the cleanup rule (see *Automatic cleanup*); orchestrators neither set nor read it.
+- Directories are `0755`, attribute files `0644`: **world-readable, root-writable**. Both the root and per-device directories are created on demand on first write.
+- Writes are atomic (`mkstemp` → `fsync` → `rename`), so concurrent writers never corrupt an entry and readers never see a half-written value.
+- Controllers the kernel manages directly — PCIe and other memory-based transports — are out of scope: they are not reached over a fabric, and `disconnect-all`'s transport-type check already excludes them.
+
+## Setting an owner
+
+An orchestrator declares its identity once, then every controller it connects through that context is registered automatically on a successful connect:
+
+```c
+struct libnvme_global_ctx *ctx = libnvme_create_global_ctx(...);
+libnvme_set_owner(ctx, "stas");   /* registers owner=stas on every connect */
+```
+
+A process that does not call `libnvme_set_owner()` produces **unowned** connections. On connect, libnvme writes the entry when an owner is set — stamping it with the current `/sys/kernel/uevent_seqnum` (used by the cleanup rule below) — and clears any stale entry for a recycled instance number when it is not.
+
+From the CLI, pass `--owner NAME` to register ownership at connect time:
+
+```sh
+nvme connect     --owner discoverd ...
+nvme discover    --owner discoverd ...
+nvme connect-all --owner discoverd ...
+nvme connect-all --nbft                 # NBFT controllers, owner=nbft
+nvme connect-all --owner boot --nbft    # NBFT controllers, owner=boot (overrides nbft)
+```
+
+`nvme connect-all --nbft` records `owner=nbft` automatically to protect firmware boot volumes. That `nbft` is only a default, though: an explicit `--owner NAME` given alongside `--nbft` overrides it, so `nvme connect-all --owner NAME --nbft` records the controllers as `NAME`. A plain `--nbft` (no `--owner`) keeps `owner=nbft`, which is what lets existing boot scripts get ownership for free without being changed. Without `--owner` and without `--nbft`, connections are unowned.
+
+## Automatic cleanup
+
+A controller's entry must be removed once the controller goes away, or stale entries accumulate and a recycled device name looks owned. **Creation and removal deliberately live in different places.** Entries are *written* by libnvme on connect (see *Setting an owner* above) because only the connecting process knows the owner — it lives in that process's libnvme context, never in sysfs or the uevent — so nothing else, a udev rule included, could supply it. *Removal* needs no owner, only the device name, and it must happen even when the kernel drops a controller on its own (connectivity loss, `ctrl-loss-tmo` expiry) with no orchestrator involved, and even on a host running no orchestrator daemon at all. So removal is driven by the device-removal event and performed by a single, always-present agent — udevd: exactly one party cleans up, and it works with no daemon required.
+
+libnvme ships a udev rule for exactly this. The udev daemon — present on essentially every system — removes a controller's registry entry when the kernel removes the controller:
+
+```
+ACTION=="remove", SUBSYSTEM=="nvme", KERNEL=="nvme[0-9]*", \
+    RUN+="/bin/sh -c '[ -e /dev/%k ] || [ $env{SEQNUM} -le $$(cat /run/nvme/registry/%k/seqnum 2>/dev/null || echo 0) ] || rm -rf /run/nvme/registry/%k'"
+```
+
+(`$$` passes a literal `$` to the shell; `$env{SEQNUM}` is the remove event's sequence number, substituted by udev.)
+
+**Why removal is guarded.** udev rules run asynchronously: udevd processes a `remove` uevent some time *after* the kernel emits it. Meanwhile the kernel is free to recycle the just-freed instance number — it allocates the lowest available id, so the very next controller to connect can be handed the same `nvmeN`. This opens a rare race: a controller is removed and, at almost the same instant, a new one is connected and inherits its id, all before udevd gets around to running the remove rule for the old controller. By the time that rule finally fires, `nvmeN` already refers to a *different*, live controller that has written its own registry entry — and a blind `rm` would clobber it.
+
+Concretely, the dangerous interleaving for the name `nvme4`:
+
+1. The kernel removes the old `nvme4`, frees instance 4, and emits a `remove` uevent — now queued in udevd.
+2. Before udevd runs that queued rule, a new controller connects. The kernel allocates the lowest free instance — 4 again — creates `/dev/nvme4` (devtmpfs, synchronously), and libnvme, *in the connecting process*, writes `/run/nvme/registry/nvme4/owner`.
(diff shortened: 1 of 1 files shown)


libnvme/REGISTRY.md
```diff
@@ -0,0 +1,119 @@
...
+- Directories are `0755`, attribute files `0644`: **world-readable, root-writable**. Both the root and per-device directories are created on demand on first write.
+- Writes are atomic (`mkstemp` → `fsync` → `rename`), so concurrent writers never corrupt an entry and readers never see a half-written value.
+- PCIe / apple-nvme controllers are out of scope — they are hardware-enumerated and already excluded from `disconnect-all` by transport-type checks.
+
+## Setting an owner
+
+An orchestrator declares its identity once, then every controller it connects through that context is registered automatically on a successful connect:
+
+```c
+struct libnvme_global_ctx *ctx = libnvme_create_global_ctx(stderr, log_level);
```
igaw: ```libnvme_create_global_ctx();```

libnvme/REGISTRY.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Automatic cleanup
+
+A registry entry must be removed when its controller goes away, or stale entries would accumulate and make a recycled device name look owned. The catch is that the kernel can tear a controller down on its own — for example when connectivity is lost and `ctrl-loss-tmo` expires — with no orchestrator involved. On a daemonless system there may be nothing watching to do the cleanup, so it cannot depend on any orchestrator running; it has to be driven by the device-removal event itself.
+
+libnvme ships a udev rule for exactly this. The udev daemon — present on essentially every system — removes a controller's registry entry when the kernel removes the controller:
+
+```
+ACTION=="remove", SUBSYSTEM=="nvme", KERNEL=="nvme[0-9]*", \
+    RUN+="/bin/sh -c '[ -e /dev/%k ] || rm -rf /run/nvme/registry/%k'"
```
igaw: IIRC, I asked this in the PR, shouldn't we use nvme-cli here to remove the entry? I would like to log everything in one place, so  `journalctl -u nvme-cli` (or whatever the command is) shows what happened. Having this only in a udev rule distributes the information and makes it hard to debug.

libnvme/REGISTRY.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+**Why the `[ -e /dev/%k ]` guard.** udev rules run asynchronously: udevd processes a `remove` uevent some time *after* the kernel emits it. Meanwhile the kernel is free to recycle the just-freed instance number — it allocates the lowest available id, so the very next controller to connect can be handed the same `nvmeN`. This opens a rare race: a controller is removed and, at almost the same instant, a new one is connected and inherits its id, all before udevd gets around to running the remove rule for the old controller. By the time that rule finally fires, `nvmeN` already refers to a *different*, live controller that has written its own registry entry — and a blind `rm` would clobber it.
```
igaw: The race window still exists. The only way to ensure this works is by serializing the events, which udevd does. So I don't think this argument is correct.

libnvme/REGISTRY.md
```diff
@@ -0,0 +1,119 @@
...
+
+The guard breaks the race by asking whether the device still exists at the moment the rule runs. devtmpfs removes a controller's device node *synchronously*, before the kernel emits `KOBJ_REMOVE`, so for the controller actually being removed `/dev/nvmeN` is already gone. If `/dev/nvmeN` is absent, the entry is genuinely stale and `rm` runs. If `/dev/nvmeN` exists, the id has been recycled to a new controller: `[ -e /dev/%k ]` is true, the `||` short-circuits, and `rm` is skipped — preserving the new owner's entry.
```
igaw: As I said, udevd ensures order of rule execution. That is if everything is run through udevd all is good. The problem starts if there is another instance observing udev events and does something. With the nvme-discoverd on the horizon, this udev rule could get really nasty. Just thinking loud, maybe all is good.

libnvme/REGISTRY.md
```diff
@@ -0,0 +1,119 @@
...
+
+| Invocation | Disconnects |
+|---|---|
+| `nvme disconnect-all` | only **unowned** controllers (safe default) |
+| `nvme disconnect-all --owner NAME` | only controllers owned by `NAME` (confirmation required) |
+| `nvme disconnect-all --force` | **all** fabric controllers regardless of ownership (confirmation required) |
+
+`--owner` and `--force` are mutually exclusive. Both prompt for a typed `yes` when stdin is a terminal; non-interactive callers (scripts) proceed, since passing the flag is itself the statement of intent.
+
+In every case PCIe and apple-nvme controllers are left alone — the transport-type check excludes non-fabric controllers before ownership (or `--force`) is even considered, so `--force` cannot reach them. This means `nvme disconnect-all --force` behaves exactly like the original `nvme disconnect-all` did: it tears down every fabric controller and never touches locally-attached devices. The ownership-aware default is the only new behavior layered on top.
```
igaw: I would just say controllers managed directly by the kernel (Memory Based Transport, PCIe) instead naming apple-nvme. This is just the driver name for a workaround. It's also PCIe transport

libnvme/REGISTRY.md
martin-belanger: Fair point on debuggability, but there's a structural problem with routing it through `nvme`. The command would be `nvme registry delete`, which lives in the **registry plugin** — and that plugin is **optional**, at your own request (you asked for the registry to be a plugin rather than a built-in). So a udev rule that shells out to `nvme registry delete` would make the registry's stale-entry cleanup depend on a command that isn't present on any system that didn't build/install that optional plugin. The cleanup is a property of the registry itself, which lives in **libnvme** — so it has to work whenever libnvme is installed, with no dependency on nvme-cli or an optional plugin, and no daemon. That's exactly why it's a libnvme-shipped udev rule.

It also wouldn't change the race: a command launched from the rule races the connect path identically — the `[ -e ]` guard is still what makes it correct. So we'd take on an optional-plugin dependency and gain nothing on correctness. 

The **debuggability** concern is legit on its own, though — happy to solve "see what happened in one place" without that dependency, e.g. have the rule log via `logger`, or add a libnvme log sink.

libnvme/REGISTRY.md
martin-belanger: I think the disconnect is one specific point: **the new controller's registry entry is not written by a udev rule — it's written by libnvme inside the `connect()` path**, synchronously in the connecting process, the moment the kernel assigns the (recycled) instance number. udevd never sees that write and never orders it. So udevd serializing udev *events* doesn't help here, because the actor that races the remove rule is the connect path, which is outside udev entirely.

Concretely, the dangerous interleaving for the name `nvme4`:

1. Kernel removes the old `nvme4` → frees instance 4 → emits a `remove` uevent (now queued in udevd).
2. Before udevd gets to that uevent, a new controller connects. The kernel allocates the lowest free instance = 4 again, creates `/dev/nvme4` (devtmpfs, synchronous), and **libnvme — in the connecting process, not via udev — writes `/run/nvme/registry/nvme4/owner`**.
3. udevd now runs the queued `remove` rule for the *old* `nvme4`. With no guard it does `rm -rf /run/nvme/registry/nvme4` and deletes the **new, live** controller's entry.

udevd processing `remove` before `add` in order does not prevent step 3 from clobbering step 2, because step 2's write never came through udev. The only information available at step 3 that distinguishes "stale entry for the controller that's really gone" from "live entry for a controller that recycled the name" is whether the device node currently exists — which is exactly what `[ -e /dev/%k ]` tests. devtmpfs removes the node *synchronously before* `KOBJ_REMOVE` is emitted (`device_del()` calls `devtmpfs_delete_node()` before `kobject_uevent(KOBJ_REMOVE)`), so:

- genuinely-gone controller → `/dev/nvme4` absent → `rm` runs (correct);
- recycled name → `/dev/nvme4` exists again → `rm` skipped (correct).

So the guard isn't a weaker substitute for serialization — it's the only mechanism that works, because the competing actor (the connect path) is not a udev event udevd can serialize. The residual window is the microsecond TOCTOU between the `[ -e ]` test and the `rm`; serialization can't close that either, for the same reason (the connect write isn't a udev event), and hitting it needs a connect ioctl to complete at that exact instant — negligible.

On the discoverd worry: this is actually why the guard is robust to it. Correctness depends on **kernel devtmpfs state**, not on udevd event ordering or on who is observing events. Another observer (e.g. discoverd) reconnecting a dropped controller just produces another connect → another `/dev/nvmeN` + libnvme entry; the `[ -e ]` check reflects that regardless of event timing. So it does not rely on serialization at all — which is the property we want once there are multiple observers.

libnvme/REGISTRY.md
martin-belanger: This also answers the natural follow-up — "if a udev rule removes entries, why doesn't one create them?" The two halves are placed where the information they need actually lives:

- **Create needs the `owner`** — and the owner is purely in-process state of the *connecting* process, passed to libnvme via the context (`ctx->owner`). It is not in sysfs, not in the `add` uevent, nowhere a udev rule could read it (the kernel has no concept of "owner" — it's a libnvme-userspace notion). A rule firing for `nvme4` has no way to know *who* connected it, so it cannot write a correct `owner=`. Only libnvme, in the connect path, has the owner — which is why the entry is written there. (It's also written synchronously inside `connect()`, so there is no window where a just-connected controller is briefly unowned and exposed to `disconnect-all`.)
- **Delete needs no `owner`** — it just removes the entry for that device name (with the `[ -e ]` recycling guard). No per-process knowledge is required, so a udev rule can do it.

So the asymmetry is deliberate, not an oversight: each operation lives where its required input is available.

libnvme/REGISTRY.md
martin-belanger: Done

libnvme/REGISTRY.md
martin-belanger: Done

libnvme/REGISTRY.md
igaw: - using nvme-cli to cleanup: I haven't though about the situation where libnvme has registry support enabled. It's a bit unlikely but still possible.
- race: ideally there should only be one single component which handles all these udev events, anything else will for certain lead to a race. This rule has potential to cause some headache. But without real logs and experiments, it's hard to say if it is relevant. Let's go with it and see if everything holds together. The proof of the pudding is in the eating.

libnvme/REGISTRY.md
igaw: Many thanks for the write up. Having the interleaving steps outlined like this helps a lot. It might make sense to add this to document.

Unfortunately, something I learned the hard way is that any race window will be hit eventually, no matter how small it is. It's just a matter of scale. 

I wonder if we could make use of the sequence counter in the udev event environment. E.g. when we create the registry based on a udev event and add it to the registry, the remove rule is only allowed to remove entries which are older then the current sequence number? Just thinking loud here.

libnvme/REGISTRY.md
martin-belanger: Right — that standalone case (libnvme with the registry enabled, but nvme-cli not installed) is exactly the deployment the libnvme-shipped udev rule is meant to cover. It is unlikely, but it is the reason cleanup has to live in libnvme and not depend on the optional `nvme registry` plugin: the registry is a libnvme feature, so its stale-entry cleanup must work whenever libnvme is installed, with no daemon and no nvme-cli. Putting the rule in libnvme keeps that guarantee.

On the debuggability point you raised earlier (one place to see what happened): I'd rather solve that without taking the nvme-cli dependency — e.g. have the rule log via `logger`, or add a libnvme log sink — so the "single place to look" property doesn't reintroduce the coupling.

martin-belanger: > ideally there should only be one single component which handles all these udev events, anything else will for certain lead to a race. This rule has potential to cause some headache. ... Let's go with it and see if everything holds together.

Agreed on the principle, and thanks for being willing to go with it. The reason I'm comfortable with multiple observers here is that the guard's correctness does not depend on single-observer serialization at all: it reads kernel devtmpfs state (`/dev/nvmeN`), not udev event ordering. Another observer (discoverd reconnecting a dropped controller) just produces another connect → another `/dev/nvmeN` + libnvme entry, and the `[ -e ]` test reflects that regardless of event timing. So adding observers doesn't add races to this rule

libnvme/REGISTRY.md
martin-belanger: > It might make sense to add this to document.

Done — I've added the numbered `nvme4` interleaving to the *Automatic cleanup* section of `REGISTRY.md`.

> any race window will be hit eventually, no matter how small it is. It's just a matter of scale.

Fair, and I won't argue the window away — with the `[ -e ]` guard alone the common recycling case is safe and only a sub-shell TOCTOU remains, but rather than rely on it being tiny, let's close it.

> I wonder if we could make use of the sequence counter in the udev event environment.

Yes — and I want to adopt this; it closes the residual window cleanly. The one refinement needed is that the entry isn't created by a udev event (it's written by libnvme in the connect path, because only that process knows the owner), so we can't simply compare two udev sequence numbers. Instead, *stamp* the entry:

- At connect, libnvme reads `/sys/kernel/uevent_seqnum` (monotonic, global) and records it in the entry, e.g. a `seqnum` attribute. Because the read happens after the connect ioctl returns, the stored value is ≥ this controller's own `add`-event seqnum.
- The remove rule removes the entry only if its own `$SEQNUM` (from the remove event's environment) is greater than the stored `seqnum`.

By monotonicity this is timing-independent: a *stale* `remove` for the old incarnation of `nvme4` was emitted before the new incarnation's `add`, so `old_remove.SEQNUM < new_entry.seqnum` → the rule skips → the new entry survives. A *legitimate* remove of the current controller has `remove.SEQNUM > add.SEQNUM ≥ stored` → the rule removes. Correct in both directions, with no dependence on the instantaneous `/dev` node state.

I'd keep the `[ -e /dev/%k ]` test as a cheap fast path and add the seqnum comparison as the tiebreaker that closes the TOCTOU. This also preserves the create/remove asymmetry — creation still lives in libnvme; only the staleness test changes from "does the node exist right now" to "is this remove older than the entry."

This is a code change (libnvme writes the seqnum; the rule grows a numeric compare) beyond this doc-only PR, so I'll do it as a follow-up and update `REGISTRY.md` to describe the seqnum tiebreaker once it lands.

libnvme/REGISTRY.md
igaw: If I understand the whole situation correctly, the whole design with the ioctl interface, reuse of ids, asynchronous event handling wont allow us to get completely race free. But these interfaces/API already exist and we have to deal with them until we find a better way. I don't want to open this can of worms right now. Thus we make it as good as possible and go from there.

martin-belanger: > won't allow us to get completely race free ... make it as good as possible

Agreed, and I'm not trying to make the whole thing provably race-free. The substrate we're built on — the connect/disconnect ioctls, lowest-free instance-number reuse, and asynchronous uevent delivery — has inherent races we can't paper over from userspace, and I don't want to open that can of worms here either. "As good as possible and go from there" is the right framing.

For the one race we were actually discussing, though — a stale REMOVE clobbering a recycled-and-live entry — your seqnum idea closes it completely, and I've submitted it as #3479. The point is that the comparison is on monotonic emission order, not timing: a stale REMOVE for the old controller was emitted before the new controller's ADD, so its SEQNUM is always lower than the new entry's stamp, and the rule skips it regardless of how the `[ -e ]` test and `rm` interleave. So that specific window isn't merely made smaller — it's gone.

I tested it on real hardware (kernel nvmet TCP target, real `nvme disconnect` events, the rule installed where udev actually reads it):

- normal stamp → entry removed (legitimate cleanup);
- stamp set above any future SEQNUM → entry preserved even though the device node was already gone — this is the stale-REMOVE case, and it can only happen if udev substituted a real SEQNUM and the comparison held, so it confirms both the runtime substitution and the guard in one shot;
- no stamp → entry removed via the `[ -e ]` fallback.

#3479 keeps `[ -e /dev/%k ]` as the cheap fast path and adds the seqnum compare as the tiebreaker that closes the TOCTOU (and updates the rule's comment to explain both checks). The approach is yours — it carries a `Suggested-by:` you — thanks for the nudge to not settle for "small."

Side note, since you mention finding a better way: the race here is really forced by the kernel API — there's nowhere to attach connection metadata that lives and dies with the controller, so we keep it in a separate file and reconcile after the fact. The clean fix would be kernel-side: let the connect path carry a free-form tag (effectively the `owner`) and have the kernel echo it back in sysfs, bound to the controller's lifecycle. I floated something along those lines on linux-nvme a while back and it didn't gain traction at the time. So this userspace registry + udev cleanup is making the best of the interfaces we have — exactly your point. A kernel-side tag would make the registry file, the cleanup rule, and this seqnum guard all unnecessary, if the idea ever finds a better moment.

libnvme/REGISTRY.md
martin-belanger: Answered with previous comment. Marking as resolved.

martin-belanger:
@igaw — I force-pushed an updated REGISTRY.md that documents your proposed seqnum safety net. The code that implements it is still pending in #3479, so to be chronologically precise, #3479 should land before this PR is merged.

libnvme/REGISTRY.md
igaw: Nice! Let me review it.

If this solution works, I don't think we need to try to push for a kernel side registry. But if it turns out we can't fix it in user space, we have at least a something to show.

martin-belanger:
@igaw — I rebased onto current master and updated the doc to match the code that landed in #3479: it now describes the `seqnum` tiebreaker and the atomic entry creation/removal, and the udev rule snippet matches the shipped rule.

igaw:
Ah  sorry for the mess with the merge order. I was excited that we figure it out :)

PR #3468, opened 2026-06-22
libnvme: test that registry attribute values round-trip verbatim

martin-belanger:
Attribute values are free-form -- only device and attribute names are validated, never the value. Add a test that stores a value containing spaces, double quotes, an apostrophe and punctuation, then confirms it reads back byte-for-byte, so this guarantee cannot regress through accidental escaping or sanitization of the value.


igaw:
Thanks!

PR #3469, opened 2026-06-22
nbft: fix sizeof operand mismatch in pointer array allocations

bellis-ibm:
Four calloc() calls used sizeof(struct) instead of sizeof(struct *)
when allocating pointer arrays. The hfi_list, security_list,
discovery_list, and subsystem_ns_list objects are declared as
pointer arrays and should be allocated using the pointer type size.

Use sizeof(struct *) to match the array element type and resolve the
static analysis warnings.


igaw:
Thanks, looks good. I've updated the commit message, so it doesn't have an empty line every second line.

image

PR #3470, opened 2026-06-22
Fix uninitialized variables

fhassan-ibm:
The result variable in get_feature_id_changed() and the flags variable in passthru() were uninitialized before use, which could cause undefined behavior. Initialize both to 0 to prevent this.


igaw:
Thanks!

PR #3471, opened 2026-06-22
libnvme: fix uninitialized argument in libnvme_get_lba_status_log()

fhassan-ibm:
The log argument was passed uninitialized to nvme_init_get_log_lba_status(). Pass the allocated buf instead to prevent undefined behavior.


igaw:
Good catch! Thanks!

PR #3472, opened 2026-06-22
Use persistent discovery controller by default

Mr-Bossman:


igaw:
The idea of https://github.com/linux-nvme/nvme-cli/issues/3112 is to make the DC persistent based on the info from the DC. The global setting is basically wrong. It should always be off by default and only when the DC announces it wants be used as persistent, nvme-cli should do it.

Mr-Bossman:
@igaw is more along the lines of what the feature request meant?

igaw:
Yes as far I can tell from a quick glance. Did you have a chance to test this against a target? 

@martin-gpy in case you have a cycle for testing, I would appreciate it!

Mr-Bossman:
@igaw Yes i did and it does seems to work as expected. Though one thing that doesn't make sense to me is why `_nvmf_discovery`, which uses very similar code, checks for `NVMF_DISC_EFLAGS_DUPRETINFO` and if `subnqn === NVME_DISC_SUBSYS_NAME`. For both arrays that have EPCSD i tested on the `NVMF_DISC_EFLAGS_DUPRETINFO`  is set for all the discovery entry. And for one the  `subnqn` is not `NVME_DISC_SUBSYS_NAME`

hreinecke:
> The idea of #3112 is to make the DC persistent based on the info from the DC. The global setting is basically wrong. It should always be off by default and only when the DC announces it wants be used as persistent, nvme-cli should do it.

Well, thing it that the decision to create a persistent discovery controller is pretty much an admin choice. For some fabrics (like FC) creating a PDC is pretty much pointless as the fabrics always knows which ports are attached, and will always be notified if new ports are accessible. And creating a PDC eats up resources on the target, so admins might decide not to use them to save resources on the target. So yes, an admin-settable option is still required.

igaw:
Oops, pushed the wrong button. Sorry.

Mr-Bossman:
@igaw is there a way to generate the docs and the accesors for `libnvme_host`?

hreinecke:
Why is the 'pdc supported' flag tagged to the host? It is the subsystem (well, the description of the subsystem in the discovery log page) which specifies whether PDC is supported. I think it would be better to move the flag to the subsystem...

Mr-Bossman:
> Why is the 'pdc supported' flag tagged to the host? It is the subsystem (well, the description of the subsystem in the discovery log page) which specifies whether PDC is supported. I think it would be better to move the flag to the subsystem...

Same reason `libnvme_host_set_pdc_enabled` exists. This flag only exists for discovery controllers which means the `nqn` is not known. To have it in the subsystem would require the json file to allow an empty `nqn` parameter

igaw:
fwiw, I think with your change we can drop the global pdc setting. From what I learned in this discussion, it never made sense.

hreinecke:
The current implementation is the result of the request to have a compile-time setting on whether PDC should be created per default or not. But yes, I do agree

> > Why is the 'pdc supported' flag tagged to the host? It is the subsystem (well, the description of the subsystem in the discovery log page) which specifies whether PDC is supported. I think it would be better to move the flag to the subsystem...
> 
> Same reason `libnvme_host_set_pdc_enabled` exists. This flag only exists for discovery controllers which means the `nqn` is not known. To have it in the subsystem would require the json file to allow an empty `nqn` parameter

But then we should rather have this as an option in 'discovery.conf' ...

fabrics.c
```diff
@@ -322,13 +331,17 @@ static int set_fabrics_options(struct libnvmf_context *fctx,
...
+	struct libnvme_host *h = NULL;
 	struct nvmf_args fa;
 	char *ptr, *p;
 	static char line[4096];
 	int argc, ret = 0;
 	bool force = false;
 
 	NVMF_ARGS(opts, fa,
+		  OPT_FLAG("epcsd",          0, &epcsd_enabled,       "Enable persistent discovery for controllers that support it"),
+		  OPT_FLAG("no-epcsd",       0, &epcsd_disabled,      "Disable persistent discovery for controllers that support it"),
```
igaw: There is `argconfig_parse_seen(opts, "epcsd")` which returns true if `epcsd` had been set by the user. This would avoid the `no-epcsd` I think. but i have to head around the whole epcsd thingy, again something simple which is utterly complex in the details...

Mr-Bossman:
> fwiw, I think with your change we can drop the global pdc setting. From what I learned in this discussion, it never made sense.

If it is enabled there is no way to disable it too. Imo it's good to keep as the logic for it makes significantly more sence than the logic that uses the fabric context. Though having the ability to set the default value probably should be removed

Mr-Bossman:
> The current implementation is the result of the request to have a compile-time setting on whether PDC should be created per default or not. But yes, I do agree
> 
> > > Why is the 'pdc supported' flag tagged to the host? It is the subsystem (well, the description of the subsystem in the discovery log page) which specifies whether PDC is supported. I think it would be better to move the flag to the subsystem...
> > 
> > 
> > Same reason `libnvme_host_set_pdc_enabled` exists. This flag only exists for discovery controllers which means the `nqn` is not known. To have it in the subsystem would require the json file to allow an empty `nqn` parameter
> 
> But then we should rather have this as an option in 'discovery.conf' ...

I added it as both. as of now there isn't really a way to have global settings in `discovery.conf`. There was talks on matrix about having the ability to have global configs, which would remove the need for a compile tile setting.

fabrics.c
Mr-Bossman: well its to override the default setting in `config.json` or the build variable, so there should be two options.

igaw: okay, now I am confused. Let met try to figure this out...

igaw: ah just saw what we do for PDC already. this is just doing the samething.

fabrics.c
```diff
@@ -361,6 +379,13 @@ static int hook_parser_next_line(struct libnvmf_context *fctx, void *user_data)
 	libnvmf_context_set_discovery_hooks(fctx, hook_discovery_log,
 		hook_parser_init, hook_parser_cleanup, hook_parser_next_line);
 
+	libnvme_get_host(hfd->ctx, fa.hostnqn, fa.hostid, &h);
+	if (epcsd_enabled)
+		libnvme_host_set_epcsd_enabled(h, true);
+
+	if (epcsd_disabled)
+		libnvme_host_set_epcsd_enabled(h, false);
```
igaw: tempted to argue for

```c
if (epcsd_enabled || epcsd_disabled)
	libnvme_host_set_epcsd_enabled(h, epcsd_enabled);
```

but this might be a bit too obfuscated...

libnvme/src/nvme/fabrics.c
```diff
@@ -2351,6 +2351,20 @@ static int _nvmf_discovery(struct libnvme_global_ctx *ctx,
...
+	for (int i = 0; i < numrec; i++) {
+		struct nvmf_disc_log_entry *e = &log->entries[i];
+		uint16_t eflags = le16_to_cpu(e->eflags);
+
+		if ((e->subtype == NVME_NQN_DISC ||
+		    e->subtype == NVME_NQN_CURR) &&
+		    (eflags & NVMF_DISC_EFLAGS_EPCSD) &&
+		    libnvme_host_is_epcsd_enabled(h, DEFAULT_EPCSD_ENABLED)) {
+			libnvmf_context_set_persistent(fctx, true);
+			break;
```
igaw, line 2622: could you move this part into the existing loop where we already have the `NVMF_DISC_EFLAGS_EPCSD` check?

igaw:
Looks good. Just a couple of nitpicks.

Could add a prefix to the subject, e.g. "fabrics: ..." and drop the links to github. The commit message should contain everything it needs to make it self explenatory. We might have to move eventually to a different hosters and then all those links are pretty useless. Thanks!

hreinecke:
I wonder: do we really need this complexity? IE is it really worth it to have _two_ additional settings?
We already have a global setting 'pdc enabled'; why do we need _another_ setting 'pdc enabled if supported'?
If 'pdc enabled' is set, why do we need to set 'pdc enabled if supported' additionally? (We already said we want a PDC with the first option ...) And if PDC is not supported, why should we allow the user to create a PDC?

igaw:
`-Dpdc-enabled` sets the default in libnvmf_context->persistent to `true`. 

```c
			/*
			 * Are we supposed to keep the discovery
			 * controller around?
			 */
			disconnect = !nfctx.persistent;

			if (strcmp(e->subnqn, NVME_DISC_SUBSYS_NAME)) {
				/*
				 * Does this discovery controller doesn't
				 * support explicit persistent connection?
				 */
				if (!(eflags & NVMF_DISC_EFLAGS_EPCSD))
					disconnect = true;
				else
					disconnect = false;
			}
```

Thus `disconnect` will be `false` on for any DC with an unique nqn. I suppose we could reuse `-Dpdc-enabled` and make the logic work correctly?

hreinecke:
> `-Dpdc-enabled` sets the default in libnvmf_context->persistent to `true`.
> 
> ```c
> 			/*
> 			 * Are we supposed to keep the discovery
> 			 * controller around?
> 			 */
> 			disconnect = !nfctx.persistent;
> 
> 			if (strcmp(e->subnqn, NVME_DISC_SUBSYS_NAME)) {
> 				/*
> 				 * Does this discovery controller doesn't
> 				 * support explicit persistent connection?
> 				 */
> 				if (!(eflags & NVMF_DISC_EFLAGS_EPCSD))
> 					disconnect = true;
> 				else
> 					disconnect = false;
> 			}
> ```
> 
> Thus `disconnect` will be `false` on for any DC with an unique nqn. I suppose we could reuse `-Dpdc-enabled` and make the logic work correctly?

That would be preferable. I would think the decision matrix should be:

-> Does the controller support explicit persistent connections (ie NVMF_DISC_EFLAGS_EPCSD is set)?
    -> No -> No PDC
-> Does the subsystem have the 'pdc_enabled' flag specified?
    -> No -> Use default.
    -> Yes -> Use PDC

This requires unique discovery subsystems to work correctly. Which is fine as I would not create a PDC for the default discovery NQN; the kernel will create a new discovery controller for _every_ default discovery NQN, so we're having a hard time identifying which of the tons of discovery subsystems we should be using.

libnvme/src/nvme/fabrics.c
Mr-Bossman:
```c
if (!connect)
		return 0;
```
is before the loop which would cause this to only work for the connect sub-command and not discover

igaw:
Thanks a lot!

PR #3473, opened 2026-06-23
Fix two Command Specific Status Codes

MaisenbacherD:


igaw:
Excellent stuff. Thanks!

PR #3475, opened 2026-06-23
micron: Add Windows support to Micron plugin commands

brgoing:
Adds Windows support to Micron plugin commands and updates plugin version to 2.1.0.
Adds new micron-utils files for implementation of utility methods that have platform-specific implementations.

Updated functionality:
* Adds cross-platform support for getting the drive model.
* Adds cross-platform handling of PCIe-register-based correctable error functionality. Windows does not support accessing PCIe registers and will return an error if correctable error commands are used.
* Adds cross-platform support for logging OS configuration information.
* Fixes Windows .zip file generation.


igaw:
Thanks!

PR #3476, opened 2026-06-24
nvme: fix ignored interruption in format command

g0l4:
Add nvme_sigint_received wrapper around sleep, so it will exit instead of continue when a SIGINT is received.


igaw:
Thanks!

PR #3477, opened 2026-06-24
libnvme: update diag attribute paths

shroffni:
The upstream kernel commit 37afebc79a11 ("nvme: add diag attribute group under sysfs") groups all diag attributes under a new directory /sys/block/.../diag/.

Update the libnvme diag attribute accessors to use the new paths. The following attributs path have been updated:
- command_retry_count
- command_error_count
- multipath_failover_count
- reset_count
- reconnect_count

In addition, the following attributes were renamed as well as their path were also changed. So update the corresponding attribute names and paths for:
- io_requeue_no_usable_path_count
- io_fail_no_available_path_count

These attributes are used by the nvme-top dashboard to display real-time controller and multipath diagnostic statistics.


shroffni:
I will resubmit this PR again. It seems I also need to update (auto-generated) accessors for the renamed attributes. Also it needs corresponding changes in nvme-cli which uses those accessors.

PR #3478, opened 2026-06-24
nvme top diag attributes

shroffni:
The diag attributes are used by the nvme-top dashboard to display real-time controller and multipath diagnostic statistics.

The upstream kernel commit 37afebc79a11 ("nvme: add diag attribute group under sysfs") now groups all diag attributes under a new directory /sys/block/.../diag/. So we need to also make relevant changes in libnvme which export these diag attributes .

The first patch updates the diag attribute paths.  In addition few other diag attributes were renamed and their paths were also changed. So second patch renames those attributes and update their paths.


igaw:
Thanks!

PR #3479, opened 2026-06-24
libnvme: guard registry cleanup against instance reuse with uevent_seqnum

martin-belanger:
## Background

The controller ownership registry under `/run/nvme/registry/` records which orchestrator owns each connected NVMe-oF controller. Entries are written by libnvme on connect (only the connecting process knows the `owner`), and removed by a shipped udev REMOVE rule when the kernel removes the controller. The rule guards the removal with `[ -e /dev/%k ]` so that a recycled instance number — the kernel hands out the lowest free id, so a freed `nvmeN` can be immediately reused — is not clobbered: if the device node exists again, a new controller has already claimed the name and written its own entry, and the stale REMOVE must leave it alone.

## The problem this fixes

The `[ -e /dev/%k ]` check handles the common recycling case, but it leaves a residual sub-shell TOCTOU: a new controller can create `/dev/nvmeN` and write its entry in the window between the `[ -e ]` test returning false and `rm` running, after which the stale REMOVE deletes a live entry. The window is tiny, but "tiny" is not "closed" — at scale it will eventually be hit.

## The fix

Stamp each registry entry at connect with the kernel's global, monotonic `/sys/kernel/uevent_seqnum`, and have the REMOVE rule delete an entry only when the REMOVE event's own `SEQNUM` exceeds that stamp.

This closes the window completely, because it replaces a timing-sensitive test with a monotonic ordering comparison:

- A stale REMOVE for the old controller was emitted *before* the new controller's ADD, so its `SEQNUM` is necessarily lower than the new entry's stamp — the comparison holds and the entry survives, regardless of scheduling or how the test and `rm` interleave.
- A genuine REMOVE of the current controller carries a `SEQNUM` higher than that controller's own connect-time stamp, so its entry is removed as intended.

The stamp is written before `owner` so an owned entry always carries the guard value; a missing stamp (e.g. an entry written by an older libnvme) reads as `0` and the rule falls back to the device-existence check alone. The `[ -e /dev/%k ]` check is kept as a cheap fast path; the seqnum comparison is the tiebreaker that closes the race.

The stamp is best-effort on the write side: if `/sys/kernel/uevent_seqnum` cannot be read or written, the connect is never failed over it — the rule simply falls back to the device-existence guard.

## Testing

Validated end-to-end against a kernel nvmet TCP target with four fabric controllers, using the rule installed where udev actually reads it and entries stamped by this build of libnvme:

- Real `nvme disconnect` of a controller with a normal stamp → kernel-driven REMOVE → entry removed (remove `SEQNUM` greater than the stamp).
- Real disconnect of a controller whose stamp was set to a value above any future `SEQNUM` → entry preserved. The device node was gone, so this can only happen if the rule evaluated `[ <SEQNUM> -le <stamp> ]` to true — which simultaneously proves that udev substitutes a real, non-empty `SEQNUM` at runtime and that a "stale" lower-SEQNUM REMOVE does not clobber a live entry.
- Real disconnect of a controller with no stamp → entry removed via the device-existence fallback.
- A controller left connected kept its entry.

Also checked: the libnvme unit test (`libnvme - registry`) asserts the stamp is written and numeric, `udevadm verify` accepts the rule, and checkpatch is clean.

### DIFF
--- a/libnvme/src/nvme/registry.c
+++ b/libnvme/src/nvme/registry.c
@@ -343,29 +343,184 @@ static int write_device_attr(const char *device, const char *attr,
 }
 
 /*
+ * Read the kernel's global uevent sequence number from
+ * /sys/kernel/uevent_seqnum.  This monotonic counter is used to stamp a
+ * registry entry at connect time so the udev REMOVE rule can tell a stale
+ * entry apart from a recycled-and-live one without relying on event ordering
+ * (see 70-nvmf-registry.rules).  Returns a newly allocated decimal string the
+ * caller must free, or NULL on any failure -- the stamp is best-effort, and a
+ * missing stamp simply makes the rule fall back to its device-existence guard.
+ */
+static char *read_uevent_seqnum(void)
+{
+	char buf[32];
+	ssize_t n;
+	int fd;
+
+	fd = open("/sys/kernel/uevent_seqnum", O_RDONLY | O_CLOEXEC);
+	if (fd < 0)
+		return NULL;
+	n = read_full(fd, buf, sizeof(buf) - 1);
+	close(fd);
+	if (n <= 0)
+		return NULL;
+
+	/* Strip the trailing newline the kernel appends. */
+	while (n > 0 && (buf[n - 1] == '\n' || buf[n - 1] == '\r'))
+		n--;
+	if (n == 0)
+		return NULL;
+	buf[n] = '\0';
+
+	return strdup(buf);
+}
+
+/*
+ * Remove a registry entry directory and its (flat) attribute files.  Shared by
+ * libnvmf_registry_delete() and the create_instance() error/cleanup paths.
+ * Returns 0 on success, -ENOENT if @path does not exist, or a negative errno.
+ */
+static int delete_dir(const char *path)
+{
+	struct dirent *de;
+	int dir_fd, ret = 0;
+	DIR *d;
+
+	d = opendir(path);
+	if (!d)
+		return -errno;
+
+	dir_fd = dirfd(d);
+	while ((de = readdir(d)) != NULL) {
+		if (de->d_name[0] == '.')
+			continue;
+		if (unlinkat(dir_fd, de->d_name, 0) < 0 && errno != ENOENT)
+			ret = -errno;
+	}
+	closedir(d);
+
+	if (ret)
+		return ret;
+
+	if (rmdir(path) < 0 && errno != ENOENT)
+		return -errno;
+	return 0;
+}
+
+/*
+ * Atomically remove a *visible* registry entry: rename it to a hidden sibling
+ * so it vanishes from observers in a single step, then purge the renamed copy
+ * at leisure.  This keeps a reader (nvme registry list, another orchestrator)
+ * from ever seeing a half-emptied entry.  Returns 0 on success, -ENOENT if
+ * @path does not exist, or a negative errno.
+ *
+ * The udev REMOVE rule deletes entries with a plain "rm -rf", which is not
+ * atomic; this guarantee therefore covers only library-driven deletes.
+ */
+static int delete_dir_atomic(const char *path)
+{
+	__cleanup_free char *trash = NULL;
+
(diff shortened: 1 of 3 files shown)


martin-belanger:
JFYI: I believe that the musl/Alpine build failure is due to build container changes.

libnvme/sysfiles/udev/rules.d/70-nvmf-registry.rules
```diff
@@ -5,9 +5,21 @@
...
+#      emitted before the new controller's ADD, so its SEQNUM is lower than the
+#      new entry's stamp; removing only when SEQNUM is greater closes the
+#      sub-shell race between the [ -e ] test and rm.  A missing stamp reads as
+#      0, falling back to check 1 alone.
+#
+# $$ yields a literal '$' for the shell; $env{SEQNUM} is the REMOVE event's
+# sequence number, substituted by udev.
 ACTION=="remove", SUBSYSTEM=="nvme", KERNEL=="nvme[0-9]*", \
-    RUN+="/bin/sh -c '[ -e /dev/%k ] || rm -rf /run/nvme/registry/%k'"
+    RUN+="/bin/sh -c '[ -e /dev/%k ] || [ $env{SEQNUM} -le $$(cat /run/nvme/registry/%k/seqnum 2>/dev/null || echo 0) ] || rm -rf /run/nvme/registry/%k'"
```
igaw, line 25: I think this reduces the race window but there is still one: The library has created the directory but hasn't  written the seqnum attribute file. In this case the test evaluates to true and the shell continues with deleting which could now remove stuff added by the library in the meantime. 

I wonder if we could first create a new temp directory, populate it and then let it appear atomically.

martin-belanger: The temp dir + atomic rename is a great idea. Let me add that to this PR.

martin-belanger: > I wonder if we could first create a new temp directory, populate it and then let it appear atomically.

Good catch — you're right, and that's exactly the fix. I've made entry creation atomic: `create_instance` now builds the entry in a temporary directory (`.nvmeN.tmp.XXXXXX`), writes `seqnum` and `owner` into it, and then `rename()`s it into place. A directory rename is atomic, so the entry is only ever observed **complete or absent** — never as a directory that exists but is missing its `seqnum` stamp. That removes the window where the rule could read `seqnum` as `0` and delete the entry the library is still populating.

I also made the symmetric case atomic: removing a live entry now renames it to a hidden sibling and purges that copy, so a reader (`nvme registry list`, another orchestrator) never sees a half-emptied entry either. One honest caveat: the udev rule's own `rm -rf` is non-atomic by nature, so this guarantee covers library-driven deletes; the rule's recursive delete of a single entry that is already going away is benign (no live entry is at risk).

Both are folded into the existing commit (force-pushed). The rule itself is unchanged — the fix lives entirely in the library's create/delete paths.

Verified on real hardware as before (kernel nvmet TCP, real `nvme disconnect`): legitimate REMOVE deletes the entry, a stale REMOVE with a SEQNUM below the stamp preserves it, and a missing stamp falls back to `[ -e ]`. The unit test gained checks that the entry carries a numeric stamp and that no temporary directory is left behind after a create.

libnvme/src/nvme/registry.c
```diff
@@ -342,30 +342,187 @@
...
+	if (asprintf(&trash, "%s/.trash.XXXXXX", registry_dir()) < 0)
+		return -ENOMEM;
+	if (!mkdtemp(trash))
+		return -errno;
+
+	/*
+	 * On failure nothing moved into the temp dir, so drop it.  The common
+	 * case is @path being absent (-ENOENT), e.g. no stale entry to replace.
+	 */
+	if (rename(path, trash) < 0) {
```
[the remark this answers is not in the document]
martin-belanger: Fixed — removed the `access()` pre-check that triggered this. The existence check was only an optimization; `rename()` is atomic and already handles the absent case (ENOENT), so there's no longer a check-then-use on the path.

igaw:
I think this should do the trick. Thanks a lot with my nitpicking here.

PR #3480, opened 2026-06-24
tree: minor refactor of subsystem init

bcapener:
Move libnvme_init_subsystem from tree.c to tree-win.c and tree-linux.c.


igaw:
Thanks!

PR #3487, opened 2026-06-24
nvme: fix inverted null check in is_ns_mgmt_support()

fhassan-ibm:
The is_ns_mgmt_support() function allocates @ctrl and checks the result before calling nvme_identify_ctrl(). However, the null check was inverted, causing the function to return early on successful allocation and continue with a NULL @ctrl.

Accessing @ctrl->oacs with a NULL pointer causes a crash.

Fix the inverted null check to correctly return on allocation failure and only proceed when @ctrl is valid.


igaw:
Uff, good catch! Thanks!

PR #3490, opened 2026-06-25
plugins/ymtc: update additional smart display info for YMTC PE511

Robin-YMTC:


igaw:
Please add the prefix "plugins/ymtc" to the commit message. Also rebase it to the current master instead merging the master into your branch.

Obviously, it also this should build without failure.

Thanks!

PR #3492, opened 2026-06-25
python-ci: run the unit tests a second time without valgrind

martin-belanger:
The build matrix runs the test suite under valgrind (`build.sh -x`), which skips the fork-based tests: the python `multiprocessing` registry test is `@unittest.skipIf(_under_valgrind)`, because forked children misbehave under valgrind. Nothing else exercised that path, so a regression in it — for example Python 3.14 making `forkserver` the default `multiprocessing` start method — could land unnoticed.

This adds a single job that builds on `ghcr.io/linux-nvme/fedora:latest` and runs the suite without valgrind (no `-x`), so the skipped tests execute. Fedora tracks the newest Python, where interpreter-default changes surface before the other CI distros.

Notes:

- A dedicated job rather than a second `meson test` step across the build matrix: the un-skipped tests are Python-only and their outcome depends on the Python runtime, not on the compiler or buildtype, so one build on the newest-Python image gives the coverage without multiplying CI cost. The C side continues to run across the full matrix under valgrind.
- Uses the pre-built `ghcr.io/linux-nvme/fedora:latest` container, which now tracks the latest Fedora release (and its Python), so no manual dependency install is needed.

Verified locally that the gap is real and the new pass closes it: `test_parallel_writes_no_corruption` is skipped when `VALGRIND_OPTS` is set and runs otherwise.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,6 +70,40 @@ jobs:
           path: |
             .build-ci/meson-logs/*.txt
 
+  # Second test pass: the build matrix above runs the suite under valgrind,
+  # which skips the fork-based tests (e.g. the python multiprocessing registry
+  # test is skipIf(_under_valgrind)).  Run the suite once more WITHOUT valgrind
+  # so those tests actually execute.  fedora:latest tracks the newest Python,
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3493, opened 2026-06-25
build: use fedora container for analyzer workflow

shroffni:
Run the analyzer workflow inside the Fedora container image instead of the Debian image. Fedora ships a newer Clang Static Analyzer release, allowing the workflow to use the latest analyzer improvements.


igaw:
Thanks!

I'll trigger a run now.

PR #3497, opened 2026-06-25
nvme: fix potential use of uninitialized flags in passthru()

fhassan-ibm:
The passthru() command derives the file open mode from the data transfer direction encoded in the opcode. However, not all opcodes imply a data transfer direction, leaving the local @flags variable uninitialized.

If an input or metadata file is specified for such commands, nvme_open_rawdata() may be called with an uninitialized @flags value.

Initialize @flags to an invalid value and fall back to the optional user-specified direction provided through "--read" or "--write" when the direction cannot be inferred from the opcode. Reject input or metadata files if no data transfer direction can be determined and log the appropriate error.


igaw:
Nice catch! Thanks!

PR #3499, opened 2026-06-25
libnvme/json: store the config as an object, not a bare top-level array

martin-belanger:
`config-schema.json` documents the libnvme config file as an object with a `hosts` array, but the code actually read and wrote a bare top-level array of hosts. The schema and the implementation had simply drifted apart, and nothing in the test suite caught it.

This series fixes that, plus a couple of pre-existing bugs found along the way, and adds the schema-validation coverage that was missing.

The format change makes the writer emit the object form the schema already describes:

    { "hosts": [ ... ] }

The reader accepts both: an object (it reads the `hosts` array) or a legacy bare array (it reads it directly). Only the object form is ever written, so a pre-3.0 config is migrated to the new layout the next time it is written out (e.g. `nvme config -U`). No separate migration tool or install-time step is needed, which keeps it safe on immutable/image-based systems where package scriptlets can't reliably touch the live `/etc`. Supporting both formats on read is ~15 lines.

Beyond matching the schema, moving to an object gives the config a place to carry future top-level/global keys (e.g. a global `owner`, `pdc`, etc.) that a bare array could not — a follow-up will add `owner` there.

Commits:

- `libnvme/json: free the subsystem object when skipping discovery subsystems` — `json_update_subsys()` allocated a json_object up front but returned early for discovery subsystems without releasing it, leaking one object per discovery subsystem on every dump.
- `libnvme/json: fix misplaced "required" keys in the config schema` — in the host and subsystem definitions, `"required"` was nested inside `"properties"`, where JSON Schema treats it as an ordinary property rather than a constraint, so the schema never actually required `hostnqn`/`nqn`. Moved to the object level, matching the port definition.
- `libnvme/json: write the config as an object, keep reading the legacy array` — the format change above, the converted/regenerated test fixtures, and a new `config-schema-validate` test.

Notes:

- The config tests use the existing read→dump→diff harness; `config-pcie-with-tcp-config.json` is intentionally left as a legacy bare array so the suite exercises the backward-compatible read path (its expected dump is the new object form, so it also doubles as a migration check). `config-pcie.json`, previously a 0-byte file the reader rejected, becomes a valid empty config.
- The new `config-schema-validate` test validates the dumped fixtures against `config-schema.json` so a future drift between the dump format and the schema fails CI. It is skipped where the Python `jsonschema` module is unavailable.
- The series is bisectable: each commit builds and the test suite passes at every step.


libnvme/src/nvme/json.c
```diff
@@ -354,8 +354,10 @@ static void json_update_subsys(struct json_object *subsys_array,
 	struct json_object *port_array;
 
 	/* Skip discovery subsystems as the nqn is not unique */
-	if (!strcmp(subsysnqn, NVME_DISC_SUBSYS_NAME))
+	if (!strcmp(subsysnqn, NVME_DISC_SUBSYS_NAME)) {
+		json_object_put(subsys_obj);
 		return;
+	}
```
igaw, line 360: The usual way to do this, do the allocation after the unique test. But I see there is more of this pattern already. Not to mention we don't check if is a valid pointer. 

Let's fix this after this series. Rather not mixing too many things into one series.

igaw:
Nice! Thanks a lot for fixing this up!

PR #3500, opened 2026-06-25
nbft: capture errno immediately after I/O failures

bellis-ibm:
The libnvmf_read_nbft() function uses errno in error messages after
failed I/O operations. However, intervening function calls may overwrite
errno before it is read, leading to incorrect error reporting.

If fseek(), ftell(), or fread() fails, libnvme_msg() or fclose() may be
called before errno is captured, potentially overwriting the original
error value.

Capture errno immediately after each failed I/O operation before any
other function calls. Also add missing error check for ftell() which
can return -1L on failure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ tests/*.pyc
 dist/
 
 .vscode/
+*.save
+*.save.1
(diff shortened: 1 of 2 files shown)


igaw:
-  Two unrelated files are added 
- The commit message could contain the information which is  in this PR, but ...
- ...I don't see why errno needs to be saved at all. `ibnvme_strerror(errno)` will be executed first before anything else and then we always return `-EINVAL`. 

The additional check for `ftell` make sense.

PR #3503, opened 2026-06-26
libnvme/test: fix exit-code capture in config-diff.sh

martin-belanger:
The test runner wrapped the command in "if ! cmd; then rc=$?". Because $? reflects the result of the negated expression rather than the command itself, it was always 0 when the branch was taken: a test binary that genuinely failed was reported as passing, masking real command-level failures.

Run the command first, capture its status, then test it, so the real exit code is preserved and propagated.


igaw:
Thanks!

PR #3504, opened 2026-06-26
libnvme/json: add registry owner to the json schema

martin-belanger:
Let the JSON config file carry the orchestrator identity — the `owner` stamped on connections in the controller ownership registry — so that connections made from a config file are owned without naming the owner on every command line.

The motivating case is `nvme connect-all -J <config>`: the config already describes the hosts/subsystems/ports to connect to, and it should be able to describe who owns the resulting connections too. The owner is a global-context property (`ctx->owner`); `connect-all` sets a CLI `--owner` before reading the config, so with no `--owner` on the command line the config's value is adopted and used to stamp every connection it creates.

- Read: `json_read_config` applies the config's `owner` only when the context has none, so an explicit CLI `--owner` still wins. Precedence stays CLI > config > default.
- Write: complementarily, `json_update_config` emits `owner` when the context has one, so an owner set on the context round-trips back to the file.
- The published schema (`config-schema.json{,.in}`) gains the `owner` property, and a round-trip test fixture (`config-owner`) exercises read → context → write.

Background: this is the change Daniel originally asked for (the registry-owner follow-up). When I went to add it I found that the schema and the code disagreed on the top-level shape of the config file — the schema described an object with a `hosts` array while the code read and wrote a bare top-level array, so there was no object to hang a global `owner` key on. That had to be fixed first, which became #3499 (write the config as an object, keep reading the legacy array, plus a schema `required` fix and a leak fix). With that merged, this is the small change it was always meant to be.

Testing: `meson test` is green, including the new `config-owner` round-trip and the `config-schema-validate` check that the dumped output conforms to the schema. checkpatch is clean.


igaw:
Thanks!

PR #3507, opened 2026-06-26
telemetry-log: Fix telemetry-log invalid argument error

jeff-lien-sndk:
Use a dynamically determined transfer size when retrieving the host and controller telemetry logs to avoid invalid argument error.

### DIFF
--- a/nvme-cmds.h
+++ b/nvme-cmds.h
@@ -565,7 +565,7 @@ nvme_get_log_telemetry_host(struct libnvme_transport_handle *hdl,
 
 	nvme_init_get_log_telemetry_host(&cmd, lpo, log, len);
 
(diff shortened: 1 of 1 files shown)


nvme-cmds.h
```diff
@@ -565,7 +565,7 @@ nvme_get_log_telemetry_host(struct libnvme_transport_handle *hdl,
 
 	nvme_init_get_log_telemetry_host(&cmd, lpo, log, len);
 
-	return libnvme_get_log(hdl, &cmd, false, len);
+	return libnvme_get_log_dynamic_chunk(hdl, &cmd, false, len);
```
igaw, line 568: makes me wonder if we should do the dynamic chunking on default...

igaw:
Thanks!

PR #3508, opened 2026-06-26
nvme: fix null pointer dereference in get_log_offset()

fhassan-ibm:
The get_log_offset() function stores the command result via args->result after retrieving the log page. However, the pointer args->result was dereferenced without first checking if it is NULL.

Dereferencing a NULL args->result pointer causes a crash when callers pass NULL to indicate they do not need the result value.

Check if args->result is non-NULL before dereferencing it to store the command result.


igaw:
Thanks!

PR #3509, opened 2026-06-26
plugins/sed: fix null pointer dereference in sedopal_print_locking_fe…

fhassan-ibm:
sedopal_print_locking_features() dereferences @ld without checkingwhether it is NULL. Add a NULL check before accessing ld->features and print an error message when no locking feature information is available.


igaw:
Thanks!

PR #3510, opened 2026-06-26
libnvme: validate context in libnvme_rescan_ctrl()

fhassan-ibm:
libnvme_rescan_ctrl() derives the global context from the controller. The derived @ctx is then passed to libnvme_ctrl_scan_namespaces() and libnvme_ctrl_scan_paths(), both of which expect a valid context.

Instead of adding NULL checks in each callee, validate @ctx in libnvme_rescan_ctrl() and return early if no valid context is available.

This avoids propagating an invalid context to the namespace and path scan helpers.


igaw:
Thanks!

PR #3511, opened 2026-06-26
fabrics: make uuid_from_dmi_entries more robust

ikegami-t:
Increase the buf length and read only the length minus one.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -120,46 +120,52 @@ static bool is_dmi_uuid_valid(const char *buf, size_t len)
 	return true;
 }
 
+static int read_file(char *filename, char *buf, size_t size)
+{
+	__cleanup_fd int f = -1;
+	int len;
+
+	f = open(filename, O_RDONLY);
+	if (f < 0)
+		return -errno;
+	len = read(f, buf, size - 1);
+	if (len < 0)
+		return -errno;
+	buf[len] = 0;
+
+	return len;
+}
+
 static int uuid_from_dmi_entries(char *system_uuid)
 {
 	__cleanup_dir DIR *d = NULL;
 	const char *entries_dir = libnvme_dmi_entries_dir();
-	int f;
+	char filename[PATH_MAX];
 	struct dirent *de;
 	char buf[513] = {0};
+	int len, type;
 
 	system_uuid[0] = '\0';
 	d = opendir(entries_dir);
 	if (!d)
 		return -ENXIO;
 	while ((de = readdir(d))) {
-		char filename[PATH_MAX];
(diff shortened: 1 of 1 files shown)


libnvme/src/nvme/fabrics.c
```diff
@@ -142,7 +142,8 @@ static int uuid_from_dmi_entries(char *system_uuid)
 		f = open(filename, O_RDONLY);
 		if (f < 0)
 			continue;
-		len = read(f, buf, 512);
+		len = read(f, buf, sizeof(buf) - 1);
+		buf[len] = 0;
 		close(f);
 		if (len <= 0)
 			continue;
```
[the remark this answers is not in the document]
ikegami-t: Just fixed the errors. Thank you.

libnvme/src/nvme/fabrics.c
```diff
@@ -120,43 +120,52 @@ static bool is_dmi_uuid_valid(const char *buf, size_t len)
...
 }
 
+static int read_file(char *filename, char *buf, size_t size)
+{
+	int len;
+	int f;
+
+	f = open(filename, O_RDONLY);
+	if (f < 0)
+		return -1;
```
igaw: This should be `return -errno`

libnvme/src/nvme/fabrics.c
```diff
@@ -120,43 +120,52 @@ static bool is_dmi_uuid_valid(const char *buf, size_t len)
...
+	len = read(f, buf, size - 1);
+	close(f);
+	if (len <= 0)
+		return -1;
```
igaw: Either save the errno before executing close or use __cleanup_fd here and then you could directly return -errno;

libnvme/src/nvme/fabrics.c
```diff
@@ -120,42 +120,49 @@ static bool is_dmi_uuid_valid(const char *buf, size_t len)
...
+	if (f < 0)
+		return f;
+	len = read(f, buf, size - 1);
+	close(f);
+	if (len <= 0)
+		return len;
+	buf[len] = 0;
+
+	return len;
+}
```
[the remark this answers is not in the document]
igaw: this suggestion is stupid. better to use __cleanup_fd and we don't need to do this silly dance with `safed_errno`

libnvme/src/nvme/fabrics.c
ikegami-t: Just fixed as mentioned. By the way I have also realized as the function `read_file` added is similar with the function `nvmf_read_file` existed but the parameters and return value are difference. So if needed I will do change the commit to use the function `nvmf_read_file` instead.

igaw:
Thanks!

PR #3512, opened 2026-06-26
micron: Fix command injection vulnerabilities

brgoing:
Fixes shell and path-related command injection vulnerabilities in the micron plugin:
- Replaces system() calls that rely on user-supplied or drive-supplied strings with spawning a new process with a fixed argv array.
- Trims and sanitizes the drive-provided serial number string before using it in path strings.
- Checks the user-provided zip/tar package path for unsafe or unsupported characters before using it.


igaw:
Thanks!

PR #3513, opened 2026-06-26
plugins/wdc: fix typed malloc size expressions for log buffers

bellis-ibm:
Use pointee-sized allocations for structured log buffers in nvme_get_hw_rev_log() and nvme_get_print_ocp_cloud_smart_log(), and simplify raw byte-buffer allocation in wdc_enc_submit_move_data().

This eliminates clang-analyzer allocator-size mismatch warnings and improves maintainability by typing allocation sizes directly to the destination pointer type.

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -2257,7 +2257,7 @@ static int wdc_create_log_file(const char *file, const __u8 *drive_log_data,
 		return -1;
 	}
 
-	fd = nvme_open_rawdata(file, O_WRONLY | O_CREAT | O_TRUNC, 0666);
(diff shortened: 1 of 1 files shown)


plugins/wdc/wdc-nvme.c
```diff
@@ -2257,7 +2257,7 @@ static int wdc_create_log_file(const char *file, const __u8 *drive_log_data,
 		return -1;
 	}
 
-	fd = nvme_open_rawdata(file, O_WRONLY | O_CREAT | O_TRUNC, 0666);
+	fd = open(file, O_WRONLY | O_CREAT | O_TRUNC, 0666);
```
igaw, line 2260: No, that is a step backwards. There is a reason why this code uses `nvme_open_rawdata`. Look into the git history why.

PR #3515, opened 2026-06-26
plugins/sandisk: fix memory leak in sndk_get_enc_drive_capabilities()

LJaved-IBM:
Fix memory leak in sndk_get_enc_drive_capabilities() where the
dev_mng_log pointer allocated by sndk_get_dev_mgment_data() was
never freed before the function returns.

Add free(dev_mng_log) at the out label to properly release the
allocated memory in both success and error paths.

Signed-off-by: LaraibJaved <LaraibJaved@ibm.com>

### DIFF
--- a/plugins/sandisk/sandisk-utils.c
+++ b/plugins/sandisk/sandisk-utils.c
@@ -843,6 +843,7 @@ __u64 sndk_get_enc_drive_capabilities(struct libnvme_global_ctx *ctx,
 	}
 
 out:
+	free(dev_mng_log);
 	return capabilities;
 }
 


plugins/sandisk/sandisk-utils.c
igaw, line 764: This just works because `sndk_get_dev_mgment_data` is always setting dev_mng_log to NULL even in the error case. Function are supposed not to have any side effects when the fail...

plugins/sandisk/sandisk-utils.c
```diff
@@ -843,6 +843,7 @@ __u64 sndk_get_enc_drive_capabilities(struct libnvme_global_ctx *ctx,
 	}
 
 out:
+	free(dev_mng_log);
```
igaw, line 846: or  we just could use `__cleanup` and don't have to deal with the funky error path at all...

igaw:
The plugin's error handling is a not really great but I don't see how this change here will make it worse. So let's go with it.

igaw:
Thanks!

PR #3516, opened 2026-06-26
solidigm: Apply fixes to allow compilation on Windows

brgoing:
No functional changes.

Changes pointer casts and format strings to
support type size differences accross platforms.
- Uses %"PRIu64" for platform-agnostic 64-bit value handling.
- Uses %zu for platform-agnostic size_t handling
- Uses uintptr_t for pointer casts

Renames internal stricmp to sldgm_stricmp to avoid conflicts with native stricmp implementations on some platforms.


igaw:
Thanks!

PR #3518, opened 2026-06-29
tree: replace strdup with xstrdup in nvme_alloc_subsystem()

sahmed-ibm:
Replace strdup() with xstrdup() when duplicating the subsystem NQN.
Unlike strdup(), xstrdup() safely returns NULL when passed a NULL
pointer, avoiding undefined behavior if subsysnqn is NULL.

Also check the return value from xstrdup() and bail out early if
duplicating the subsystem NQN fails, either because the input pointer
is NULL or because memory allocation was unsuccessful.


igaw:
Could you merge the two patches into one. Thanks!

sahmed-ibm:
Sure!

igaw:
Thanks!

PR #3519, opened 2026-06-29
wdc-nvme: fix null dereference of sph in get_dev_mgmt_log_page_lid_data()

sahmed-ibm:
wdc_get_dev_mng_log_entry() may be able to find valid log entries but
fail to find desired log page id leaving @sph to NULL.

Continuing to use @sph in this case may result in a NULL pointer
dereference.

Validate both the return value from wdc_get_dev_mng_log_entry() and
the @sph pointer before dereferencing it. Split the error handling
into separate paths so callers receive an appropriate return code and
a meaningful diagnostic depending on the failure.


igaw:
Thanks!

PR #3521, opened 2026-06-29
micron: Fix more security vulnerabilities

brgoing:
Fixes security vulnerabilityies in the Micron plugin.
- Fixes buffer overflow vulnerabilities due to sprintf, read, and readlink handling.
- Fixes NULL pointer dereference vulnerabilities.
- Makes temporary data directories write-protected.


igaw:
Thanks!

PR #3522, opened 2026-06-29
wdc-nvme: validate PCI ID lookup in wdc_log_page_directory()

sahmed-ibm:
wdc_get_pci_ids() returns an error if it fails to retieve the PCI device ID. In that case, device_id remains uninitialized, and using it later in wdc_log_page_directory() results in undefined behavior.

Check the return value from wdc_get_pci_ids() and bail out early if the PCI ID lookup fails.

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -10864,6 +10864,9 @@ static int wdc_log_page_directory(int argc, char **argv, struct command *acmd,
 
 
 		ret = wdc_get_pci_ids(ctx, hdl, &device_id, &read_vendor_id);
+		if (ret) 
+			return ret;
+		
 		log_id = wdc_is_zn350(device_id) ?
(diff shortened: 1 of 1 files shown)


plugins/wdc/wdc-nvme.c
```diff
@@ -10867,6 +10864,9 @@ static int wdc_log_page_directory(int argc, char **argv, struct command *acmd,
 
 
 		ret = wdc_get_pci_ids(ctx, hdl, &device_id, &read_vendor_id);
+		if (ret) {
+			return ret;
+		}
```
igaw: drop the brackets to follow the coding style in this file (see a few lines above for example). thanks!

sahmed-ibm: Will do!

sahmed-ibm:
Dropped the braces around the single-statement if to match the surrounding coding style

igaw:
Thanks!

PR #3527, opened 2026-06-30
nvme: Open files using nvme_open_rawdata to fix Windows issues

brgoing:
Reading files that may have raw binary data on Windows without using the O_BINARY flag when opening can cause some bytes to be misinterpreted or mishandled.  This can lead to bad data or truncated reads. Use nvme_open_rawdata to ensure that O_BINARY is added if supported.

Fixes firmware file reading issues found during testing on Windows, as well as other potential issues on Windows.


igaw:
Thanks!

PR #3529, opened 2026-07-01
nvme-top: add paging support and improve topology update handling

shroffni:
This series contains three patches that improve usability and robustness of the nvme-top dashboard.

The first patch fixes an issue when returning from the subsystem topology dashboard to the subsystem list dashboard. If the NVMe topology changes while the subsystem dashboard is active, the previously selected subsystem may move, particularly when the subsystem list spans multiple page frames. Since the subsystem list is rebuilt when switching back and nvme-top does not maintain page-to-screen mappings, preserving the previous selection is not feasible. Instead, restore the selected subsystem as the first entry in the list. This change lays the groundwork for the paging support introduced in the next patch.

The second patch adds Page Up and Page Down support to the nvme-top dashboard. This makes navigation significantly more efficient when the host contains a large number of NVMe subsystems or when a subsystem contains tens or hundreds of namespaces.

The final patch fixes the netlink uevent handling by treating ENOBUFS as a non-fatal condition. An ENOBUFS error indicates that one or more kernel kobject uevents were dropped because the socket receive buffer overflowed. Rather than terminating the event processing, nvme-top synthesizes an NVMe kobject change event and triggers a topology rescan, ensuring that the displayed topology remains consistent even when individual uevents are missed.

### DIFF
--- a/nvme-print-stdout-top.c
+++ b/nvme-print-stdout-top.c
@@ -1475,20 +1475,43 @@ wait_for_event:
 			s = subsys_arr[subsys_idx];
 			quit = stdout_top_draw_subsys_topology_screen(db_ctx,
 					stream, s);
-			scroll = 0;
 			if (quit)
 				break;
-			fallthrough;
+
+			scroll = 0;
+			free(subsys_arr);
+			subsys_arr = NULL;
+			libnvme_free_global_ctx(ctx);
+
+			ctx = stdout_top_rescan_topology();
+			if (!ctx) {
+				quit = 1;
+				break;
+			}
+
+			/*
+			 * The topology may have changed while the subsystem
+			 * dashboard was active. Restart from the first
+			 * subsystem instead of trying to restore the previous
+			 * screen position.
+			 */
+			subsys_idx = 0;
+			subsys_arr = stdout_top_build_subsys_arr(ctx,
+					&num_subsys);
+			if (!subsys_arr)
+				quit = 1;
+
+			break;
 		case EVENT_TYPE_NVME_UEVENT: {
 			__cleanup_free char *subsys_name = NULL;
-			libnvme_subsystem_t s;
+			libnvme_subsystem_t s = subsys_arr[subsys_idx];
 
-			s = subsys_arr[subsys_idx];
 			subsys_name = strdup(libnvme_subsystem_get_name(s));
 			if (!subsys_name) {
 				quit = 1;
 				break;
 			}
+
 			free(subsys_arr);
 			subsys_arr = NULL;
 			libnvme_free_global_ctx(ctx);
@@ -1511,6 +1534,7 @@ wait_for_event:
 			if (subsys_idx < 0)
 				subsys_idx = 0;
 
+			scroll = 0;
 			break;
 		}
 		case EVENT_TYPE_KEY_DOWN:
--- a/nvme-print-stdout-top.c
+++ b/nvme-print-stdout-top.c
@@ -1069,7 +1069,7 @@ static void stdout_top_print_subsys_topology_footer(
 		struct dashboard_ctx *db_ctx, FILE *stream)
 {
 	fprintf(stream, "\n--------------------------------------\n");
-	fprintf(stream, "[ESC to go back to the previous screen, q to quit]\n");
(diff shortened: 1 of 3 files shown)


nvme-print-stdout-top.c
```diff
@@ -1475,20 +1475,43 @@ void stdout_top(int refresh_interval)
...
 			if (quit)
 				break;
-			fallthrough;
+
+			scroll = 0;
+			free(subsys_arr);
+			subsys_arr = NULL;
+			libnvme_free_global_ctx(ctx);
+
+			ctx = stdout_top_rescan_topology();
```
igaw, line 1552: I haven't noticed during the initial review. I assume you have to free the global context so that all resources are freed from `libnvme_scan_topology`? The idea about the global context that it is allocated only once, so we need to look into a `libnvme_rescan_topology` I think.

igaw: So for this series this is fine, but we should really make the library able to deal with this use case.

igaw: There is `libnvme_refresh_topology`, which should do the trick,

shroffni: Yes this looks good. Should I fix it in this series? Or shall I post it in a separate patch?

igaw: you can post a separate patch for this or update the series. whatever you prefer. Could you look through the copilot findings below though? No need for long explanation if you think it's wrong.

util/dashboard.c
```diff
@@ -509,6 +519,48 @@ static int wait_for_event(struct dashboard_ctx *db_ctx,
...
+	unsigned char c;
+	enum event_type event;
+
+	event = wait_for_event(db_ctx, &c, 1);
+	switch (event) {
+	case EVENT_TYPE_ERROR:		/* fall through */
+	case EVENT_TYPE_KEY_QUIT:	/* fall through */
+	case EVENT_TYPE_TIMEOUT:
+		return event;
+	default:
```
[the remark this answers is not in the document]
shroffni: I agree we may receive EVENT_TYPE_NVME_UEVENT or EVENT_TYPE_SIGWINCH while waiting for escape key sequence. I'd rather update wait_for_esc_seq() to not include fds which monitor uevents and sigwinch. 

As the name wait_for_esc_seq() suggests we're waiting for escape sequence, it makes more sense to only include key stroke monitor fd. I'll update the patch and resend.

util/dashboard.c
```diff
@@ -509,6 +519,48 @@ static int wait_for_event(struct dashboard_ctx *db_ctx,
...
+		else if (c == 53 || c == 54) {
+			char prev = c;
+
+			event = wait_for_event(db_ctx, &c, 1);
+			switch (event) {
+			case EVENT_TYPE_ERROR:		/* fall through */
+			case EVENT_TYPE_KEY_QUIT:	/* fall through */
+			case EVENT_TYPE_TIMEOUT:
+				return event;
+			default:
```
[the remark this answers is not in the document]
shroffni: Same comment as above. I will exclude monitor fds which doesn't wait for or monitor key press.

util/dashboard.c
```diff
@@ -538,21 +590,12 @@ enum event_type dashboard_wait_for_event(struct dashboard_ctx *db_ctx)
 					return EVENT_TYPE_KEY_ESC;
```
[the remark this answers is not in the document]
shroffni: Same comment as above.

nvme-print-stdout-top.c
shroffni: Okay so I have addressed comments from copilot and per the review comment I will update the code. I am glad that copilot-pro is really able to find subtle issues which human eyes may miss.  

Regarding libnvme_refresh_topology(), I will handle it in a separate patch.

util/dashboard.c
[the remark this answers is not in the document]
shroffni: Addressed this now in the updated PR.

util/dashboard.c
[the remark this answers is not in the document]
shroffni: Addressed in this updated PR.

igaw:
Thanks a lot!

PR #3536, opened 2026-07-02
Make persistent discovery controller configurable for the subsystem

hreinecke:
Some subsystems / vendors might choose to enable or disable persistent discovery controllers for particular subsystems, independent on the global option to enable PDCs.
This patch adds a new subsystem configuration option 'persistent_discovery_controller', which controls whether a PDC should be created for this subsystem. The value defaults to the host PDC setting if not specified in the configuration file.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -2519,9 +2519,8 @@ static int _nvmf_discovery(struct libnvme_global_ctx *ctx,
 				if (!(eflags & NVMF_DISC_EFLAGS_EPCSD))
 					disconnect = true;
 				else
-					disconnect = false;
+					disconnect = libnvme_subsystem_get_pdc_enabled(s);
 			}
-
 			set_discovery_kato(&nfctx);
 		} else {
 			/* NVME_NQN_NVME */
--- a/libnvme/doc/config-schema.json.in
+++ b/libnvme/doc/config-schema.json.in
@@ -58,6 +58,10 @@
 		    "type": "string",
 		    "maxLength": 223
 		},
+		"persistent_discovery_ctrl": {
+		    "description": "Enable/disable Persistent Discovery Controller",
+		    "type": "boolean"
+		},
 		"ports": {
 		    "description": "Array of NVMe subsystem ports",
 		    "type": "array",
--- a/libnvme/libnvme3/accessors.i
+++ b/libnvme/libnvme3/accessors.i
@@ -176,6 +176,7 @@ struct libnvme_subsystem {
 	const char * firmware;
 	%immutable subsystype;
 	const char * subsystype;
+	bool pdc_enabled;
 	%extend {
 		%immutable iopolicy;
 		const char * iopolicy;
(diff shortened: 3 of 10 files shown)


hreinecke:
This is a somewhat competing implementation to https://github.com/linux-nvme/nvme-cli/pull/3472; I'm not trying to change the default, but rather control the creation of a PDC on a per-subsystem basis.

libnvme/src/nvme/fabrics.c
```diff
@@ -2416,9 +2416,8 @@ static int _nvmf_discovery(struct libnvme_global_ctx *ctx,
 				if (!(eflags & NVMF_DISC_EFLAGS_EPCSD))
 					disconnect = true;
 				else
-					disconnect = false;
+					disconnect = libnvme_subsystem_get_pdc_enabled(s);
```
igaw, line 2522: yeah, this makes a lot of sense to me.

igaw:
As far I understand the implementation it makes sense to handle the PDC on subsys level. @Mr-Bossman, @johnmeneghini  WDYT?

libnvme/src/nvme/fabrics.c
Mr-Bossman: everything in this for loop only runs if `_nvmf_discovery` gets called with `connect = true` so running `nvme connect`. The initial goal was to not only support default pdc for `nvme connect`, but also `nvme discover`. The goal post then moved a bit to support default EPCSD for `nvme discover` as nvme connect supports it already. If the goal of this PR is to support default pdc for `nvme connect` and not `nvme discover` then this is fine either way its a step in the right direction.

Sorry for the late reply i somehow missed this until now

Mr-Bossman (approved):

igaw:
> The initial goal was to not only support default pdc for nvme connect, but also nvme discover

Oh I've missed this important fact. Sorry about that. It's kind of obvious now you mentioned it. I blame my distraction due the 3.x work


Ah, this case it I think Hannes patch should just be updated. The 

```c
	if (!connect)
		return 0;
```

check should be after the PDC check. My idea is to avoid to iterate over the entries twice. The logic spread across the function. I suppose it would help to refactor this into smaller functions with a clear scope what they do. We have `connect` and `disconnect`...

Mr-Bossman:
> > The initial goal was to not only support default pdc for nvme connect, but also nvme discover
> 
> Oh I've missed this important fact. Sorry about that. It's kind of obvious now you mentioned it. I blame my distraction due the 3.x work
> 
> Ah, this case it I think Hannes patch should just be updated. The
> 
> ```c
> 	if (!connect)
> 		return 0;
> ```
> 
> check should be after the PDC check. My idea is to avoid to iterate over the entries twice. The logic spread across the function. I suppose it would help to refactor this into smaller functions with a clear scope what they do. We have `connect` and `disconnect`...

at that point its almost identical to my PR https://github.com/linux-nvme/nvme-cli/pull/3472/ exept in the `subsystem` instead of `host`.

igaw:
The JSON part has been replaced with the ini config. #3698 added support to read/write the persistant flag from the ini config in the discovery section. @Mr-Bossman is looking into hooking the flag into code back. Thus I'll close this PR.

PR #3538, opened 2026-07-02
ioctl-win: Implement reset_ctrl_device using a PnP-level reset.

brgoing:
Windows doesn't have an IOCTL for resetting an NVMe controller, but a plug-and-play-level reset can be used to accomplish an equivalent reset of the controller.  Implements PnP-level reset for reset_ctrl_device on Windows and adds helper methods to ctrl-map.


igaw:
Thanks!

PR #3523, opened 2026-06-29
libnvmf: Add the NVMe-oF exclusion list

martin-belanger:
## Summary

This series adds a **system-wide NVMe-oF exclusion list**: a persistent, administrator-authored set of files under `/etc/nvme/exclusions/` naming controllers that no orchestrator may auto-connect. It is a small, cooperative coordination layer — the administrator's opt-out — and the natural companion to the ownership registry that landed earlier.

The two are mirror images: the **registry** records *who owns* a controller and guards against **accidental disconnect** (so `disconnect-all` does not tear down a connection some component depends on); the **exclusion list** guards against **accidental connect** (so an orchestrator does not auto-connect a controller the administrator wants left alone). Together with the transport ID introduced here, they are the building blocks that let independent NVMe-oF orchestrators — the initramfs (NBFT/FC boot), a human running `connect-all`, `nvme-discoverd`, `nvme-stas` — coexist on one host.

## Why an exclusion list (and not just editing a config file)

The motivating cases are connections whose *intent comes from a source the administrator cannot edit*:

- **NBFT controllers** are described by a firmware boot table configured in UEFI — there is no host-side entry to comment out. Excluding the controller is the only way to keep an orchestrator off an NBFT path, e.g. to take it out of service during testing.
- **Discovered controllers** — a target returned as a DC's discovery log page entry (DLPE), or a controller announced over mDNS — are advertised by the fabric at runtime and listed in no file on the host.

For all of these the exclusion list is the only local lever: the one place an administrator can say "not this one", regardless of where the connection request originated. Like the registry, it is **cooperative, not enforcement** — every orchestrator runs as root and could connect anything; the list simply lets well-behaved tools refrain by agreement.

## What's in the series

1. **`libnvme: add shared fabrics helpers in preparation for exclusion list`** — a preparation refactor: extract the private `mkdir -p`, atomic-cloexec-tempfile, and test-directory-validation helpers from `registry.c` into `util-fabrics.c` as internal `libnvmf_` helpers, and convert `registry.c` to use them. No functional change.
2. **`libnvme: add the NVMe-oF exclusion list`** — the feature: the `libnvmf_exclusion_*` API in `exclusion.c`, built on a new transport-ID type (`libnvmf_tid_*` in `tid.c`) that canonicalizes a connection's identifying tuple and derives a stable hash from it; an `nvme exclusion` plugin (create/delete/edit/list/add/remove); and SWIG bindings exposing lists/entries/match to Python.
[...]

PR #3539, opened 2026-07-02
libnvme/registry: substitute RUNDIR into the registry udev rule

martin-belanger:
Follow-up to #3523 (the round-6 `RUNDIR` change to `REGISTRY_DIR_DEFAULT`). The library now derives the registry path from the build's `rundir` setting (`RUNDIR "/nvme/registry"`), but the udev cleanup rule still hardcoded `/run/nvme/registry` — the rule file was installed as static text, so a build that relocates `rundir` left udev cleaning a directory the library no longer writes.

Fix: the rule becomes a `70-nvmf-registry.rules.in` template with `@RUNDIR@` in the `RUN+=` line, generated via `configure_file(configuration: substs)` — the same pattern `libnvme.spec.in` already uses two stanzas above. Both sides of the registry (library and rule) now derive the path from the single meson setting.

No behavior change for a default build: the generated rule is identical modulo the prefix the build already applies everywhere else.

### DIFF
--- a/libnvme/meson.build
+++ b/libnvme/meson.build
@@ -42,9 +42,11 @@ if want_libnvme
     endif
 
     if want_fabrics
-        install_data(
-            'sysfiles/udev/rules.d/70-nvmf-registry.rules',
-            install_dir: udevrulesdir,
+        configure_file(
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #3540, opened 2026-07-02
plugins/exclusion: operate on the default list when --name is omitted

martin-belanger:
Bug fix for the `nvme exclusion` CLI (merged in #3523). The library treats a NULL list name as the **default list** (`/etc/nvme/exclusions.conf` — the one `nvme disconnect --exclude` writes to), but every mutating `nvme exclusion` subcommand rejected a missing `--name`, so the CLI could not manage the most commonly written list at all: the only ways to touch it were `disconnect --exclude` or a hand edit.

`--name` is now optional on `create`, `delete`, `add`, `remove` and `edit`; omitting it targets the default list. The help texts and the five man pages say so, and the synopsis shows the option as optional.

Also fixes a misleading error: an invalid list *name* was reported as `invalid entry (unknown key): <entry>`, pointing the user at a perfectly valid entry. The entry is now validated first (`invalid entry: …`) and a remaining `-EINVAL` is attributed to the name (`invalid list name: …`).

New unit tests cover the NULL-name operations end to end (create/add/match/remove/delete on the default list), alongside the existing exclusion suite. All tests green on glibc and musl.

### DIFF
--- a/Documentation/nvme-exclusion-add.txt
+++ b/Documentation/nvme-exclusion-add.txt
@@ -8,7 +8,7 @@ nvme-exclusion-add - Add an entry to an NVMe-oF exclusion list
 SYNOPSIS
 --------
 [verse]
-'nvme' [<global-options>] 'exclusion add' --name=<NAME> | -N <NAME>
+'nvme' [<global-options>] 'exclusion add' [--name=<NAME> | -N <NAME>]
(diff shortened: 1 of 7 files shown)


igaw:
This what I was trying to figure out yesterday in the chat, how we handle the defaults. Thanks for making it a nice user experience!

PR #3541, opened 2026-07-02
libnvme: make discovery connects honor the exclusion list

martin-belanger:
The exclusion list (merged in #3523) is aimed squarely at orchestrators: it names the controllers no auto-connecting agent may touch, so independent agents can coexist with each other and with the administrator's wishes. `nvme connect-all` is such an orchestrator — it connects whatever a Discovery Log Page, the NBFT table, or a configuration file enumerates — yet `libnvmf_exclusion_match()` had no caller in nvme-cli, so an excluded controller was happily reconnected by the very next `connect-all` run. This PR closes that gap (called out as a deliberate follow-up in the #3523 discussion).

One static helper consults the list right before an *enumerated* controller is connected: DLP entries and referral DCs (`nvmf_connect_disc_entry`), NBFT records (`nbft_connect`), and controllers enumerated from `config.json` / `discovery.conf`. A match is skipped and reported at INFO level, so `connect-all --verbose` shows which controllers were left alone and why.

Controllers named explicitly by the user (`nvme connect`, `nvme discover` with an address) are deliberately not checked — a targeted human action overrides the list; only the entries such a DC enumerates are gated. This matches the "not allowed to block targeted human actions" rule already documented in EXCLUSIONS.md, which now also names `connect-all` explicitly.

**Testing.** Unit: the existing exclusion/TID match suites (green on glibc and musl; the connect paths themselves need a kernel). Functional, against a local nvmet target with multiple subsystems: with `exclusion = nqn=<subsysnqn>` in `exclusions.conf`, `nvme connect-all -v` connected every subsystem except the excluded one and logged the skip; after `nvme exclusion remove`, the next `connect-all` connected it again.

### DIFF
--- a/libnvme/design/EXCLUSIONS.md
+++ b/libnvme/design/EXCLUSIONS.md
@@ -96,6 +96,8 @@ if (libnvmf_exclusion_match(ctx, tid))
 
 `libnvmf_exclusion_match()` re-reads the files on every call (no caching), so an edit takes effect on the next connection attempt without restarting anything. This is the only API an orchestrator needs; the create/add/remove calls are for the management tooling.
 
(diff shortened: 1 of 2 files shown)


igaw:
Looks good, I've spend some time trying to figure out if the return code was correct in all places. As far I can tell it all makes sense. Tricky stuff, wouldn't be surprised if this is not 100% right. But let's add those hooks into place and see what the testing reveals. 

Thanks!

PR #3542, opened 2026-07-02
Enable testing support on Windows

brgoing:
Enables testing support on Windows and updates tests to run on Windows.
- Adds support for enabling tests and nvme-tests options
- Excludes linux-specific unit tests when building on Windows
- Adds support in nvme tests for detecting Windows and skipping tests for commands that are not supported on Windows
- Updates some tests to work on Windows

### DIFF
--- a/libnvme/test/meson.build
+++ b/libnvme/test/meson.build
@@ -70,25 +70,27 @@ if cxx_available
 
 endif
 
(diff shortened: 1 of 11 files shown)


igaw:
Nice! Thanks a lot!

PR #3543, opened 2026-07-02
nvme_test: Update run_ns_io to be platform agnostic.

brgoing:
Instead of using linux-specific commands to run reads and writes, run_ns_io now uses nvme commands that will work regardless of platform. Pulls some functionality from nvme_test_io into nvme_test to allow for code reuse and to avoid code duplication when detecting LBA IO parameters and configuring read and write commands.

### DIFF
--- a/tests/nvme_test.py
+++ b/tests/nvme_test.py
@@ -666,6 +666,87 @@ class TestNVMe(unittest.TestCase):
 
         return 0 if err_log_entry_count == entry_count else 1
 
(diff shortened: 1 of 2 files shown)


igaw:
Yes, makes totally sense to drop the `dd`.

igaw:
Thanks!

brgoing:
@igaw Is this waiting for something to be able to merge?  Are there any changes you need me to make?

igaw:
No, I thought I pressed the merge button.... let me do it again now.

igaw:
Thanks again :)

PR #3544, opened 2026-07-02
nvme: enable cli for windows build

bcapener:
Remove windows dummy cli target and add all relevant source files to the windows cli build.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -553,29 +553,22 @@ if want_nvme
     subdir('plugins')   # declares: plugin_sources
     subdir('util')      # declares: util_sources
 
-    sources = []
-    if host_system == 'windows'
-        sources += [
-            'nvme-dummy.c', # Dummy source file for Windows port bring up.
-        ]
-    else
-        sources += [
-            'logging.c',
-            'nvme-cmds.c',
-            'nvme-models.c',
-            'nvme-print-binary.c',
-            'nvme-print-stdout.c',
-            'nvme-print.c',
-            'nvme-rpmb.c',
-            'nvme.c',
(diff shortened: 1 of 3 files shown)


nvme-rpmb.c
```diff
@@ -131,6 +133,10 @@ unsigned char *create_hash(const char *algo,
 	close(infd);
 
 	return hash;
+#else
+	/*Windows support will be added later*/
```
bcapener: See this PR https://github.com/linux-nvme/nvme-cli/pull/3545

igaw: I've merged #3545. Could you rebase it? Thanks!

meson.build
```diff
@@ -553,30 +553,27 @@ if want_nvme
...
+        'nvme-print-binary.c',
+        'nvme-print-stdout.c',
+        'nvme-print.c',
+        'nvme-rpmb.c',
+        'nvme.c',
+        'plugin.c',
+    ]
+    if host_system != 'windows'
         sources += [
             'libnvme-wrap.c',
```
igaw: I've removed `libnvme-wrap.c` as it is dead code anyway: #3547 

This should result in only on feature adding for source files. Yay!

igaw:
Very nice to see this landing soon! Fantastic job!

nvme-rpmb.c
bcapener: done.

igaw:
Thanks a lot!

I’m glad we got it working. After the heavy refactoring, the platform-specific code is now well compartmentalized, so I’d call this a win. 🙂

PR #3545, opened 2026-07-02
nvme: add windows support for rpmb hash functions

bcapener:
Moved hash functions used in RPMB into new util/hash.h with separate linux and windows implementations. Also renamed hash functions hmac_sha256 to create_hmac_sha256 and rpmb_md5 to create_md5.


igaw:
Thanks!

PR #3546, opened 2026-07-02
Proposed test structure for adding plugin tests

brgoing:
Expands the python testing framework to support defining and enabling tests for individual plugins.  Adds simple Micron and OCP plugin tests as examples.

### DIFF
--- a/tests/plugins/README.md
+++ b/tests/plugins/README.md
@@ -1,0 +1,121 @@
+# Plugin Tests
+
+This directory contains hardware tests for nvme-cli vendor plugins. Each plugin
+has its own subdirectory with tests that exercise plugin commands against real
+hardware.
+
+## Architecture
+
+```
+tests/plugins/
+├── plugin_test.py            # Base class for all plugin tests
+└── <plugin>/
+    ├── meson.build           # Meson test definitions for a specific plugin
+    ├── <plugin>_test.py      # plugin-specific base class
+    └── *_test.py             # Individual test files
+```
+
+### Class Hierarchy
+
+```
+unittest.TestCase
+└── TestNVMe (tests/nvme_test.py)
+    └── TestPlugin (tests/plugins/plugin_test.py)
+        └── Test<Plugin> (tests/plugins/<plugin>/<plugin>_test.py)
+```
+
+- **TestNVMe** — provides device config loading, `run_cmd()`, `exec_cmd()`,
+  JSON parsing helpers, and log directory management.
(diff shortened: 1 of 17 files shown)


tests/plugins/README.md
```diff
@@ -0,0 +1,136 @@
...
+|------------------|---------|------------------------------------------|
+| `<plugin>-tests` | false   | Run tests against real hardware compliant with the plugin |
+
+A plugin test suite is included in the build when these three conditions
+are met:
+
+1. `nvme-tests=true` (gates the entire `tests/` subtree)
+2. The per-plugin option is enabled (e.g., `ocp-tests=true`)
+3. The plugin is listed in the `plugins` build option
+
```
igaw, line 49: Could we do hook into the build configuration? Instead listing every single plugin to test?

```meson
option(
  'plugins',
  type : 'array',
  choices: ['amzn', 'dapustor', 'dell', 'dera',
            'exclusion', 'fdp', 'feat', 'huawei', 'ibm', 'innogrit', 'inspur',
            'intel', 'lm', 'mangoboost', 'memblaze', 'micron', 'nbft',
            'netapp', 'nvidia', 'ocp', 'registry', 'sandisk', 'scaleflux',
            'seagate', 'sed', 'shannon', 'solidigm', 'ssstc', 'toshiba',
            'transcend', 'virtium', 'wdc', 'ymtc', 'zns'],
  description : 'comma-separated list of plugins to build (all plugins are included if unset, [] = no plugins)'
)
```

My idea is to use the selection here and run the tests if `-Dplugin-tests` is enabled.

brgoing: My thought was that even though you may be building all of the plugins, you are only connected to certain hardware, so you'd only want the tests for that hardware to be enabled when you run 'meson test'. If we change this and configure the tests for all of the available plugins, we couldn't include them when we run 'meson test'. We'd have to run the specific plugin test suite for the connected hardware by itself. 

Or are you proposing that when testing plug-in functionality for specific hardware, we'd only build the plugins that we wanted to test, and enable that testing with a 'plugin-tests' option? It seems like either way we'd have to specify something per plug-in that we wanted to test.

igaw: Okay, understood. I would like to avoid to have explicit `--$NAME-tests`. If you could model the option in the same way as the `plugins` option, that is as array option. This should give the flexibility your are asking for, right?

brgoing: Thanks for the suggestion.  I've updated the design to use a plugin-tests array option.

igaw:
Thanks!

PR #3549, opened 2026-07-06
nbft: fix missing error checks for ftell() and rewind()

bellis-ibm:
The libnvmf_read_nbft() function did not check whether ftell()
returned (size_t)-1L, leaving a case where an invalid file size could
be used silently. Additionally, rewind() failures were not detected.

Check if ftell() returns (size_t)-1L and detect rewind() failures via 
errno, closing the file and returning -EINVAL in both cases.


igaw:
Thanks!

PR #3550, opened 2026-07-06
libnvme: fix null pointer dereference in _discovery_config_json()

LJaved-IBM:
The _discovery_config_json() function retrieves the transport and
traddr parameters from a controller. The early return only triggers
when both @transport and @traddr are NULL, allowing execution to
continue with a NULL @transport.

Passing a NULL @transport to strcmp(), which has a nonnull attribute,
causes undefined behavior.

Change the condition to return early if either @transport or @traddr
is NULL, preventing the NULL pointer from being passed to strcmp().


igaw:
Thanks!

PR #3551, opened 2026-07-06
Additional Windows build support

brgoing:
These changes add the rest of the build support required for Windows and adds information and a helper utility for configuring a build system on Windows.
- Adds support for building on Windows with statically linked libraries.
- Improves build support for running and testing builds that use shared libraries.
- Adds build configuration for building plugins on Windows.
- Adds readme section regarding building on Windows, as well as a utility for configuring an MSYS2 environment on Windows for building nvme-cli.

### DIFF
--- a/plugins/meson.build
+++ b/plugins/meson.build
@@ -7,39 +7,47 @@ all_plugins = {
   'dell': ['plugins/dell/dell-nvme.c'],
   'dera': ['plugins/dera/dera-nvme.c'],
   'fdp': ['plugins/fdp/fdp.c'],
-  'huawei': ['plugins/huawei/huawei-nvme.c'],
   'ibm': ['plugins/ibm/ibm-nvme.c'],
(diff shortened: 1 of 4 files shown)


plugins/meson.build
```diff
@@ -7,40 +7,48 @@ all_plugins = {
...
+    'zns': ['plugins/zns/zns.c'],
+  }
 else
-  selected_plugins = []
+  all_plugins += {
+    'micron': [
+      'plugins/micron/micron-nvme.c',
+      'plugins/micron/micron-utils.c',
+      'plugins/micron/micron-utils-win.c'
+      ],
```
igaw, line 45: I was starting to wonder why the micron shows up in both branches... the difference is `micron-utils-linux.c` vs `micron-utils-win.c`

For now this is okay. One thing we could do in future is to add a meson.build to each plugin if this helps to make it a bit more streamlined.

igaw:
Thanks!

PR #3552, opened 2026-07-06
nvme-print: Update generic name listing to be platform independent

brgoing:
Refactors generic path discover in nvme-print and nvme-print-stdout to fall back to using livnvme_ns_get_generic name.  Allows platform independent generic name behavior, and removes duplicated "ng%dn%d" string building on Linux.  Also widens Generic list section of detailed list slightly to provide enough space for Windows generic names.


igaw:
Nice! Thanks!

PR #3553, opened 2026-07-06
nvme: do not print status message on default

### LINKED ISSUE #2025: admin-passthru with raw_binary prepends status message
https://github.com/linux-nvme/nvme-cli/commit/414f1eda0bb7540c58cbb1f85dee6c6d84abba43#commitcomment-122245811

Per discussion in this update, currently status message of admin-passthru command is written to stdout and prepended to the data returned from the device if -i option is not used.  I'd like to request separating status message from stdout like it used to be if possible.

dwsuse:
Unix tools should remain silent on success and produce output only when
    it is part of their functional result. Emitting status messages by
    default violates the long-established convention that stdout is reserved
    for machine-consumable data, while diagnostics and optional progress
    reporting belong on stderr or behind an explicit verbose flag.

    A core strength of Unix tooling is composability: commands are routinely
    used in pipelines, shell scripts, automation frameworks, and other
    programs. Unsolicited status output contaminates stdout, forcing
    consumers to implement fragile filtering logic and making the tool
    harder to integrate reliably. The absence of default status messages is
    therefore not merely a stylistic preference but an interoperability
    requirement.

    This follows the traditional Unix principle that "no news is good news":
    successful execution is indicated by the exit status, not by
    human-readable confirmation text. Users who need operational visibility
    can opt into it through verbose or progress-reporting options, while
    automation and existing workflows continue to receive clean,
    deterministic output.

Fixes: https://github.com/linux-nvme/nvme-cli/issues/2025
Replaces: https://github.com/linux-nvme/nvme-cli/pull/3416

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -176,7 +176,7 @@ static void save_discovery_log(char *raw, struct nvmf_discovery_log *log)
 
 	fd = open(raw, O_CREAT | O_RDWR | O_TRUNC, 0600);
 	if (fd < 0) {
-		fprintf(stderr, "failed to open %s: %s\n", raw, libnvme_strerror(errno));
+		nvme_show_error("failed to open %s: %s", raw, libnvme_strerror(errno));
 		return;
 	}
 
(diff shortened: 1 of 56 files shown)


igaw:
While this not addressing all the issues the code base has, it is going into the right direction. There are more cleanups needed. Let's get this in and fixup the rest piece by piece. There is no point trying to fix everything in one  go.

PR #3554, opened 2026-07-06
libnvme/tid: numeric-only, sanitizing Transport ID (TID)

martin-belanger:
# libnvme/tid: numeric-only, sanitizing Transport ID (TID)

Cleanup/prep for the upcoming INI connection-config parser, not a feature. That parser needs a TID that's a pure function of numeric input. Removes `get_hash()`/`equal()` (upstream API, zero production callers) and formalizes numeric-only, sanitizing construction. A `hostname` member with deferred resolution was prototyped while designing the config parser but never merged upstream -- rejected, see `TID.md`.

**Follow-up (not in this PR):** drop `hostname2traddr()`, the last real `getaddrinfo()` caller in libnvme (the other 3 call sites use `AI_NUMERICHOST`, no DNS). Push resolution to CLI/discoverd, drop `NVME_HAVE_NETDB`/`<netdb.h>` entirely. `libnvmf_traddr_is_numeric()` (added here) is what that follow-up will use.

## Commits (review order)

1. `design/INTEGRATION.md` -- library/app boundary: libnvme owns protocol I/O, caller owns policy/non-determinism/concurrency.
2. `design/TID.md` -- design authority for the TID.
3. `tid: reject hostnames, canonicalize numeric addresses` -- `traddr`/`host_traddr` canonicalized via `inet_pton`/`inet_ntop`; hostname fails construction. New `libnvmf_traddr_is_numeric()`.
4. `tid: drop get_hash() and equal()` -- dead public API; hashing moves to the one caller that needs it (discoverd, via `get_canonical()`).
5. `tid: replace per-field setters with a set_identity triplet` -- addressing construction-only; `set_identity()` owns host-identity rules.
6. `fabrics: use str(), not get_canonical(), in the exclusion log` -- fixes a log line that could exceed 500 chars.
7. `tid: group functions by role` -- pure reorder, no content change.

## Numbers

- Code: +409/-328 | Tests: +171/-101 | Docs: +170
- `meson test`: 81/0 fail | checkpatch + valgrind clean

### DIFF
--- a/libnvme/design/INTEGRATION.md
+++ b/libnvme/design/INTEGRATION.md
@@ -1,0 +1,39 @@
+# Integrating libnvme into an application
+
+> **libnvme owns NVMe protocol I/O. The caller owns policy, non-determinism, and its own concurrency.**
+
+The code is the source of truth; this document explains the boundary and the reasoning behind it. For exact signatures see the kdoc in `src/nvme/*.h`.
+
+## Why this needs stating
+
+The CLI/library split extracted code by mechanical convenience, not by a stated rule, so some caller responsibilities ended up inside the library. Hostname resolution is the clearest example — it sits inside libnvme today even though it was always the caller's job. This document is the rule that the boundary should have had from the start.
+
+## The boundary
+
+- **NVMe protocol I/O** — ioctls to `/dev/nvmeX`, the `/dev/nvme-fabrics` connect write, log pages, Identify, sysfs enumeration. If it *is* NVMe, it belongs in the library.
+- **Policy** — any choice with more than one defensible answer (which resolved address to use, which source address to bind, how hard to retry). The right answer depends on the deployment, not on NVMe, so policy belongs to the caller.
+- **Non-determinism** — anything not a pure function of its inputs (DNS is the archetype). It poisons anything downstream that needs to be reproducible.
+- **Concurrency** — *when* and on which thread a call runs relative to the application's event loop. Only the application knows its loop.
+
+Hostname resolution fails the policy and non-determinism tests outright, and (below) also blocks — three independent reasons it belongs to the caller, not the library.
+
+## libnvme is synchronous and blocking
+
+Every libnvme API blocks the calling thread by default: connect blocks on the `/dev/nvme-fabrics` write, Get Log Page and Identify block on ioctls, and even device-tree enumeration is real sysfs I/O. The one opt-in exception is I/O passthru over io_uring: `libnvme_submit_io_passthru()` queues a command without blocking, though `libnvme_reap_passthru()` still blocks to collect the result — a caller has to explicitly build that path. No file descriptors to poll for the default synchronous path, no callbacks, no loop of its own — deliberately: integrating with every event-loop implementation in existence (sd_event, GLib, raw epoll, libuv, ...) doesn't scale, and would lock every caller into whichever one libnvme picked.
+
+## The main-loop contract
+
+A process with a main loop must never call libnvme from the loop thread. Offload it:
+
+- **Worker thread** — for in-process blocking calls (Discovery Log Page, Identify, sysfs scan); post the result back via an eventfd/pipe the loop already watches. This is how nvme-stas drives nearly every libnvme call.
+- **Subprocess or systemd unit** — for connect, when isolation and lifecycle supervision are wanted too.
+
+nvme-discoverd uses both: it retrieves Discovery Log Pages on a worker thread but performs connects by starting a systemd unit, so the fully-blocking call never touches its loop.
+
+## Worked example: hostname resolution
+
+A hostname can only enter through `traddr`, from a CLI argument or a config file — a Discovery Log Page's address family is numeric-only by spec, `host_traddr` is a source address a hostname can't sensibly name, and `host_iface` is a kernel-resolved interface name. The caller resolves `traddr` before calling libnvme — inline for a one-shot CLI, on a worker thread for a daemon — and picks among the results itself (e.g. IPv4 vs. IPv6). Internally, libnvme's Transport ID (TID) constructors enforce this by rejecting hostnames outright (`design/TID.md`).
+
+## Rule of thumb
+
+*Is it NVMe protocol I/O?* → the library. *Is it policy, non-deterministic, or event-loop-aware?* → the caller. Something that's both: split it, keep the protocol part in the library, leave policy and scheduling to the caller.
--- a/libnvme/design/TID.md
+++ b/libnvme/design/TID.md
@@ -1,0 +1,87 @@
+# The NVMe-oF Transport ID (TID)
+
+A **TID** identifies one NVMe-oF connection: the transport tuple plus the subsystem and host identifiers that together name one host-to-controller association. `libnvmf_tid_get_canonical()` renders it as a deterministic, fixed-field-order string. A caller that wants a compact name (e.g. nvme-discoverd deriving a systemd unit name) hashes that string itself — libnvme doesn't hash it (see "Naming is the caller's job").
+
+> The code is the source of truth; for exact signatures see the kdoc in `src/nvme/tid.h`.
+
+## Fields
+
+A TID carries eight fields:
+
+- **Addressing** — `transport`, `traddr`, `trsvcid`, `host_traddr`, `host_iface`.
+- **Subsystem** — `subsysnqn`.
+- **Host identity** — `hostnqn`, `hostid`.
+
+`libnvmf_tid_get_canonical()` renders these in fixed field order (`transport=tcp;traddr=1.2.3.4;trsvcid=8009;…`), omitting unset fields, so the same logical TID always yields the same string regardless of set order.
+
+## Addressing is numeric-only
+
+`traddr`/`host_traddr` must be numeric IP addresses; a hostname makes construction fail — libnvme never resolves, defers, or carries one (rationale: `design/INTEGRATION.md`). `libnvmf_traddr_is_numeric(traddr)` checks a candidate address up front, using the same numeric definition the constructors enforce internally (an IPv6 scope suffix like `fe80::1%eth0` counts as numeric).
+
+## Sanitization
+
+Sanitization normalizes an address that *will* be accepted into one consistent spelling — it is not validation. It makes the canonical form deterministic per producer, and gives the connection-identity matcher (below) cheap string equality instead of a semantic address comparison.
+
+This is structural, not a convention to remember: addressing fields are only ever set via `libnvmf_tid_from_fields()`, `libnvmf_tid_parse()`, or `libnvmf_tid_dup()` — the per-field addressing setters are gone from the public API — and `transport` is an immutable constructor argument, since sanitizing an address requires knowing the transport.
+
+| Field | tcp / rdma | fc | loop |
+|---|---|---|---|
+| `traddr` | numeric → canonicalize; hostname → **fails** | as-is | as-is |
+| `host_traddr` | numeric → canonicalize; hostname → **fails** | as-is | (n/a) |
+| `trsvcid` / `host_iface` | as-is | (n/a) | (n/a) |
+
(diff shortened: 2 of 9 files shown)


libnvme/design/INTEGRATION.md
```diff
@@ -0,0 +1,55 @@
+# Integrating libnvme into an application
+
+libnvme is a synchronous NVMe protocol engine. This document draws the line between what the library does and what the application that uses it must do, and it explains how a process built around an event loop is expected to call libnvme without wrecking that loop. The single sentence the rest of this document unpacks:
```
igaw: I am fine with those documents but they are very lengthy and repetitive IMO, this leads to an overload for human beings. Though my understanding we have these documents around for the LLMs to understand how the code is organized. If we this gets really out of hand we can still look into condensing them down.

libnvme/design/INTEGRATION.md
```diff
@@ -0,0 +1,55 @@
...
+
+The clearest symptom is that the CLI/library split, done roughly six years ago, extracted code by mechanical convenience rather than by principle. Nobody wrote down what belonged on each side, so responsibilities that are really the *application's* ended up inside the library. Hostname resolution is the canonical example: it sits deep inside libnvme today, even though — as this document argues — it was always the caller's job. That is not a criticism of any one change; it is what happens when a boundary is never given a stated rule to be judged against. This document is that rule.
```
igaw: `six years ago` will be stale next year.

libnvme/design/TID.md
```diff
@@ -0,0 +1,115 @@
...
+  - `0h` — the ratified default (TP4110); legal for a fabrics Connect where the controller supports it (Base 2.3 §5.2.12.3). But `0h` means "anonymous" — the kernel treats two `0h` hosts with different NQNs as a collision — so it does not scale past one persona.
+  - **derive from the HostNQN UUID** — for a UUID-format HostNQN (`nqn.2014-08.org.nvmexpress:uuid:<X>`), the deterministic HostID is `<X>` (TP4126 method b). This is unique per persona and reproducible, and libnvme already has `libnvme_hostid_from_hostnqn()`.
+
+  The system-wide resolution order for an absent hostid is: explicit → `/etc/nvme/hostid` → derive from the HostNQN UUID → `0h`, **never random**. Of these the TID itself performs only the UUID derivation; the other steps reach it as the explicit `hostid` the caller passes (a value read from `/etc/nvme/hostid`, or a deliberate `0h`). `set_identity()` never reads a file and never invents `0h` on its own — if it is handed no hostid and the hostnqn is not a UUID, it leaves the field unset for the caller to fill.
```
igaw: sure, it can be random.

libnvme/design/TID.md
```diff
@@ -0,0 +1,115 @@
...
+- `libnvmf_tid_is_empty(tid)` — true if the TID (or NULL) sets no fields.
+- The field getters.
+
+Removed in 3.0: all eight per-field setters — the five addressing ones (`set_transport/traddr/trsvcid/host_traddr/host_iface`) and the three identity ones (`set_subsysnqn/hostnqn/hostid`). Addressing is construction-only; identity goes through `set_identity`. Also removed: `get_hash`, `is_hashable`, `equal`, `resolve_hostname`, and the `hostname` member — hostnames, hashing, and TID-to-TID equality are not the library's concern (see above; `equal` in particular had no production caller and its exact-string semantics diverged from the address-aware comparators `libnvme_ipaddrs_eq`/`addr_equal` and the connection-identity matcher already in the tree).
```
igaw: I would remove this, not really helping

libnvme/src/nvme/tid.c
```diff
@@ -255,6 +272,96 @@ __libnvme_public bool libnvmf_tid_is_empty(const struct libnvmf_tid *tid)
...
+	return -EINVAL;
+}
+
+/*
+ * Sanitize the addressing fields once the transport is known: canonicalize a
+ * numeric traddr/host_traddr.  A hostname is rejected outright -- resolving
+ * one can block on DNS and picking among several results is a policy call,
+ * both of which belong to the caller (see design/INTEGRATION.md), so the TID
+ * never holds one.  IP addressing applies only to the IP transports; fc and
+ * loop are left untouched.
```
igaw: please reduce the number of repetition in the documentation. This is exactly what the doc on `canon_ip` says and several times in `design/*.md`.

libnvme/src/nvme/tid.h
```diff
@@ -67,9 +67,13 @@ bool libnvmf_tid_is_empty(const struct libnvmf_tid *tid);
  * @hostnqn:     Host NQN, or NULL.
  * @hostid:      Host Identifier, or NULL.
  *
- * Convenience constructor.  NULL fields are stored as NULL.
+ * Convenience constructor.  NULL fields are stored as NULL.  For an IP
+ * transport (tcp, rdma) a numeric traddr/host_traddr is canonicalized; a
+ * hostname is rejected -- resolving one is the caller's job (see
+ * design/INTEGRATION.md and libnvmf_traddr_is_numeric()), not libnvme's.
```
igaw: again, the AI is adding a lot of documentaiton repetition. This overloads any human being eventually. Please tune this down a bit, thanks.

libnvme/src/nvme/tid.c
```diff
@@ -139,37 +135,6 @@ __libnvme_public const char *libnvmf_tid_get_canonical(
...
-	 * keeps the derived unit name short while keeping collisions negligible
-	 * at realistic scale: for ~200 concurrently-connected controllers the
-	 * birthday-bound collision probability is on the order of 1 in 1e10.
-	 */
-	h = libnvmf_fnv1a_64(canon, strlen(canon)) & 0xffffffffffffULL;
-	if (asprintf(&p->_hash, "%012" PRIx64, h) < 0) {
-		p->_hash = NULL;
-		return NULL;
-	}
-	return p->_hash;
```
igaw, line 168: It is good that this is being removed, it was a bit odd from the beginning, but I couldn't really figure out why I don't like it. now it's clear, it's policy.

igaw:
Looks all fine, my biggest concern is the amount of documentation which is bloaty/repetitive and sometimes too specific (e.g. time references, change logs,...). It overloads not just me, I got feedback that people are just not reading the lengthy commit messages and documentation. While it is nice to document the code base, we also have to keep it balanced, it is not just for the machines.

martin-belanger:
> Looks all fine, my biggest concern is the amount of documentation which is bloaty/repetitive and sometimes too specific (e.g. time references, change logs,...). It overloads not just me, I got feedback that people are just not reading the lengthy commit messages and documentation. While it is nice to document the code base, we also have to keep it balanced, it is not just for the machines.

Agreed, I'll trim the docs.

libnvme/design/INTEGRATION.md
martin-belanger: Agreed, but these docs are not for LLMs only. It's for whoever picks up the fabrics code next and avoid bringing policy code back into libnvme.

libnvme/design/INTEGRATION.md
martin-belanger: Good catch. All I wanted to say is that the CLI/library split predates the boundary rule of keeping policy out of libnvme.

libnvme/design/TID.md
martin-belanger: You're right. The actual requirement is that the hostid stays *stable* across connects, not that it's chosen deterministically — `/etc/nvme/hostid` itself is typically populated with a randomly-generated UUID the first time, and that's fine because it's then persisted and reused for every connect after. 

What breaks trackability is regenerating a fresh random value on *every* connect, not the fact that the value's origin was random once. I'll reword "hostid must be deterministic — never random" to "hostid must be stable across connects".

libnvme/design/TID.md
martin-belanger: Will do

libnvme/src/nvme/tid.c
martin-belanger: Will do

libnvme/src/nvme/tid.c
martin-belanger: Agreed — at first I thought this could be used in different places, but then it's only discoverd that needs this and as you said, it's policy.

martin-belanger:
@igaw - trimmed the repeated rationale in INTEGRATION.md, TID.md, and the matching kdoc. Reworded the hostid "never random" claim. Docs are now roughly 40% of their original length.

libnvme/design/INTEGRATION.md
martin-belanger: Done

libnvme/design/INTEGRATION.md
martin-belanger: Done

libnvme/design/TID.md
martin-belanger: Done

libnvme/src/nvme/tid.h
martin-belanger: Done

libnvme/src/nvme/tid.c
martin-belanger: Done

libnvme/src/nvme/tid.c
martin-belanger: Done

libnvme/design/TID.md
martin-belanger: Done

libnvme/design/INTEGRATION.md
```diff
@@ -0,0 +1,39 @@
...
+- **Policy** — any choice with more than one defensible answer (which resolved address to use, which source address to bind, how hard to retry). The right answer depends on the deployment, not on NVMe, so policy belongs to the caller.
+- **Non-determinism** — anything not a pure function of its inputs (DNS is the archetype). It poisons anything downstream that needs to be reproducible.
+- **Concurrency** — *when* and on which thread a call runs relative to the application's event loop. Only the application knows its loop.
+
+Hostname resolution fails the policy and non-determinism tests outright, and (below) also blocks — three independent reasons it belongs to the caller, not the library.
+
+## libnvme is synchronous and blocking
+
+Every libnvme API may block the calling thread: connect blocks on the `/dev/nvme-fabrics` write, Get Log Page and Identify block on ioctls, and even device-tree enumeration is real sysfs I/O. No file descriptors to poll, no callbacks, no loop of its own — deliberately: integrating with every event-loop implementation in existence (sd_event, GLib, raw epoll, libuv, ...) doesn't scale, and would lock every caller into whichever one libnvme picked.
```
igaw: With the proper integration of io_uring, it's also possible to submit NVMe commands asynchronously. Maybe this should noted here or maybe rewritten that the standard behavior is blocking. Because the caller needs to do some work to get exposed to the asynchronous behavior.

libnvme/design/INTEGRATION.md
```diff
@@ -0,0 +1,39 @@
...
+
+## The main-loop contract
+
+A process with a main loop must never call libnvme from the loop thread. Offload it:
+
+- **Worker thread** — for in-process blocking calls (Discovery Log Page, Identify, sysfs scan); post the result back via an eventfd/pipe the loop already watches. This is how nvme-stas drives nearly every libnvme call.
+- **Subprocess or systemd unit** — for connect, when isolation and lifecycle supervision are wanted too.
+
+discoverd uses both: it retrieves Discovery Log Pages on a worker thread but performs connects by starting a systemd unit, so the fully-blocking call never touches its loop.
```
igaw: Mabye we should consistently call it `nvme-discoverd`, just to avoid name collision with anything else?

igaw:
Looks good, just my usual nitpicking. 

and the reduction/cleanup in the docs is way better, thanks a lot!

libnvme/design/INTEGRATION.md
martin-belanger: Updated

libnvme/design/INTEGRATION.md
martin-belanger: Done

igaw:
Thanks a lot!

PR #3556, opened 2026-07-07
sfx-nvme: fix sizeof mismatch in calloc in nvme_dump_evtlog()

sahmed-ibm:
The nvme_dump_evtlog() function allocates memory for @pevent using calloc() with the nmemb and size arguments swapped. sizeof(*pevent) was passed as the element count instead of as the allocation size.

Fix this by using calloc(1, sizeof(*pevent)) to correctly allocate a single instance of the struct.


igaw:
Thanks!

btw, you should probably not using '@varname1223447441144555441' annotation as this links to random people which happen to have this handle on github.

PR #3557, opened 2026-07-07
Re-Scan(refresh) topology using libnvme_refresh_topology()

shroffni:
The first patch updates the signature of libnvme_refresh_topology() so that it returns the status of refresh operation. The callers of this function can evaluate the return status and based on that take necessary action.

The second patch avoids recreating nvme global context whenever nvme top detects topology change and instead now uses  libnvme_refresh_topology() to rebuild the topology. The nvme global context is now only created once at the beginning during nvme top execution.


igaw:
Thanks!

PR #3558, opened 2026-07-07
libnvme: move hostname resolution out of the library and drop netdb

martin-belanger:
## Summary

Pulls hostname resolution out of libnvme entirely, along with the netdb dependency itself -- no `<netdb.h>`, no `NVME_HAVE_NETDB`, anywhere in libnvme or its tests. The CLI resolves `-a`/`--host-traddr` before libnvme ever sees the address. libnvme rejects a hostname at connect time, canonicalizing a numeric address through the TID constructor.

This is policy moving where it belongs: which address a hostname resolves to (IPv4/IPv6), and when to block on DNS to find out, is a caller decision, not a library one.

## Why now

Prep-patch for the INI config parser: config entries are user-authored and may legitimately contain a hostname. The consumer side needs to resolve raw entries itself before building a TID, and that only works if the resolve-in-the-caller model already exists. This PR establishes it.

## What changed

- `libnvme_ipaddrs_eq()` and friends: `getaddrinfo(AI_NUMERICHOST)` → `inet_pton()`. Side effect: these silently no-op'd on every static build (`NVME_HAVE_NETDB` is always false there) -- now they work.
- `nvmf_sanitize_addrs()`: rejects a hostname traddr/host_traddr at connect time, canonicalizes a numeric one via the TID constructor.
- CLI `nvmf_resolve_addr()`: resolves before handing off to libnvme. Crude and sequential, one blocking `getaddrinfo()` at a time -- nvme-cli is a one-shot tool.
- Test gating moved off `NVME_HAVE_NETDB` (always false on static builds) onto `want_fabrics`, which is what these tests actually depend on now.

## Behavior deltas

- `nvme connect`/`connect-all` with a hostname: unchanged for users, resolution just happens earlier.
- Passing a hostname straight into a libnvme fabrics API (Python bindings, `libnvmf_add_ctrl()`) now fails with a clear error instead of resolving. nvme-stas resolves before calling, so no real impact there.
- A numeric address is now canonicalized before reaching the kernel, so sysfs reports the same spelling the TID does.

## Verified

Build (default, `-Dfabrics=disabled`, `--default-library=static`), full test suite, checkpatch, valgrind, and manual connect/connect-all against a live nvmet target with a hostname traddr.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -36,6 +36,13 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#ifdef NVME_HAVE_NETDB
+#include <netdb.h>
+
+#include <arpa/inet.h>
+#include <sys/socket.h>
+#endif
+
 #include <libnvme.h>
 
 #ifdef NVME_HAVE_LIBKMOD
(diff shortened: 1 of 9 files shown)


igaw:
Looks good. Getting this blocking DNS lookup out of the library is good thing!

Thanks!

Issue #3530, opened 2026-07-01
Replace environmant variables settings with global context


igaw:
There are a handful of environment variables which control the library. With the introduction of the global context it should be possible to use it to configure the library. 

library behavior:

- [x] - `LIBNVME_FORCE_4K`
- [x] - `LIBNVME_MI_PROBE_ENABLED`

testing: 

- [x] - `LIBNVME_SYSFS_PATH` 
- [x] - `LIBNVME_HOSTNQN` 
- [x] - `LIBNVME_HOSTID`

PR #3566, opened 2026-07-08
libnvme/config: add an INI-format NVMe-oF config parser

martin-belanger:
Hi Daniel,

Here's the big PR I promised — ~4,400 lines total, but only ~1,400 of that is actual code and another ~1,400 tests; the rest is kdoc/comments, blank lines, and the design doc. It's self-contained: nothing calls into it yet, so no regression risk today. It's this size because it does two things the JSON blob didn't: real validation with `file:line` diagnostics, and a drop-in file layout.

Suggested order:

1. `CONFIG.md` first (the design contract), 
2. then `config.h` (the whole public surface), 
3. then the commits in order: 
3.1 **tokenizer**: libnvme/config: add an internal INI reader
3.2 **param bag**: libnvme/config: add the parameter bag and key table 
3.3 **raw parser**:  libnvme/config: parse a config file into its raw model
3.4 **cascade resolver**: libnvme/config: resolve the cascade into the flat connection list
3.5 **public API**:  libnvme/config: publish the resolved configuration
3.6 **arg emitter**:  libnvme/config: add the connect argument emitter
4. Last commit is unrelated — a pre-existing `host_iface`/`trsvcid` copy-paste mixup in `libnvmf_add_ctrl()`, fixed while I was in the area.

A second (much smaller) PR (write side + `nvme config convert`) follows once this one lands.


libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,353 @@
...
+
+[Discovery Controller Defaults]
+keep-alive-tmo = 30                  # DC keep-alive default
+
+[I/O Controller Defaults]
+keep-alive-tmo = 5                   # IOC keep-alive default (kernel default)
+
+[Host]
+# hostnqn comes from /etc/nvme/hostnqn by default
+# hostid comes from /etc/nvme/hostid by default
```
igaw: I wonder if we should not move them also into the global section. WDYT?

though I haven't figured out what the impact of this would be... if those stay separate, it would be good to document it here why this is the case.

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,353 @@
...
+
+A drop-in typically adds a new host persona — opening with its own `[Host]` block naming that persona's `hostnqn` — plus the connections made under that identity. (A drop-in may also omit `[Host]` entirely, adding more connections under the default persona instead.) It may also carry its own copy of `[Global]` or either `*Defaults` section, each a scoped overlay limited to that drop-in's own connections (see [Precedence](#precedence-and-the-merge-model)).
+
+```ini
+# /etc/nvme/nvme-fabrics.conf.d/prod-fabric.conf
+[Host]
+hostnqn       = nqn.2014-08.org.nvmexpress:uuid:1111…-A
+hostid        = 46ba5037-7ce5-41fa-9452-48477bf00080
+hostsymname   = Hydra
+dhchap-secret = DHHC-1:00:…
```
igaw: Maybe remove the dhcap example here, we should not encourage people to use this method for loading the keys into the kernel. Another big TODO on the list is to make the secret loading a bit more useful for production systems.

Or just rename the example testing-fabrics.conf :)

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,353 @@
...
+- a key left unset inherits from the top-level section;
+- nothing leaks to sibling drop-ins or back to the top level.
+
+The `[Global]` / type-default rungs are computed per file this way, so the result never depends on the order drop-ins are read in.
+
+**Exception — per-link security stays at the section level.** The TLS and DH-CHAP parameters (see [Security parameters](#security-parameters)) are *not* overridable on a `controller =` line: the PSK and authentication secrets are bound to the `(hostnqn, subsysnqn)` pair, so they are constant across all paths to an endpoint and belong on the endpoint section. Everything else — `ctrl-loss-tmo`, `keep-alive-tmo`, digests, queue counts, `reconnect-delay`, … — can be overridden per path on the `controller =` line.
+
+### Security parameters
+
+A connection's security settings fall into two families — **authentication** (DH-CHAP, which proves the host and controller identities in-band) and **encryption** (TLS, which protects the data on the wire) — plus one parameter that bridges them. Because they are bound to the host+subsystem relationship, they live on the **endpoint section**, not on a `controller =` path (the exception above).
```
igaw: Maybe add something adding secrets to the config file are there for testing purposes. It's not a good idea to do it in a production system. The idea is that the keys are handled by a 3rd party which loads them into the kernel keyring. I was looking into how this works with the systemd ecosystem. Unfortunately, never got anywhere. That's why currently we have this very simple udev rule which loads the file content in /etc/nvme/keys into the kernel. Not really good either, it's stop gap solution.

libnvme/src/nvme/ini.c
```diff
@@ -0,0 +1,187 @@
...
+
+	return callback(LIBNVMF_INI_KV, *section, key, value, line,
+			user_data);
+}
+
+/*
+ * The tokenizing loop, over a buffer the caller already owns and may
+ * modify in place.  libnvmf_ini_parse_buf() dups its const input once to
+ * get one of these; libnvmf_ini_parse_file() already allocated its own and
+ * passes it straight through, with no second copy.
```
igaw: I'd say the second part should be removed, that is just reasoning not really helping the reader to understand this function.

libnvme/src/nvme/ini.c
```diff
@@ -0,0 +1,187 @@
...
+	if (!*s || *s == '#')
+		return 0;
+
+	/* Case 2: a section header... */
+	if (*s == '[') {
+		end = strchr(s, ']');
+
+		/* ...unless unterminated, empty, or followed by text. */
+		if (!end || end == libnvmf_trim(s + 1) ||
+		    *libnvmf_trim(end + 1)) {
```
igaw: running libnvmf_trim twice is a bit excessive.

libnvme/src/nvme/ini.c
```diff
@@ -0,0 +1,187 @@
...
+
+int libnvmf_ini_parse_buf(struct libnvme_global_ctx *ctx, const char *text,
+		libnvmf_ini_fn callback, void *user_data)
+{
+	__cleanup_free char *copy = NULL;
+
+	if (!ctx || !text || !callback)
+		return -EINVAL;
+
+	copy = xstrdup(text);
```
igaw: on line 128 we test for `!text` so a simple `strdup` should be fine or drop the check in line 128

libnvme/src/nvme/ini.c
```diff
@@ -0,0 +1,187 @@
...
+
+	return parse_lines(copy, callback, user_data);
+}
+
+int libnvmf_ini_parse_file(struct libnvme_global_ctx *ctx, const char *path,
+		libnvmf_ini_fn callback, void *user_data)
+{
+	__cleanup_free char *text = NULL;
+	struct stat st;
+	FILE *f;
```
igaw: `__cleanup_file FILE *f = NULL` would avoid those explicit `fclose` in the error path

libnvme/src/nvme/ini.c
```diff
@@ -0,0 +1,187 @@
...
+	}
+	fclose(f);
+	text[len] = '\0';
+
+	/* An embedded NUL means this is not the text file we expect. */
+	if (memchr(text, '\0', len))
+		return -EINVAL;
+
+	return parse_lines(text, callback, user_data);
+}
```
igaw, line 180: Nice little parser. I like :)

libnvme/test/ini.c
```diff
@@ -0,0 +1,302 @@
...
+	return check(ctx, "CRLF line endings", "[s]\r\nk = v\r\n", want, 2);
+}
+
+static int abort_second(enum libnvmf_ini_event event, const char *section,
+			const char *key, const char *value, unsigned int line,
+			void *user_data)
+{
+	int *count = user_data;
+
+	if (++(*count) == 2)
```
igaw: This is not that easy to read (remembering the operator preference rules...), maybe make this a bit explicit.

libnvme/src/nvme/config-ini.c
```diff
@@ -0,0 +1,285 @@
...
+struct kv {
+	struct kv *next;
+	char *key;
+	char *value;
+};
+
+struct libnvmf_params {
+	struct kv *head;
+	struct kv **tail;
+};
```
igaw, line 151: fwiw, there is also a ccan list implementation, would avoid another open coded version...

libnvme/src/nvme/config-ini.c
```diff
@@ -0,0 +1,285 @@
...
+	if (!e)
+		goto fail;
+	e->key = strdup(key);
+	if (!e->key) {
+		free(e);
+		goto fail;
+	}
+	e->value = copy;
+	*p->tail = e;
+	p->tail = &e->next;
```
igaw: this part could go into a smaller function something like `libnvmf_params_new_add` (a good name obviously is needed), would document what this snippet does.

libnvme/src/nvme/config-ini.c
```diff
@@ -0,0 +1,285 @@
...
+
+	if (!p || !key)
+		return NULL;
+	e = kv_find(p, key);
+
+	return e ? e->value : NULL;
+}
+
+int libnvmf_params_merge(struct libnvmf_params *dst,
+		const struct libnvmf_params *src)
```
igaw, line 249: is the order of the arguments sane? I usually prefer it to have it `src`, `dst` but that's me. If there are other interfaces which follow this pattern, than okay, but if not, I'd rather have the other way around.

libnvme/src/nvme/config-ini.c
```diff
@@ -0,0 +1,285 @@
...
+	return -EINVAL;
+}
+
+static int check_int(const char *value)
+{
+	char *end;
+	long v;
+
+	errno = 0;
+	v = strtol(value, &end, 10);
```
igaw: could be also support hex values?

libnvme/src/nvme/config-ini.c
```diff
@@ -79,7 +86,7 @@ int libnvmf_params_set(struct libnvmf_params *p, const char *key,
 	if (!p || !key || !value)
 		return -EINVAL;
 
-	copy = strdup(value);
+	copy = xstrdup(value);
```
igaw: value is NULL checked, so strdup works just fine.

libnvme/src/nvme/config-ini.c
```diff
@@ -93,7 +100,7 @@ int libnvmf_params_set(struct libnvmf_params *p, const char *key,
 	e = calloc(1, sizeof(*e));
 	if (!e)
 		goto fail;
-	e->key = strdup(key);
+	e->key = xstrdup(key);
```
igaw: same here key is NULL checked.

libnvme/src/nvme/config-ini.c
```diff
@@ -176,37 +183,37 @@ void libnvmf_params_for_each(const struct libnvmf_params *p,
...
-	{ "hostsymname",		LIBNVMF_KEY_STRING,	LIBNVMF_KEY_IDENTITY },
+	{ "hostnqn",		LIBNVMF_KEY_STRING,	LIBNVMF_KEY_IDENTITY },
+	{ "hostid",		LIBNVMF_KEY_STRING,	LIBNVMF_KEY_IDENTITY },
+	{ "hostsymname",	LIBNVMF_KEY_STRING,	LIBNVMF_KEY_IDENTITY },
 
 	/* endpoint sections only */
-	{ "nqn",			LIBNVMF_KEY_STRING,	LIBNVMF_KEY_NQN },
-	{ "controller",			LIBNVMF_KEY_STRING,	LIBNVMF_KEY_CONTROLLER },
+	{ "nqn",		LIBNVMF_KEY_STRING,	LIBNVMF_KEY_NQN },
+	{ "controller",		LIBNVMF_KEY_STRING,	LIBNVMF_KEY_CONTROLLER }
```
igaw: these whitespace changes are not really necessary. could all go into the previous patch.

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +88,68 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+};
+
+struct libnvmf_conf_file {
+	char *path;			/* provenance, diagnostics */
+	struct libnvmf_params *global;	/* [Global]; empty when absent */
+	/* [Discovery Controller Defaults] */
+	struct libnvmf_params *dc_defaults;
+	struct libnvmf_params *ioc_defaults; /* [I/O Controller Defaults] */
+	bool has_host;			/* file carries a [Host] section */
+	char *hostnqn;			/* "" records an explicit reset */
```
igaw: this comment is redundant, no? I though this is the model for all parameters.

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +88,68 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+	 * raw strings, not a TID: a config file is a human interface and may
+	 * legitimately name a host, and TID construction rejects hostnames
+	 * (libnvme/design/TID.md) -- resolving them is the consumer's job.
+	 * subsysnqn and the host identity are filled in at resolve time.
+	 */
+	char *transport;
+	char *traddr;
+	char *trsvcid;		/* NULL when the file left it unset */
+	char *host_traddr;	/* NULL when the file left it unset */
+	char *host_iface;	/* NULL when the file left it unset */
```
igaw: do we need those comments, I find this obvious.

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +88,68 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+struct libnvme_global_ctx;
+
+struct libnvmf_conf_path {
+	struct libnvmf_conf_path *next;
+	/*
+	 * Addressing as the file spelled it.  traddr/host_traddr are kept as
+	 * raw strings, not a TID: a config file is a human interface and may
+	 * legitimately name a host, and TID construction rejects hostnames
+	 * (libnvme/design/TID.md) -- resolving them is the consumer's job.
+	 * subsysnqn and the host identity are filled in at resolve time.
```
igaw: The first part of the documntation is good, the reasoning why is already lenghty explained in the documents.

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +88,68 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+	 */
+	char *transport;
+	char *traddr;
+	char *trsvcid;		/* NULL when the file left it unset */
+	char *host_traddr;	/* NULL when the file left it unset */
+	char *host_iface;	/* NULL when the file left it unset */
+	/* per-path tunables; never security */
+	struct libnvmf_params *overrides;
+	unsigned int line;	/* 1-based config-file line, for diagnostics */
```
igaw: the comment doesn't really help me at all, need to look at the code. maybe just say used 'for diagnostics'?

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +88,68 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+ * The raw model of ONE parsed file -- sections and lines faithfully
+ * recorded, nothing resolved.  The cascade resolver (which merges the
+ * top-level file with the drop-ins and expands endpoints into connections)
+ * consumes a list of these.  Struct members are open because this is an
+ * internal type; nothing here crosses the public API boundary.
+ */
+
+struct libnvme_global_ctx;
+
+struct libnvmf_conf_path {
```
igaw, line 126: The main issue I see with this interface is, that it's makes this part of the API fixed. We can't really change it after the release without breaking existing users. Do we really want to do this?

libnvme/src/nvme/config-ini.c
```diff
@@ -283,3 +290,543 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value)
...
+		return -ENOMEM;
+	ep->is_dc = is_dc;
+	ep->line = line;
+	ep->params = libnvmf_params_new();
+	if (!ep->params) {
+		free(ep);
+		return -ENOMEM;
+	}
+
+	for (tail = &pc->f->endpoints; *tail; tail = &(*tail)->next)
```
igaw: if we move the structs into the implementation, we just could use the ccan list APIs...

libnvme/src/nvme/config-ini.c
```diff
@@ -283,3 +290,543 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value)
...
+	return 0;
+}
+
+static int enter_section(struct conf_parse *pc, const char *name,
+			 unsigned int line)
+{
+	pc->ep = NULL;
+
+	if (!strcmp(name, "Global")) {
+		if (pc->f->global)
```
igaw:
```
struct libnvmf_conf_file *f = pc->f;
```

would make it a bit simpler to read IMO.

libnvme/src/nvme/config-ini.c
```diff
@@ -0,0 +1,285 @@
...
+	{ "reconnect-delay",		LIBNVMF_KEY_INT,	LIBNVMF_KEY_TUNABLE },
+	{ "ctrl-loss-tmo",		LIBNVMF_KEY_INT,	LIBNVMF_KEY_TUNABLE },
+	{ "fast-io-fail-tmo",		LIBNVMF_KEY_INT,	LIBNVMF_KEY_TUNABLE },
+	{ "tos",			LIBNVMF_KEY_INT,	LIBNVMF_KEY_TUNABLE },
+	{ "duplicate-connect",		LIBNVMF_KEY_BOOL,	LIBNVMF_KEY_TUNABLE },
+	{ "disable-sqflow",		LIBNVMF_KEY_BOOL,	LIBNVMF_KEY_TUNABLE },
+	{ "hdr-digest",			LIBNVMF_KEY_BOOL,	LIBNVMF_KEY_TUNABLE },
+	{ "data-digest",		LIBNVMF_KEY_BOOL,	LIBNVMF_KEY_TUNABLE },
+
+	/* security -- bound to (hostnqn, subsysnqn); never per-path */
```
igaw: So that means our config.json is actually wrong, no? There it's model into the port section, which is 'per-path' or do I get this wrong again... confused...

igaw: just for reference

Image

libnvme/src/nvme/config-resolve.c
```diff
@@ -0,0 +1,410 @@
...
+
+static int dropin_filter(const struct dirent *d)
+{
+	const char *dot = strrchr(d->d_name, '.');
+
+	return dot && dot != d->d_name && !strcmp(dot, ".conf");
+}
+
+struct libnvmf_conf *libnvmf_conf_load(struct libnvme_global_ctx *ctx,
+		const char *path, int *err)
```
igaw: I would suggest to follow the error return code pattern used. That is the result is the last argument and not the return code.

libnvme/src/nvme/config-ini.h
igaw: unless this header is not exposed as API... :)

libnvme/src/nvme/config.h
```diff
@@ -0,0 +1,323 @@
...
+ *
+ * Reads the main file (if it exists) and the drop-ins (if the directory
+ * exists) and resolves them into the flat connection list.  An absent
+ * configuration is valid and yields an empty list.  Any validation error
+ * in any file fails the read as a unit; diagnostics carry file:line and
+ * are reported through @ctx.
+ *
+ * Return: the resolved configuration, or NULL on error.
+ */
+struct libnvmf_config *libnvmf_config_read(struct libnvme_global_ctx *ctx,
```
igaw: I'd prefer to return an error code here and have the result as last argument.

libnvme/src/nvme/config-ini.c
igaw: and after reading through all (and remembering the cover letter...) I'd say it makes sense to use the ccan lists which also give some level of type information.

igaw:
Generally, I really like this. It's a big PR but that was to be expected. And it's good to see the whole picture. Also the diagnostic part is a nice feature to have. The public API is not that big after realizing what is exposed and what not. Good job!

The main points from my first review pass is mostly nitpicking (function signatures and not using ccan list) 

My big question is about config.json: is this correctly modeled? From my current understanding it is not... and the other question is: do we  want to keep `/etc/nvme/hostnqn` and `/etc/nvme/hostid` alive?

libnvme/design/CONFIG.md
martin-belanger: Deliberately separate: `[Global]` can pass down to drop-ins (cascading default), `[Host]` cannot (a persona is a complete identity, isolated per file — no `[Host]` means the system default, never another file's `[Host]`). Will add a line here explaining that.

libnvme/design/CONFIG.md
martin-belanger: Will do — dropping the literal secret from the worked example. Separately drafting a real proposal for production secret handling (TPM-backed, reusing the kernel-keyring pattern the TLS PSK path already has) — will share once it's in shape.

libnvme/design/CONFIG.md
martin-belanger: Will do — will add that note. This is the same thread as the production-secrets proposal above; it'll cover both `dhchap-secret`/`dhchap-ctrl-secret` here and the `/etc/nvme/tls-keys` udev-import path, since they're the same underlying problem.

I'm also looking into what systemd has to offer for this.

libnvme/src/nvme/ini.c
martin-belanger: Fair — the outer `parse_lines()` already trims the whole line, so this second call only needs to strip trailing whitespace before `=`, not redo the leading trim. Will tighten it.

libnvme/test/ini.c
martin-belanger: Will do

libnvme/src/nvme/config-ini.c
martin-belanger: This one's `dst, src` to match the C library's copy-function convention — `memcpy`, `memmove`, `strcpy`, `strcat` are all destination-first, and that's the closest precedent to what this function does (copy src's keys into dst). No strong objection to `src, dst` if you'd rather standardize the other way.

libnvme/src/nvme/config-ini.c
martin-belanger: None of the current integer keys (queue counts, timeouts) have a real hex use case, and decimal-only avoids the classic `strtol` leading-zero-means-octal footgun for free. Would rather add it if a concrete key needs it than support it speculatively — easy to add later if you have one in mind.

libnvme/src/nvme/config-ini.h
martin-belanger: Correct — `config-ini.h` is internal-only, not installed, not part of the public header set (`config.h` is the public surface). Nothing here crosses the ABI boundary, so the struct layout is free to change any time.

libnvme/src/nvme/config-ini.c
martin-belanger: Will do — and since it's touching the same code as the ccan `list` migration, I'll fold this into that same commit rather than doing it twice: once `struct kv` moves to `ccan/list` the manual tail-splice disappears anyway, replaced by `list_add_tail()`, so the extracted helper will be small by construction. Following this file's existing static-helper naming (`kv_find()`, no `libnvmf_` prefix since it's internal) — will call it `kv_append()` unless you'd rather something else.

libnvme/src/nvme/ini.c
martin-belanger: **CORRECTION**:  This isn't actually duplicate computation — the two calls check different things on different substrings: one tests for an empty bracketed name, the other for trailing garbage after `]`. But you're right it reads dense as one compound condition — will split it into two clearly-named checks.

libnvme/src/nvme/config-ini.h
martin-belanger: Fair — the "why raw strings, not a TID" reasoning duplicates what TID.md already explains at length. Will trim this to state the fact plus a doc reference, rather than re-derive the reasoning inline.

libnvme/src/nvme/config-ini.c
martin-belanger: Checked the spec on this one — Base Spec 2.3-2025.08.01 §8.3.5.5.7 and TCP Transport 1.2-2025.08.01 §3.6.1.4 both tie the secret/PSK to the (host, subsystem) pair, and the TCP spec explicitly prohibits a different PSK per controller when a retained one exists. The kernel's own nvmet target agrees: `nvmet_port` is genuinely the transport endpoint (one port, many subsystems), `nvmet_ctrl` is the controller instance, and the administratively configured secret lives on `nvmet_host`, not the port. So `config.json`'s per-port storage was extra flexibility the protocol doesn't actually allow using — looks like it just mirrored the host driver's sysfs layout (`dhchap_secret` only exists in `nvme_dev_attrs[]`, not `nvme_subsys_attrs[]`) rather than a deliberate design choice. Pinning security to the endpoint section here is correct, not a regression, and `controller=` matches the kernel's own terminology (nvmet keeps port/controller/secret as three distinct things too).

libnvme/design/CONFIG.md
igaw: got it. then the question is, should we opt for something like globals.conf to have only one style of configurations?

libnvme/design/CONFIG.md
igaw: systemd has ways to store secrets protected by TPM. In the end we should first figure out what use cases are actually relevant before jumping into the design/implementation phase. So far nvme-cli tried to stay away from key management and provisioning of systems. IIRC, IHVs have solutions for fleet management, so the question is do we want to replicated that or do we need to provide some hooks/plugins etc?

libnvme/src/nvme/ini.c
igaw: oops, I read twice the same argument....

libnvme/src/nvme/config-ini.c
igaw: Me neither, it's more a question of consistency. I need to check what we do in our functions. more API review :)

libnvme/src/nvme/config-ini.c
igaw: I ask because the kernel accepts hex as arguments and we have some dance to do anyway in build_args.

libnvme/src/nvme/config-ini.h
igaw: Very nice!

libnvme/src/nvme/config-ini.c
igaw: Thanks for the review. I'd say when doing the conversion those secrets are moved to the endpoint section. There shouldn't be any problems with this approach. Just something to remember when writing the release notes.... hopefully AI helps here.

martin-belanger:
> and the other question is: do we want to keep /etc/nvme/hostnqn and /etc/nvme/hostid alive?

hostnqn/hostid — before answering "keep or drop the `/etc/nvme` files," I think there's a more basic question: should libnvme be picking a hostnqn/hostid by itself at all? Traced it — `lookup_host()` (core connect path, `fabrics.c`) calls `libnvmf_host_get_ids()`, which has a whole baked-in precedence chain: DMI/device-tree → /etc/nvme files → config.json's first host → CLI args → random-generate if still nothing. That's libnvme deciding policy, same category of thing we already pulled out for hostname resolution. Feels like it should go the same way: libnvme requires an explicit hostnqn+hostid to connect, full stop, and every orchestrator (CLI included) is on the hook to supply one — however it gets it, `/etc/nvme`, a config file, DMI, or freshly generated. What's your read?

igaw:
> > and the other question is: do we want to keep /etc/nvme/hostnqn and /etc/nvme/hostid alive?
> 
> hostnqn/hostid — before answering "keep or drop the `/etc/nvme` files," I think there's a more basic question: should libnvme be picking a hostnqn/hostid by itself at all? Traced it — `lookup_host()` (core connect path, `fabrics.c`) calls `libnvmf_host_get_ids()`, which has a whole baked-in precedence chain: DMI/device-tree → /etc/nvme files → config.json's first host → CLI args → random-generate if still nothing. That's libnvme deciding policy, same category of thing we already pulled out for hostname resolution. Feels like it should go the same way: libnvme requires an explicit hostnqn+hostid to connect, full stop, and every orchestrator (CLI included) is on the hook to supply one — however it gets it, `/etc/nvme`, a config file, DMI, or freshly generated. What's your read?

Generally, I'd say yes. Though there is a lot of complexity currently backed into `libnvmf_host_get_ids` and the code around it. The idea behind this function is that not every client of the API has to implement the same kind of lookup code, e.g. for the Python binding users. I agree it might be better to move this out and expect the correct hostnqn/hostid is provided on the 'connect/discover' command. Would it make sense to provide this function as util function?

libnvme/design/CONFIG.md
martin-belanger: I thought about this some more. I agree there are too many sections, which makes it hard to see the overall picture and understand how values cascade from the top level down to the drop-ins.

I don't think `globals.conf` (or `nvme-fabrics.conf.globals`) is the right solution either. We also need multi-persona support, which is exactly what `[Host]` sections are for (one file, one `[Host]` section, one identity).

I think the better approach is to drop `[Global]` entirely. `[Discovery Controller Defaults]` and `[I/O Controller Defaults]` become the only two default sections, and any default parameter that needs to apply to both (DC and IOC) simply gets written twice.

That's a genuine simplification. Today, determining the effective value for an I/O controller requires mentally merging three sources: the `controller=` line, `[I/O Controller Defaults]`, and `[Global]`. Without `[Global]`, there are only two. The duplication cost is small in practice, and security keys never lived in `[Global]` anyway, so there's no security advantage either way.

I'll rework this before merging. It'll touch the raw model, the cascade resolver, and the docs, examples, and tests.

libnvme/src/nvme/config-ini.c
martin-belanger: Good catch, and you're right — checked both ends: the kernel's `match_int()` uses base-0 parsing (`lib/parser.c`), and nvme-cli's own `argconfig.c` already does the same (`strtol(optarg, &endptr, 0)`) for every equivalent CLI flag. So decimal-only here was actually the inconsistency, not the safe choice — CONFIG.md's own rule is "what you write here is what you'd pass on the command line," and the command line already accepts hex/octal via base 0. Switching `check_int()` to match.

libnvme/src/nvme/config-ini.c
martin-belanger: Agreed, and noted for the converter — will make sure JSON-to-INI conversion (next PR) collapses any per-port secrets to one endpoint-level value during conversion, and calls that out explicitly in the release notes rather than doing it silently.

libnvme/design/CONFIG.md
martin-belanger: Fair pushback — I was heading toward "nvme-cli builds this itself" before actually settling whether that's the right posture. Given IHVs already have fleet-management/provisioning tooling, and nvme-cli's history of staying out of key management, I think the hooks/plugins framing deserves real consideration rather than a quick answer here. Let me think through the use-cases-first angle properly and come back with something more concrete rather than reply off the cuff.

martin-belanger:
> Would it make sense to provide this function as util function?

Like the util-function idea — makes sense to promote `libnvmf_host_get_ids()` to the public API (it's currently internal-only, `private-fabrics.h`, not even SWIG-visible) rather than duplicate that lookup logic per consumer. 

Plan, as a separate PR (not part of this one): `lookup_host()` stops calling it automatically — `connect`/`discover` require an explicit hostnqn/hostid, full stop, `-EINVAL` otherwise. The function itself moves to `fabrics.h` as a normal public util any caller can opt into, and the CLI becomes its first consumer: when `--hostnqn`/`--hostid` aren't given on the command line, nvme-cli will explicitly call this helper itself before invoking connect, instead of relying on it happening silently inside libnvme. 

nvme-stas actually built its own version of this exact lookup years ago; would have reused this if it had existed then.

libnvme/design/CONFIG.md
martin-belanger: Done

libnvme/design/CONFIG.md
martin-belanger: Done

libnvme/design/CONFIG.md
martin-belanger: Will follow up - marking as resolved

libnvme/src/nvme/ini.c
martin-belanger: Done

libnvme/src/nvme/ini.c
martin-belanger: Done

libnvme/src/nvme/ini.c
martin-belanger: Done

libnvme/src/nvme/ini.c
martin-belanger: Done

libnvme/test/ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: marking as resolved - we can always revisit later

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done - will follow up with the JSON-to-INI converter

libnvme/src/nvme/config-resolve.c
martin-belanger: Done

libnvme/src/nvme/config.h
martin-belanger: Done

martin-belanger:
@igaw — I saw your message about `libnvmf_host_get_ids()` and #3530. I can take a look at that once this PR is merged.

BTW, I've addressed all of the review comments. Let me know if you spot anything else that needs to be addressed.

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,360 @@
+# NVMe-oF Connection Configuration
+
+This is the host-side configuration that says **which NVMe over Fabrics (NVMe-oF) controllers a host should connect, and with what parameters**. It is the INI-format successor to `/etc/nvme/config.json`: the same information — a host identity, its subsystems and discovery controllers, and the per-path transport addresses — expressed in a `[Section]` / `key = value` format instead of JSON.
```
igaw: I find this a bit hard to read overall. The excessive use of dashes to add additional information is just tiresome. These should be simple sentences, not complex structures. Furthermore, I wouldn't mention the previous configuration format so often, especially not at the very beginning.

With my broken English, I'd rather see something like this:

`
This is the host-side configuration that specifies **which NVMe over Fabrics (NVMe-oF) controllers a host should connect to, and with what parameters**. The configuration file uses the INI format.
`

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,360 @@
...
+
+The format is parsed by **libnvme**, so every cooperating consumer reads one format through one parser: `nvme connect-all` / `nvme config` in nvme-cli, the **nvme-discoverd** daemon, and (through the Python bindings) **nvme-stas**. It sits alongside the two other host-wide coordination files — the [ownership registry](REGISTRY.md) and the [exclusion list](EXCLUSIONS.md) — and shares their transport-ID identity model (`src/nvme/tid.h`).
+
+> This document describes the configuration **format and intent**. Once the parser lands, the header kdoc and the `nvme-*` man pages are the authoritative reference; the code is the source of truth.
+
+## Why INI, and the move from `config.json`
```
igaw: I don't know if we really should add this section at all. It's good for the commit message. If you want to keep it, I'd moved to the end of the documentation. This files should really document about the current version of the configuration. The historical stuff can go to the NOTES section as with an good man page.

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,360 @@
...
+
+nvme-cli has historically stored saved connections in `/etc/nvme/config.json` (via the optional `json-c` library). That file is being phased out in favor of the INI format described here, for three reasons.
+
+**No third-party dependency.** This file is read by libnvme itself, with no external library required. JSON ties the config path to `json-c`, which is *optional* in libnvme (there are no-json builds); INI keeps the everyday config working on minimal and embedded targets where `json-c` may be unavailable or unwanted.
+
+**Ergonomics.** JSON has no comment support. An administrator cannot temporarily comment out a single controller, and the shipped file cannot be self-documenting. The commented-out lines and inline `#` notes shown throughout this document are not expressible in JSON at all.
+
+**It matches the systemd ecosystem.** An administrator editing `/etc/nvme/*.conf` is already working alongside systemd `.conf` / `.network` / `.service` files; the INI idiom is the familiar, hand-edit-friendly one — no braces, quoting, or indentation pitfalls.
+
+### How the transition works
```
igaw: We might want to have this section in a release notes for the package managers but not really here I think. It really distracts from the important part.

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,360 @@
...
+- **The JSON dump stays an optional nvme-cli feature.** Serializing the live tree to JSON has one real consumer — testing, and the "snapshot current state to a file" convenience. That dump lives in nvme-cli (which already links `json-c` for its other output) and stays `json-c`-gated; libnvme needs no `json-c` dependency to read connection config at all.
+
+> `discovery.conf` is also superseded. By its own man page it is "a list of connect-all commands to run" — limited to Discovery Controllers (DCs) by construction, with no way to express an I/O Controller (IOC) or global defaults. A line can carry `--hostnqn`/`--hostid`, but there is no `[Host]`-style persona: the pair has to be repeated on every line for one identity, and nothing validates the result. The format here covers the full set, so it replaces `discovery.conf` as well.
+
+## Single vs. multiple host personalities
+
+Before the mechanics, it helps to know the two shapes a configuration takes, because the second is less widely known and it drives much of the layout below.
```
igaw: What does this sentence try to tell me?

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,360 @@
...
+
+**Multiple host personalities.** A single host can deliberately present *different* identities — different `(Host NQN, Host Identifier)` pairs — to different fabrics, as though it were several independent hosts. Each such identity is a *personality*. Each one gets its own file under `nvme-fabrics.conf.d/`, opening with its own `[Host]` block; keeping one persona per file is what stops them from getting tangled.
+
+This second shape becomes important the moment **authentication and Transport Layer Security (TLS)** enter the picture, because credentials — a TLS Pre-Shared Key (PSK) or a DH-CHAP (Diffie-Hellman HMAC-CHAP) secret — are bound to a specific host identity and subsystem, not to the machine. Separate personalities let a host hold a *distinct* credential set per fabric, cleanly scoped: a key provisioned for one personality cannot be reused to reach another, and an administrator can rotate or revoke one fabric's credentials by editing just that persona's file. A host fronting several independently-administered or multi-tenant fabrics — each with its own access control and key material — is the canonical case.
```
igaw: I've asked chatgpt to rewrite this paragraph with 'techinical documantation, man page style, simple sentences, precise language'

`
This distinction is important when using authentication or Transport Layer Security (TLS). Authentication credentials, such as a TLS Pre-Shared Key (PSK) or a DH-CHAP (Diffie-Hellman HMAC-CHAP) secret, are associated with a host identity and an NVMe subsystem, not with the machine itself.

Separate host identities allow different credential sets to be configured for different NVMe over Fabrics (NVMe-oF) deployments. A credential configured for one host identity cannot be used by another. Credentials for a single deployment can be updated or revoked by modifying only the corresponding host configuration.

This configuration is intended for systems that connect to multiple independently administered or multi-tenant NVMe-oF deployments, each with its own access control policy and credentials.
`

I have a hard time understanding the original, the rewrite is way simpler to grasp. Not sure if it fully correct though.

martin-belanger: Using ChatGPT is a good idea. I will have it review the whole doc to see if there are other things that can be improved.

libnvme/src/nvme/ini.h
```diff
@@ -0,0 +1,43 @@
...
+#pragma once
+
+/*
+ * Internal INI reader for nvme-fabrics.conf.  It only tokenizes: section
+ * headers, "key = value" lines (an empty value is reported as "", distinct
+ * from an absent key), '#' comments and blank lines.  All meaning -- which
+ * sections and keys exist, singleton rules, precedence -- belongs to the
+ * consumer's callback, so the engine can serve consumers with different
+ * policies (fail-safe readers, strict validators).
+ */
```
igaw, line 21: What about something more in `man page` style of documentation?

```
/*
 * Internal INI parser for nvme-fabrics.conf.
 *
 * The parser handles only the INI syntax. It recognizes section headers,
 * key-value pairs in the form "key = value", comments starting with '#',
 * and empty lines. Empty values are returned as an empty string and are
 * distinguished from missing keys.
 *
 * The parser does not interpret sections or keys and does not enforce
 * configuration rules. Validation, duplicate handling, and precedence rules
 * are implemented by the caller through the consumer callback.
 */
```

libnvme/src/nvme/ini.h
```diff
@@ -0,0 +1,43 @@
...
+
+/*
+ * Called once per meaningful line (comments and blank lines are skipped).
+ * @section is the current section name, or NULL before the first header; for
+ * a SECTION event it is the newly entered section.  A malformed section
+ * header clears the current section, so the lines that follow it report
+ * @section == NULL rather than being misattributed -- the fail-safe
+ * direction.  @line is 1-based.  A non-zero return stops the parse and is
+ * returned to the libnvmf_ini_parse_*() caller.
+ */
```
igaw, line 46:
```
/*
 * Called for each parsed line. Comments and empty lines are ignored.
 *
 * @section contains the current section name. It is NULL before the first
 * section header. For a SECTION event, @section contains the new section
 * name.
 *
 * If a section header is malformed, the current section is cleared. Following
 * lines are reported with @section == NULL to avoid assigning them to an
 * invalid section.
 *
 * @line contains the line number.
 *
 * A non-zero return value stops parsing and is returned by
 * libnvmf_ini_parse_*().
 */
```

igaw:
the commit message for the second patch is also difficult understand. It mentions 'bag' without really defining what it is and the the code it doesn't really show up, only in a comment. I've gave chatgpt the commit message to transform into more technical tone:

```
Connection parameters need to distinguish between inheriting a value,
resetting to the kernel default, and setting an explicit value. Keep these
states separate so the configuration precedence rules can be represented
without ambiguity.

Define each supported configuration key in a single key table. The table
captures the value type and the configuration class that controls where the
key is valid (tunable, security, [Host] identity, or endpoint).

Do not carry forward the legacy numeric --tls_key option. Use the
hyphenated fast-io-fail-tmo spelling introduced with the 3.0 rename.
```

libnvme/src/nvme/config-ini.c
```diff
@@ -0,0 +1,303 @@
...
+
+#include "config-ini.h"
+
+/*
+ * The recognized keys.  One row per key the format accepts; the CLI option
+ * name is the only spelling (CONFIG.md "Parser conventions").  Note
+ * fast-io-fail-tmo is the hyphenated form the 3.0 CLI rename settles on, not
+ * today's --fast_io_fail_tmo spelling, and the legacy numeric --tls_key
+ * option is deliberately absent: its hyphenated name is taken by the current
+ * --tls-key and it is slated for removal with the key-command consolidation.
```
igaw: The whole Note section is something for the commit message. We don't need to document the past here.

libnvme/src/nvme/config-ini.h
```diff
@@ -0,0 +1,90 @@
...
+ * three states, which is what makes the precedence cascade and the
+ * reset-to-kernel-default form ("key =") expressible:
+ *
+ *   key absent      ->  unset: inherit from the next cascade level;
+ *   value == ""     ->  reset: stop inheriting, let the kernel default apply;
+ *   value == "..."  ->  set to that value.
+ *
+ * Values are kept as the strings the file spelled; they are validated on the
+ * way in (libnvmf_key_check_value) and interpreted on the way out (the
+ * emitter).  Iteration preserves first-insertion order.
```
igaw:
```c
/*
 * Stores connection parameters used during configuration cascading.
 * Each cascadeable parameter has one of three states:
 *
 *   key absent      -> unset: inherit from the next cascade level;
 *   value == ""     -> reset: stop inheriting and use the kernel default;
 *   value == "..."  -> set to the specified value.
 *
 * Keeping these states separate allows the precedence rules and the
 * reset-to-kernel-default form ("key =") to be represented without ambiguity.
 *
 * Values are stored as strings from the configuration file. They are
 * validated when added (libnvmf_key_check_value) and interpreted when
 * emitted. Iteration order follows the order in which keys were inserted.
 */
```

libnvme/src/nvme/config-ini.h
```diff
@@ -0,0 +1,90 @@
...
+ * Authors: Martin Belanger <Martin.Belanger@dell.com>
+ */
+#pragma once
+
+#include <stdbool.h>
+
+/*
+ * Internal building blocks for the INI connection config
+ * (libnvme/design/CONFIG.md): the connection-parameter bag with its
+ * three-state values, and the table of recognized keys.
```
igaw: maybe something like? it avoids to use of 'bag' which is undefined and the internal pointing around. This will run out sync eventually. 
```
/*
 * Internal building blocks for the INI connection configuration:
 * the connection-parameter store that tracks unset, reset, and explicit
 * values, and the table of recognized keys.
 */
```

libnvme/src/nvme/config-ini.h
```diff
@@ -0,0 +1,90 @@
...
+typedef void (*libnvmf_params_fn)(const char *key, const char *value,
+		void *user_data);
+void libnvmf_params_for_each(const struct libnvmf_params *p,
+		libnvmf_params_fn callback, void *user_data);
+
+/*
+ * The recognized configuration keys.  The key class encodes where a key may
+ * appear; the type drives value validation.  Addressing and identity fields
+ * that live in the TID (transport, traddr, ...) are deliberately NOT in this
+ * table -- controller= lines route them to the TID parser instead.
```
igaw: Because the 'why' is missing for the absence of the TID. I would just leave 'Addressing ...' away. I think everyone gets it, that is still the generic part of parser which tracks the connection parameters for the cascading resolving.

libnvme/src/nvme/config-ini.h
```diff
@@ -0,0 +1,90 @@
...
+ */
+enum libnvmf_key_type {
+	LIBNVMF_KEY_STRING,
+	LIBNVMF_KEY_INT, /* decimal integer; -1 is legal (e.g. ctrl-loss-tmo) */
```
igaw: If you need to document this, I would expected the type as comment, e.g. `int_t`, not where it is used.

libnvme/src/nvme/config-ini.h
```diff
@@ -0,0 +1,90 @@
...
+	LIBNVMF_KEY_BOOL /* 1/yes/y/true/t/on and 0/no/n/false/f/off */
```
igaw: Also this documentation seems a bit too much. If someone really cares, look at the parser.

libnvme/src/nvme/config-ini.c
```diff
@@ -0,0 +1,303 @@
...
+ */
+static const struct libnvmf_key keys[] = {
+	/* connection tunables -- cascade + per-path override */
```
igaw: Maybe add the reference to `@LIBNVMF_KEY_TUNABLE`? It connects the keys to the `enum libnvmf_key_class`

libnvme/src/nvme/config-ini.h
```diff
@@ -0,0 +1,90 @@
...
+};
+
+enum libnvmf_key_class {
+	LIBNVMF_KEY_TUNABLE,	/* any section; per-path override allowed */
+	LIBNVMF_KEY_SECURITY,	/* any section; NEVER on a controller= line */
+	LIBNVMF_KEY_IDENTITY,	/* [Host] only: hostnqn, hostid, hostsymname */
+	LIBNVMF_KEY_NQN,	/* endpoint sections only: nqn */
+	LIBNVMF_KEY_CONTROLLER,	/* endpoint sections only: ctrl (repeatable) */
+};
```
igaw, line 66: As mentioned in the c file, I think linking the `keys` definition to this enum would make sense.

igaw:
> the commit message for the second patch is also difficult understand. It mentions 'bag' without really defining what it is and the the code it doesn't really show up, only in a comment. I've gave chatgpt the commit message to transform into more technical tone:

Ah, the term "bag" is from compiler terminology. It's been a while. I suppose it can stay then, I just forgot about compiler building, something other people do :)

igaw:
I let chatgpt translate the commit message for `libnvme/config: add an INI-format NVMe-oF config parse`

```
libnvme/config: add an INI-format NVMe-oF config parser

Store sections and entries from a configuration file without applying
precedence or inheritance rules. The cascade resolver consumes these
records later, allowing parsing and policy decisions to remain separate.

Global and per-type default sections are merged with a warning when
duplicated. Host sections are singletons, while endpoint sections may be
repeated. Controller entries keep the raw addressing strings and separate
per-path tunable overrides. Unknown keys and security options on path
entries are rejected.

Syntax and structural errors fail parsing with a file:line diagnostic.
Non-fatal policy issues produce warnings and allow parsing to continue.

Keep addressing values as raw strings instead of converting them to TIDs.
Configuration files are a user-facing interface and may contain hostnames
in traddr or host-traddr fields. TID creation has stricter requirements and
belongs to the consumer that interprets the parsed configuration.
```

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +90,65 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+ * The raw model of ONE parsed file -- sections and lines faithfully
+ * recorded, nothing resolved.  The cascade resolver (which merges the
+ * top-level file with the drop-ins and expands endpoints into connections)
+ * consumes a list of these.  Struct members are open because this is an
+ * internal type; nothing here crosses the public API boundary.
+ */
+
+struct libnvme_global_ctx;
+
+struct libnvmf_conf_path {
```
igaw, line 126: There is `struct libnvmf_conf_path` and `struct libnvmf_conf_file`. Is the `struct libnvmf_conf_path` the representation for the files in `*.conf.d`? Anyway a short comment on both structures would be greatly appreciated. 

(I stopped to reading too early, after reading the members things are clear, though see my reorder request on libnvmf_conf_file).

igaw: ah this represents the connection/path ('LIBNVMF_KEY_CONTROLLER').

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +90,65 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+struct libnvmf_conf_endpoint {
+	struct list_node entry;
+	bool is_dc;
+	char *nqn;			/* NULL on a DC = well-known NQN */
+	struct libnvmf_params *params;	/* section keys incl. security */
+	struct list_head paths;		/* struct libnvmf_conf_path */
+	unsigned int line;
+};
+
+struct libnvmf_conf_file {
```
igaw: So `file` is used for the DC, IOC, HOST, ... section. I think I wouldn't have such a confusion if the order these structs would be in the order they are also in the file (example). i think it would be possible to move `struct libnvmf_conf_file` up. Also after looking at the variables in the struct it more clear what these struct do.

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +90,65 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+	char *path;			/* provenance, diagnostics */
+	/* [Discovery Controller Defaults] */
+	struct libnvmf_params *dc_defaults;
+	struct libnvmf_params *ioc_defaults; /* [I/O Controller Defaults] */
```
igaw: the doc on dc_defaults and ioc_defaults can be dropped. the variable name contains the exact same information.

libnvme/src/nvme/config-ini.h
```diff
@@ -88,3 +90,65 @@ int libnvmf_key_check_value(const struct libnvmf_key *key, const char *value);
...
+ * Parse one file into its raw model.  Structural errors (CONFIG.md Tier 1:
+ * junk lines, a second [Host], a known key in the wrong section, a bad value,
+ * a broken controller= line, ...) fail the file as a unit: NULL is returned
+ * and *err is set, with a file:line diagnostic through @ctx.  Hygiene issues
+ * (Tier 2: unknown keys, unknown sections, a repeated singleton) warn through
+ * @ctx and parsing continues.
+ */
+struct libnvmf_conf_file *libnvmf_conf_file_parse(
+		struct libnvme_global_ctx *ctx,
+		const char *path, int *err);
```
igaw:
```c
/*
 * Parse one file into its raw configuration model.
 *
 * Structural errors, such as keys in invalid sections, invalid values, or
 * malformed controller= lines, cause the entire file to fail parsing.
 *
 * On success, the output file pointer is initialized.
 *
 * Returns:
 *   0 on success;
 *   a negative error code on failure.
 */
int libnvmf_conf_file_parse(
		struct libnvme_global_ctx *ctx,
		const char *path, struct libnvmf_conf_file **file);
```

libnvme/test/config-ini.c
```diff
@@ -236,14 +244,296 @@ static bool test_value_check(void)
...
+		"\n"
+		"[Subsystem]\n"
+		"nqn           = nqn.2014-08.org.nvmexpress:vol1\n"
+		"tls           = true\n"
+		"ctrl-loss-tmo = 1800\n"
+		"controller    = transport=tcp;traddr=10.0.0.9;trsvcid=4420;host-iface=eth0\n"
+		"controller    = transport=tcp;traddr=10.0.0.10;trsvcid=4420;keep-alive-tmo=10\n";
+	struct libnvmf_conf_file *f;
+	struct libnvmf_conf_endpoint *dc, *ss;
+	struct libnvmf_conf_path *p, *p2;
```
igaw: `p1` instead `p`?

igaw:
```
libnvme/config: resolve the cascade into the connection list

Resolve the configuration hierarchy into the flat connection list consumed
by users of the API. Load the main configuration file and its sorted
drop-ins, then apply the precedence rules to produce the final set of
(role, addressing, identity, params) connections.

Keep default parameter scopes attached to resolved discovery controllers.
This ensures controllers discovered through a DC (such as referrals or DLP
entries) inherit the correct DC and IOC defaults. mDNS-discovered DCs use the
top-level scope.

Enforce relational validation rules during resolution. Reject drop-in
personas without a name, duplicate host IDs assigned to different personas,
and host NQNs associated with multiple host IDs.

Keep addressing and identity fields as raw strings. traddr may still contain
a hostname at this stage because parsing does not perform resolution. TID
construction happens later in the consumer after address resolution.
```

libnvme/src/nvme/config-ini.h
```diff
@@ -152,3 +152,60 @@ struct libnvmf_conf_file *libnvmf_conf_file_parse(
 		struct libnvme_global_ctx *ctx,
 		const char *path, int *err);
 void libnvmf_conf_file_free(struct libnvmf_conf_file *f);
+
+/*
+ * The resolved configuration: the flat list of connections the files add up
+ * to, every cascade applied.  This is what CONFIG.md means by "the grouping
+ * evaporates" -- consumers see (role, addressing, identity, params), never
+ * files or sections.
```
igaw: I would not mention the CONFIG.md reference. The comment should be written in this way it is clear within this context.

libnvme/src/nvme/config-ini.h
```diff
@@ -152,3 +152,60 @@ struct libnvmf_conf_file *libnvmf_conf_file_parse(
...
+	struct libnvmf_conf_conn *next;
+	bool is_dc;
+	/*
+	 * Complete addressing + identity, as raw strings rather than a TID:
+	 * traddr/host_traddr may still be a hostname here (the parser never
+	 * resolves), so building a TID -- which rejects one -- is the
+	 * consumer's job, after it resolves.  subsysnqn is filled from the
+	 * endpoint (or the well-known discovery NQN); hostnqn/hostid from the
+	 * file's persona.  An identity the persona leaves to the system
+	 * default is NULL; the connect path applies its usual fallback.
```
igaw:
```
/*
 * Complete addressing and identity information.
 *
 * Address and identity fields are stored as raw strings. TID construction is
 * deferred to the consumer.
 */
```

libnvme/src/nvme/config-ini.h
```diff
@@ -152,3 +152,60 @@ struct libnvmf_conf_file *libnvmf_conf_file_parse(
...
+	 */
+	char *transport;
+	char *traddr;
+	char *trsvcid;
+	char *host_traddr;
+	char *host_iface;
+	char *subsysnqn;
+	char *hostnqn;
+	char *hostid;
+	struct libnvmf_params *params; /* fully merged; "" entries mean reset */
```
igaw: such a random comment. just remove it.

libnvme/src/nvme/config-ini.h
```diff
@@ -152,3 +152,60 @@ struct libnvmf_conf_file *libnvmf_conf_file_parse(
...
+	char *hostsymname;	       /* persona extra; not connect material */
```
igaw: Wouldn't it make sense to reference Discovery Information Entry (DIE) and this is the symbolic name? The comment itself is not helping.

libnvme/src/nvme/config-ini.h
```diff
@@ -152,3 +152,60 @@ struct libnvmf_conf_file *libnvmf_conf_file_parse(
...
+	struct libnvmf_params *top_ioc_params;
+};
+
+/*
+ * Load and resolve @path plus the .conf drop-ins under <path>.d.  A missing
+ * main file or drop-in directory is not an error (an absent configuration is
+ * empty); any Tier 1 problem in any file fails the load as a unit.
+ */
+struct libnvmf_conf *libnvmf_conf_load(struct libnvme_global_ctx *ctx,
+		const char *path, int *err);
```
igaw:
```c
/*
 * Load and resolve the configuration file at @path and any drop-in files
 * under <path>.d.
 *
 * A missing configuration file or drop-in directory is not an error. In that
 * case, an empty configuration is returned.
 *
 * If any file contains a fatal validation or parsing error, loading fails and
 * no configuration is returned.
 *
 * On success, @conf is initialized.
 *
 * Returns:
 *   0 on success;
 *   a negative error code on failure.
 */
int libnvmf_conf_load(struct libnvme_global_ctx *ctx,
		      const char *path,
		      struct libnvmf_conf **conf);
```

libnvme/src/nvme/config-resolve.c
```diff
@@ -0,0 +1,410 @@
...
+ *   [Global] (top-level, overlaid by the file's own)
+ *   < per-type defaults (top-level, overlaid by the file's own)
+ *   < [Host]
+ *   < endpoint section
+ *   < controller= line overrides
+ *
+ * Identity is per file: a drop-in's [Host] must name its hostnqn (a persona
+ * without a name would silently conflate with the default persona); the
+ * relational Tier 1 rules reject a hostid shared by two personas and one
+ * hostnqn appearing with two hostids.
```
igaw:
```c
/*
 * Resolve the raw configuration models of the main file and its drop-ins
 * into the final connection list.
 *
 * Configuration is applied from least to most specific:
 *
 *   [Global]
 *   < per-type defaults
 *   < [Host]
 *   < endpoint section
 *   < controller= line overrides
 *
 * The resolver merges inherited parameters, validates the configuration, and
 * produces the flat list of resolved connections.
 */
```

libnvme/src/nvme/config-resolve.c
igaw: I thought I commented on this before. Please be a bit more careful, I noticed in other PRs the same problem, that you said you updated it and I didn't double check it and it was still the original version.

libnvme/src/nvme/config.h
```diff
@@ -0,0 +1,296 @@
...
+ * under <file>.d/.  At read time the files resolve into a flat list of
+ * connections -- each one a complete (role, addressing, identity,
+ * parameters) tuple, every default and override already applied.  Files,
+ * sections and drop-ins are authoring conveniences; they never cross this
+ * API boundary.
+ *
+ * Routine operation only reads the configuration; there are no setters.
+ * Connection configuration support is only available when fabrics support
+ * is enabled.
+ */
```
igaw, line 34:
```
/**
 * DOC: config.h
 *
 * NVMe-oF connection configuration.
 *
 * The configuration consists of a main INI file and optional .conf drop-ins
 * under <file>.d/. At load time, these inputs are resolved into a flat list
 * of connections. Each connection contains the complete role, addressing,
 * identity, and parameter set after all defaults and overrides have been
 * applied.
 *
 * Files, sections, and drop-ins are internal representation details and are
 * not exposed through this API.
 *
 * The configuration API is read-only. It provides access to resolved
 * connection data but does not provide setters.
 *
 * Connection configuration support is available only when NVMe-oF fabrics
 * support is enabled.
 */
```

libnvme/src/nvme/config.h
```diff
@@ -0,0 +1,296 @@
...
+ *                                   draws the top-level IOC scope
+ *
+ * Return: borrowed parameter set, valid while @config is alive.  @via_dc must
+ * be a connection of @config; an I/O-controller @via_dc yields NULL.  Passing
+ * a connection that does not belong to @config is undefined.
+ */
+const struct libnvmf_params *libnvmf_config_resolve_discovered(
+		const struct libnvmf_config *config,
+		const struct libnvmf_config_conn *via_dc,
+		bool is_dc);
```
igaw, line 275: should this function be able to return any error codes?

after looking at it, probably not necessary.

igaw:
```
libnvme/config: publish the resolved configuration API

Expose the resolved configuration through a public API. The API provides
configuration load, validation, and destruction, iteration over the resolved
connection list, per-connection accessors, and lookup of the default
parameters for controllers discovered at runtime (for example through DLP,
AEN, or mDNS).

Keep the parser and configuration hierarchy internal. Consumers see only the
resolved connection list, where each entry provides its role, addressing,
identity, and connection parameters.

Expose addressing and identity through per-field accessors rather than as a
TID. Address fields may still require name resolution, so TID construction is
left to the caller.

Publish the read-only connection parameter API because resolved connections
reference shared parameter sets instead of copying individual values.

Link the test against the installed shared library instead of the internal
test library to verify the exported symbols defined by the version script.
```

igaw:
```
libnvme/config: add the connect argument emitter

Emit an NVMe connect argument list from a resolved (TID, parameters) pair.
The resolved configuration model provides the information needed to generate
the arguments for a single "nvme connect" invocation, which can be used by
consumers creating transient connections or displaying reproducible commands.

Emit addressing and identity fields first, followed by connection parameters.
Parameters with unset or reset values are omitted so the kernel defaults
remain in effect. Boolean parameters are emitted as flags.

Accept a TID and parameter set instead of a resolved connection object. This
allows the emitter to be used for both configured connections and
runtime-discovered controllers, where the caller constructs the TID and
obtains the parameters separately.
```

libnvme/src/nvme/config-resolve.c
martin-belanger: Ah yes! In the pervious pass I only converted `libnvmf_conf_load()`. I forgot that `libnvmf_conf_file_parse()` had the same signature. I will fix that one as well.

libnvme/design/CONFIG.md
martin-belanger: Done

martin-belanger:
@igaw -- Think I've addressed everything from this round, including the `libnvmf_conf_file_parse()` miss — I'd only converted `libnvmf_config_load()` last time and missed its sibling; it's now on the same `int fn(..., struct foo **out)` convention, both call sites and the test updated. Also reworded the commit messages for the INI parser, cascade resolver, publish, and emitter commits to your suggested text, and renamed the "parameter bag" to "parameter store" (fair point that "bag" isn't common terminology here).

Pushed. Let me know if I missed anything.

libnvme/design/CONFIG.md
martin-belanger: Done

libnvme/design/CONFIG.md
martin-belanger: Done

libnvme/design/CONFIG.md
martin-belanger: Done

libnvme/design/CONFIG.md
martin-belanger: Done. I used ChatGPT and it found a few things that could be clarified.

libnvme/src/nvme/ini.h
martin-belanger: Done

libnvme/src/nvme/ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/test/config-ini.c
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-ini.h
martin-belanger: Done

libnvme/src/nvme/config-resolve.c
martin-belanger: Done

libnvme/src/nvme/config.h
martin-belanger: Done

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,364 @@
...
+
+Separate host identities let different credential sets be configured for different NVMe-oF deployments: a credential configured for one host identity cannot be used by another, and credentials for a single deployment can be updated or revoked by editing only that identity's configuration.
+
+This matters most for hosts that connect to multiple independently administered or multi-tenant NVMe-oF deployments, each with its own access control policy and credentials.
+
+The identity rules that make personalities valid (why each needs a distinct Host NQN *and* Host Identifier) are in [Host identity and multiple personalities](#host-identity-and-multiple-personalities); the file mechanics follow next.
+
+## File layout
+
+A configuration is a **main file plus an optional drop-in directory**, the same `foo.conf` + `foo.conf.d/` pattern systemd uses. The default main file is `/etc/nvme/nvme-fabrics.conf`, and its drop-in directory is the main file's name with `.d` appended: `/etc/nvme/nvme-fabrics.conf.d/`.
```
igaw, line 35: One thing we might want to do, is to have some more naming rules, e.g. `fabrics-*.conf`. This would avoid clashes with potentially new conf files in the future, e.g. nvme-discoverd.conf or so.

For the time, I'm fine with this version and we can change this later in the tree.

libnvme/design/CONFIG.md
```diff
@@ -0,0 +1,364 @@
...
+
+## Top-level config file (`nvme-fabrics.conf`)
+
+`[Discovery Controller Defaults]` and `[I/O Controller Defaults]` hold the default connection parameters (`ctrl-loss-tmo`, `reconnect-delay`, `keep-alive-tmo`, digests, queue counts, …) for connections of that *type*, unless something more specific overrides them (see [Precedence](#precedence-and-the-merge-model)). There is deliberately no third, type-agnostic defaults section: some parameters do want a different default per controller class — most clearly the keep-alive timeout: `keep-alive-tmo = 30` for a Discovery Controller, `keep-alive-tmo = 5` (the kernel default) for an I/O Controller — and a value that happens to apply to both is simply written in both sections. That's a deliberate choice, not an oversight: checking one `*Defaults` section shows the *whole* effective default set for that class, rather than requiring a reader to also mentally merge in a separate global scope; the occasional duplicated line is a small price for that directness.
```
igaw, line 68: and we can always introduce this later if needed anyway.

igaw:
Looks good. I've decided to add a doc update patch on top of it. I hope this is okay for you. I really want to avoid over documenting the code and reduce the complexity in the documentation language. It takes mental capacity away when reading the code. 

Anyway, thanks a lot for this huge PR! Excellent work.

PR #3568, opened 2026-07-09
nvme: remove dead initialization of data_written in get_telemetry_log()

LJaved-IBM:
The get_telemetry_log() function declares @data_written with an
initial value of 0. Before @data_written is read, it is always
assigned again inside the write loop. The initial assignment is
never read and has no effect.

Remove the dead initialization to clean up the code.


igaw:
Thanks!

PR #3569, opened 2026-07-09
nvme-models: Add cross-platform support for reading PCI IDs and use in nvme-models

brgoing:
Adds new, cross-platform support for reading PCI IDs and refactors nvme-models to use the new functionality. This allows nvme_product_name to work on Windows if a PCI IDs file is specified in the PCI_IDS_PATH environment variable.  The cross-platform PCI ID utility will also be useful for plugins that currently implement this functionality internally for Linux only.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -543,6 +543,7 @@ if want_nvme
         'logging.c',
         'nvme-cmds.c',
         'nvme-models.c',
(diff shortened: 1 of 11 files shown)


igaw:
Nice! Thanks!

PR #3572, opened 2026-07-10
test: replace rewind() with fseek() to enable error detection

bellis-ibm:
rewind() has a void return type and signals failure only through errno,
which subsequent stdio calls can overwrite before it is read. Replace
both rewind() calls with fseek(..., SEEK_SET) which returns non-zero on
failure and can be checked directly.


igaw:
Thanks!

Issue #3474, opened 2026-06-23
`nvme disconnect` should take the same arguments as `nvme connect`


dwsuse:
By extending `nvme disconnect` accept the same arguments as `nvme connect`, it's possible to write symatrical command lines. Currently it is necessary to provide the device created by the connect step to disconnect, which makes scripting etc unnecessary complex

PR #3573, opened 2026-07-10
cli/libnvme: add --idempotent and --devid-file to nvme connect

martin-belanger:
Prep patch for nvme-discoverd (not yet upstreamed). discoverd manages each fabrics connection as a systemd transient unit whose `ExecStart=` is `nvme connect`:

- `--idempotent` exits 0 when the target is already connected, instead of 1 — otherwise a harmless race with an existing connection fails the unit.
- `--devid-file FILE` names a file `libnvmf_connect()` writes the kernel-assigned device name (e.g. `nvme0`) to on success, so the unit's `ExecStop=` can find it again to disconnect.

Three commits:

1. **libnvme/fabrics**: write the connected device name to a file on request. 
2. **cli/fabrics**: add `--idempotent` and `--devid-file` to `nvme connect` — the CLI options themselves.
3. **meson**: stop deriving `RUNDIR` from `--prefix` — found while testing: `RUNDIR` was `join_paths(prefix, get_option('rundir'))`, so any build silently got `<prefix>/run` instead of the real `/run` tmpfs. `/run` isn't an install path, so it's now hardcoded, the same way systemd does it.

### DIFF
--- a/Documentation/nvme-connect.txt
+++ b/Documentation/nvme-connect.txt
@@ -9,6 +9,8 @@ SYNOPSIS
 --------
 [verse]
 'nvme' [<global-options>] 'connect' [--config=<filename>]
+			[--devid-file=<filename>]
+			[--idempotent]
 			[--owner=<NAME>]
 			[<fabrics-options>]
 
@@ -22,6 +24,22 @@ OPTIONS
 -------
 --config=<filename>::
 	Use the specified JSON configuration file instead of the default.
+
+--devid-file=<filename>::
+	Write the kernel-assigned device name (for example, nvme0) to
+	<filename> once the controller is available.  If the controller
+	is already connected, the existing device name is written
+	instead.  When used with --idempotent, the existing device name
+	is written before the command exits successfully.  This allows a
+	supervising process (for example, a systemd unit) to determine
+	which device to disconnect later.  The output file is created
+	before attempting the connection.  If it cannot be created (for
+	example, because the parent directory does not exist), the
+	command fails without attempting the connection.
+
+--idempotent::
+	Exit with status 0 if the requested controller is already
+	connected instead of reporting an error.
 
 --owner=<NAME>::
 	Record NAME as the owner of the connected controller in the NVMe
--- a/fabrics.c
+++ b/fabrics.c
@@ -209,6 +209,7 @@ struct hook_fabrics_data {
 	char *raw;
 	char **argv;
 	FILE *f;
+	bool idempotent; /* nvme connect only, false when unused */
 };
 
 static bool hook_decide_retry(struct libnvmf_context *fctx, int err,
@@ -256,8 +257,23 @@ static void hook_already_connected(struct libnvmf_context *fctx,
 		const char *transport, const char *traddr,
 		const char *trsvcid, void *user_data)
 {
+	struct hook_fabrics_data *hfd = user_data;
+
 	if (quiet)
 		return;
+
+	/*
+	 * Under --idempotent this is the success path (libnvmf_connect()
+	 * already wrote the devid file, if requested) -- downgrade from an
+	 * error to an informational note.
+	 */
+	if (hfd->idempotent) {
+		nvme_show_verbose_info(
+			"already connected to hostnqn=%s,nqn=%s,transport=%s,traddr=%s,trsvcid=%s",
+			libnvme_host_get_hostnqn(host), subsysnqn,
+			transport, traddr, trsvcid);
+		return;
+	}
 
 	nvme_show_error("already connected to hostnqn=%s,nqn=%s,transport=%s,traddr=%s,trsvcid=%s",
 		libnvme_host_get_hostnqn(host), subsysnqn,
@@ -785,6 +801,8 @@ int fabrics_connect(const char *desc, int argc, char **argv)
 	__cleanup_free char *hid = NULL;
 	char *config_file = NULL;
 	char *owner = NULL;
+	char *devid_file = NULL;
+	bool idempotent = false;
 	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
 	__cleanup_nvmf_context struct libnvmf_context *fctx = NULL;
 	__cleanup_nvme_ctrl libnvme_ctrl_t c = NULL;
@@ -795,6 +813,10 @@ int fabrics_connect(const char *desc, int argc, char **argv)
 	NVMF_ARGS(opts, fa,
 		  OPT_STRING("config",             'J', "FILE", &config_file, nvmf_config_file),
 		  OPT_STRING("owner",                0, "NAME", &owner,           "record this owner in the registry"),
+		  OPT_STRING("devid-file", 0, "FILE", &devid_file,
+			     "write connected device name to FILE"),
+		  OPT_FLAG("idempotent", 0, &idempotent,
+			   "exit 0 if already connected"),
 		  OPT_FLAG("dump-config",          'O', &dump_config,             "Dump JSON configuration to stdout"));
 
 	nvmf_default_args(&fa);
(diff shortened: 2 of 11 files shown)


meson.build
```diff
@@ -102,7 +102,13 @@ udevrulesdir   = join_paths(prefixdir, get_option('udevrulesdir'))
 dracutrulesdir = join_paths(prefixdir, get_option('dracutrulesdir'))
 systemddir     = join_paths(prefixdir, get_option('systemddir'))
 nmdispatchdir  = join_paths(prefixdir, get_option('nmdispatchdir'))
-rundir         = join_paths(prefixdir, get_option('rundir'))
+
+# /run is a kernel-mounted tmpfs whose location the FHS fixes system-wide; it
```
tbzatek: I believe this is not a kernel-mounted tmpfs, it is a regular tmpfs often created by an init system or any of its service. I.e. I didn't have `/run` on my system for a long time, everything was in `/var/run`. Not all distributions do have `/run` available, I remember adding fstab line explictly for this.

martin-belanger: I thought `/var/run` had been deprecated in favor of `/run` because `/var` was not always available during early boot. However, if there are still distros that haven't made the switch to `/run`, we should at least have an option to configure the runtime data dir (with the default equal to `/run`).

I will add the option and update the PR.

martin-belanger: Fixed.

igaw:
Can take a quick timeout and figure out if we really need `--devid-file`? As far I understand it is used to handle the `nvme disconnect` case in a systemd service, right? 

Couldn't we first try to make `nvme disconnect` to accept the same arguments as `nvme connect` and use this info to lookup the correct device? If this doesn't work we can go down this route if needed.

https://github.com/linux-nvme/nvme-cli/issues/3474

igaw:
On `--idempotent`, I haven't really thought thru what we currently do and why. Maybe we should make the idempotent case default? On the other hand tools like 'mkdir' do:

```shell
$ mkdir /tmp/test
$ mkdir /tmp/test
mkdir: cannot create directory ‘/tmp/test’: File exists
$ mkdir -p /tmp/test
``` 

Hmm, with this in mind `--idempotent` seems the right way.

martin-belanger:
I looked into this before adding `--devid-file`, and I don't think it fully replaces it, for two reasons:

1. Duplicate connections are allowed (`--duplicate-connect`/`-D`), and the matcher we'd use to find "the" controller for a given set of connect args (`libnvme_ctrl_find()`) just returns the first match it finds — no ambiguity detection. With two identical connections up, disconnecting by args would silently tear down *a* matching controller, not necessarily the one the caller meant.

2. `traddr`/`host_traddr` get sanitized before connecting, not just resolved. A hostname goes through `getaddrinfo()` (`nvmf_resolve_addr()`); a numeric address still gets canonicalized (`canon_ip()` in `tid.c`, an `inet_pton()`/`inet_ntop()` round-trip) — IPv6 always ends up in its compact form, and IPv4 loses any zero-padding (`010.000.000.001` becomes `10.0.0.1`). So even a numeric address that never touches DNS isn't necessarily what ends up in sysfs byte-for-byte. `disconnect` would need to reproduce both steps to match reliably, and if either drifts (DNS answers differently, or a script has the address in a non-canonical form), it'd either match nothing or match the wrong controller.

`--devid-file` sidesteps both: it's the literal device name from the actual connect call, so there's no ambiguity and nothing to re-resolve at disconnect time.

I still think #3474 is worth doing for the common case (no duplicates, static addressing) — it just doesn't cover the general case, so I'd rather keep `--devid-file` for that and treat #3474 as a separate, later improvement rather than a replacement.

igaw:
Okay, that makes sense. Though I am trying to figure out what the semantics are when the user calls with `--idempotent` and `--devid-file` together. In this case, wouldn't we 'steal' the device?

btw, the rational from your explanation should be in the commit message.

Documentation/nvme-connect.txt
```diff
@@ -9,6 +9,8 @@ SYNOPSIS
 --------
 [verse]
 'nvme' [<global-options>] 'connect' [--config=<filename>]
+			[--devid-file=<filename>]
+			[--idempotent]
```
igaw, line 13: the devid-file bits in the commit  should be either a separate patch or merged into the previous one

fabrics.c
```diff
@@ -256,9 +257,24 @@ static void hook_already_connected(struct libnvmf_context *fctx,
...
+	struct hook_fabrics_data *hfd = user_data;
+
 	if (quiet)
 		return;
 
+	/*
+	 * Under --idempotent this is the success path (libnvmf_connect()
+	 * already wrote the devid file, if requested) -- downgrade from an
+	 * error to an informational note.
+	 */
```
igaw, line 269: over documented. that's what the code tells me already.

libnvme/src/nvme/fabrics.c
```diff
@@ -3669,12 +3742,23 @@ __libnvme_public int libnvmf_connect(
...
 	if (err) {
+		/*
+		 * A genuine kernel-level race: some other connect beat us to
+		 * it between our topology scan above and this ioctl. @c is
+		 * only our own unconnected draft, not the winning
+		 * controller, so rescan and re-resolve to find it.
+		 */
+		if (err == -ENVME_CONNECT_ALREADY &&
+		    libnvme_scan_topology(ctx, NULL, NULL) == 0)
+			write_devid_file(fctx, devid_fd, lookup_ctrl(h, fctx));
```
igaw, line 3753: But doesn't mean we have now two owners for the same device?

libnvme/src/nvme/fabrics.h
```diff
@@ -420,6 +420,33 @@ int libnvmf_context_set_crypto(struct libnvmf_context *fctx,
...
+ * connection is live. The containing directory must already exist. The file
+ * is opened with O_NOFOLLOW (a symlink at the final path component is not
+ * followed) and is truncated only when a device name is actually written,
+ * so a name recorded by an earlier successful connect survives a failed
+ * attempt.
+ *
+ * Return: 0 on success, negative error code otherwise.
+ */
+int libnvmf_context_set_devid_file(struct libnvmf_context *fctx,
+		const char *devid_file);
```
igaw, line 448: Shouldn't we just create the files in /run/nvme* ourself and let the user provide just an id? Ideally tracked via the owner if used.

In this current form it's more or less just a 'nvme connect -v | grep device > /tmp/devid-file'

libnvme/src/nvme/fabrics.c
martin-belanger: Right — this is by design and already spec'd in `rfc-nvme-discoverd.md` §9: `--idempotent --owner=<name>` is how discoverd adopts a controller it didn't create (NBFT, FC kickstart before switch_root). No arbitration in nvme-cli itself — discoverd checks the registry owner before claiming.

One thing that spec calls for that this branch doesn't do yet: the registry entry should be written on the already-connected path too, not just on a fresh connect. Right now only the devid-file write is wired into `libnvmf_connect()` — the registry update still only fires on a fresh connect. This will be added in a separate PR.

libnvme/src/nvme/fabrics.h
martin-belanger: I considered using what you suggested:

```
nvme connect -v | grep device > /tmp/devid-file
```

My only concern was how stable the stdout is. We need the stdout to print exactly `nvmeX`, or at least   that it can be parsed so that we can retrieve the "nvmeX" easily. 

Let me spend more time on this approach. That would eliminate the need for `--devid-file`. We still need `--idempotent` though.

igaw: My point is that I want to avoid feature creap. If the devid thing really helps that's fine.

BTW, just saw your email on systemd mailing list (took me a while to see it, I know...) And from this

```
3. A behavior that surprised us at first: systemd does not expand specifiers
   (%t, %N, %n) in the Exec* arguments of a transient unit created through
   the StartTransientUnit D-Bus API -- the arguments reach the program
   verbatim. For example a path passed as

       %t/nvme/discoverd/units/%N.devid

   arrives literally instead of becoming

       /run/nvme/discoverd/units/<unit-name>.devid
```

So I see how this devid argument would help, but I still wonder if a simple `id` instead of a full file path would be better.

about: `nvme connect --output-format json` would be the better

martin-belanger:
Closing this. It will be easier to submit a clean PR with `--idempotent` in its own commit, and `--devid-file` in a separate commit.

PR #3574, opened 2026-07-10
netapp: fix dead store in netapp_smdevices_get_info() and netapp_ontapdevices_get_info()

LJaved-IBM:
Both netapp_smdevices_get_info() and netapp_ontapdevices_get_info()
call libnvme_get_nsid() and store the result in @err, but never
check @err before proceeding. The return value is immediately
overwritten by the next assignment, making it a dead store.

If libnvme_get_nsid() fails, the functions silently continue
with an uninitialized @item->nsid, leading to undefined behavior.

Check the return value of libnvme_get_nsid() in both functions,
log the error with nvme_show_error() and return 0 on failure to
match the existing error handling pattern in those functions.


igaw:
Thanks!

PR #3576, opened 2026-07-13
ioctl-linux: fix uninitialized cmd32.result in ioctl_passthru32()

sahmed-ibm:
When hdl->ctx->dry_run is true, the goto skips the memcpy and explicit assignment of cmd32.result, leaving it uninitialized when assigned to cmd->result at the out label.

Fix by zero-initializing cmd32 at declaration to ensure deterministic behavior on all code paths.


igaw:
Thanks!

PR #3578, opened 2026-07-14
micron: Use new nvme-pci-ids to get PCI IDs

brgoing:
Removes PCI ID reading utility and related methods from micron-utils and updates the micron-nvme plugin to use nvme-pci-ids to read the IDs.


igaw:
Thanks!

PR #3579, opened 2026-07-14

LJaved-IBM:
@sph irrelevant and leaving a potentially misleading non-NULL
value before the call.

The Clang static analyzer flags this as a null pointer dereference
since @sph may be NULL when dereferenced after the function call.

Initialize @sph to NULL before the call to make the intent clear
and remove the now-redundant null check in wdc_get_dev_mng_log_entry().


igaw:
Could you rebase this change, there is merge conflict. Thanks!

PR #3582, opened 2026-07-15
wdc-nvme: fix dead assignment in wdc_get_pci_ids()

sahmed-ibm:
After successfully reading the vendor sysfs file, ret is assigned 0, but this value is never read. The variable is
unconditionally overwritten by the subsequent read() call for the device ID file before it is used again.

Remove the dead assignment. No functional change.

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -40,6 +40,7 @@
 #include "plugin.h"
 #include "util/cleanup.h"
 #include "util/types.h"
+#include "nvme-pci-ids.h"
 
 #define CREATE_CMD
 #include "wdc-nvme.h"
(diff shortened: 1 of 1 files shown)


igaw:
What about replacing the `wdc_get_pci_ids` implementation with the newly added:

```c
int nvme_get_pci_ids(struct libnvme_global_ctx *ctx,
		struct libnvme_transport_handle *hdl,
		__u32 *vid, __u32 *did,
		__u32 *subsys_vid, __u32 *subsys_did,
		__u32 *class_code);
```

more common code and works also on windows...

sahmed-ibm:
That should work I believe. I'll work on making this change

igaw:
I've drop the first patch, as the change was removed afterwards anyway.

Thanks!

PR #3583, opened 2026-07-15
NBFT API overhaul

tbzatek:
This is a sync of the Boot Specification rev. 1.3 into `nvme-types-nbft.h` and the `nbft.h` parser API.
- ditched the parsed bools in favour of untranslated flags bitfields - a futureproof approach since `nbft.h` don't include any reserved fields
- updated the Python interface file to follow the flags instead of bools
- in contrast the `nvme nbft show` JSON output preserves all the original bool flags to maintain compability with dracut
- the DHCP DUID fields are tough to swallow as these are composite and variable values, I still wanted to provide human readable representation in the JSON output; the parser API provides original bytes though
- added Security Descriptor parser API without actual parser implementation, as agreed on the last Timberland SIG call - added only to avoid breaking stable API in the future, follows actual rev. 1.3. spec pretty closely

Assisted by Claude, it's surprisingly easy to let it craft synthetic binary NBFT table for testing, if needed.

Cc: @mwilck @stuarthayes @martin-belanger @johnmeneghini @MichaelRabek


igaw:
From a quick glance it looks good. Does every patch for itself compile? I think I saw code refactoring which will break the build, e.g. `this_hfi_is_default_route`.

tbzatek:
> From a quick glance it looks good. Does every patch for itself compile? I think I saw code refactoring which will break the build, e.g. `this_hfi_is_default_route`.

The commits were divided in functional blocks for easier review, I've squashed them to avoid bisecting compile failures.

igaw:
Thanks a lot!

PR #3584, opened 2026-07-15
meson: stop deriving RUNDIR from --prefix

martin-belanger:
`RUNDIR` was `join_paths(prefix, get_option('rundir'))`, with `rundir` defaulting to the bare string `run` -- so a default build silently got `<prefix>/run` (e.g. `/usr/local/run`) instead of the real `/run` tmpfs. `/run`'s location is fixed by the FHS, not something `--prefix` should scale. `rundir` stays a configurable option (some distros only have `/var/run`), it just no longer joins with `--prefix`.


igaw:
Thanks!

PR #3586, opened 2026-07-15
fabrics: add --idempotent and --devid-file to nvme connect

martin-belanger:
nvme connect gets two new options for supervised use (systemd transient units, nvme-discoverd): `--idempotent` makes connect succeed instead of failing when the target is already connected, and `--devid-file` saves the connected device name to a file so a supervising process can find it again later, e.g. to disconnect.

Two commits, each a complete, independent feature:
- `nvme-cli/fabrics: add --idempotent to nvme connect`
- `libnvme/fabrics: add --devid-file to nvme connect`

Replaces #3573, closed in favor of this PR: same two features, split more deliberately (one commit each, rather than bundled).

### DIFF
--- a/Documentation/nvme-connect.txt
+++ b/Documentation/nvme-connect.txt
@@ -9,6 +9,7 @@ SYNOPSIS
 --------
 [verse]
 'nvme' [<global-options>] 'connect' [--config=<filename>]
+			[--idempotent]
 			[--owner=<NAME>]
 			[<fabrics-options>]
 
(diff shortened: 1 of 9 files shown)


igaw:
Updated the API documentation so it omitted implementation details such as O_NOFOLLOW, truncation behavior, and race handling. This is too much details for a public API. While at it, I also gave the man page doc a chatgpt treatment.

igaw:
Thanks!

PR #3587, opened 2026-07-15
libnvme/config: add the INI configuration write side

martin-belanger:
New code, not yet wired into anything: `libnvmf_config_emit_new/add/install()` builds and atomically installs an INI configuration tree from a flat list of (role, TID, params, persona) connections.

This exists specifically to support converting an old-format config (JSON `config.json`/`discovery.conf`) to the new INI format. Given a legacy config parsed into that flat connection list, this is the write side that turns it into a valid INI format (e.g. `nvme-fabrics.conf` + drop-ins) -- the actual JSON reader and the `nvme config convert` command that will call this are a separate, later PR.

Write-side counterpart to #3566 (merged, the read side).

### DIFF
--- a/libnvme/src/nvme/config.h
+++ b/libnvme/src/nvme/config.h
@@ -304,6 +304,119 @@ int libnvmf_connect_args_emit(const struct libnvmf_tid *tid,
 		void *user_data);
 
 /**
+ * struct libnvmf_config_emitter - Configuration emitter.
+ *
+ * Opaque type used to build and write an NVMe Fabrics configuration.
+ */
+struct libnvmf_config_emitter;
+
+/**
+ * libnvmf_config_emit_new() - Create a configuration emitter.
+ * @ctx: libnvme global context. Must not be NULL.
+ *
+ * Return: A new emitter on success, or NULL on failure.
+ * Free the emitter with libnvmf_config_emit_free().
+ */
+struct libnvmf_config_emitter *libnvmf_config_emit_new(
+		struct libnvme_global_ctx *ctx);
+
+/**
+ * libnvmf_config_emit_free() - Free a configuration emitter.
+ * @emitter: Emitter to free. NULL is ignored.
+ */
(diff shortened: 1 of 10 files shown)


libnvme/src/nvme/config.h
```diff
@@ -303,6 +303,131 @@ int libnvmf_connect_args_emit(const struct libnvmf_tid *tid,
...
+		const struct libnvmf_params *params, const char *hostsymname);
+
+/**
+ * libnvmf_config_emit_install() - write the collected configuration.
+ * @emitter: the configuration builder
+ * @file:    main configuration file, or NULL for the default
+ *           (/etc/nvme/nvme-fabrics.conf).  The drop-in directory is
+ *           always derived from the file name: <file>.d/.
+ *
+ * Refuses when a configuration already exists at the target -- the main
```
igaw: This is a policy. Maybe a `bool force` flag instead a  flat refusal?

igaw:
Looks good, except the documentation and commit messages. Could you run it through chatgpt 'technical documentation, man page style, simple language, precise language'? Thanks!

martin-belanger:
@igaw - Just pushed an update addressing both comments:

- Added a `force` parameter to `libnvmf_config_emit_install()`. When set, it overwrites an existing configuration at the target instead of returning `-EEXIST`. Covered by a new test (`test_force_overwrite`).
- Ran the documentation and commit messages through ChatGPT for a clearer, more concise pass, then reviewed the result by hand to make sure nothing load-bearing got lost in translation.

Full series re-verified after these changes: build clean, `meson test`, checkpatch 0 errors, valgrind clean on the new test, and the first commit builds/tests standalone in an isolated worktree.

libnvme/src/nvme/config.h
martin-belanger: Adding a `bool force` is a good idea.
Done.

igaw:
Thanks!

PR #3590, opened 2026-07-16
libnvme/nvme-cli: add `nvme config-convert`, the legacy config to INI migration tool

martin-belanger:
This is the first thing that actually wires up to the INI read/write API added in #3566 and #3587: a one-shot, explicit command that converts an existing `config.json` and/or `discovery.conf` into the new INI format.

It's also just the first step. The plan is for a follow-on PR to add *implicit* conversion: `connect-all`/`discover` converting legacy config to INI automatically on first invocation, so INI becomes the format libnvme actually reads, with the conversion happening transparently at the CLI layer before any libnvme call. That implicit path is separate work, not part of this PR -- this one only adds the explicit command: `nvme config-convert`.

Commits:
- `nvme-cli: add the config.json reader for INI conversion`
- `fabrics, nvme-cli: add the discovery.conf reader for INI conversion`
- `nvme-cli: add the config-convert command`
- `libnvme/design: fix stale config-convert references in CONFIG.md`
- `doc: add nvme-config-convert man page`

`nvme config-convert` converts `config.json` and/or `discovery.conf` to the INI format read by `nvme connect-all`/`nvme discover`. With no `--config`, it converts whichever of the two well-known legacy files exist, matching what a bare `connect-all` already reads today; `--config` restricts it to one JSON file. Refuses to overwrite an existing target unless `--force` is given. On success, the converted legacy file(s) are renamed to `<name>.converted` so a repeated run doesn't convert them again.

Almost all of this lives in `config-convert.c`/`config-convert.h`, deliberately. Since support for the legacy configuration files is temporary, the conversion logic is kept isolated so it can eventually be removed with minimal impact. Unfortunately, not all the code could be put in `config-convert.c`. The only exception is the `discovery.conf` argv-style parser (`nvmf_convert_discovery_line`), which stays in `fabrics.c` to reuse the same option table as nvme discover; everything else lives in `config-convert.c`.

Ran the comments and man page through ChatGPT for a "man page style, simple/precise language" pass, then reviewed the result by hand.

Depends on #3587 (merged).

### DIFF
--- a/config-convert.c
+++ b/config-convert.c
@@ -8,13 +8,16 @@
 #include <errno.h>
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 
 #include <libnvme.h>
 
 #include "common.h"
 #include "config-convert.h"
 #include "fabrics.h"
+#include "logging.h"
 #include "nvme-print.h"
+#include "util/argconfig.h"
 #include "util/cleanup.h"
 
 #ifdef CONFIG_JSONC
@@ -418,4 +421,151 @@ int nvme_config_convert_discovery(struct libnvmf_config_emitter *emitter,
 	}
 
 	return 0;
+}
+
+/* Best effort. The configuration has already been installed. */
+static void rename_to_converted(const char *path)
+{
+	__cleanup_free char *dst = NULL;
+
+	if (asprintf(&dst, "%s.converted", path) < 0)
+		return;
+
+	if (rename(path, dst))
+		nvme_show_error(
+			"converted %s but failed to rename it to %s: %s",
+			path, dst, strerror(errno));
+}
+
+/* True if @path is gone because a prior run already renamed it away. */
+static bool already_converted(const char *path)
+{
+	__cleanup_free char *converted = NULL;
+
+	if (asprintf(&converted, "%s.converted", path) < 0)
+		return false;
+
+	return !access(converted, F_OK);
+}
+
+static int install_converted(struct libnvmf_config_emitter *emitter,
+		const char *output_file, const char *json_path,
+		const char *disc_path, bool converted_json,
+		bool converted_disc, bool force)
+{
+	int ret;
+
+	ret = libnvmf_config_emit_install(emitter, output_file, force);
+	if (ret == -EEXIST) {
+		nvme_show_error("%s already exists; refusing to overwrite",
+				 output_file);
+		return ret;
+	}
+	if (ret) {
+		nvme_show_error("failed to write %s: %s", output_file,
+				 libnvme_strerror(-ret));
+		return ret;
+	}
+
+	if (converted_json)
+		rename_to_converted(json_path);
+	if (converted_disc)
+		rename_to_converted(disc_path);
+
+	return 0;
+}
+
+int nvme_config_convert(const char *desc, int argc, char **argv)
+{
+	char *config_file = NULL;
+	char *output_file = NULL;
+	const char *target;
+	const char *json_path;
+	bool verbose = false;
+	bool force = false;
+	bool converted_json = false, converted_disc = false;
+	bool json_already_done = false, disc_already_done = false;
+	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
+	struct libnvmf_config_emitter *emitter = NULL;
+	int ret;
+
+	OPT_ARGS(opts) = {
+		OPT_STRING("config", 'J', "FILE", &config_file,
+			   "convert this JSON file (default: config.json)"),
+		OPT_STRING("output", 'o', "FILE", &output_file,
+			   "write result here (default: nvme-fabrics.conf)"),
+		OPT_FLAG("force", 0, &force,
+			 "overwrite an existing target"),
+		OPT_FLAG("verbose", 'v', &verbose, "increase output verbosity"),
+		OPT_END()
+	};
+
+	ret = argconfig_parse(argc, argv, desc, opts);
+	if (ret)
+		return ret;
+
+	log_level = map_log_level(verbose ? 1 : 0, false);
+
+	ret = nvme_create_global_ctx(&ctx);
+	if (ret) {
+		nvme_show_error("Failed to create topology root: %s",
+				 libnvme_strerror(-ret));
+		return ret;
+	}
+	libnvme_set_logging_level(ctx, log_level, false, false);
+
+	emitter = libnvmf_config_emit_new(ctx);
+	if (!emitter)
+		return -ENOMEM;
+
+	json_path = config_file ? config_file : PATH_NVMF_CONFIG;
+	if (!access(json_path, F_OK)) {
+		ret = nvme_config_convert_json(emitter, json_path);
+		if (ret)
+			goto out;
+		converted_json = true;
+	} else if (already_converted(json_path)) {
+		json_already_done = true;
+	} else if (config_file) {
+		/*
+		 * An explicit --config to a file that neither exists nor was
+		 * ever converted: let the JSON parser produce its own
+		 * "failed to parse" error instead of silently no-op'ing,
+		 * since this was an explicit ask.
+		 */
+		ret = nvme_config_convert_json(emitter, json_path);
+		if (ret)
+			goto out;
+		converted_json = true;
+	}
+
+	if (!access(PATH_NVMF_DISC, F_OK)) {
+		ret = nvme_config_convert_discovery(emitter, PATH_NVMF_DISC);
+		if (ret)
+			goto out;
+		converted_disc = true;
+	} else if (already_converted(PATH_NVMF_DISC)) {
+		disc_already_done = true;
+	}
+
+	if (!converted_json && !converted_disc) {
+		if (json_already_done || disc_already_done) {
+			nvme_show_result("already converted; nothing to do");
+			ret = 0;
+			goto out;
+		}
+		nvme_show_error("nothing to convert: neither %s nor %s exists",
+				 json_path, PATH_NVMF_DISC);
+		ret = -ENOENT;
+		goto out;
+	}
+
+	target = output_file ? output_file : PATH_NVMF_INI;
+	ret = install_converted(emitter, target, json_path, PATH_NVMF_DISC,
+				 converted_json, converted_disc, force);
+
+out:
+	libnvmf_config_emit_free(emitter);
+
+	return ret;
 }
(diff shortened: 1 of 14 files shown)


config-convert.c
```diff
@@ -419,3 +422,127 @@ int nvme_config_convert_discovery(struct libnvmf_config_emitter *emitter,
...
+			converted_json = true;
+		}
+		if (!access(PATH_NVMF_DISC, F_OK)) {
+			ret = nvme_config_convert_discovery(emitter,
+							     PATH_NVMF_DISC);
+			if (ret)
+				goto out;
+			converted_disc = true;
+		}
+		if (!converted_json && !converted_disc) {
```
igaw: this check is in the else block instead after the `if/else` block.

config-convert.c
```diff
@@ -419,3 +422,127 @@ int nvme_config_convert_discovery(struct libnvmf_config_emitter *emitter,
...
+			nvme_show_error(
+				"nothing to convert: neither %s nor %s exists",
+				PATH_NVMF_CONFIG, PATH_NVMF_DISC);
+			ret = -ENOENT;
+			goto out;
+		}
+	}
```
igaw, line 561: The documentation says 'one shot' migration, but in fact we have to call it twice, once for the json config and once for discovery.conf. I think it should do a one shot operation.

igaw:
Looks good. Nice to see how simple the convert function is. Great job!

igaw:
Could we also have some end to end tests for this feature (those in the `tests` directory?) We need to get this feature as correct as possible, including the corner cases if possible.

config-convert.c
martin-belanger: Good catch. I'll have that fixed.

config-convert.c
martin-belanger: Fixed

config-convert.c
martin-belanger: fixed

martin-belanger:
Thanks for the review.

- Moved the "nothing to convert" check out of the `else` block, per your comment on config-convert.c:531. It fell out naturally from the fix below rather than needing a standalone change.
- Fixed the "one shot" issue (config-convert.c:538): discovery.conf conversion is no longer gated by `--config`. It matches `nvme connect-all`'s existing behavior now -- the well-known discovery.conf path is always converted when it exists, regardless of whether `--config` points at a non-default config.json. `--config` only ever affects which JSON file gets converted.
- Added end-to-end tests in `tests/nvme_config_convert_test.py`: DC/subsystem conversion, the legacy bare-array JSON format, dhchap-secret default inheritance and per-port override, and the error/force/rename paths.

One limitation on the test coverage worth flagging: discovery.conf's own conversion isn't exercised end-to-end. Its path has no `--config`-style override, so a test can only hit it by touching the real `/etc/nvme/discovery.conf` -- and that file exists on essentially every system, even freshly installed ones, since we ship an empty (comment-only) stub by default. Converting it end-to-end in a test would mean renaming that file on whatever machine runs the suite, which isn't something a test should do. The suite instead skips itself whenever that file is present, and I verified the logic separately against the real binary with the guard bypassed.

Related note, not for this PR: we should probably stop installing that empty discovery.conf stub by default (`meson.build` and `nvme.spec.in` both ship it unconditionally today). It's what made the test-coverage gap above universal rather than a corner case. Tracking as a separate follow-up.

igaw:
Yes, we should remove the discovery.conf install step now. This should also be documented eventually. My current plan is to start documenting this for a release note file. All the stuff which could regress should be mentioned there.

martin-belanger:
Sounds good — happy to help once you've got a structure in mind. What format are you picturing (a single `RELEASE_NOTES.md`, one entry per PR, something else)? I can contribute entries for the changes on my end (this discovery.conf removal, and whatever else from the INI/config pipeline is worth flagging) once there's a shape to fit them into.

tests/nvme_config_convert_test.py
```diff
@@ -0,0 +1,265 @@
...
+    else 'nvme'
+
+_REAL_DISCOVERY_CONF = '/etc/nvme/discovery.conf'
+
+
+@unittest.skipIf(os.path.exists(_REAL_DISCOVERY_CONF),
+                  f'{_REAL_DISCOVERY_CONF} exists on this machine; '
+                  'config-convert always converts and renames it, so '
+                  'skipping to avoid touching real system configuration')
+class ConfigConvertCLITest(unittest.TestCase):
```
igaw: Is there a reason why this is not based on the TestNVMe class? It brings some helpers such run_cmd and all the logging infra.

tests/nvme_config_convert_test.py
```diff
@@ -0,0 +1,265 @@
...
+        self.assertIn('nqn = nqn.2024-01.com.example:data.vol5', content)
+
+    def test_source_file_renamed_to_converted_on_success(self):
+        self._write_json({'hosts': []})
+        self._convert()
+        self.assertFalse(os.path.exists(self.config_json))
+        self.assertTrue(os.path.exists(self.config_json + '.converted'))
+
+
+if __name__ == '__main__':
```
igaw, line 272: I one test would also be if the config has been already be converted and the user calls 'nvme config-convert -J config.json' again. I think in this case config-convert should not error out instead doing nothing as there is already a config.json.converted,.e.g  the idempotent behavior.

tests/nvme_config_convert_test.py
```diff
@@ -0,0 +1,265 @@
...
+import tempfile
+import unittest
+
+
+# Capture the nvme binary path from argv before unittest.main() strips it.
+_NVME_BIN = sys.argv[1] \
+    if len(sys.argv) > 1 and not sys.argv[1].startswith('-') \
+    else 'nvme'
+
+_REAL_DISCOVERY_CONF = '/etc/nvme/discovery.conf'
```
igaw, line 42: 60:#define PATH_NVMF_DISC		SYSCONFDIR "/nvme/discovery.conf"
294:	hfd->f = fopen(PATH_NVMF_DISC, "r");

we could use the new --set-options here:

`nvme --set-options discovery_conf=SOMEPATH/discovery.conf convert-config ...`

tests/nvme_config_convert_test.py
martin-belanger: `TestNVMe.setUp()` unconditionally requires a real controller. None of that applies here. Subclassing it would mean overriding `setUp()` to skip essentially everything the base class does.

If `run_cmd`/the logging infra are the specific parts worth reusing, I could pull those out of `TestNVMe` into a small shared base class and use that. Let me know if that's worth doing, or if plain `subprocess.run()` (what's there now) is fine for a test this size.

tests/nvme_config_convert_test.py
martin-belanger: Good idea. Will do.

tests/nvme_config_convert_test.py
martin-belanger: `--set-options discovery_conf=...` needs the same plumbing a dedicated flag would: adding new param to the global ctx + setter/getter, etc. 

**There is a no-new-API alternative**: `unshare --user --map-root-user --mount` + a bind-mount over /etc/nvme, so the test's own discovery.conf shadows the real one ([man page](https://man7.org/linux/man-pages/man1/unshare.1.html)). No real root needed, but it depends on unprivileged user namespaces being allowed in whatever environment runs it -- not guaranteed everywhere, so it'd need a graceful skip where unavailable.

**Another option**: Since this only really needs to work in CI, and our CI containers are fully ours (ephemeral, thrown away after the job), we may not need either: the test can just back up /etc/nvme/discovery.conf, write known content, assert, and restore it afterward, gated behind an explicit CI-only env var. Locally it stays exactly as-is -- self-skips if a real discovery.conf exists. Simplest option: no new C code, no namespaces.

martin-belanger: Forget the `unshare` approach. This requires elevated privileges.

martin-belanger: Dug into where `tests/` actually runs before assuming any CI container is the right target. Turns out it doesn't run in the per-PR build matrix at all today -- `-Dnvme-tests=true` is only ever passed in `run-nightly-tests.yml`, which runs on a schedule against a privileged podman container (`ghcr.io/linux-nvme/debian:latest`), not on every PR push. `meson install` doesn't run there either, so I pulled that image (and the fedora one) directly to check: neither has nvme-cli installed, and /etc/nvme doesn't even exist. So discovery.conf genuinely isn't present in that container unless we put it there ourselves.

Given that, the env-var-gated backup/write/restore approach from above should work cleanly there: no nvme-cli package to conflict with, a privileged/root container, thrown away after each nightly run. I'd suggest we only wire this into the nightly workflow for now rather than try to get it into the per-PR matrix -- nightly-only discovery.conf coverage seems like a reasonable bar given the constraints, and it's already the same tier `registry-cli` gets today (that one doesn't run per-PR either, same infrastructure gap, unrelated to this PR).

What do you think?

igaw: Hmm, I think it's not worth the effort for discovery.conf. As far I can tell we can cover all scenarios with config.json and the only gain with discovery.conf is the parsing bits, which is already covered in libnvme as far I remember

Let's go without it I'd say.

tests/nvme_config_convert_test.py
igaw: Ah, didn't realize that. I would like to use the same infra for command logging we have currently. So yeah, best idea is to move the common code in a base class I'd say.

martin-belanger: Will do.

igaw:
> Sounds good — happy to help once you've got a structure in mind. What format are you picturing (a single `RELEASE_NOTES.md`, one entry per PR, something else)? I can contribute entries for the changes on my end (this discovery.conf removal, and whatever else from the INI/config pipeline is worth flagging) once there's a shape to fit them into.

I don't have an exact idea how it should be handled, I think when a new feature or so lands. I'd say we just copy the strategy/format from an existing project, systemd? 

What about RELEASE.md

Anyway w oould make my life simpler :)

tests/nvme_config_convert_test.py
martin-belanger: Done

tests/nvme_config_convert_test.py
martin-belanger: Done

martin-belanger:
All done. 
I will have a closer look at what systemd does when they have a new relase. From what I can tell they use a "NEWS" file.

igaw:
Okay, let's go with this and see how it goes. Thanks a lot!

PR #3591, opened 2026-07-16
libnvme/python: add config_read() and config_validate()

martin-belanger:
Adds Python bindings for the read side of the INI connection configuration API added in #3566/#3587: `config_read()` and `config_validate()`.

`config_read(ctx, file=None)` returns the resolved connection list as a list of dicts (one per connection, in file order), the same shape as `libnvmf_config_conn_for_each()`'s fields. 

`config_validate(ctx, file=None)` dry-runs the same read for diagnostics only.

Only the reader is bound -- the configuration emitter (write) API from #3587 has no Python binding here.


igaw:
Thanks!

PR #3592, opened 2026-07-17
doc: add NEWS.md

martin-belanger:
@igaw: Proposed `NEWS.md`, not a final decision -- modeled on systemd's own NEWS file per your suggestion. Seeded with a few real entries (backward-incompatible changes and new features from this cycle) to show what it'd look like in practice, not meant to be the definitive list yet.

One thing I'm not sure about: it's Markdown. Let me know if you'd rather have plain text instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,121 @@
+<!-- SPDX-License-Identifier: GPL-2.0-only -->
+# NEWS
+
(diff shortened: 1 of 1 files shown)


igaw:
Really nice, let's go with this, we improve/extend in the tree. There is quite a bit of missing entries from my API cleanup rampage...

PR #3594, opened 2026-07-20
json: pdc related fixes while using JSON config file

martin-gpy:
Few PDC related fixes while running nvme connect-all using the JSON config file.


igaw:
This code is gone now and libnvme doesn't use json for the configuration anymore.

I am closing this one as we have #3112 for tracking the PDC work.

Thanks anyway.

PR #3595, opened 2026-07-20
nvme-cli/fabrics: consume the INI config directly from the CLI

martin-belanger:
## Summary

- Add `libnvmf_context_apply_params()`, a table-driven helper that applies a resolved parameter set (from INI) to a connect context (fctx).
- Switch `connect-all`/`discover`/`connect -J` to read the INI configuration directly and iterate connections in the CLI itself, instead of going through libnvme's JSON config loops. Hostname resolution and hostnqn/hostid selection are policy decisions and now live in the CLI. The CLI enforces the exclusion list on every auto-connected entry.
- Old style config (config.json, discovery.conf) automatically converted to INI on first invocation.
- Retire the `/run/nvme` volatile-config reader and the discovery.conf line-parser hooks: nothing reads the former, and the INI config replaces the latter's only caller.

First slice of moving nvme-cli off the legacy JSON config path onto the INI config from #3566/#3587/#3590/#3591. Self-contained: `-O`/`--dump-config`, `nvme config`'s `-M`/`-R`/`-U`, and the now-unused JSON loops in libnvme are untouched here, addressed in follow-up PRs.

## Test plan

- [x] `meson test -C .build` — 86/86 passing
- [x] `make checkpatch` — 0 errors
- [x] Each commit builds standalone
- [x] Manual: `nvme connect -J <ini-file>`, `nvme discover`, `nvme connect-all` against scratch INI configs (default path, custom `-J`, legacy-JSON auto-convert), under valgrind — no leaks

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -19,7 +19,6 @@
  * Fabrics specification standard.
  */
 
-#include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
@@ -58,7 +57,6 @@
 #include "logging.h"
 #include "util/sighdl.h"
 
-#define PATH_NVMF_RUNDIR	RUNDIR "/nvme"
 #define MAX_DISC_ARGS		32
 #define MAX_DISC_RETRIES	10
 
@@ -99,7 +97,7 @@ static const char *nvmf_hdr_digest	= "enable transport protocol header digest (T
 static const char *nvmf_data_digest	= "enable transport protocol data digest (TCP transport)";
 static const char *nvmf_tls		= "enable TLS";
 static const char *nvmf_concat		= "enable secure concatenation";
-static const char *nvmf_config_file	= "Use specified JSON configuration file or 'none' to disable";
+static const char *nvmf_config_file	= "Use specified INI configuration file (auto-converts a legacy .json) or 'none' to disable";
 
 #define NVMF_ARGS(n, f, ...)                                                                  \
 	NVME_ARGS(n,                                                                              \
@@ -168,12 +166,9 @@ static int setup_common_context(struct libnvmf_context *fctx,
 		struct nvmf_args *fa);
 
 struct hook_fabrics_data {
-	struct nvmf_args *fa;
 	nvme_print_flags_t flags;
 	bool quiet;
 	char *raw;
-	char **argv;
-	FILE *f;
 	bool idempotent;
 };
 
@@ -250,33 +245,6 @@ static void hook_discovery_log(struct libnvmf_context *fctx,
 		save_discovery_log(hfd->raw, log);
 	else if (!connect)
 		nvme_show_discovery_log(log, numrec, hfd->flags);
-}
-
-static int hook_parser_init(struct libnvmf_context *dctx, void *user_data)
-{
-	struct hook_fabrics_data *hfd = user_data;
-
-	hfd->f = fopen(PATH_NVMF_DISC, "r");
-	if (hfd->f == NULL) {
-		nvme_show_error("No params given and no %s", PATH_NVMF_DISC);
-		return -ENOENT;
-	}
-
-	hfd->argv = calloc(MAX_DISC_ARGS, sizeof(char *));
-	if (!hfd->argv)
-		return -1;
-
-	hfd->argv[0] = "discover";
-
-	return 0;
-}
-
-static void hook_parser_cleanup(struct libnvmf_context *fctx, void *user_data)
-{
-	struct hook_fabrics_data *hfd = user_data;
-
-	free(hfd->argv);
-	fclose(hfd->f);
 }
 
 /*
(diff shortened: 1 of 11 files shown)


libnvme/src/nvme/config.c
```diff
@@ -244,3 +247,111 @@ __libnvme_public int libnvmf_connect_args_emit(const struct libnvmf_tid *tid,
...
+	else if (!strcmp(key, "nr-write-queues"))
+		cfg->nr_write_queues = strtol(value, NULL, 0);
+	else if (!strcmp(key, "nr-poll-queues"))
+		cfg->nr_poll_queues = strtol(value, NULL, 0);
+	else if (!strcmp(key, "queue-size"))
+		cfg->queue_size = strtol(value, NULL, 0);
+	else if (!strcmp(key, "keep-alive-tmo"))
+		/* Applied verbatim; the NQN-conditional default is a
+		 * connect-time decision, see set_discovery_kato().
+		 */
```
igaw, line 283: 'this comment doesn't make sense to me this is KATO who does the nqn play a role?

libnvme/src/nvme/config.c
```diff
@@ -244,3 +247,111 @@ __libnvme_public int libnvmf_connect_args_emit(const struct libnvmf_tid *tid,
...
+	} else if (!strcmp(key, "dhchap-secret")) {
+		state->hostkey = value;
+		state->have_crypto = true;
+	} else if (!strcmp(key, "dhchap-ctrl-secret")) {
+		state->ctrlkey = value;
+		state->have_crypto = true;
+	}
+	/* Identity/addressing keys never reach here: get_params() only
+	 * surfaces tunable and security keys.
+	 */
```
igaw, line 316: I understand the first part of the comment, the second part is what? this is apply_params?

libnvme/src/nvme/config.c
```diff
@@ -244,3 +247,111 @@ __libnvme_public int libnvmf_connect_args_emit(const struct libnvmf_tid *tid,
...
+	bool bval;
+
+	/* An explicit reset is not applied: @fctx already carries the
+	 * kernel default it was created with (libnvmf_default_config()).
+	 */
+	if (!*value)
+		return;
+
+	/* strtol(..., 0) matches check_int()'s hex/octal-aware parsing
+	 * (config-ini.c); atoi() would mis-parse "0x10" as 0.
```
igaw: no need to add man page documentation on strtol.

libnvme/src/nvme/config.c
```diff
@@ -244,3 +247,111 @@ __libnvme_public int libnvmf_connect_args_emit(const struct libnvmf_tid *tid,
...
+ * unconditionally, so they're buffered here and applied once at the
+ * end instead of clobbering each other per key.
+ */
+struct apply_state {
+	struct libnvmf_context *fctx;
+	const char *hostkey;
+	const char *ctrlkey;
+	const char *keyring;
+	const char *tls_key;
+	const char *tls_key_identity;
```
igaw: libnvmf_context has 
```c
	const char *hostkey;          // !access:write=custom
	const char *ctrlkey;          // !access:write=custom
	const char *keyring;          // !access:write=custom
	char *tls_key;                // !access:write=custom
	const char *tls_key_identity; // !access:write=custom
```

Why do we need to treat them special. The comment mentiones 'clobbering' how? There is nothing special about these keys, we can either pass them via the connect string or we load them into the kernel keystore.

libnvme/src/nvme/config.c
```diff
@@ -244,3 +247,111 @@ __libnvme_public int libnvmf_connect_args_emit(const struct libnvmf_tid *tid,
...
+static void apply_param(const char *key, const char *value, void *user_data)
+{
+	struct apply_state *state = user_data;
+	struct libnvmf_context *fctx = state->fctx;
+	struct libnvme_fabrics_config *cfg = &fctx->ctrl_params.cfg;
+	bool bval;
+
+	/* An explicit reset is not applied: @fctx already carries the
+	 * kernel default it was created with (libnvmf_default_config()).
+	 */
```
igaw, line 266: but only two keys have defaults
```c
void libnvmf_default_config(struct libnvme_fabrics_config *cfg)
{
	cfg->tos = -1;
	cfg->ctrl_loss_tmo = NVMF_DEF_CTRL_LOSS_TMO;
}
```
what about the rest? and these are not kernel defaults. these are library defaults.

igaw:
For the first commit could you update the commit message. It's really hard to read. Maybe
```
libnvme/config: apply a parameter store to a connect context

Add libnvmf_context_apply_params() to apply the parameters read from an
INI file to an fctx. Parameters corresponding to unset keys are left
unchanged, while parameters explicitly reset to the empty string are
skipped instead of being applied as zero.
```
?

fabrics.c
```diff
@@ -765,45 +820,102 @@ int fabrics_discovery(const char *desc, int argc, char **argv, bool connect)
...
+	}
+
+	ret = libnvmf_config_read(ctx, config_file, &cfg);
+	if (ret) {
+		nvme_show_error("failed to read %s: %s", config_file,
+			libnvme_strerror(-ret));
+		return ret;
+	}
+
+	st = (struct consume_state){
```
igaw, line 911: this is a very unusual style. didn't even know this is possible.

```c
struct consume_state st = {
...
};
```

igaw:
Okay, not an easy change, but given the mess before, let's got with this. It makes sense while reading it. So the code is understand. Any chance to add a end to end test for this?

libnvme/src/nvme/config.c
martin-belanger: Fixed

libnvme/src/nvme/config.c
martin-belanger: Sorry, that was `libnvmf_config_conn_get_params()`, abbreviated badly. Spelled it out.

libnvme/src/nvme/config.c
martin-belanger: Fair, trimmed to one line.

libnvme/src/nvme/config.c
martin-belanger: You're right, and there was a simpler way to do this I'd missed: `libnvmf_params_get()` already exists for a direct by-name lookup. Replaced the whole buffer-in-`apply_state`-then-`set_crypto()`-once approach with five direct `libnvmf_params_get()` calls after the loop -- no more special-casing these five keys in the per-key callback at all. The "clobbering" concern was real (`libnvmf_context_set_crypto()` does unconditionally overwrite all five fields), it just didn't need the machinery I built around it.

libnvme/src/nvme/config.c
martin-belanger: That comment was wrong on both counts, thanks for catching it. `libnvmf_default_config()` only sets `tos` and `ctrl_loss_tmo` -- everything else is just zero from `calloc`, not some default this was "already carrying." And `ctrl_loss_tmo`'s value is a library choice (`NVMF_DEF_CTRL_LOSS_TMO`), not anything the kernel dictates. Reworded to just say what's actually true: an explicit reset is left unapplied, `@cfg` keeps whatever it already had. Also caught and fixed the same "kernel default" wording in `config-api.c`'s test output strings while I was in there.

martin-belanger:
> For the first commit could you update the commit message. It's really hard to read. Maybe
> 
> ```
> libnvme/config: apply a parameter store to a connect context
> 
> Add libnvmf_context_apply_params() to apply the parameters read from an
> INI file to an fctx. Parameters corresponding to unset keys are left
> unchanged, while parameters explicitly reset to the empty string are
> skipped instead of being applied as zero.
> ```
> 
> ?

Done, used your wording verbatim -- it's clearer than what I had.

fabrics.c
martin-belanger: That's `struct consume_state st` declared at the top of the function (per our own "declarations before code" convention) but not initializable until after a few earlier steps that can each fail (`nvme_config_convert_auto()`, `libnvmf_host_get_ids()`, `libnvmf_config_read()`) -- so it can't be a declare-with-initializer. The compound-literal assignment is how I reconciled "declare at the top" with "the values aren't ready yet." Not attached to the style if you'd rather see it a different way, but that's the reasoning.

martin-belanger:
> Okay, not an easy change, but given the mess before, let's got with this. It makes sense while reading it. So the code is understand. Any chance to add a end to end test for this?

There's already a real test for the part of this that's testable without hardware: `test_apply_params()` in `libnvme/test/config-api.c` (same commit) exercises `libnvmf_context_apply_params()` directly -- hex/decimal ints, bools, explicit resets, and all five crypto keys, including the exact no-clobber scenario from the comment above. It's been running as part of `meson test` this whole time, and it passing after the crypto-key rework confirms that refactor preserves the same verified behavior.

A true end-to-end test (actually connecting over the wire) is a bigger ask than it sounds, though. I checked whether `--dry-run` could get us there: it can't -- `libnvme_set_dry_run()` only gates IOCTL passthru on an already-open device handle, not the connect operation itself, which still needs a real `open("/dev/nvme-fabrics")` regardless. That needs either real hardware + root, or the NVMe-oF loop transport with `nvmet` configured -- neither of which fits `tests/`'s current no-hardware tier (what `nvme_config_convert_test.py` uses). It'd fit the nightly workflow's privileged tier instead, alongside the loop/rdma/tcp blktests runs that already happen there. Happy to scope that out as a follow-up if you want it, but wanted to flag it's nightly-tier work, not a quick addition to this PR.

igaw:
Thanks a lot.

testing: we should have something in the nightly builds. this is something we need to be sure it works. people will be very unhappy if we break their systems. but one thing at a time.

PR #3599, opened 2026-07-21
doc: add libnvme changes to NEWS.md

igaw:
Add some of the biggest libnvme API changes to NEWS.md. There are likely some more.


igaw:
what is missing?

- [ ] support for Windows
- [ ] splitting of the previously single command initiazation and submission into:  nvme_cmd_init & nvme_submit
- [ ] python bindings changes
- [ ] moving fabrics code into the library
- [ ] introduction of nvmf_context (hooks etc)
- [ ] preference to use 64bit ioclt instead the 32bit version
- [ ] diagnostic API
- [ ] nvme top
- [ ] global ini configuration
- [ ] nvme utils - introspection for auto generation completion
- [ ] nvme disconnect accepts now all fabrics command line options
- [ ] PDC/EPCSD support extended/added
- [ ] Refactoring/cleanups in nvme discover code path
- [ ] Read sysfs attribute lazy -> avoid overhead when scanning sysfs tree
- [ ] nvme keys: new plugins to handle TLS/KXCHAP secrets/keys (breaking change)
- [ ] nbft parser extended/hardened
- [ ] introduction of internal library for shared/common code

martin-belanger:
Maybe I can ask Claude to check all the changes since the last release and make a list of all the user-facing API changes/additions.

igaw:
I've asked copilot and this is an updated version of it. it's okay but it lacks a bit...

PR #3600, opened 2026-07-21
libnvme: fix regressions from the implicit host-lookup removal

martin-belanger:
Follow-up fixes for #3596. I was partway through reviewing it when it merged, and kept finding things — sorry these are only showing up now instead of as review comments.

## What broke

`25d6c95a7` ("libnvme/fabrics: do not look up hostnqn/hostid implicitly") correctly moved hostnqn/hostid resolution out of connect/discover and into the caller's hands, but three places didn't get updated to match the new contract, and one gap was worth closing while in there.

**PCIe controllers vanish from `nvme list`.** `libnvme_get_host()` lost its fallback entirely. A PCIe controller has no `hostnqn`/`hostid` sysfs attribute, so it now hits `libnvme_lookup_host(ctx, NULL, NULL)`, which returns `NULL` at its own `if (!hostnqn)` guard, and `scan_topology()` just logs the failure at DEBUG and silently drops the controller. Reproduced on real hardware: a drive that lists fine on `master` disappears entirely after this merge.

**Python bindings can crash the interpreter.** `libnvmf_read_hostnqn()`, `libnvmf_read_hostid()`, and `libnvmf_host_get_ids()` all dereference `ctx` unconditionally. They're public API now and reachable from Python, which can pass `None`. `None` maps to a NULL pointer, and all three segfault instead of raising.

**The Python binding reintroduces the exact thing the PR removes.** `set_fctx_host_params()` (nvme.i) called `libnvmf_host_get_ids()` unconditionally, so a `Ctrl(ctx, {...})` built from a dict with no `hostnqn`/`hostid` got one invented via the full fallback-and-generate chain — the same implicit lookup this PR takes out of the C side, just resurfacing one layer up in the SWIG glue.

## The fixes (2 commits)

1. **`libnvme: fix regressions from the implicit host-lookup removal`** — all three regressions above, plus one new check: a fabrics-transport `Ctrl` (tcp/rdma/fc/loop) with no resolvable hostnqn anywhere used to construct successfully and only fail later, confusingly, the first time `connect()`/`discover()` was called. It now rejects at construction instead, since the caller already knows which transport it asked for at that point. PCIe and apple-nvme are unaffected — neither has a host identity concept.

   For the PCIe fix specifically: rather than resolving a hostnqn/hostid for `libnvme_get_host()` (which would just relocate the same policy-in-the-library problem this PR is fixing), it falls back to a fixed placeholder. `tree-win.c` already did exactly this; the two platforms now share one `#define` instead of Linux resolving and Windows hardcoding a duplicate literal.

2. **`libnvme: consolidate libnvme_get_host() into tree.c`** — a separate commit, not a regression fix. Once both platforms fall back to the same placeholder, `tree-linux.c`'s and `tree-win.c`'s `libnvme_get_host()` are byte-identical, and neither has any platform-specific dependency left (`libnvme_lookup_host()` and `libnvme_host_set_hostsymname()` already live in shared code). Moved the one implementation into `tree.c`, dropped both copies.

## Testing

Full `meson test` suite passes (86/86). Manually verified `nvme list` on real hardware, side by side with a pre-merge build, to confirm the PCIe regression and the fix. Exercised the identity-resolution paths directly against the built Python module — every combination of dict-supplied / ctx-default / neither, for both a UUID-derivable and a non-derivable hostnqn, plus the `None`-ctx crash paths before and after.

### DIFF
--- a/libnvme/libnvme3/nvme.i
+++ b/libnvme/libnvme3/nvme.i
@@ -175,10 +175,10 @@ typedef struct { const char *key; const char **val; } str_fields_t;
  */
 static int set_fctx_host_params(struct libnvme_global_ctx *ctx,
 				struct libnvmf_context *fctx,
+				const char *transport,
 				PyObject *dict)
 {
 	const char *hostnqn = NULL, *hostid = NULL;
-	char *hnqn = NULL, *hid = NULL;
 	const char *hostkey = NULL, *ctrlkey = NULL;
 	const char *keyring = NULL, *tls_key = NULL, *tls_key_identity = NULL;
 	int err;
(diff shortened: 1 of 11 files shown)


libnvme/libnvme3/nvme.i
```diff
@@ -202,16 +202,53 @@ static int set_fctx_host_params(struct libnvme_global_ctx *ctx,
...
-	free(hid);
+	if (hostnqn || hostid) {
+		bool hostid_missing = !hostid;
+
+		err = libnvmf_context_set_hostnqn(fctx, hostnqn, hostid);
+		if (err) {
+			/* The only way this specific combination fails is a
+			 * hostid that can't be derived from hostnqn (it's
+			 * only derivable from a uuid:-form hostnqn) -- say so,
+			 * instead of surfacing the bare errno from below.
```
dwsuse, line 239: `libnvmf_context_set_hostnqn` has currently no sanity checks. we should probably add them. currently the function will only return success.

libnvme/libnvme3/nvme.i
```diff
@@ -202,16 +202,53 @@ static int set_fctx_host_params(struct libnvme_global_ctx *ctx,
...
 	}
 
-	libnvmf_context_set_hostnqn(fctx, hnqn, hid);
+	/* PCIe (and apple-nvme) have no host identity concept -- a fabrics
+	 * transport is a policy decision the caller must have already made
+	 * by the time it constructs a Ctrl, not something to discover only
+	 * when connect()/discover() is later called on it.
+	 */
+	if (!hostnqn && strcmp(transport, "pcie") &&
+	    strcmp(transport, "apple-nvme")) {
```
dwsuse: I wonder if we should introduce a version of `bool libnvme_ctrl_is_transport_fabric(libnvme_ctrl_t c);` which takes the transport string.

martin-belanger: Done -- added `libnvme_transport_is_fabric(const char *transport)`, and `libnvme_ctrl_is_transport_fabric()` now just delegates to it (`return c && libnvme_transport_is_fabric(c->transport);`). Same classification logic in one place instead of the inline `strcmp` chain I'd written in nvme.i.

libnvme/libnvme3/nvme.i
martin-belanger: It's not quite that -- it already returns `-EINVAL` for a NULL `hostnqn`, and again if `hostid` is missing and can't be derived from a UUID-form `hostnqn` (verified directly, and this is exactly what the error path a few lines up in this same diff is handling). What it doesn't have is *format* validation: nothing checks that `hostnqn` looks like a real NQN, or that `hostid` looks like a real UUID -- any non-empty string passes. That's probably what you're getting at. Adding that is a reasonable idea, but it's a separate, bigger design question (what format to actually enforce, and where) -- I'd rather not fold it into this regression-fix PR. Happy to scope it as its own follow-up if you want it.

martin-belanger:
All review comments addressed.

libnvme/libnvme3/nvme.i
igaw: stupid me, I looked at the old version... sorry for the noise.

igaw:
Thanks a lot!

PR #3602, opened 2026-07-21
nvme-cli: retire --dump-config from discover/connect-all/connect

martin-belanger:
## Summary

Retires `--dump-config`/`-O` from `discover`/`connect-all`/`connect`. It dumps the live topology tree as JSON, which is already out of step now that nvme-cli reads its own configuration as INI -- but it has real, independent problems beyond that:

- Anything discovered via a DC's discovery log page shows up indistinguishable from something deliberately configured. Feeding the dump back in as a new `config.json` would freeze today's dynamic discovery results into static entries that silently go stale the moment the DC's subsystem list changes.
- It conflates every orchestrator's connections with no way to attribute ownership.
- Its per-connection parameters (queue counts, timeouts, TLS settings) are only ever populated for a connection *this process* just made -- `libnvme_scan_topology()` never reads them back from sysfs, so anything pre-existing shows fabricated defaults.

Retiring it here also clears the way for a later commit to delete libnvme's JSON config machinery entirely. A replacement, INI-native and scoped to what a given invocation was actually told to connect (excluding anything a DC's DLP resolves on its own), is planned separately.

`nvme config`'s own `-M`/`-O`/`-R`/`-U` are untouched here -- they retire in a later commit when that command itself is replaced.

## Test plan

- [x] `meson compile -C .build` clean, default and `-Dwerror=true`
- [x] `meson test -C .build` 86/88 (2 expected fail, unchanged)
- [x] `-O` confirmed gone from `--help` output on all three commands
- [x] checkpatch clean (aside from known-noisy false-positive shapes)


igaw:
Thanks!

PR #3603, opened 2026-07-21
nvme: add dump-command-metadata command

jimmunn:
Add `nvme dump-command-metadata`, which walks the live plugin/command tree and emits every command and its options as JSON describing the CLI surface. The command needs no device and is gated on CONFIG_JSONC.

This is the first step toward generating shell completion scripts: the JSON output is intended to be consumed by a generator that produces completions for bash, zsh, and PowerShell.

To capture options, dump-command-metadata invokes each command's fn, which calls argconfig_parse(). A hook there copies the command's options array and returns a sentinel so the command unwinds before opening a device, rather than actually running it.

Includes command-metadata-schema.json and a Python unit test (unit-py/) that validates the output against the schema and cross-checks it against --help and `nvme help`.

### DIFF
--- a/plugins/utils/command-metadata.c
+++ b/plugins/utils/command-metadata.c
@@ -21,10 +21,14 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#ifdef _WIN32
+#include <windows.h>
+#endif
 
 #include "command-metadata.h"
 #include "common.h"
@@ -63,6 +67,139 @@ static struct command_metadata_command *command_metadata_cur_command;
  * for the parser-unwind sentinel; checked after each command fn returns.
  */
 static int command_metadata_capture_error;
+
+/*
+ * capture_saved_stderr_fd holds a dup of stderr from before the NUL redirect,
+ * so crash messages still reach the user.
+ */
+static const char *capture_current_command;
+static int capture_saved_stderr_fd = -1;
+
+/*
+ * (void)write() does not suppress GCC's warn_unused_result under
+ * _FORTIFY_SOURCE; assigning to a discarded variable does.
+ */
+static void write_raw(int fd, const char *buf, size_t len)
+{
+	ssize_t ret = write(fd, buf, len);
+	(void)ret;
+}
+
+/* strlen() is not async-signal-safe; measure inline for the crash handler. */
+static void write_str(int fd, const char *s)
+{
+	size_t len = 0;
+
+	if (!s)
+		return;
+
+	while (s[len])
+		len++;
+
+	write_raw(fd, s, len);
+}
+
+static void write_uint(int fd, unsigned int n)
+{
+	char buf[10];	/* enough for any 32-bit unsigned int */
+	int i = sizeof(buf);
+
+	do {
+		buf[--i] = '0' + (n % 10);
+		n /= 10;
+	} while (n && i);
+
+	write_raw(fd, buf + i, sizeof(buf) - i);
+}
+
+static void capture_crash_handler(int sig)
+{
+	int fd = capture_saved_stderr_fd >= 0 ? capture_saved_stderr_fd : STDERR_FILENO;
+
+	write_str(fd, "dump-command-metadata: fatal: '");
+	write_str(fd, capture_current_command ? capture_current_command : "(unknown)");
+	write_str(fd, "' crashed during option capture (signal ");
+	write_uint(fd, sig);
+	write_str(fd, ")\n");
+
+	signal(sig, SIG_DFL);
+	raise(sig);
+}
+
+/*
+ * On Windows, OS-level exceptions (stack overflow, access violations) bypass
+ * the C signal() mechanism entirely and terminate the process silently.
+ * SetUnhandledExceptionFilter intercepts these so we can print the same
+ * diagnostic before dying.
+ */
+#ifdef _WIN32
+static void write_hex(int fd, unsigned int n)
+{
+	char buf[8];
+	int i = sizeof(buf);
+
+	do {
+		buf[--i] = "0123456789abcdef"[n & 0xf];
+		n >>= 4;
+	} while (n && i);
+
+	write_raw(fd, buf + i, sizeof(buf) - i);
+}
+
+static LONG WINAPI capture_exception_filter(EXCEPTION_POINTERS *ep)
+{
+	int fd = capture_saved_stderr_fd >= 0 ? capture_saved_stderr_fd : STDERR_FILENO;
+
+	write_str(fd, "dump-command-metadata: fatal: '");
+	write_str(fd, capture_current_command ? capture_current_command : "(unknown)");
+	write_str(fd, "' crashed during option capture (exception 0x");
+	write_hex(fd, ep->ExceptionRecord->ExceptionCode);
+	write_str(fd, ")\n");
+
+	return EXCEPTION_CONTINUE_SEARCH;
+}
+
+static LPTOP_LEVEL_EXCEPTION_FILTER capture_prev_filter;
+#endif
+
+static void (*capture_prev_sigsegv)(int);
+static void (*capture_prev_sigabrt)(int);
+#ifdef SIGBUS
+static void (*capture_prev_sigbus)(int);
+#endif
+
+/* Normalize SIG_ERR to SIG_DFL so restoration is always a valid disposition. */
+static void (*set_crash_handler(int sig))(int)
+{
+	void (*prev)(int) = signal(sig, capture_crash_handler);
+
+	return prev == SIG_ERR ? SIG_DFL : prev;
+}
+
+static void install_crash_handlers(void)
+{
+	capture_prev_sigsegv = set_crash_handler(SIGSEGV);
+	capture_prev_sigabrt = set_crash_handler(SIGABRT);
+#ifdef SIGBUS
+	capture_prev_sigbus = set_crash_handler(SIGBUS);
+#endif
+#ifdef _WIN32
+	capture_prev_filter = SetUnhandledExceptionFilter(capture_exception_filter);
+#endif
+}
+
+static void remove_crash_handlers(void)
+{
+	signal(SIGSEGV, capture_prev_sigsegv);
+	signal(SIGABRT, capture_prev_sigabrt);
+#ifdef SIGBUS
+	signal(SIGBUS, capture_prev_sigbus);
+#endif
+#ifdef _WIN32
+	SetUnhandledExceptionFilter(capture_prev_filter);
+	capture_prev_filter = NULL;
+#endif
+}
 
 /* ------------------------------------------------------------------ */
 /* Pass 1: capture                                                    */
@@ -215,7 +352,9 @@ static int capture_command(struct command_metadata_command *mc, struct command *
 
 	command_metadata_capture_error = 0;
 	command_metadata_cur_command = mc;
+	capture_current_command = cmd->name;
 	(void)cmd->fn(2, argv, cmd, plugin);
+	capture_current_command = NULL;
 	command_metadata_cur_command = NULL;
 
 	/*
@@ -281,12 +420,14 @@ static struct command_metadata_program *build_model(struct program *prog)
 	if (devnull >= 0) {
 		saved_stdout = dup(STDOUT_FILENO);
 		saved_stderr = dup(STDERR_FILENO);
+		capture_saved_stderr_fd = saved_stderr;
 		if (saved_stdout >= 0)
 			dup2(devnull, STDOUT_FILENO);
 		if (saved_stderr >= 0)
 			dup2(devnull, STDERR_FILENO);
 	}
 
+	install_crash_handlers();
 	argconfig_set_parse_hook(metadata_capture_hook);
 
 	for (pi = 0, plugin = prog->extensions; plugin; plugin = plugin->next, pi++) {
@@ -316,6 +457,8 @@ static struct command_metadata_program *build_model(struct program *prog)
 	}
 
 	argconfig_set_parse_hook(NULL);
+	remove_crash_handlers();
+	capture_saved_stderr_fd = -1;
 
 	fflush(stdout);
 	fflush(stderr);
--- a/unit-py/test_command_metadata_schema.py
+++ b/unit-py/test_command_metadata_schema.py
@@ -63,8 +63,8 @@ def json_c_available():
 
 
 def dump_metadata():
-    """Run `nvme utils dump-command-metadata` and return parsed JSON, or None if
-    it produced no usable output."""
+    """Run `nvme utils dump-command-metadata` and return (parsed JSON, stderr).
+    Returns (None, stderr) if it produced no usable output."""
     proc = subprocess.run(
         [NVME_BIN, "utils", "dump-command-metadata"],
         stdout=subprocess.PIPE,
@@ -73,8 +73,8 @@ def dump_metadata():
     )
     text = proc.stdout.strip()
     if proc.returncode != 0 or not text:
-        return None
-    return json.loads(text)
+        return None, proc.stderr
+    return json.loads(text), proc.stderr
 
 
 def iter_commands(data):
@@ -128,10 +128,12 @@ class TestCommandMetadataSchema(unittest.TestCase):
     def setUpClass(cls):
         if not json_c_available():
             raise unittest.SkipTest("nvme built without json-c")
-        cls.data = dump_metadata()
+        cls.data, stderr = dump_metadata()
         if cls.data is None:
-            raise AssertionError(
-                "dump-command-metadata produced no output (built with json-c)")
+            msg = "dump-command-metadata produced no output (built with json-c)"
+            if stderr:
+                msg += "\nstderr: " + stderr.rstrip()
+            raise AssertionError(msg)
         with open(SCHEMA_PATH) as f:
             cls.schema = json.load(f)
 


util/argconfig.c
```diff
@@ -318,6 +318,13 @@ static bool argconfig_check_verbose(struct argconfig_commandline_options *s)
 	return false;
 }
 
+static argconfig_parse_hook_fn argconfig_parse_hook;
+
+void argconfig_set_parse_hook(argconfig_parse_hook_fn hook)
+{
+	argconfig_parse_hook = hook;
+}
```
igaw, line 326: The change is a bit big and there are a few things which could go in first before the main feature. That would make it a bit simpler to review. The hook here could go independent (the commit message should explain what it is for).

util/argconfig.h
```diff
@@ -180,6 +180,19 @@ int argconfig_parse(int argc, char *argv[], const char *program_desc,
...
 int argconfig_parse_global(int argc, char *argv[],
 			   struct argconfig_commandline_options *options);
+
+/*
+ * Optional hook invoked at the very top of argconfig_parse(). When set, it
+ * fully replaces parsing: argconfig_parse() returns the hook's value without
+ * running getopt or touching any state. Used by dump-command-metadata to
+ * capture each command's options array. NULL (the default) means normal
+ * parsing.
+ */
```
igaw: The LLMs tend to overdocument stuff in the source code. This information is good for the commit message, I don't think we should have it here. It's kind of obvious from the name of the function. I a big a fan of self explaining code. Comments are the last resort IMO.

util/json.h
```diff
@@ -12,6 +12,7 @@
 #define json_free_array(a) json_object_put(a)
 #define json_object_add_value_uint(o, k, v) json_object_object_add(o, k, json_object_new_uint64(v))
 #define json_object_add_value_int(o, k, v) json_object_object_add(o, k, json_object_new_int(v))
+#define json_object_add_value_bool(o, k, v) json_object_object_add(o, k, json_object_new_boolean(v))
```
igaw, line 17: Another thing which could be added independently.

plugins/utils/command-metadata.c
```diff
@@ -0,0 +1,564 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
```
igaw, line 2: Could also add the copyright statement line? e.g. something like

```c
 * Copyright (c) 2025 Micron Technology, Inc.
```

plugins/utils/command-metadata-schema.json
```diff
@@ -0,0 +1,129 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/linux-nvme/nvme-cli/command-metadata-schema.json",
```
igaw, line 3: Could we place this file somewhere else? I still want to cleanup the root directory eventually, it's crowded with code, configuration files, compliance files, project documentation, etc.

For the config-schema.json.in is currently in libnvme/doc/ what about having it in Documentation?

.gitignore
```diff
@@ -16,6 +16,8 @@ compile_commands.json
 tests/**/__pycache__
 tests/nvmetests
 tests/**/*.pyc
+unit-py/**/__pycache__
+unit-py/**/*.pyc
```
igaw, line 20: sure, could go separately.

plugins/utils/command-metadata.c
```diff
@@ -0,0 +1,564 @@
...
+#include "common.h"
+#include "nvme.h"
+#include "util/json.h"
+
+/*
+ * The whole command is JSON-only, so it is compiled out entirely without
+ * json-c support: nvme-builtin.h does not register it and nvme.c does not
+ * define its handler, so dump_command_metadata() is never referenced.
+ */
+#ifdef CONFIG_JSONC
```
igaw, line 43: I would suggest to make this command a build option instead coupling it to CONFIG_JSONC being around. This way we can disable in default builds and only enable when updating the tab completion files.

plugins/utils/command-metadata.h
```diff
@@ -0,0 +1,62 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+/*
```
igaw, line 2: also here a copyright statement please.

command-metadata.c
```diff
@@ -0,0 +1,564 @@
...
+		if (!model->plugins) {
+			free(model);
+			return NULL;
+		}
+	}
+
+	/*
+	 * Suppress stdout/stderr while invoking command fns: a few commands
+	 * print before they reach the parser (e.g. gen-hostnqn), and some emit
+	 * parse-error diagnostics in reaction to the capture sentinel.
```
igaw: I was looking at gen-hostnqn  (and another one) recently, I a noticed it doesn't use the arg parser. I think we should add this to those commands. Every command should invoke the arg parser otherwise the globals argument options are not supported.

fabrics.c
```diff
@@ -761,12 +761,12 @@ int fabrics_discovery(const char *desc, int argc, char **argv, bool connect)
...
 	nvmf_default_args(&fa);
 
-	load_nvme_fabrics_module();
-
 	ret = argconfig_parse(argc, argv, desc, opts);
 	if (ret)
 		return ret;
 
+	load_nvme_fabrics_module();
+
```
igaw, line 790: looks good, add it as preparation patch.

meson.build
```diff
@@ -583,6 +583,7 @@ if want_nvme
     if json_c_dep.found()
         sources += [
             'nvme-print-json.c',
+            'command-metadata.c',
```
igaw: Let's add a meson build option for this feature.

igaw:
Looks good overall.

I'd like to have some of the stuff as preparation patches. And making an dedicated build option, this is not useful for normal users.

jimmunn:
> Looks good overall.
> 
> I'd like to have some of the stuff as preparation patches. And making an dedicated build option, this is not useful for normal users.

I think this could be useful for normal users. It would allow consuming scripts to determine if a newer command exists, or change how it calls a command based on changes to its options. I've seen too much code in other projects that checks for version numbers and make decisions based on that. This provides usees a machine readable structured representation of the API to use to make decisions within their scripts.

igaw:
Fair enough. I haven't thought about the runtime introspection use case. I suppose this would be useful when wrapping a GUI on top?

What about moving the command into a plugin. I'd like to avoid cluttering the top name space and gives the user the opportunity decided if it's enabled or not. Furthermore, it would bundle the code in one directory.

jimmunn:
Rebased onto current master. The prep changes this depended on are now merged upstream, so the remaining diff is just the command, schema, and test. Also reworded the subject to the nvme: prefix.

jimmunn:
Added the missing copyright lines in command-metadata.c and command-metadata.h.

util/argconfig.h
jimmunn: Removed comment in other PR that has been merged.

util/json.h
jimmunn: Added in another PR

.gitignore
jimmunn: Added in another PR.

command-metadata.c
jimmunn: There are a few other commands that don't use the arg parser. I think this is the definitive list:

- gen-hostnqn
- show-hostnqn
- zns list
- micron plugin-version
- micron cloud-SSD-plugin-version
- seagate plugin-version
- seagate cloud-SSD-plugin-version

Do we want all of these to call the argument parser. I am pretty sure at the very least the --output-format will work on all of them. If so, this should probably be in it's own PR.

plugins/utils/command-metadata.h
jimmunn: Done

plugins/utils/command-metadata.c
jimmunn: Done

command-metadata.c
jimmunn: I modified this comment to make it more generic. Once we change these commands, the comment would have been stale.

fabrics.c
jimmunn: Added in another PR

jimmunn:
> What about moving the command into a plugin. I'd like to avoid cluttering the top name space and gives the user the opportunity decided if it's enabled or not. Furthermore, it would bundle the code in one directory.

I like the idea of a plugin. Now comes the hardest part of programming, naming. We bounced around some ideas here and we came up with "utils", "cli" and "cli-utils" (plus a few others that didn't make the cut). So, it would look like one of these:

```
nvme utils dump-command-metadata
nvme cli dump-command-metadata
nvme cli-utils dump-command-metadata
```

I don't know what other commands, if any, may end up in this plugin. The concern with "utils" is that it could become a junk drawer of miscellaneous commands, but that may be OK. I am leaning towards "utils" but am fine with the others. Do you have a preference?

Do you want the creation of the new plugin in its own PR, or I could just create a new commit in this PR that just adds the plugin then another commit to move the dump-command-metadata command and related files to the new plugin.

command-metadata.c
igaw: Yes, I think we should fix those commands up to use the nvme arg parser. I agree this should go into a separate PR.

igaw:
Naming is difficult. I didn't propose any names in my original comment because I could only think of bad ones, and I didn't want to influence your suggestions. My initial idea was `meta`, which is obviously not a great name.

Both `utils` and `cli` sound fine to me. I think `utils` is a bit more self-explanatory and user-friendly.

It's fine if you create the plugin within this PR or as a separate PR. It's both fine. What is simpler for you works for me.

plugins/utils/command-metadata-schema.json
jimmunn: Moved to new plugin directory.

igaw:
I was about to merge the PR but then I noticed that the Windows build fails:

```
=================================== 31/31 ====================================
test:         nvme-cli:nvme-cli - command-metadata-schema
start time:   22:07:39
duration:     0.21s
result:       exit status 1
command:      MALLOC_PERTURB_=36 MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MESON_TEST_ITERATION=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 PATH=D:/a/nvme-cli/nvme-cli/.build-ci/libnvme/src;D:/a/nvme-cli/nvme-cli/.build-ci/;D:\a\_temp\msys64\ucrt64\bin;D:\a\_temp\msys64\usr\local\bin;D:\a\_temp\msys64\usr\bin;D:\a\_temp\msys64\usr\bin;C:\Windows\System32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;D:\a\_temp\msys64\usr\bin\site_perl;D:\a\_temp\msys64\usr\bin\vendor_perl;D:\a\_temp\msys64\usr\bin\core_perl D:\a\_temp\msys64\ucrt64\bin/python3.EXE D:/a/nvme-cli/nvme-cli/.build-ci/../unit-py/test_command_metadata_schema.py ./nvme.exe D:/a/nvme-cli/nvme-cli/plugins/utils/command-metadata-schema.json
----------------------------------- stderr -----------------------------------
E
======================================================================
ERROR: setUpClass (__main__.TestCommandMetadataSchema)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:/a/nvme-cli/nvme-cli/.build-ci/../unit-py/test_command_metadata_schema.py", line 133, in setUpClass
    raise AssertionError(
        "dump-command-metadata produced no output (built with json-c)")
AssertionError: dump-command-metadata produced no output (built with json-c)

----------------------------------------------------------------------
```
From the build logs, it looks like there is json-c support enabled. So not clear what's going on here.

BTW, I think you can squash all three patches together. I would have done that anyway :)

jimmunn:
@igaw , the Windows build was failing due to memblaze perf-stats-print-x crashing on Windows. PR #3654 fixes that. I also added some better error handling to this PR to provide visibility to a crash like that. Once PR #3654 is merged, the Windows build for this PR should pass.

plugins/utils/command-metadata.c
```diff
@@ -64,6 +68,105 @@ static struct command_metadata_command *command_metadata_cur_command;
...
+
+	write_str(fd, "dump-command-metadata: fatal: '");
+	write_str(fd, capture_current_command ? capture_current_command : "(unknown)");
+	write_str(fd, "' crashed during option capture (signal ");
+	/* Print signal number as 1-3 ascii digits; async-signal-safe. */
+	if (sig >= 100)
+		(void)write(fd, &"0123456789"[sig / 100], 1);
+	if (sig >= 10)
+		(void)write(fd, &"0123456789"[(sig / 10) % 10], 1);
+	(void)write(fd, &"0123456789"[sig % 10], 1);
```
igaw: uff... please use something like

```c
static void write_uint(int fd, unsigned int n)
{
	char buf[3];
	int i = sizeof(buf);

	do {
		buf[--i] = '0' + (n % 10);
		n /= 10;
	} while (n && i);

	write(fd, buf + i, sizeof(buf) - i);
}

write_str(fd, "' crashed during option capture (signal ");
write_uint(fd, sig);
write_str(fd, ")\n");
```

igaw:
I'm going to merge the first part of PR, then you can rebase and update the last patch.

plugins/utils/command-metadata.c
jimmunn: I added your write_uint function. I also added a corresponding write_hex function to follow the same pattern.

jimmunn: I added write_raw to avoid a build failure.

plugins/utils/command-metadata.c
```diff
@@ -64,6 +68,123 @@ static struct command_metadata_command *command_metadata_cur_command;
...
+	char buf[3];
+	int i = sizeof(buf);
+
+	do {
+		buf[--i] = '0' + (n % 10);
+		n /= 10;
+	} while (n && i);
+
+	write_raw(fd, buf + i, sizeof(buf) - i);
+}
```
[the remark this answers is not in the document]
igaw: I didn't know that strlen is not safe to use in the signal handler...

igaw:
I've tried this on Linux and it doesn't produce any output:

For this test I added a `signal(SIGTERM);` to `smart-log`

```
› meson test -C .build -v "nvme-cli:nvme-cli - command-metadata-schema"
ninja: Entering directory `/home/wagi/work/nvme-cli-upstream/.build'
ninja: no work to do.
1/1 nvme-cli:nvme-cli - command-metadata-schema RUNNING
>>> UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 LD_LIBRARY_PATH=/home/wagi/work/nvme-cli-upstream/.build/libnvme/src MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MALLOC_PERTURB_=110 MESON_TEST_ITERATION=1 /usr/bin/python3 /home/wagi/work/nvme-cli-upstream/.build/../unit-py/test_command_metadata_schema.py ./nvme /home/wagi/work/nvme-cli-upstream/plugins/utils/command-metadata-schema.json
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
E
======================================================================
ERROR: setUpClass (__main__.TestCommandMetadataSchema)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/wagi/work/nvme-cli-upstream/.build/../unit-py/test_command_metadata_schema.py", line 136, in setUpClass
    raise AssertionError(msg)
AssertionError: dump-command-metadata produced no output (built with json-c)

----------------------------------------------------------------------
Ran 0 tests in 0.004s

FAILED (errors=1)
```

jimmunn:
> I've tried this on Linux and it doesn't produce any output:
> 
> For this test I added a `signal(SIGTERM);` to `smart-log`
> 

The lack of output here is by design and expected. If a command crashes while capturing its options, then we didn't capture them — so any output would be silently incomplete. Rather than emit a partial/misleading dump, we emit nothing.

Also, SIGTERM isn't a good signal for testing this. SIGTERM can't be raised by the running code itself — it's delivered from outside the process (a kill, a timeout, a service manager). So we don't handle it, and we don't print that a command crashed, because one didn't: the process was terminated externally.

To exercise the crash path, inject raise(SIGSEGV) or abort() (SIGABRT) instead — those are the fault signals the capture handler catches, and you'll see the dump-command-metadata: fatal: '<cmd>' crashed during option capture diagnostic.

plugins/utils/command-metadata.c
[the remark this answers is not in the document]
jimmunn: I addressed the Copilot reported issues in the latest push:

- write_str now computes the length inline instead of calling strlen, which isn't async-signal-safe.
- write_uint's buffer is widened to 10 bytes so any 32-bit unsigned int formats without dropping leading digits.

write_hex already sized its buffer correctly, so it's unchanged.

igaw:
ah okay, that explains it. btw, copilot was unhappy about the `signal, raise` sequence: https://github.com/linux-nvme/nvme-cli/pull/3686#discussion_r3674559844

haven't really spend time on trying to make sense of it. if you say it is bogus, I'll merge this version.

jimmunn:
> ah okay, that explains it. btw, copilot was unhappy about the `signal, raise` sequence: [#3686 (comment)](https://github.com/linux-nvme/nvme-cli/pull/3686#discussion_r3674559844)
> 
> haven't really spend time on trying to make sense of it. if you say it is bogus, I'll merge this version.

Yeah, it's bogus. Both signal() and raise() are explicitly listed as async-signal-safe in signal-safety(7). The diagnostic already uses write(2), and the reset-to-SIG_DFL-then-raise() is intentional — it preserves the core dump and the correct exit status.

Copilot's other comment on that branch was legit, though: install_crash_handlers() didn't guard against signal() returning SIG_ERR, which could then get passed back to signal() on restore. I've pushed a fix that normalizes SIG_ERR to SIG_DFL at install time (via a small set_crash_handler() helper), so restoration is always valid. That's the only delta since you last looked.

igaw:
Alright, thanks for looking into it. I try to stay away from signals, this is such a stupid and horrible API.

igaw:
Thanks a lot!

PR #3604, opened 2026-07-21
doc: add nvme-fabrics.conf(5) man page and sample file

martin-belanger:
## Summary

Adds a `nvme-fabrics.conf(5)` man page documenting the INI connection-configuration format (sections, keys, controller-line syntax, precedence), plus an example file at `/etc/nvme/nvme-fabrics.conf.sample`.

This restores -- at a path the config reader never touches -- what removing the old `discovery.conf` install stub took away: a way to see the expected file format without reading source or the network.

- `Documentation/nvme-fabrics.conf.txt` -- the man page source. Keeps to a quick-reference shape (every key listed, minimal prose per key) rather than CONFIG.md's design-rationale depth; pulls its key list and worked example from CONFIG.md directly.
- `Documentation/meson.build` -- only installs man1 pages today; adds a man5 install path.
- `etc/nvme-fabrics.conf.sample.in` -- installed as a plain `%files` entry (not `%ghost` -- static reference content, not runtime state). 

## Test plan

- [x] Man page actually built and rendered via `-Ddocs=man -Ddocs-build=true` (real asciidoc/xmlto toolchain) and checked with `man --local-file`; `-Ddocs=all` (html) also builds
- [x] Precompiled-fallback install path isolate-tested (man1 and man5 both land in the right `mandir`)
- [x] Refactored man1/man5 meson blocks verified to produce byte-identical output against the pre-refactor two-block version
- [x] checkpatch clean (0 errors, 0 warnings)

### DIFF
--- a/Documentation/meson.build
+++ b/Documentation/meson.build
@@ -273,6 +273,12 @@ adoc_sources = [
     'nvme-mangoboost-id-ctrl',
 ]
 
+# File-format references (man section 5), installed separately from the
+# man1 command pages above.
+adoc_sources_man5 = [
+    'nvme-fabrics.conf',
+]
+
 adoc_includes = [
     'cmd-plugins.txt',
     'cmds-main.txt',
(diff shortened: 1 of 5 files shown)


igaw:
Thanks! Very cool that we have this now even properly documented!

PR #3605, opened 2026-07-21
libnvme/test: check fopen() before fclose() in config-emit

martin-belanger:
Coverity CID [561980](https://scan9.scan.coverity.com/#/project-view/64948/14256?selectedIssue=561980)/[561979](https://scan9.scan.coverity.com/#/project-view/64948/14256?selectedIssue=561979): fopen() can return NULL, and passing that to fclose() is undefined behavior. Split each into a checked fopen() followed by fclose(), matching the pattern already used elsewhere in the test suite (config-api.c's write_file()).


igaw:
Thanks!

PR #3606, opened 2026-07-22
nvme-cli: rework "nvme config" into config-validate/config-show

martin-belanger:
## Summary

Replaces the JSON-era `nvme config` (`-J`/`-R`/`-M`/`-O`/`-U`) editor with two flat, read-only commands:

- `nvme config-validate` -- resolves the connection configuration without connecting, reporting `file:line` diagnostics on failure and `<file>: OK` on success. Wraps `libnvmf_config_validate()`.
- `nvme config-show` -- renders each resolved connection as its equivalent command line: `nvme connect-all` for a discovery controller entry, `nvme connect` for an I/O controller entry. A connection with no explicit host persona is annotated as using the system default. Never resolves addressing -- a hostname `traddr` is shown as configured, not looked up.

This is the last nvme-cli caller of `libnvme_read_config()`/`libnvme_dump_config()`/`libnvmf_config_modify()`, so dropping those plus `json.c`/json-c from libnvme is unblocked next.

Neither command auto-converts legacy `config.json`/`discovery.conf`, unlike `connect`/`discover`/`connect-all` -- a read-only inspection command silently rewriting files on disk would be a surprising side effect.

## Test plan

- [x] `nvme config-validate`/`nvme config-show` exercised against valid and malformed config files, including entries with unresolvable hostname `traddr`
- [x] checkpatch clean (0 errors, 3 warnings -- pre-existing description-string length + two intentionally-unwrapped `OPT_ARGS` table rows)

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -66,7 +66,6 @@
 static char *raw;
 static bool persistent;
 static bool quiet;
-static bool dump_config;
 
 static const char *nvmf_tport		= "transport type";
 static const char *nvmf_traddr		= "transport address";
@@ -98,6 +97,7 @@ static const char *nvmf_data_digest	= "enable transport protocol data digest (TC
 static const char *nvmf_tls		= "enable TLS";
 static const char *nvmf_concat		= "enable secure concatenation";
 static const char *nvmf_config_file	= "Use specified INI configuration file (auto-converts a legacy .json) or 'none' to disable";
+static const char *nvmf_config_file_ro	= "INI configuration file (default: " PATH_NVMF_INI ")";
 
 #define NVMF_ARGS(n, f, ...)                                                                  \
 	NVME_ARGS(n,                                                                              \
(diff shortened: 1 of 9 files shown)


fabrics.c
```diff
@@ -1381,64 +1372,141 @@ int fabrics_config(const char *desc, int argc, char **argv)
 	}
 	libnvme_set_logging_level(ctx, log_level, false, false);
 
-	libnvme_read_config(ctx, config_file);
+	ret = libnvmf_config_validate(ctx, config_file);
+	if (ret)
+		nvme_show_error("%s: invalid configuration", config_file);
+	else
+		printf("%s: OK\n", config_file);
```
igaw: this should used nvme_show_result, so that the output is in the correct format.

martin-belanger: fixed

martin-belanger:
I did the plugin migration as a separate second commit rather than folding it into the first, since it's a distinct change. This commit also migrates the already-merged config-convert into the plugin alongside config-validate/config-show, so all three end up under nvme config [subcommand] together.

fabrics.c
```diff
@@ -1345,33 +1345,24 @@ int fabrics_disconnect_all(const char *desc, int argc, char **argv)
...
 
-	NVMF_ARGS(opts, fa,
-		  OPT_STRING("config",             'J', "FILE", &config_file, nvmf_config_file),
-		  OPT_FLAG("scan",                 'R', &scan_tree,           "Scan current NVMeoF topology"),
-		  OPT_FLAG("modify",               'M', &modify_config,       "Modify JSON configuration file"),
-		  OPT_FLAG("dump",                 'O', &dump_config,         "Dump JSON configuration to stdout"),
-		  OPT_FLAG("update",               'U', &update_config,       "Update JSON configuration file"));
-
-	nvmf_default_args(&fa);
+	OPT_ARGS(opts) = {
```
igaw, line 1355: Okay, this means the globals arguments are here not valid. I was wondering how the printfs will work with `--output-format json`.

With `OPT_ARGS` only stdout is supported. Should we support json as well?


And  there is a bug in the current parser

```shell
› ./nvme --output-format json config validate
nvme: /builddir/build/BUILD/json-c-0.18-build/json-c-json-c-0.18-20240915/json_object.c:564: json_object_object_add_ex: Assertion `json_object_get_type(jso) == json_type_object' failed.
[1]    170368 IOT instruction (core dumped)  ./nvme --output-format json config validate
```

igaw:
Let's go with this and figure out the rest later.

igaw:
Thanks!

PR #3608, opened 2026-07-22

LJaved-IBM:
@sph irrelevant. Avoid this by setting @sph to NULL before
calling wdc_get_dev_mng_log_entry() to make the intent clear.

Also remove the bogus null-pointer check on *p_p_found_log_entry
in wdc_get_dev_mng_log_entry(). The check tested the wrong thing
— it dereferenced p_p_found_log_entry to read the output value,
rather than checking p_p_found_log_entry itself. Since the very
next line unconditionally writes to *p_p_found_log_entry, the
check provided no protection and was misleading.


igaw:
Would you mind to update the commit message so it doesn't use the `@` for the variable names. Very unusual and there is no `DaTa` anyway. Thanks!

PR #3609, opened 2026-07-22
nvme: fabrics_config_validate: reject a missing config file

martin-belanger:
libnvmf_config_validate() treats an absent file as a valid, empty configuration -- correct for connect-all/discover, where no config just means nothing to connect, not an error. "nvme config validate" inherited that behavior and reported "OK" for a file that does not exist at all, which is misleading: the whole point of the command is confirming the file itself is correct.

Check for the file explicitly before validating and report a clear error instead.


igaw:
Thanks for the quick fix!

PR #3613, opened 2026-07-22
libnvme: drop the legacy JSON config path and json.c/json-c

martin-belanger:
## Summary

Drops the legacy JSON config path from libnvme now that `config-validate`/`config-show` (merged in #3606) confirmed there's no remaining nvme-cli caller:

- Removes `libnvme_read_config()`, `libnvme_dump_config()`, `libnvmf_config_modify()`, `json.c`/`no-json.c`, and libnvme's own `json-c` build dependency. The top-level `nvme` binary keeps linking `json-c` directly for `nvme-print-json.c` -- unaffected, confirmed via `nvme list -o json`.
- Also removes `libnvme_dump_tree()`: it has no caller outside its own test, and `nvme list -vvv -o json` already walks the same host/subsystem/controller/namespace tree and renders it as JSON, so there's nothing to port.
- Drops the JSON-config-only test coverage this obsoletes (`libnvme/test/config/`, the `config-schema.json` doc/validator pair) and replaces `libnvme/test/sysfs/tree-dump.c` with a JSON-free `test-tree-scan.c` -- same sysfs fixtures and diff harness, reused unchanged, just a plain-text dump instead of JSON. No longer gated on `json-c` availability, so this scan-topology regression test now runs in every build configuration, including no-fabrics.

## Test plan

- [x] `meson compile -C .build` / `meson test -C .build` clean (79/81, 2 expected fail unchanged)
- [x] `-Dfabrics=disabled -Dwerror=true` build clean
- [x] Both commits verified standalone (isolated worktree) and combined
- [x] `nvme list -o json` smoke-tested to confirm the CLI's own JSON output is unaffected
- [x] Accessor regeneration confirmed no drift
- [x] checkpatch clean (0 errors; 1 pre-existing false positive on a byte-identical restored file, confirmed by replaying checkpatch against the same file's own original merged commit)


igaw:
The tests in libnvme/test/config are still valid and should stay. They test if the topology scan works or the hostnqn lookup order etc. So these should be migrated over to the new tree-scan output.

martin-belanger:
> The tests in libnvme/test/config are still valid and should stay. They test if the topology scan works or the hostnqn lookup order etc. So these should be migrated over to the new tree-scan output.

I guess I used a sledge hammer when I should have used a scalpel. :smile: 

The topology-scan piece, and the half of `hostnqn-order.c` that checks CLI-style args, port cleanly onto `tree-scan.c` — I'll migrate those.

The JSON-config half doesn't have an equivalent anymore, and the reason is architectural, not just mechanical. Part of what this series does is push policy out of libnvme: the loop that reads and parses the config file is moving into the caller. By the time a connection reaches libnvme now, its hostnqn/hostid/addresses are already fully resolved by the caller — libnvme just consumes a finished config, it no longer cascades through Subsystem/Host/global-ctx state to fill in missing pieces. That cascade (`libnvmf_host_get_ids()` falling back to whatever `libnvme_read_config()` had staged as "the first host") was JSON-specific machinery, and it goes away with `json.c`.

That said, one piece of that same fallback is still alive, just reachable from a different direction: `libnvmf_host_get_ids()` still falls back to whatever host `libnvme_first_host()` returns, and that's now populated purely from sysfs-discovered controllers rather than from a config file. `tree-scan.c` already scans sysfs fixtures, so I'll add a scenario there with an existing connected controller and check that `libnvmf_host_get_ids()` picks up its hostnqn/hostid — same libnvme code path as the old test, exercised from the topology-scan angle instead of the JSON-config angle.

For the config-driven precedence itself, I'll add a new libnvme-only test fixture that drives it end to end through public API: parse an INI file with `libnvmf_config_read()`, pull a connection's hostnqn/hostid via `libnvmf_config_conn_get_hostnqn()`/`get_hostid()`, and feed the result into `libnvmf_tid_from_fields()` — with fixtures covering both "connection sets its own hostnqn/hostid" and "connection inherits the default" cases. That's real libnvme code path coverage through the new config-reading API, not a copy of nvme-cli's logic.

Between the tree-scan.c addition and this new fixture, I don't expect this to cost any libnvme coverage — if anything, the amount of code coming out with json.c should move the percentage up, not down.

martin-belanger:
@igaw - Pushed a commit that migrates the coverage as described above:

- `libnvme/test/config-api.c`: new `test_hostnqn_precedence()`, driving the real INI-driven precedence (a connection's own hostnqn/hostid wins, else the caller's default) end to end through public API (`libnvmf_config_read()` → `libnvmf_config_conn_get_hostnqn()`/`get_hostid()` → `libnvmf_tid_from_fields()`), reusing the existing two-persona fixture.
- `libnvme/test/sysfs/tree-scan-hostnqn.c` (new, fabrics-gated): confirms `libnvmf_host_get_ids()`'s host-tree fallback still resolves correctly against the existing `tree-pcie`/`tree-apple-nvme` sysfs snapshots, now that it's reached purely via sysfs-discovered hosts rather than JSON config. Kept as a separate binary from `tree-scan.c`, since that one builds unconditionally (no fabrics needed for PCIe topology scanning) while `libnvmf_host_get_ids()`/`<nvme/fabrics.h>` aren't available in a `-Dfabrics=disabled` build.
- Found and fixed a latent bug in `tree-diff.sh` along the way: two tests sharing the same `.tar.xz` fixture raced on the same extraction directory. Keyed the directory off the expected-output filename instead, which is unique per test.

Verified: full suite green, checkpatch clean, and confirmed both the fabrics-enabled and `-Dfabrics=disabled` builds behave correctly (the new pieces are correctly absent from the latter).

libnvme/test/config-api.c
```diff
@@ -564,6 +564,93 @@ static bool test_apply_params(struct libnvme_global_ctx *ctx)
...
+
+	assert(!libnvmf_config_read(ctx, fx->main_path, &config));
+	libnvmf_config_conn_for_each(config, collect, &conns);
+	assert(conns.n == 3);
+	mv = conns.conn[1];
+	pv = conns.conn[2];
+
+	hostnqn = libnvmf_config_conn_get_hostnqn(mv);
+	if (!hostnqn)
+		hostnqn = default_hostnqn;
```
igaw: This looks wierd. I'd expect either `assert(hostnqn)` or `assert(!hostnqn)`. What's the purpose to fallback this way?

igaw:
Thanks. TBH, there was some redundancy (which also confused me) in the test framework I've never clean up. So this streamlines the testing quiet a bit. With the extensive unit tests for the new INI configuration I think we are covering very well our bases. 

There is only this assert question I have the rest looks good.

libnvme/test/config-api.c
martin-belanger: Fixed.

igaw:
Thanks a lot! This was a huge task and I am very glad you got this sorted out. Finally we have proper testing of the configuration side, no more dependency on system configuration etc. This was always hard to tests, reproduce and to fix.

Nice to see this code gone! I owe a few beers!

PR #3614, opened 2026-07-22
micron: Add tests for some micron plugin commands and fix errors found in testing

brgoing:
Adds tests for the following micron plugin commands:
- vs-internal-log
- vs-temperature-stats
- vs-pcie-stats
- clear-pcie-correctable-errors
- vs-drive-info

Also fixes micron plugin issues uncovered during testing:
- Fixes return value issues.
- Refactors PCI BDF discovery to fix issues and to simplify and standardize.
- Refactors BDF reading to spawn processes instead of using popen for security.
- Improves error messaging for accuracy and consistency.

### DIFF
--- a/plugins/micron/micron-nvme.c
+++ b/plugins/micron/micron-nvme.c
@@ -648,6 +648,44 @@ exit_status:
 	return err;
 }
 
+static int micron_validate_output_format(const char *format,
+	nvme_print_flags_t supported_formats,
+	nvme_print_flags_t *output_format)
+{
+	nvme_print_flags_t f;
+	int err = validate_output_format(format, &f);
+
+	if (err)
+		return err;
+
+	/* NORMAL is 0, so it is always supported and handled separately */
+	if (f == NORMAL || (f & supported_formats)) {
+		*output_format = f;
+		return 0;
+	}
+
+	return -ENOTSUP;
+}
+
+static int micron_get_output_format(struct argconfig_commandline_options *opts,
+	const char *global_format, const char *local_format,
+	nvme_print_flags_t supported_formats, nvme_print_flags_t default_format,
+	nvme_print_flags_t *format)
+{
+	if (!format)
+		return -EINVAL;
+
+	if (global_format && argconfig_parse_seen(opts, "output-format"))
+		return micron_validate_output_format(global_format,
+			supported_formats, format);
+	else if (local_format && argconfig_parse_seen(opts, "format"))
+		return micron_validate_output_format(local_format,
+			supported_formats, format);
+
+	*format = default_format;
+	return 0;
+}
+
 /*
  * Plugin Commands
  */
@@ -891,7 +929,7 @@ static int micron_temp_stats(int argc, char **argv, struct command *acmd,
 	unsigned int tempSensors[SensorCount] = { 0 };
 	const char *desc = "Retrieve Micron temperature info for the given device ";
 	const char *fmt = "output format normal|json";
-	nvme_print_flags_t flags;
+	nvme_print_flags_t format = NORMAL;
 	struct format {
 		char *fmt;
 	};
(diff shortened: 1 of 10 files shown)


plugins/micron/micron-utils-linux.c
```diff
@@ -19,157 +22,330 @@
...
 
-int micron_get_pcie_aer_errors(struct libnvme_transport_handle *hdl,
-	__u32 *correctable_errors, __u32 *uncorrectable_errors)
+/*
+ * Validates that a string is a canonical PCI address in the
+ * "DDDD:BB:DD.F" form (domain:bus:device.function), e.g. "0000:03:00.0".
+ *
+ * Return: true if valid, false otherwise.
+ */
+static bool pcie_bdf_is_valid(const char *bdf)
```
igaw, line 33: should we have this as helper in nvme-pci-ids?

plugins/micron/micron-nvme.c
```diff
@@ -912,14 +950,13 @@ static int micron_temp_stats(int argc, char **argv, struct command *acmd,
 		return -1;
 	}
 
-	err = validate_output_format(nvme_args.output_format, &flags);
-	if (err) {
+	err = micron_get_output_format(opts, nvme_args.output_format, cfg.fmt,
```
igaw, line 931: Could we make the validate_output_format better instead adding another wrapper on top?

NVME_ARGS will announce it supports:

```c
#ifdef CONFIG_JSONC
#define DESC_OUTPUT_FORMAT "Output format: normal|json|binary|tabular"
#else /* CONFIG_JSONC */
#define DESC_OUTPUT_FORMAT "Output format: normal|binary|tabular"
#endif /* CONFIG_JSONC */

#define NVME_ARGS(n, ...)                                                            \
	struct argconfig_commandline_options n[] = {                                   \
		OPT_GROUP("Options"),                                                  \
		OPT_GROUP("Global options"),                                           \
		OPT_INCR("verbose",      'v', &nvme_args.verbose,                      \
                         "Increase output verbosity"),                                 \
		OPT_FMT("output-format", 'o', &nvme_args.output_format,                \
                         DESC_OUTPUT_FORMAT),                                          \
[...]
```

and which will not match with what this plugin supports. `tabular` is also only supported by very few commands. What about keeping NVME_ARGS with the default output format settings (-tabular) and introduce a new NVME_ARGS_OUTPUT_FORMAT (better name wouldn't hurt) and this one takes the help text and a support ed flags argument which is handled by validate_output_format?

plugins/micron/micron-utils-linux.c
brgoing: Do you think it would be useful to have it available for use outside of this specific case?  I didn't find anything else that was using the BDF strings in a way that might need to be validated like this, which is why I implemented it locally.

plugins/micron/micron-nvme.c
brgoing: Great idea.  This seems like it is only semi-related to the changes in this pull request.  Would it be better to make the changes here, or as a separate refactoring PR after this one?

brgoing: The micron_get_output_format method is still needed to handle the format setting coming from either the local "format" parameter or the global "output-format" parameter.  I'm pushing for removing the local format parameter in favor of the global output-format, but there are concerns about breaking users of the existing command parameters.  I'll push toward standardizing those, but that will be in a future PR.  The micron_validate_output_format that micron_get_output_format calls is the wrapper for validate_output_format that can be removed in favor of enhancements to validate_output_format.

brgoing: I went ahead and added the changes to this PR with a new commit.

plugins/micron/micron-utils-linux.c
igaw: okay, let's leave it here. if something else needs this we can still refactor.

igaw:
Nice one! Thanks!

Issue #3607, opened 2026-07-22
Passing -J <JSON_config_file> option results in error


martin-gpy:
I'm aware that INI config file changes are in place now. But reading a JSON config file now results in a _malformed line_ error which used to work properly in the past. For e.g.

```
# cat /usr/local/etc/nvme/config.json
[
  {
    "hostnqn":"nqn.2014-08.org.nvmexpress:uuid:e6550026-173e-4959-ba74-be367844bd8a",
    "hostid":"e6550026-173e-4959-ba74-be367844bd8a",
    "subsystems":[
      {
        "nqn":"nqn.1992-08.com.netapp:sn.48391d66c0a611ecaaa5d039ea165514:subsystem.subsys_CLIENT116_0",
        "ports":[
          {
            "transport":"tcp",
            "traddr":"192.168.1.117",
            "host_traddr":"192.168.1.6",
            "trsvcid":"4420"
          }
        ]
      }
    ]
  }
]
```

```
# nvme connect-all -J /usr/local/etc/nvme/config.json
Failed to lookup host 'nqn.2014-08.org.nvmexpress:uuid:e6550026-173e-4959-ba74-be367844bd8a'
/usr/local/etc/nvme/config.json:1: malformed line "["
failed to read /usr/local/etc/nvme/config.json: Invalid argument
#
```
Is this expected? Or it no longer handles the -J <JSON_config_file> option anymore?

igaw:
This is supposed to work, especially with the automatic conversion. During the work on the configuration format we found out that the were bugs in the json parser which didn't follow the schema definition. I think your config example is a version before the `hosts` level was introduced. The 37a7247723f3 ("libnvme/json: write the config as an object, keep reading the legacy array") change is likely causing this regression. We should add your example to the tests and fix the parser.

@martin-belanger

martin-belanger:
Thanks for the report — you're right that this used to work and doesn't anymore. Confirmed: the current INI parser doesn't detect a legacy JSON file by its .json extension, so it tries to read it as INI syntax and fails with a confusing "malformed line" error instead of a clear one pointing you at nvme config-convert. Working on a fix.

martin-belanger:
Update: I tried to reproduce this directly with your exact config.json content on current master and couldn't — nvme connect-all -J config.json correctly detects the .json extension, auto-converts it, and proceeds (renaming the original to .converted). So my first read was wrong; that detection logic is there and working.

[...]

PR #3615, opened 2026-07-22
nvme, libnvme: fix two config-conversion regressions found via #3607

martin-belanger:
## Summary
- Fix `libnvmf_discovery_nbft()` logging a bogus "Failed to lookup host" error on every `connect-all`/`discover` invocation that doesn't pass `--no-nbft`, regardless of whether any NBFT tables exist. It looked up the host using local `hostnqn`/`hostid` variables that are only assigned later, per NBFT-table entry -- still `NULL` at the point of this first lookup.
- Fix `nvme_config_convert_auto()` naming an explicit `--config` pointing at the legacy default `config.json` path as `config.conf` instead of `nvme-fabrics.conf`, unlike a bare invocation converting the same file. Beyond the inconsistent naming, this orphaned the config: a later bare invocation found neither `nvme-fabrics.conf` nor `config.json` (renamed to `.converted`) and silently proceeded empty.

Both surfaced while following up on #3607.

## Test plan
- [x] `meson compile` clean
- [x] `meson test`: 86/88 (2 expected-fail unchanged)
- [x] Live repro + fix verification for both bugs via `nvme connect-all --dry-run`

### DIFF
--- a/config-convert.c
+++ b/config-convert.c
@@ -488,7 +488,8 @@ int nvme_config_convert_auto(struct libnvme_global_ctx *ctx,
 
 	*ini_path = NULL;
 
-	is_default = !strcmp(config_file, PATH_NVMF_INI);
+	is_default = !strcmp(config_file, PATH_NVMF_INI) ||
+		     !strcmp(config_file, PATH_NVMF_CONFIG);
 	if (is_default) {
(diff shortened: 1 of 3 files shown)


igaw:
Thanks!

FWIW, the commit message are still hard to read.

PR #3616, opened 2026-07-22
fabrics: fix host_traddr hostname resolution + add FC WWN validation

martin-belanger:
## Summary
- `nvmf_resolve_args()` resolved both `traddr` and `host_traddr` via DNS; `host_traddr` is always one of the host's own local addresses, never a remote hostname. Only `traddr` is resolved now. Fixed the matching kdoc on `libnvmf_config_conn_get_host_traddr()`, which had copy-pasted the "may be a hostname" wording from `get_traddr()`'s doc.
- `tid_sanitize_addr()` validated tcp/rdma addressing (`canon_ip()`) but left FC untouched -- any string was accepted as `traddr`/`host_traddr` for `transport=fc`. Added a lightweight structural check for the `nn-0x<hex>:pn-0x<hex>` WWN-pair shape, the same format `sanitize_discovery_log_entry()` already normalizes discovery-log traddr into.

## Test plan
- [x] `meson compile` clean
- [x] `meson test`: 86/88 (2 expected-fail unchanged)
- [x] valgrind on `test-tid`: 0 errors, 0 leaks
- [x] Standalone check covering 10 FC-address edge cases (valid pair, wrong-case prefix, trailing garbage, short/empty hex, legacy colon-byte format, cross-transport) -- all as expected

### DIFF
--- a/libnvme/src/nvme/tid.c
+++ b/libnvme/src/nvme/tid.c
@@ -78,21 +78,52 @@ static int canon_ip(const char *in, char **out)
 }
 
 /*
- * Sanitize the addressing fields once the transport is known: canonicalize a
- * numeric traddr/host_traddr via canon_ip(); a hostname is rejected outright
- * -- resolving one is the caller's job, not libnvme's.  IP addressing
- * applies only to the IP transports; fc and loop are left untouched.
+ * A lightweight structural check for FC addressing: "nn-0x<hex>:pn-0x<hex>"
+ * (node name + port name, each up to a 64-bit WWN in hex) -- the same shape
+ * sanitize_discovery_log_entry() already normalizes discovery-log traddr
+ * into.  Not a deep WWN semantic check (OUI bits etc.), the same spirit as
+ * canon_ip() below for tcp/rdma: catch a malformed value, not police it.
+ */
+static bool fc_wwn_is_valid(const char *in)
+{
+	char nn[17], pn[17];
+	int len = 0;
+
+	if (sscanf(in, "nn-0x%16[0-9a-fA-F]:pn-0x%16[0-9a-fA-F]%n",
+		   nn, pn, &len) != 2)
+		return false;
+
+	return (size_t)len == strlen(in);
+}
+
+/*
+ * Sanitize the addressing fields once the transport is known.  IP
+ * transports: canonicalize a numeric traddr/host_traddr via canon_ip(); a
+ * hostname is rejected outright -- resolving one is the caller's job, not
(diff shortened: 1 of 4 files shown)


libnvme/src/nvme/tid.c
```diff
@@ -78,21 +78,52 @@ static int canon_ip(const char *in, char **out)
...
- * numeric traddr/host_traddr via canon_ip(); a hostname is rejected outright
- * -- resolving one is the caller's job, not libnvme's.  IP addressing
- * applies only to the IP transports; fc and loop are left untouched.
+ * A lightweight structural check for FC addressing: "nn-0x<hex>:pn-0x<hex>"
+ * (node name + port name, each up to a 64-bit WWN in hex) -- the same shape
+ * sanitize_discovery_log_entry() already normalizes discovery-log traddr
+ * into.  Not a deep WWN semantic check (OUI bits etc.), the same spirit as
+ * canon_ip() below for tcp/rdma: catch a malformed value, not police it.
+ */
+static bool fc_wwn_is_valid(const char *in)
```
igaw, line 87: yes! thanks!

igaw:
> * `nvmf_resolve_args()` resolved both `traddr` and `host_traddr` via DNS; `host_traddr` is always one of the host's own local addresses, never a remote hostname.

What about the use case where an admin wants to use hostnames in the config, because the host gets the IP address via DHCP and it is not fixed? Is this a valid use case?

martin-belanger:
For TCP, host-traddr only selects a secondary address on the interface. The primary address is already the default when host-traddr is omitted, so the only thing host-traddr ever picks is one of the interface's other (secondary) addresses. DHCP hands out one address per interface, and that's the primary one, so a hostname lookup for host-traddr on TCP would just resolve back to the primary address every time -- not the secondary one the option exists to select. So I don't think this is a real case for TCP, and host-traddr should stay numeric there.

RDMA might be different -- the driver seems to use host-traddr to select both the interface and the source address. If that's right, a multi-homed RDMA host could plausibly want to pin a specific NIC by address, and that address could be DHCP-assigned. Is DHCP actually used on RDMA fabrics in your experience? If so it might be worth allowing a hostname there, even if not for TCP.

FC is moot either way -- host-traddr there is a WWN, which was never DNS-resolvable to begin with.

martin-belanger:
Further reading: https://github.com/linux-nvme/nvme-cli/wiki/Using-%E2%80%90%E2%80%90host%E2%80%90traddr-and-%E2%80%90%E2%80%90host%E2%80%90iface

igaw:
Thanks for double checking. I remembered that we had this discussion but couldn't remember that the result was this wiki entry.

igaw:
Thanks!

PR #3617, opened 2026-07-22
util: add json_object_add_value_bool() macro

jimmunn:
This adds a boolean value into the json object.


igaw:
Thanks!

PR #3618, opened 2026-07-22
gitignore: ignore unit-py build artifacts

jimmunn:
Ignore the __pycache__ and *.pyc files generated by the unit-py tests.


igaw:
Thanks!

PR #3619, opened 2026-07-22
fabrics: load fabrics module after parsing args

jimmunn:
Defer load_nvme_fabrics_module() until after argconfig_parse() in fabrics_discovery() and fabrics_connect().

This is necessary for the new dump-command-metadata command.


igaw:
Thanks!

PR #3620, opened 2026-07-22
util: add optional argconfig parse hook

jimmunn:
Add argconfig_set_parse_hook() to register a hook that, when set, replaces argconfig_parse() entirely and supplies its return value. NULL (the default) keeps normal parsing.


igaw:
Thanks!

PR #3621, opened 2026-07-22
build: add python-jsonschema to Windows UCRT64 deps

jimmunn:
Install mingw-w64-ucrt-x86_64-python-jsonschema in the Windows setup script and CI job so schema validation can run there.

This will be used by the tests for the new dump-command-metadata command.


igaw:
Thanks!

PR #3624, opened 2026-07-23
nvme: add JSON output to config show

martin-belanger:
## Summary
- Add `-o json` support to `nvme config show`. New `print_ops` member `config_conn_list`: the existing text renderer moved from `fabrics.c` into `nvme-print-stdout.c` (`stdout_config_conn_list()`), and a new JSON renderer was added in `nvme-print-json.c` (`json_config_conn_list()`), both `#ifdef CONFIG_FABRICS`-gated following the existing `discovery_log` pattern. `fabrics_config_show()` switches from a bare `OPT_ARGS` list to `NVME_ARGS` to gain `-o`/`--output-format` (it had none before).
- JSON reports each connection's fields as configured, not canonicalized through a TID -- an array of connection objects under `"connections"`, with a nested `"params"` object for per-connection tunables. `hostnqn`/`hostid` are `null` when a persona falls back to the system default, matching the text renderer's existing "don't show a fixed identity the config doesn't actually pin" rule.

## Test plan
- [x] `meson compile` clean, both default and `-Dfabrics=disabled`
- [x] Manual verification against real INI fixtures: text output unchanged, JSON output correct for both default-identity and pinned-identity configs, malformed-INI error path returns a valid JSON error object with no crash

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -1379,108 +1379,17 @@ int fabrics_config_validate(const char *desc, int argc, char **argv)
 	return ret;
 }
 
-/*
- * libnvmf_connect_args_emit() callback for "nvme config-show": print each
- * formatted "--option=value" straight to stdout as part of the running
- * "nvme connect" line.
- */
-static void print_conn_arg(const char *arg, void *user_data)
-{
-	printf(" %s", arg);
-}
-
-static void print_conn_field(const char *name, const char *value)
-{
-	if (value)
-		printf(" --%s=%s", name, value);
-}
-
-/*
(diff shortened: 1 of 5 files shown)


igaw:
Thanks!

igaw:
FWIW,

```shell
$ nvme config show --output-format json
{
  "connections":[]
}
```

martin-belanger:
An empty "connections" is expected if a default config (nvme-fabrics.conf) is missing.

igaw:
yes, that is what I kind of expected :) sorry for not being more verbose.

PR #3625, opened 2026-07-23
doc: add man pages for config show and validate

martin-belanger:
The `config` plugin's `show` and `validate` subcommands had no documentation; only `config convert` did. Cover them the same way, and cross-link all three from cmds-main.txt and each other's SEE ALSO.


igaw:
Thanks a lot!

PR #3626, opened 2026-07-23
fabrics: check registry owner before connect-all/discover

martin-belanger:
## Summary
- Bug fix: `connect-all`/`discover` enumerated a discovery controller's log page regardless of who owns it, letting `70-nvmf-autoconnect.rules`' udev-triggered fan-out act on a device some other orchestrator already manages. `disconnect-all` already checks registry ownership before acting (`disconnect_all_match()`); this aligns `connect-all` with that existing behavior.
- Covers both the `--device=nvmeX` path and a bare `--transport`/`--traddr` invocation that resolves to an existing controller through `libnvmf_discovery()`'s own fallback matcher (transport/traddr/subsysnqn/host identity, including kernel-reported `src_addr` for TCP when `host-traddr`/`host-iface` aren't given) -- the latter needed a new public libnvme function, `libnvmf_lookup_ctrl()`, exposing that same matcher so the CLI can preview which controller (if any) a connection request resolves to before deciding whether to proceed.
- `REGISTRY.md` gets a new "`connect-all`/`discover --device` behavior" section documenting it.

## Test plan
- [x] `meson compile` clean: default, `-Dwerror=true`, `-Dfabrics=disabled`
- [x] `make checkpatch-diff`: 0 errors (2 pre-accepted long string-literal warnings)
- [x] Manual verification with a scratch registry (`--set-options test-base-dir=`): owned device skipped with no `--owner`, proceeds with matching `--owner`, unregistered device proceeds
- [ ] The "no `--device`, matches an existing ctrl" branch is code-reviewed but not live-tested -- needs a real NVMe-oF connection

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -42,6 +42,8 @@
 #include <sys/socket.h>
 #endif
 
+#include <ccan/str/str.h>
+
 #include <libnvme.h>
 
 #ifdef NVME_HAVE_LIBKMOD
@@ -716,6 +718,43 @@ static int fabrics_discovery_config(struct libnvme_global_ctx *ctx,
 
 #define NBFT_SYSFS_PATH		"/sys/firmware/acpi/tables"
 
+/*
+ * A controller's ownership extends to everything done through it, so the
+ * caller's operation may not proceed unless the invoking owner matches
+ * (mirrors disconnect_all_match()'s registry check). No ownerless
+ * exemption -- a mismatched or missing --owner is skipped the same way;
+ * the escape hatch is passing the owner's own identity.
+ *
+ * --force skips the check entirely: it means the caller will never reuse
+ * an existing controller, so there is nothing to check ownership against.
+ *
+ * Returns 0 to proceed, 1 to skip, or a negative errno on a registry
+ * read failure.
+ */
+static int check_ctrl_owner(struct libnvme_global_ctx *ctx,
+			     struct libnvmf_context *fctx,
+			     const char *owner, bool force)
+{
+	__cleanup_free char *reg_owner = NULL;
+	int ret;
+
+	if (force)
+		return 0;
+
+	ret = libnvmf_get_owner_from_fctx(ctx, fctx, &reg_owner);
+	if (ret)
+		return ret;
+	if (!reg_owner)   /* no owner in registry */
+		return 0;
+
+	if (owner && streq(reg_owner, owner))
+		return 0;
+
+	nvme_show_error("owned by '%s'; skipping -- owner handles discovery",
+			 reg_owner);
+	return 1;
+}
+
 int fabrics_discovery(const char *desc, int argc, char **argv, bool connect)
 {
 	__cleanup_free char *hnqn = NULL;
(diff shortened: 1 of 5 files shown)


libnvme/src/nvme/fabrics.h
```diff
@@ -535,6 +535,23 @@ int libnvmf_context_set_io_queues(struct libnvmf_context *fctx,
...
+ * Searches @h's subsystems for a controller matching @fctx's transport,
+ * address, subsysnqn and host identity -- the same match libnvmf_connect()
+ * and libnvmf_discovery() use internally to decide whether to reuse an
+ * existing connection instead of creating a new one. Does not itself
+ * connect, disconnect, or otherwise change anything; a caller may use it to
+ * preview which controller (if any) an operation on @fctx would resolve to.
+ *
+ * Return: the matching controller, or NULL if none exists.
+ */
+libnvme_ctrl_t libnvmf_lookup_ctrl(libnvme_host_t h,
```
igaw: I've remove the lookup functions on purpose, because the caller has different requirements and then we have to implement policy in the library. That's why there is non in the API left at the moment. 

Also if there is a lookup function it should be `libnvme_lookup_ctrl` but this function is really not for the public API.

fabrics.c
```diff
@@ -716,6 +718,62 @@ static int fabrics_discovery_config(struct libnvme_global_ctx *ctx,
...
+ * An explicit --device is checked by name directly, since
+ * libnvmf_discovery() reuses that device regardless of whether it
+ * matches @fctx's other parameters. Without --device, unless --force,
+ * libnvmf_lookup_ctrl() previews the same match libnvmf_discovery()
+ * falls back to.
+ *
+ * Returns 0 to proceed, 1 to skip, or a negative errno on a registry
+ * read failure.
+ */
+static int check_discovery_owner(struct libnvme_global_ctx *ctx,
```
igaw: This seems to be a function which could be part of the library, this would avoid to expose the lookup function.

martin-belanger: I like this idea, and I'm working on something even more generic. Still woking out the details. Will keep you posted.

libnvme/src/nvme/fabrics.h
martin-belanger: fixed

fabrics.c
martin-belanger: fixed

martin-belanger:
Addressed both review comments. Added two new public APIs. Same functionality, different signature. Need the first signature  for nvme-discoverd and the second for the CLI.

- `libnvmf_get_owner_from_tid(ctx, tid, &owner)`
- `libnvmf_get_ctrl_owner(ctx, fctx, &owner)`

`libnvmf_lookup_ctrl()` is gone entirely -- back to the internal-only `lookup_ctrl()` it was before this PR.

`check_ctrl_owner()` on the CLI side (renamed from `check_discovery_owner()` -- nothing discovery-specific about it) is down to just the actual policy now: bail out on `--force`, one library call, compare the result to `--owner`, log and skip on a mismatch. Everything mechanical moved into libnvme.

igaw:
Looks good. The commit message is outdated and the documentation should be more in technical language (man page style) please.

martin-belanger:
Fixed function headers and commit message

igaw:
Thanks a lot!

PR #3627, opened 2026-07-23
libnvme: write registry entry on an already-connected controller

martin-belanger:
## Summary
- Bug fix: `libnvmf_connect()` only wrote the ownership registry entry on a fresh connect. Its two "already connected" exit paths -- a tree-level dup match, and a kernel-race rescan -- skipped it, so `nvme connect --owner X --idempotent` against an already-connected controller left the registry stale. The registry must be authoritative "regardless of who made the initial connection" (RFC's own spec for this feature), and nvme-discoverd's `--idempotent`-based adoption of pre-existing connections (NBFT, FC boot) depends on it.
- Added a small `ctrl_instance()` helper (parses `"nvme3"` -> `3` from a ctrl's name, matching the existing `sscanf(..., "nvme%d", ...)` pattern in `scan-linux.c`) and reused the existing `registry_update_on_connect()` in both paths, unconditionally -- so the registry reflects the current invocation's `--owner` regardless of which path resolved the connection. No CLI-side change needed: `libnvme_set_owner()` already runs before `libnvmf_connect()` regardless of `--idempotent`.

## Test plan
- [x] `meson compile` clean: default, `-Dwerror=true`, `-Dfabrics=disabled`
- [x] `update-accessors`: no drift (no public API surface changed)
- [ ] Not live-tested -- exercising either already-connected exit path needs a real NVMe-oF connection

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -1441,6 +1441,22 @@ static int __nvmf_supported_options(struct libnvme_global_ctx *ctx)
 	return 0;
 }
 
+/* Parse the kernel instance number out of a ctrl's name ("nvme3" -> 3). */
+static int ctrl_instance(libnvme_ctrl_t c)
+{
+	int instance = -1;
+	const char *name;
+
+	if (!c)
+		return instance;
+
+	name = libnvme_ctrl_get_name(c);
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3629, opened 2026-07-23

LJaved-IBM:
@sph irrelevant. Avoid this by setting @sph to NULL before
calling wdc_get_dev_mng_log_entry() to make the intent clear.

As @sph is now set to NULL before it's passed to wdc_get_dev_mng_log_entry(),
remove the redundant NULL pointer check from wdc_get_dev_mng_log_entry().

### DIFF
--- a/plugins/wdc/wdc-nvme.c
+++ b/plugins/wdc/wdc-nvme.c
@@ -2390,11 +2390,6 @@ bool wdc_get_dev_mng_log_entry(__u32 log_length, __u32 entry_id,
 	__u32 offset = 0;
 	struct wdc_c2_log_subpage_header *p_next_log_entry = NULL;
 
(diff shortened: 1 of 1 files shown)


igaw:
The commit message is still the same as the last time. Drop the `@` from it. Something like this:

```

The get_dev_mgmt_log_page_lid_data() function initializes sph to
a pointer derived from data before passing it to
sph irrelevant. Avoid this by setting sph to NULL before
calling wdc_get_dev_mng_log_entry() to make the intent clear.

As sph is now set to NULL before it's passed to wdc_get_dev_mng_log_entry(),
remove the redundant NULL pointer check from wdc_get_dev_mng_log_entry().
```

I'll fix it up myself.

igaw:
Thanks!

PR #3630, opened 2026-07-23
libnvme: fix SIGSEGV in __libnvme_transport_handle_open_mi

chen-gz:
When opening an MI connection over MCTP via __libnvme_transport_handle_open_mi, the returned endpoint pointer 'ep' was not saved into the transport handle 'hdl->ep', and the handle was not added to the endpoint's controllers list.

This led to a SIGSEGV later when attempting to submit commands using the handle, as hdl->ep remained NULL.

Fix this by correctly assigning hdl->ep, hdl->id, and adding the handle to the controllers list.

### DIFF
--- a/libnvme/src/nvme/mi.c
+++ b/libnvme/src/nvme/mi.c
@@ -91,6 +91,10 @@ int __libnvme_transport_handle_open_mi(struct libnvme_transport_handle *hdl, con
 	if (!ep)
 		return -EINVAL;
 
+	hdl->ep = ep;
+	hdl->id = ctrl_id;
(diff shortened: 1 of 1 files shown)


igaw:
Ooops, this went wrong when introducing the transport handle abstraction. Thanks for fixing it.

igaw:
Unfortunately, I don't have a MI setup for testing, so I can't really test if all works fine. If you could test a bit that would be awesome. Maybe run it with valgrind etc to catch any leaks or UAF. We have some ideas to improve the situation for UAF but currently we don't have anything in place.  

Thanks a lot!

Issue #3598, opened 2026-07-21
generate accessors for global context


igaw:
At least generate the accessors for these members:

```c
	/* global options to steer libnvme behavior or overwrite defaults */
	bool force_4k;
	bool mi_probe_enabled;
	bool ioctl_probing;
	char *test_base_dir;
	char *test_sysfs_dir;
```

PR #3632, opened 2026-07-24
libnvme: accessor generator naming-prefix override (#3598)

martin-belanger:
## Summary
- Follow-up on the #3598 discussion:

- Commit 1: adds the `prefix=` key. `libnvme_global_ctx`'s `hostnqn`/`hostid` accessors now come out as `libnvme_set_hostnqn()`/`libnvme_set_hostid()` instead of `libnvme_global_ctx_set_*`.
- Commit 2: four hand-written `libnvme_global_ctx` setters (`force_4k`, `dry_run`, `ioctl_probing`, `mi_probe_enabled`) were just field assignments with no getters, so replaced them with generated accessors. `libnvme_set_probe_enabled()` becomes `libnvme_set_mi_probe_enabled()` to match its field.

## Test plan
- [x] `meson compile` clean on both commits: default, `-Dwerror=true`, `-Dfabrics=disabled`
- [x] `update-accessors`: no drift
- [x] `make checkpatch-diff`: 0 errors, 2 known warnings (generator annotation line, multi-commit duplicate signoff)


igaw:
Thanks a lot!

PR #3635, opened 2026-07-24
nvme: Fix get-log xfer-len parameter handling

brgoing:
Passes the cfg.xfer_len value to libnvme_get_log instead of the hard-coded NVME_LOG_PAGE_PDU_SIZE. The xfer-len parameter was already defined and validated, but the value wasn't being used.


igaw:
Thanks!

PR #3636, opened 2026-07-24
shared: add a small static utility library

martin-belanger:
## Summary

Adds a small `shared/` static library for utility code with no natural home in any one of nvme-cli, libnvme, python-bindings, or nvme-discoverd -- string helpers, a growable pointer-array, a boolean parser, and `mkdir -p`. Modeled loosely on systemd's `src/basic/`: one topic per file, header-only where the whole thing is a static inline function, a `.c`/`.h` pair where it isn't. No umbrella header, matching the precedent already set by libnvme's own `nvme/cleanup.h`.

**This PR only adds the library. Nothing consumes it yet.** No existing file is touched other than one line in the root `meson.build` wiring up the new `subdir('shared')`. The intent is to get the framework itself reviewed and merged first, since it's a small, low-risk, purely additive diff; a follow-up PR will then port to the existing code base.

## What's included

- `shared/string-util.h` -- `streq0()`, `streqcase0()`, `xstrdup()`, `trim()` (static inline).
- `shared/parse-util.h`/`.c` -- `parse_bool()`, the systemd `parse_boolean()` convention (1/yes/y/true/t/on, 0/no/n/false/f/off).
- `shared/array-util.h`/`.c` -- `struct ptrarray` / `ptrarray_append()` / `ptrarray_free()`, a generic growable pointer array.
- `shared/fs-util.h`/`.c` -- `mkdir_p()`.
- `shared/meson.build` -- builds `libshared` (static), declares `shared_dep`.
- One line in the root `meson.build`: `subdir('shared')` alongside the existing `subdir('ccan')`.

## Test plan

- [x] `meson compile` clean (default options)
- [x] `meson test`: 81/83 (2 expected-fail, unchanged from master)
- [x] checkpatch: clean
- [x] Confirmed `libshared.a` builds standalone with nothing yet linking against `shared_dep`

### DIFF
--- a/shared/array-util.h
+++ b/shared/array-util.h
@@ -1,0 +1,38 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+/*
+ * This file is part of nvme-cli.
+ * Copyright (c) 2026 Dell Technologies Inc. or its subsidiaries.
+ *
+ * Authors: Martin Belanger <martin.belanger@dell.com>
+ */
+#pragma once
+
+#include <stddef.h>
+
+/*
+ * A growable array of untyped pointers, doubling capacity (starting at 8)
+ * as items are appended. Neither owns nor frees the pointed-to items --
+ * only the backing array itself (see ptrarray_free()).
+ */
+struct ptrarray {
+	void **items;
+	size_t len;
+	size_t cap;
+};
+
+/*
+ * Append item to a, growing the backing array as needed.
+ *
+ * To build a NULL-terminated array (e.g. to hand back to a caller that
+ * expects one), append every real item, then append NULL last -- that
+ * final call reserves the extra slot itself, same as any other append.
+ *
+ * Return: 0 on success, -ENOMEM on allocation failure.
+ */
+int ptrarray_append(struct ptrarray *a, void *item);
+
+/*
+ * Free the backing array only, leaving *a zeroed. Caller is responsible
+ * for freeing each item first if the array owns them.
+ */
+void ptrarray_free(struct ptrarray *a);
--- a/meson.build
+++ b/meson.build
@@ -552,6 +552,7 @@ substs.set('LIBNVME_SO_VERSION', libnvme_so_major)
 doc_tgts = []
 
 subdir('ccan')      # declares: ccan_dep
+subdir('shared')    # declares: shared_dep
 subdir('libnvme')   # declares: libnvme_dep
 
 if want_nvme
--- a/shared/array-util.c
+++ b/shared/array-util.c
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * This file is part of nvme-cli.
+ * Copyright (c) 2026 Dell Technologies Inc. or its subsidiaries.
+ *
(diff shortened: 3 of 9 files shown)


shared/array-util.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+#pragma once
+
+#include <stddef.h>
+
+/*
+ * A growable array of untyped pointers, doubling capacity (starting at 8)
+ * as items are appended. Neither owns nor frees the pointed-to items --
+ * only the backing array itself (see ptrarray_free()).
+ */
```
igaw, line 16: What about using darray from ccan instead?

igaw:
What's the long term plan with shared? Do you aim for replacing `util`?

shared/array-util.h
martin-belanger: Asked Claude to analyze `darray`. A few things push me toward keeping `ptrarray`:

- `darray_growalloc`/`darray_realloc` store `realloc()`'s result into `.item` unconditionally, so on OOM `.item` becomes `NULL` and `darray_append` writes through it -- there's no way for a caller to detect or handle the failure. `ptrarray_append()` returns `-ENOMEM` and every current call site propagates it.
- The shape of the struct `ptrarray` fits better with the use cases than `darray`. I plan to use `ptrarray` for `nvme-discoverd` and the exclusion list, and `ptrarray` fits better those use cases than `darray`.
- darray is not currently in our `ccan` copy. We would need to pull it from upstream.

Happy to revisit if there's a case `darray` covers that `ptrarray` doesn't, but as-is I'd rather keep the explicit error return.

martin-belanger:
> What's the long term plan with shared? Do you aim for replacing `util`?

No -- `util/` is CLI-specific (argconfig, json wrapper, base64, table, dashboard, ...) and nothing outside the CLI touches it, so there's nothing to replace there.

The one exception is `util/cleanup.h`, which is byte-for-byte the same `__cleanup`/`__cleanup_free` boilerplate as `libnvme/src/nvme/cleanup.h`. That duplication predates this PR and is exactly the kind of thing `shared` is for. Plan is to move it in as a follow-up once this lands, not as part of this PR -- wanted the framework in first before touching either existing file.

igaw:
Sure, not all of util is for shared, but there a a few candidates IMO:

At least these to APIs could be shared between libnvme and nvme-cli.
- `base64`
- `crc32`

Maybe there are some more function inside the headers which could/should be shared (`*_to_cpu` etc).

A shared cleanup header would be indeed nice to have. Also the compiler attribute could be a candidate.

martin-belanger:
Agreed. I can do an inventory of the candidates and move them into shared -- base64 and crc32 are definitely good ones. Like I said, I wanted the infrastructure in place first (this PR), then reshuffle existing things into it.

shared/array-util.h
igaw: Fair enough. Though one thing `darray` does better is the type checks. Could you steal this idea from it? Also tests cases for such things would be good.

martin-belanger: Good idea to add type checks. It should be possible to steal the idea.

igaw:
As discussed on the matrix channel, let's merge this version and improve from here.

PR #3639, opened 2026-07-24
sfx-nvme: fix dead assignment in sfx_dump_evtlog()

sahmed-ibm:
The sfx_dump_evtlog() function calls nvme_dump_evtlog() and stores the return value in err. However, the function unconditionally returns 0 instead of returning err, so the result of the call is never used. Any error returned by nvme_dump_evtlog() is silently discarded, causing the caller to always see success even when the operation failed.

Return err instead of 0 so that errors are correctly propagated to the caller.


igaw:
Thanks!

PR #3643, opened 2026-07-27
libnvme: validate NBFT descriptor ranges

### LINKED ISSUE #3641: NBFT descriptor list bounds check can wrap and crash connect-all
## Summary

We found a crash in the NBFT table reader. NBFT is the ACPI table through which firmware describes remote NVMe storage to the operating system.

A table can place the Discovery Descriptor list offset near the end of the 32-bit range. The reader adds the list size before checking the table boundary. That addition wraps to a small value. The check passes, then `nvme connect-all --nbft` reads beyond the table and terminates with `SIGSEGV`.

## Affected revision

The crash is present on nvme-cli master at `962163f836cfe29be90a1766585da21e6f483210`.

The result was confirmed with an ordinary release build on 64-bit Arm Ubuntu 24.04. No sanitizer or preload library was present. The isolated host had no NVMe fabrics device.

## Reproduction

Pull request #3643 contains the 700-byte NBFT table used for this result. Its SHA-256 is `d20406de6a987a5694cedaf3c0150bc10900ed52a311f98e8e494429518507c6`.

The production command is:

```sh
nvme connect-all --nbft --nbft-path=<directory>
```

All three runs terminated with `SIGSEGV`.

## Impact

This command is commonly started during early boot when the root filesystem resides on remote NVMe storage. A crash can stop the root connection and prevent the host from booting.

Firmware normally supplies the table. A virtual machine monitor can supply it as well. The command also accepts a caller-selected table directory.

## Cause

The range check adds an untrusted 32-bit offset to a descriptor list size. The result can wrap before comparison with the table length. Descriptor lengths also need a lower bound because the reader accesses a full C structure for each entry.

## Suggested fix

Check the offset against the table length first. Compare the list size with the remaining bytes afterward. Reject any descriptor length smaller than the structure read by the parser.

Pull request #3643 adds those checks and regression tests.

flaviens:
## Summary

NBFT descriptor list checks previously added a 32-bit offset to the list size. A large offset could wrap before the result was compared with the table length. The parser would then read beyond the supplied table.

This change checks each offset before subtracting it from the table length. It also verifies that every descriptor length can hold the structure consumed by the reader. Entry stepping follows the descriptor length recorded in the Control Descriptor through one shared helper. A zero list offset is rejected.

An invalid Control Descriptor now causes the table to be rejected. This prevents callers from receiving a partly initialized NBFT object.

## Result

The wrapped Discovery Descriptor case terminated with `SIGSEGV` in all three runs before this change. It is rejected cleanly afterward.

The same check protects the Security Profile list used by pull request #3528. Its wrapped-offset case is also rejected cleanly. A second case with an invalid Control Descriptor no longer reaches a null Host ID access.

## Tests

Three malformed tables cover the Discovery list wrap and the Security list wrap. The third table has an invalid Control Descriptor. Each test requires a normal parser rejection and fails if the process receives a signal.

The production command completed without a signal in three runs for every covered case. The added tests pass in a release build and under AddressSanitizer with UndefinedBehaviorSanitizer.

Checkpatch reports zero errors and zero warnings.

Closes #3641


igaw:
Very nice! This looks good to me. @tbzatek any objections?

igaw:
Thanks a lot! Really nice work!

libnvme/test/nbft/meson.build
```diff
@@ -105,3 +111,14 @@ foreach table: tables_bad
         )
     endif
 endforeach
+
+foreach table: tables_rejected
+    test(
+        'libnvme - @0@'.format(table),
+        files('nbft-reject.sh'),
```
tbzatek, line 118: Instead of introducing new test helper, could this be done through meson test [expected_exitcode](https://mesonbuild.com/Reference-manual_functions_test.html#test_expected_exitcode)?

tbzatek:
Good catch! I was actually thinking about fuzzing binary NBFT tables using AI - check my #3670 that were created purely based on the Spec pdf file. I haven't attempted the fuzzing yet, feel free to submit more testing tables.

libnvme/test/nbft/meson.build
igaw: That would pull the min meson requirement to 1.11.0. Currently the min version is at 0.62.0.

flaviens: Agreed. The wrapper keeps compatibility with the current Meson 0.62 minimum. I will leave it in place.

tbzatek: Ah, missed that! Thanks for pointing that out.

PR #3644, opened 2026-07-27
libnvme: harden NBFT interface references

### LINKED ISSUE #3642: NBFT connection handling can dereference missing interface entries
## Summary

We found three ways for NBFT connection handling to dereference an absent Host Fabric Interface. NBFT is the ACPI table through which firmware describes remote NVMe storage to the operating system.

The first case leaves a hole in the Secondary HFI array while counting the entry as active. The second accepts a Discovery Descriptor whose HFI index cannot be resolved. The third reaches the end of the subsystem list, then reads through that completed iterator while reporting a missing interface for an unlinked Discovery Descriptor.

Each case reaches `nvme connect-all --nbft` and terminates with `SIGSEGV`.

## Affected revision

The crashes are present on nvme-cli master at `962163f836cfe29be90a1766585da21e6f483210`.

The checks used an ordinary release build on 64-bit Arm Ubuntu 24.04. No sanitizer or preload library was present. The isolated host had no NVMe fabrics device.

## Reproduction

Pull request #3644 contains the three NBFT tables and a command-level regression test.

| Case | Size | SHA-256 | Result |
| --- | ---: | --- | --- |
| Secondary HFI array hole | 641 bytes | `00f3fb551668b431644e634ec0fa5111aca38519ca4cf892b7aed526d781d707` | 3/3 `SIGSEGV` |
| Unresolved Discovery HFI | 700 bytes | `2f3587c4d636affaab12be0e7f88a0b96ff971ba47391e7813415837b6f8a980` | 3/3 `SIGSEGV` |
| Completed subsystem iterator | 700 bytes | `424049b12eb9de0e5f426d40858f7743c6c7867a5eed994428bed5c05f33b482` | 3/3 `SIGSEGV` |

The production command is:

```sh
nvme connect-all --nbft --nbft-path=<directory>
```

## Impact

This command is commonly started during early boot when the root filesystem resides on remote NVMe storage. A crash can prevent the root connection and leave the host unable to boot.

The completed-iterator case can occur with a valid Discovery Descriptor when a firmware interface is absent or has a different operating system name.

## Suggested fix

Store resolved Secondary HFI entries next to each other. Skip unresolved entries without increasing the active count.

Reject a Discovery Descriptor when its HFI cannot be resolved. Keep defensive checks in the connection code.

Use the Discovery Descriptor index in its missing-interface message.

Pull request #3644 applies these changes and adds a production-command regression test.

flaviens:
## Summary

NBFT connection handling assumed that every active interface entry had resolved during parsing. Three cases could violate that assumption and terminate `nvme connect-all --nbft`.

This change compacts Secondary HFI entries after resolution. Missing entries are skipped without increasing the active count.

Discovery Descriptors with no resolved HFI are dropped during parsing. The connection code also checks interface pointers before use.

The missing-interface message for an unlinked Discovery Descriptor now uses that descriptor index. Each table iteration resets the host ID before reading the table value.

## Result

Each of the three supplied tables terminated with `SIGSEGV` in all three runs before this change. After the change, the production command completed without a signal in all nine runs.

## Tests

A command-level regression test runs the production NBFT connection command with each table. The test uses the supported test sysfs setting. It fails if the command receives a signal.

The regression test passes in a release build and under AddressSanitizer with UndefinedBehaviorSanitizer.

Checkpatch reports zero errors and zero warnings.

Closes #3642


libnvme/src/nvme/fabrics.c
```diff
@@ -3203,7 +3203,7 @@ __libnvme_public int libnvmf_discovery_nbft(struct libnvme_global_ctx *ctx,
 
 		if (fctx->hostid)
 			hostid = fctx->hostid;
-		else if (*entry->nbft->host.id) {
+		else if (entry->nbft->host.id && *entry->nbft->host.id) {
 			ret = libnvme_uuid_to_string(entry->nbft->host.id, uuid);
 			if (!ret)
```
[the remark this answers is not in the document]
flaviens: Fixed in 7221ee0c7. Each table iteration now initializes `hostid` from the caller value before considering the table UUID. A table without either value therefore uses the default host ID.

tests/nvme_nbft_test.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
```
igaw, line 1: Could also add your copyright statement here?

tests/nvme_nbft_test.py
```diff
@@ -0,0 +1,60 @@
...
+            str(nvme),
+            f"--set-options=test-sysfs-dir={sysfs_root}",
+            "connect-all",
+            "--nbft",
+            f"--nbft-path={table_dir}",
+            "--hostnqn=nqn.2014-08.org.nvmexpress:uuid:"
+            "3a3f9475-2f25-48d3-9025-0c2f288a4e88",
+            "--hostid=3a3f9475-2f25-48d3-9025-0c2f288a4e88",
+        ]
+        result = subprocess.run(
```
igaw, line 38: Ideally, this would be based on the class `TestNVMeBase`. This would hook the tests case into the existing framework (with the logging etc). 

Though there is still the problem to solve how to hand in arguments. And I see the tables are handed in as arguments.  There is a config file for passing in arguments but I was told this is not a good thing to have. 

I suppose we got with this version here and figure out the config file eventually.

flaviens: Agreed. I kept the standalone runner for this change because it passes the table files directly. We can move it to `TestNVMeBase` once the test framework has an accepted way to supply those inputs.

tests/nvme_nbft_test.py
flaviens: Added in 636bfb2c1.

igaw:
Thanks a lot!

PR #3645, opened 2026-07-27
virtium: remove erroneous (float) cast in vt_save_smart_to_vtview_log()

fhassan-ibm:
Multiplying a double by (float)HOUR_IN_SECONDS unnecessarily downcasts the integer constant to a lower-precision float before the multiplication, introducing a spurious type conversion. The result is then assigned to a long, which Coverity flags as an overflowed integer argument.

Remove the (float) cast so the multiplication is double * int, which is well-defined and loses no precision before the long assignment.

Fixes: Coverity CID 557484 (Overflowed integer argument)


igaw:
Thanks!

PR #3647, opened 2026-07-27
wdc: free dssd_specific_ver when smart_log_ver < 3

fhassan-ibm:
In wdc_show_cloud_smart_log_json(), dssd_specific_ver is always allocated via json_create_object() regardless of smart_log_ver. However, it is only added to the root JSON object (and therefore owned and freed with it) when smart_log_ver >= 3.  When smart_log_ver < 3 the object is neither added nor freed, causing a resource leak on every such call.

Add an else branch to free dssd_specific_ver when the version check fails so the allocated object is always released.

Fixes: Coverity CID 557510 (Resource leak)


igaw:
Thanks!

PR #3648, opened 2026-07-27
ibm: fix missing break in show_ibm_smart_log() case 0x00f5

fhassan-ibm:
In show_ibm_smart_log(), the switch statement that dispatches on entry->attr is missing a break at the end of case 0x00f5. Without it, execution falls through into the default branch after printing the PCIe error counters, which is unintended behaviour.

Add the missing break to terminate case 0x00f5 cleanly so that only the default branch handles unrecognised attribute values.

Fixes: Coverity CID 557383 (Out-of-bounds read)

### DIFF
--- a/plugins/ibm/ibm-nvme.c
+++ b/plugins/ibm/ibm-nvme.c
@@ -214,6 +214,7 @@ static void show_ibm_smart_log(struct nvme_ibm_additional_smart_log *smart, cons
 				le32_to_cpu(smart->device_pcie_received_errors.split_raw2.upper));
 			printf("PCIe Recd Transitions to Recoveries : %"PRIu32"\n",
 				le32_to_cpu(smart->device_pcie_received_errors.split_raw2.lower));
+			break;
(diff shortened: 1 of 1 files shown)


igaw:
the commit description doesn't match with the code. are you sure you are fixing the right switch statement?

fhassan-ibm:
The commit description is misleading — I apologize for the confusion. The only issue is that case 0x00f5 falls through into default: unintentionally. The break simply prevents that fall-through so default: only handles genuinely unrecognised attribute values. There is no out-of-bounds risk — that part of the description was incorrect.

igaw:
Could you update the commit message accordingly then? Not sure if the web  UI shows it correctly, but it looks like the indent of the break is not correct.

fhassan-ibm:
Updated — fixed the indentation of the break statement and corrected the Fixes: tag in the commit message.

igaw:
Thanks!

PR #3649, opened 2026-07-27
sndk plugin: Fix vs-smart-add-log for NVMe OF

jeff-lien-sndk:
The current sandisk plugin vs-smart-add-log command is failing with "Unknown device id" in the NVMe OF environment.  This change will resolve that failure.


igaw:
Thanks!

PR #3650, opened 2026-07-27
Add support for global configuration

### LINKED ISSUE #3593: add global config file
The configuration for the fabrics is done via INI files now. Add support also for a global configuration for global options, e.g. `no-ioctl-probing`.

dwsuse:
Allow the user to set the global configuration via a config file.

Fixes: #3593


shared/ini.c
```diff
@@ -0,0 +1,168 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
```
martin-belanger: Should this be LGPL-2.1-or-later as you pointed out to me earlier. 😄

igaw: claude changed the license own it's own. tsts

global-config.c
```diff
@@ -0,0 +1,145 @@
...
+static int parse_uint(const char *value, unsigned int *out)
+{
+	unsigned long v;
+	char *end;
+
+	if (!*value)
+		return -EINVAL;
+
+	errno = 0;
+	v = strtoul(value, &end, 0);
```
martin-belanger, line 40: I think the intent here is to parse an unsigned. However, if a signed negative number is passed in, this function will return a very large unsigned number.

For example:
```
strtoul("-5")  = 18446744073709551611, errno=0, *end='\0'
strtoul("-1")  = 18446744073709551615, errno=0, *end='\0'
```

global-config.c
```diff
@@ -0,0 +1,145 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * This file is part of libnvme.
```
martin-belanger: Nit. The file is not part of libnvme

unit/meson.build
```diff
@@ -59,3 +59,28 @@ test_argconfig_parse = executable(
...
+
+test_global_config = executable(
+    'test-global-config',
+    ['test-global-config.c', '../global-config.c'],
+    dependencies: [
+        config_dep,
+        ccan_dep,
+        libnvme_dep,
+        shared_dep,
+        json_c_dep,
```
martin-belanger: I think json_c_dep is not needed by this test fixture

unit/test-ini.c
```diff
@@ -17,14 +17,12 @@
 #include <string.h>
 #include <unistd.h>
 
-#include <nvme/lib.h>
-
-#include "nvme/ini.h"
+#include "../shared/ini.h"
```
martin-belanger: nitpick: `#include "../shared/ini.h"` here, but `#include <ini.h>` in `global-config.c`.

meson defines `shared_dep` which sets the include dir, so `#include <ini.h>` would work here.  No need for the relative ../ path

unit/meson.build
dwsuse: it looks like it is pulled in via the `#include ../nvme.h`. hmm, maybe it's better to make this not depending on those headers...

martin-belanger:
Looks good. I just had a few nitpicks.

unit/meson.build
dwsuse: global-ini.c pulls in nvme.h and this depends on json-c and libnvme. Can't get really rid of this.

martin-belanger: Agreed. I did not see the full picture.

PR #3651, opened 2026-07-27
netapp-nvme: fix uninitialized pointer in netapp_smdevices()

sahmed-ibm:
The netapp_smdevices() function declares the devices pointer used to receive the scandir() output, but does not initialize it before passing its address to scandir().

If execution reaches the scandir() call with devices holding an indeterminate stack value, the behavior is undefined.

Initialize devices to NULL so it holds a defined value before scandir() writes to it through the pointer.


igaw:
This commit message is wrong.

The namelist parameter is an output parameter. On success, scandir() allocates an array and stores its address in *namelist. On failure, it returns -1. There is no requirement in POSIX that *namelist be initialized before the call. This is just to shut up the analyzer.

PR #3652, opened 2026-07-27
sfx-nvme: fix resource leak in sfx_status()

sahmed-ibm:
The sfx_status() function opens a file descriptor for the block device path to retrieve the sector size and capacity via ioctl. The descriptor is never closed after the ioctls complete, and fd is silently overwritten on the next open() call.

This results in a file descriptor leak on every invocation of sfx_status() when the device has a readable block node.

Close fd after the ioctls to prevent the leak.


igaw:
Thanks!

PR #3654, opened 2026-07-27
memblaze: fix stack overflow in perf-stats-print-x

jimmunn:
Heap-allocate the performance stats log buffer to avoid overflowing the stack on platforms with a small default stack size.

This crash is what was causing the Windows build to fail for PR #3603.


igaw:
Thanks!

PR #3655, opened 2026-07-27
shared: consolidate cross-project duplicates and adopt the shr_ namespace

martin-belanger:
## Why this is 9 commits

This started as the small mechanical follow-up promised in #3636's own commit message: port each project's duplicate `base64`/`crc32`/`parse_bool`/`streq0`-style helpers over to `shared/`, which landed empty ("nothing consumes this yet").

It grew past that. Once real code started including `shared/` headers, every symbol in there needed a `shr_` prefix to avoid colliding with the rest of the tree (including libc itself). The rename alone touches 40 files. Then the rename needed test coverage, since none of this had any before. Each commit is small, self-contained, and builds+tests+checkpatches clean on its own, so it should review fine one at a time even though the total looks big.

## Why `shared/` isn't restructured as its own project library

This doesn't add per-subdir license text or otherwise reshape `shared/` into an independent library layout. The project's existing `LICENSES/` directory (REUSE convention, SPDX identifiers per file) already tracks every license in the tree, including `LicenseRef-Gary-S-Brown-CRC32` for the vendored crc32 code -- so a separate license structure scoped to `shared/` alone wouldn't add anything.

## Commits, in review order

1. **`shared: relicense to LGPL-2.1-or-later`** -- `shared/` was GPL from #3636; libnvme (LGPL) needs to link it, so this comes first, before anything moves in.
2. **`shared: fix mkdir_p() silently truncating long paths`** -- an unrelated bug I found in already-merged code while reading `shared/fs-util.c`; fixed separately so it's not buried in the move commit below.
3. **`shared: move base64, crc32, and misc utility functions`** -- the actual dedup: consolidates near-duplicate/independently-sourced copies from `util/` and `libnvme/src/nvme/` into one, verified byte-identical or output-identical first.
4. **`shared: dedup cleanup.h boilerplate`** -- `util/cleanup.h` and `libnvme/src/nvme/cleanup.h` carried identical macro boilerplate; moved to `shared/cleanup.h`, each side keeps only its own additions.
5. **`shared: add PTRARRAY_DEFINE() for type-checked ptrarray wrappers`** -- the macro Daniel asked for on #3636, so a consumer gets a type-checked array instead of a load of `void *`.
6. **`shared: rename everything to the shr_ namespace`** -- every `shared/` symbol gets a `shr_` prefix; the reason the diff looks big.
7. **`shared: add test coverage for array-util, base64, crc32`** -- none of this had tests before; adds them now that the symbols are stable.
8. **`shared: add README`** -- what `shared/` is for and its conventions.
9. **`libnvme: drop test/ioctl's own freep(), use shared/cleanup.h`** -- a third copy of the same cleanup boilerplate, found late, outside the two files commit 4 targeted.

## Test plan

- `meson compile` clean with `-Dwerror=true`.
- `meson test`: 85 passing, 2 expected-fail (pre-existing, unrelated), 0 failures.


igaw:
Thanks!

PR #3665, opened 2026-07-28
nbft-plugin: fix resource leak in show_nbft()

sahmed-ibm:
The show_nbft() function calls libnvmf_nbft_read_files() to build a linked list of NBFT file entries pointed to by head. The call to libnvmf_nbft_free() to release that list is guarded by the condition that ret is zero and head is non-NULL.

If libnvmf_nbft_read_files() partially populates head before returning an error, the already-allocated entries are never freed, resulting in a resource leak on the error path.

Move libnvmf_nbft_free() outside the conditional block so head is always freed. libnvmf_nbft_free() already handles a NULL head safely, so the unconditional call is correct in all cases.


igaw:
Thanks!

PR #3666, opened 2026-07-28
discoverd: parse discoverd.conf with the shared ini parser

martin-belanger:
`config_load()` had its own line-by-line INI parser, predating the shared one `shr_ini_parse_file()` (shared/ini.c) now provides.


igaw:
Thanks!

PR #3667, opened 2026-07-28
nvme: fix to check sanitize status error

ikegami-t:
Previously the status check error not returned by the sanitize command.


ikegami-t:
Fix for the PR: #3438 comment https://github.com/linux-nvme/nvme-cli/pull/3438#discussion_r3643646383.

igaw:
Thanks!

PR #3668, opened 2026-07-28
util: fix memory leak in read_binary_file()

bellis-ibm:
The function returns NULL on all paths including success, leaking the allocated buffer. A duplicate unreachable goto also prevented the file from being closed on allocation failure.

Return buffer at the cleanup label and fix the goto target to close the file on the error path.

### DIFF
--- a/util/utils.c
+++ b/util/utils.c
@@ -67,25 +67,23 @@ char *hex_to_ascii(const char *hex)
 unsigned char *read_binary_file(char *data_dir_path, const char *bin_path,
 				long *buffer_size, int retry_count)
 {
-	char *file_path = NULL;
-	FILE *bin_file = NULL;
+	__cleanup_free char *file_path = NULL;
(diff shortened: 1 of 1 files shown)


util/utils.c
igaw: I would use `__cleanup_file` and get rid of the gotos.

igaw: while looking at this, could we just use `asprintf` for creating the `file_path`?

igaw: if we would use a `strdup` here we could also use `__cleanup_free` on `file_path` and use early returns everywhere and this would make the error handling way simpler and less error prone.

igaw:
Sorry while looking at your patch (which is correct) I realized this function should be updated to use the `__cleanup` helper to make the error handling way simpler and less error prone.

igaw:
Nice cleanup! Thanks!

PR #3669, opened 2026-07-28
libnvme: return libnvmf_tid_from_fields() as int, not a pointer

martin-belanger:
libnvmf_tid_from_fields() returned NULL on any failure, so a caller could not tell a validation error (-EINVAL, non-numeric traddr/host_traddr) from an allocation failure (-ENOMEM). Two internal callers actually got this wrong: fabrics.c's build_conn_tid() reported every failure as -EINVAL, and libnvme/src/nvme/fabrics.c's nvmf_sanitize_addrs() and the owner lookup in libnvmf_get_owner_from_fctx() reported every failure as -ENOMEM.

Convert to int with an out-param, matching libnvmf_config_read()'s existing convention, and update every call site. The SWIG binding now raises a real errno-based message instead of always PyErr_NoMemory().

Also check for a partial shr_xstrdup() failure across the struct's multi-field copy, which the pointer-returning version never checked.

### DIFF
--- a/discoverd/src/tid.c
+++ b/discoverd/src/tid.c
@@ -19,11 +19,14 @@ struct libnvmf_tid *tid_new(const char *transport, const char *traddr,
 			    const char *host_traddr, const char *host_iface,
 			    const char *hostnqn, bool is_dc)
 {
+	struct libnvmf_tid *t;
+
 	if (!trsvcid || trsvcid[0] == '\0')
 		trsvcid = libnvmf_get_default_trsvcid(transport, is_dc);
 
-	return libnvmf_tid_from_fields(transport, traddr, trsvcid, subsysnqn,
-				       host_traddr, host_iface, hostnqn, NULL);
+	libnvmf_tid_from_fields(transport, traddr, trsvcid, subsysnqn,
+				host_traddr, host_iface, hostnqn, NULL, &t);
+	return t;
 }
 
 /*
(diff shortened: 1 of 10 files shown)


igaw:
Do we have more of such functions? I tried to get rid of them and only return the result on very simple functions (pure allocations for example) Anything which does something should only return error codes and have the result as argument. I know some people hate this but we don't have the `ERR_PTR` type of macros to return values and error codes directly.

martin-belanger:
I audited for exactly this pattern before submitting: went through all 295 exported symbols looking for pointer-returning constructors that conflate `-EINVAL` and `-ENOMEM` behind a bare NULL. Found five candidates — `libnvmf_tid_from_fields()`, `libnvmf_tid_parse()`, `libnvmf_tid_parse_strict()`, `libnvmf_tid_dup()`, and `libnvmf_config_emit_new()`.

I then checked every call site to see if the specific error code was ever needed. For `tid_parse()`/`tid_parse_strict()`/`tid_dup()`/`config_emit_new()`, none of the ~40 call sites cared — every caller only checked for NULL. Converting them would have been pure mechanical churn with no realized behavioral benefit.

BTW, there's another function with the same shape: `libnvme_create_global_ctx()`. It and the four I left alone all have one thing in common: they allocate an object and return NULL if they can't. I kept all of them as pointer-returning, but if you'd rather I convert them to an error code (or 0 on success) plus an outparam, I can do it.

So I don't think there's another live candidate right now — but I agree with the general principle.

igaw:
I would suggest to update  libnvmf_tid_parse() and  libnvmf_tid_parse_strict() as both are non trivial and they should actually return error codes if something goes wrong. NULL could be no memory or the parser found something invalid.

PR #3670, opened 2026-07-28
libnvme/nbft: Add sample synthetic NBFT tables

tbzatek:
AI-generated binary tables crafted based on the Boot Spec whitepaper, without knowledge of the parser code. Various combinations of records that seemingly don't make sense, only to increase coverage of the NBFT parser tests.


igaw:
Nice! Thanks!

PR #3671, opened 2026-07-28
libnvme: fix comments that still describe removed JSON config support

martin-belanger:
json-c was dropped from libnvme entirely during the INI migration, but several comments/kdoc still described host/keyring/ctrl-merge logic in terms of a JSON config file that is no longer read. Update them to describe what the code actually does now.

libnvmf_host_get_ids()'s kdoc also had two further inaccuracies unrelated to JSON: the tiers were listed weakest-first instead of in actual lookup order, hostid-from-hostnqn derivation only works when the hostnqn has a "uuid:" component (undocumented), and "verifies" overstated a mismatch check that only logs a debug warning.

No behavior change.


igaw:
Thanks!

PR #3672, opened 2026-07-28
nvme-models: fix stream EOF state errors in __nvme_product_name and pull_class_info

bellis-ibm:
parse_vendor_device() and pull_class_info() share the same FILE
pointer as the outer fgets loop in __nvme_product_name(). Either 
function can consume the file to EOF, leaving the stream's EOF 
indicator set before the outer loop re-evaluates its fgets condition on 
the back-edge, and before pull_class_info() reads from the same stream. 

Add clearerr() at the end of the outer loop body to reset the EOF indicator 
after each iteration, ensuring the stream is in a defined state before the next
fgets call and before any delegate reads.

### DIFF
--- a/nvme-models.c
+++ b/nvme-models.c
@@ -342,9 +342,12 @@ static char *__nvme_product_name(int id)
 						device,
 						sub_device,
 						sub_vendor);
+			clearerr(file);
 		}
-		if (is_class_info(readbuf))
+		if (is_class_info(readbuf)) {
 			pull_class_info(readbuf, file, class);
+			clearerr(file);
+		}
 	}
 	fclose(file);
 


nvme-models.c
```diff
@@ -296,34 +342,46 @@ static char *__nvme_product_name(__u32 *vid, __u32 *did,
 						device,
 						sub_device,
 						sub_vendor);
+						clearerr(file);
```
igaw: one tab too much

nvme-models.c
```diff
@@ -296,34 +342,46 @@ static char *__nvme_product_name(__u32 *vid, __u32 *did,
...
 		}
 		if (is_class_info(readbuf))
 			pull_class_info(readbuf, file, class);
+		clearerr(file);
```
igaw: In this case the `clearerr` should only be part of the `if`, no?

igaw:
I am confused, I just merged the second patch via https://github.com/linux-nvme/nvme-cli/commit/baeeb08bd639c3fda806adfca54ce1766826cd73 and the first patch hasn't changed. Could you rebase this PR and address my comments. Thanks

bellis-ibm:
Ok, thanks for the feedback. I've rebased onto master and made those changes as suggested.

igaw:
Thanks a lot!

Issue #3657, opened 2026-07-28
nvme config convert: keep orignal config file


igaw:
The config conversion strategy could leave a system in a broken state on distributions that support rollback. Consider the case where the `nvme-cli` package is updated and the next invocation of `nvme-cli` performs the config conversion. The existing configuration files are renamed to `config.json.converted` and `discovery.conf.converted`. If the user then decides to roll back the package update, the original configuration files are no longer available.

Instead, it would be better to keep the existing configuration files and create a link such as `config.json.convert -> config.json`. The presence of this link would indicate that the conversion has already been performed.

For explicit conversion, it would also be useful to provide an option that automatically removes the old configuration files, so that `config.json.convert` and `config.json` are removed.

The strategy for eventually removing the old configuration files is distribution-specific. For example, a distribution may choose to remove obsolete configuration files during a package update. However, this is a packaging policy decision and therefore outside the scope of this project.

tbzatek:
This is an important point to highlight in the Release Notes.

> Consider the case where the `nvme-cli` package is updated and the next invocation of `nvme-cli` performs the config conversion.

I would strongly suggest to perform any conversion in the distro package post-scripts.

Also consider that some distributions allow package version pinning and rollback of individual packages back and forth, so any conversion or backup/restore should be supported both ways. Packaging concern likely again.

Initramfs case is probably fine as the new nvme-cli would get picked up along with converted config files. We'll also need to update the dracut nvmf module to include any new config files once the dust settles and probably add some kind of `Conflicts: <older_dracut_ver>` to ensure correct versions.

Cc: @mwilck

igaw:
FWIW, I've discussed this with our distro experts and I got the strong recommendation to implement auto migration. post-scripts install hooks are not the right choice for Tumbleweed at least. 

Though there is nothing preventing a packager to do trigger migrations in the post-script sections. As far I can tell all use cases are supported when this feature here is implement. 

Forgot to mention, we wont support config rollbacks. This is out of the scope of this project. 

Sure all will be documented in the release notes which no one will read.

PR #3681, opened 2026-07-29
nvme: preserve legacy json config for rollbacks

martin-belanger:
`rename_to_converted()` destroyed `config.json`/`discovery.conf` once converted, so rolling back to a pre-INI nvme-cli release found no usable config at all. Mark conversion with a symlink instead of a rename, leaving the original in place for rollback.

Since the source file no longer disappears after conversion, the convert triggers now check the .converted symlink before checking whether the source exists, or every invocation would reconvert forever. Update the closing messages to describe the symlink-based marker.

### DIFF
--- a/config-convert.c
+++ b/config-convert.c
@@ -423,21 +423,30 @@ int nvme_config_convert_discovery(struct libnvmf_config_emitter *emitter,
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


igaw:
Fixes #3657

igaw:
Thanks!

PR #3683, opened 2026-07-29
nvme: few mem & stale pointer fixes

martin-gpy:
Adding a couple of mem related fixes along with & a stale pointer fix

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -557,11 +557,6 @@ static int create_common_context(struct libnvme_global_ctx *ctx,
 	if (err)
 		goto err;
 
-	err = libnvmf_context_set_crypto(fctx, fa->hostkey, fa->ctrlkey,
-		fa->keyring, fa->tls_key, fa->tls_key_identity);
-	if (err)
-		goto err;
-
 	libnvmf_context_set_persistent(fctx, persistent);
 
 	*fctxp = fctx;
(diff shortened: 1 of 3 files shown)


libnvme/src/nvme/fabrics.c
```diff
@@ -711,6 +711,10 @@ __libnvme_public int libnvmf_context_set_crypto(struct libnvmf_context *fctx,
 		return 0;
 	}
 
+	if (fctx->tls_key) {
```
igaw: free() checks for NULL pointers, so need for the if here. Also setting fctx->tls_key to NULL is not needed as the very next line will assign a new pointer.

fabrics.c
```diff
@@ -557,11 +557,6 @@ static int create_common_context(struct libnvme_global_ctx *ctx,
 	if (err)
 		goto err;
 
-	err = libnvmf_context_set_crypto(fctx, fa->hostkey, fa->ctrlkey,
```
igaw, line 562: move this into a separate commit. thanks!

libnvme/src/nvme/fabrics.c
martin-gpy: Ah yes, you are right. Will make that change.

fabrics.c
martin-gpy: Will do.

igaw:
Thanks!

PR #3684, opened 2026-07-29
libnvme/nbft: tests: Regenerate reference NBFT table dumps

tbzatek:
Adapt to the recent dump tool additions and the newly added tables.

(race between #3673 and #3670)


igaw:
Thanks!

PR #3685, opened 2026-07-29
libnvme: return libnvmf_tid_parse{,_strict}() as int, not a pointer

martin-belanger:
libnvmf_tid_parse() and libnvmf_tid_parse_strict() returned NULL on any failure, conflating -EINVAL (malformed input) with -ENOMEM (allocation failure). Change both to return an error code through an int + outparam, matching libnvmf_tid_from_fields()'s convention.

Also check the per-field strdup() calls in the shared parser for allocation failure -- previously unchecked, a mid-parse OOM could silently produce a truncated TID instead of an error.

Follow up to: https://github.com/linux-nvme/nvme-cli/pull/3669


igaw:
Thanks!

PR #3687, opened 2026-07-29
shared: add more test coverage

martin-belanger:
Unit-test parse-util, hash-util, string-util, io-util, fs-util, and net-util. Internal helpers like these don't need public-API-style NULL hardening, so string-util's tests only exercise NULL where a function already documents handling it.


igaw:
Excellent stuff. Thanks!

PR #3691, opened 2026-07-29
libnvme/fabrics: skip NBFT records with Security Profile Descriptors

tbzatek:
Use of NBFT Security Profile Descriptors is currently unimplemented, print an error and skip SSNS records and Discovery Descriptors that have an associated Security Profile Descriptor to avoid connecting without the required security parameters.

At the current state the code will still make a connection attempt totally ignoring the security descriptor values. Although bypassing security like this is considered a target misconfiguration, user should be aware of what's going on. Thus deny such connections and skip them, incl. discovery. 

----

As discussed on previous Timberland SIG meetings, we intentionally want to avoid making use of Security Descriptors at the moment until we have other components ready and tested (dracut, tlshd, the secret retrieval mechanism/Redfish client). Also, we currently don't have access to any system with capable firmware, unable to develop nor test this solution. Lastly, this functionality is also missing from the reference edk2 implementation. Rather than shipping half-baked untested solution, disable the functionality for now.

Cc: @flaviens @mwilck @johnmeneghini @MichaelRabek @stuarthayes @cleech

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -3235,6 +3235,14 @@ __libnvme_public int libnvmf_discovery_nbft(struct libnvme_global_ctx *ctx,
 					continue;
 				}
 
+				if ((*ss)->security) {
+					libnvme_msg(ctx, LIBNVME_LOG_ERR,
+						"SSNS %d has Security Profile Descriptor %d associated, security profile descriptors are currently unimplemented, skipping\n",
+						(*ss)->index,
+						(*ss)->security->index);
+					continue;
+				}
+
 				nfctx.ctrl_params.host_traddr = NULL;
 				if (!fctx->ctrl_params.host_traddr &&
 				    !strncmp((*ss)->transport, "tcp", 3))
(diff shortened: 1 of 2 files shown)


flaviens:
I see two things to address before merging.

1. Checkpatch fails because the two log messages split string literals. Keeping each format string on one source line clears all three warnings. I verified that form with checkpatch.

2. The guard checks the resolved `security` pointer. Both `read_ssns()` and `read_discovery()` leave that pointer null when a nonzero security reference cannot be resolved. The connection then proceeds. Please reject unresolved security references during parsing or preserve the association state and check it here. A malformed table could otherwise still trigger an unsecured connection attempt.

A test for a resolved security reference should show that no connection is attempted. A second case should cover an unresolved reference.

All other CI jobs pass.

tbzatek:
> 2\. The guard checks the resolved `security` pointer. Both `read_ssns()` and `read_discovery()` leave that pointer null when a nonzero security reference cannot be resolved. The connection then proceeds. Please reject unresolved security references during parsing or preserve the association state and check it here. A malformed table could otherwise still trigger an unsecured connection attempt.

Good idea, let's let it fail. There might be some more unguarded unresolved dependencies in the parser between tables and objects, though not resulting in a security issue.

igaw:
@flaviens any objection?

flaviens:
Lgtm, thanks!

igaw:
Thanks!

PR #3692, opened 2026-07-30
plugins/ymtc: fix additional smart info display for YMTC PE511

Robin-YMTC:


igaw:
You got it! I've decided to squash the patches together, no need to see the devel history in the final version. I've also reordered some of the variable declarations.

Thanks!

PR #3697, opened 2026-07-30
nbft: ensure transport buffer is null terminated in read_ssns() and r…

fhassan-ibm:
strncpy() does not guarantee null termination when the source string is exactly as long as or longer than the destination buffer. Both read_ssns() and read_hfi() copy the result of trtype_to_string() into a char transport[8] field using sizeof(transport) as the length, which leaves no room for a null terminator if the source fills all 8 bytes.

Reduce the strncpy length by 1 and explicitly set the final byte to '\0' in both call sites, ensuring the buffer is always null terminated regardless of the source string length.

Fixes: Coverity CID 557381 (Buffer not null terminated, High)


igaw:
Thanks

PR #3698, opened 2026-07-30
Add support to read/convert `persistent/epcsd`to config

dwsuse:
The new config system doesn't know about PDC. Add support for `persistent` in the discovery sections.
While at it, also add a command which creates configs from the command line. 

Related to #3112 

This PR does not implement the PDC/EPCSD logic, it only adds the config related bits.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -489,14 +489,18 @@ int nvmf_convert_discovery_line(struct libnvmf_config_emitter *emitter,
 	char *argv[MAX_DISC_ARGS] = { "discovery.conf" };
 	char *ptr, *p = line;
 	int argc = 1;
-	bool persistent = false, force = false;
-	bool no_persistent = false;
+	bool persistent = false, force = false, epcsd = false;
+	bool no_persistent = false, no_epcsd = false;
 
 	NVMF_ARGS(opts, fa,
 		  OPT_FLAG("persistent", 'p', &persistent,
 			   "persistent discovery connection"),
 		  OPT_FLAG("no-persistent", 0, &no_persistent,
 			   "explicitly not a persistent discovery connection"),
+		  OPT_FLAG("epcsd",        0, &epcsd,
+			   "Explicit Persistent Connection Support for Discovery"),
+		  OPT_FLAG("no-epcsd",     0, &no_epcsd,
+			   "explicitly does not support Explicit Persistent Connection Support for Discovery"),
 		  OPT_FLAG("force",        0, &force,
 			   "Force persistent discovery controller creation"));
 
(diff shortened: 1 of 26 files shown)


martin-belanger:
LGTM. I thought that `ensure_dir_exists()` could be a shared function (`shr_mkdir_from_fname()` or somthing like that), but that would mean removing the ctx, which is nice for printing errors.

dwsuse:
> LGTM. I thought that `ensure_dir_exists()` could be a shared function (`shr_mkdir_from_fname()` or somthing like that), but that would mean removing the ctx, which is nice for printing errors.

Good idea. I think there are bunch more of little good helpers everywhere. We need to go on a raid :)

Anyway, I think by  just returning good error codes for the caller would be enough to write something useful.

libnvme/src/nvme/private-fabrics.h
```diff
@@ -50,6 +50,7 @@ struct libnvmf_context { // !generate-accessors:read=generated,write=generated
 	/* common fabrics configuration */
 	const char *device;
 	bool persistent;
+	bool epcsd;
```
Mr-Bossman: from what i can tell we arent using epcsd anywhere.
From my my pr it should be checked instead of where i call `libnvme_host_is_epcsd_enabled`
https://github.com/linux-nvme/nvme-cli/pull/3472/changes#diff-c47d250fa87bf995531d790513464dc0cc79714a94d1e6d657dd84946a88871aR2354

fabrics.c
```diff
@@ -489,12 +489,12 @@ int nvmf_convert_discovery_line(struct libnvmf_config_emitter *emitter,
...
 	char *ptr, *p = line;
 	int argc = 1;
-	bool line_persistent = false, line_force = false;
+	bool persistent = false, force = false;
 
 	NVMF_ARGS(opts, fa,
-		  OPT_FLAG("persistent", 'p', &line_persistent,
+		  OPT_FLAG("persistent", 'p', &persistent,
 			   "persistent discovery connection"),
-		  OPT_FLAG("force",        0, &line_force,
```
Mr-Bossman, line 497: NGL i never really understood what `force` did. both `-p` and `--force` seem to make any DC persistent, even ones that don't support `EPCSD`.

Mr-Bossman:
IMO this should be 2 prs one for the PDC stuff and another for the create command and earlier.

fabrics.c
igaw: IIRC, the `force` flag is to overwrite logic in `libnvmf_discovery`, basically it's a 'do what I say'.

libnvme/src/nvme/private-fabrics.h
igaw: This is correct. I've mentioned this in the commit message. This is currently only extending the config parser/logic to handle these new flags.

dwsuse:
rebased on the latest master

dwsuse:
cleanup of the series. introduce tristate from the beginning

Mr-Bossman:
LGTM, do you want me to implement the epsd control logic?

dwsuse:
> LGTM, do you want me to implement the epsd control logic?

Yes, that would be great as you have plenty of gear to play with :)

PR #3699, opened 2026-07-30
zns: fix resource leak of table in list()

sahmed-ibm:
table_init_with_columns() allocates a struct table on the heap and the caller is responsible for freeing it via table_free(). Add a null check after allocation to handle failure early.  The two early return paths on ctx creation failure and topology scan failure both bypassed table_free(), leaking the allocation.

Replace the bare return statements with goto free_table so all exit paths converge on the single table_free(t) call before returning.


plugins/zns/zns.c
```diff
@@ -105,21 +105,29 @@ static int list(int argc, char **argv, struct command *acmd,
 		{ "FW Rev", LEFT, 8 },
 	};
 	struct table *t = table_init_with_columns(columns, ARRAY_SIZE(columns));
+	if(!t) {
```
igaw: Missing space between if and the opening parenthesis.

igaw:
There was another problem in the `list` function with the proper order of arguments parsing, addressed in 6fb59f2a230e ("plugins: invoke the arg parser for the remaining commands"). This also fixed the resource leak. 

Thanks anyway.

PR #3700, opened 2026-07-30
micron: clean up and fix micron telemetry log reading

brgoing:
Cleans up and consolidates telemetry log reading within the Micron plugin.
- Takes advantage of new nvme_get_log_telemetry_ctrl/host dynamic sizing to read the logs with in the largest chunks possible.
- Uses nvme_telemetry_log struct for header decode.
- Makes all telemetry log paths in the plugin go through the same helper methods, reducing duplication.

Fixes issues found during cleanup.
- Fixes data area 4 support detection and last block calculation.
- Fixes log size calculation that was leaving the last block unread.
- Fixes telemetry bin file names.


igaw:
Thanks!

PR #3701, opened 2026-07-30
netapp: fix resource leak on alloc failure in netapp_ontapdevices()

fhassan-ibm:
The netapp_ontapdevices() function calls scandir() to build a list of NVMe device entries. After scandir() succeeds, @ontapdevices is allocated with calloc(). If the calloc() fails, the function returns -ENOMEM without freeing the @devices array returned by scandir().

This results in a memory leak each time the allocation failure path is taken.

Free each @devices entry and then @devices itself before returning -ENOMEM to prevent the leak.

### DIFF
--- a/plugins/netapp/netapp-nvme.c
+++ b/plugins/netapp/netapp-nvme.c
@@ -1006,8 +1006,8 @@ static int netapp_ontapdevices(int argc, char **argv, struct command *acmd,
 {
 	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
 	const char *desc = "Display information about ONTAP devices.";
-	struct dirent **devices;
-	int num, i, ret, fmt;
+	__cleanup_dirents struct dirents devs = {};
+	int i, ret, fmt;
 	struct ontapdevice_info *ontapdevices;
 	char path[264];
 	char *devname = NULL;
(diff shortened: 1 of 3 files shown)


plugins/netapp/netapp-nvme.c
```diff
@@ -1060,6 +1060,9 @@ static int netapp_ontapdevices(int argc, char **argv, struct command *acmd,
 	ontapdevices = calloc(num, sizeof(*ontapdevices));
 	if (!ontapdevices) {
 		nvme_show_error("Unable to allocate memory for devices");
+		for (i = 0; i < num; i++)
+			free(devices[i]);
+		free(devices);
```
igaw: Or we could move the `__cleanup_dirents` from `private-tree.h` to `shared/cleanup.h` and start using it here.

fhassan-ibm: Thanks for the suggestion @igaw! I've implemented it — moved struct dirents, cleanup_dirents(), and __cleanup_dirents from libnvme/src/nvme/private-tree.h into shared/cleanup.h (with the required #include <dirent.h>), then refactored netapp_ontapdevices() to use __cleanup_dirents struct dirents devs = {} in place of the raw struct dirent **devices. The manual for/free loop at the calloc failure path and at the end of the function are now both gone — cleanup happens automatically via the attribute.

martin-gpy:
Thanks for this PR. This same issue should exist in netapp_smdevices() as well. And ideally, one shouldn't combine the moving of dirents cleanup to shared/cleanup.h plus refactoring netapp_ontapdevices() in the same commit itself i.e. should separate out the two.

So here's what I'd suggest: create 3 commits here - 1st for moving dirents cleanup to shared/cleanup.h, 2nd and 3rd commits for netapp_ontapdevices() & netapp_smdevices() respectively to consume the __cleanup_dirents_. Separate commits are preferred for ontapdevices & smdevices because they cater to namespace devices from different arrays (though both belong to the same vendor NetApp).

fhassan-ibm:
Thanks @martin-gpy! Good catch on netapp_smdevices() — it had the exact same issue. I've restructured into 3 separate commits as suggested:

shared: move dirents cleanup from private-tree.h to shared/cleanup.h — moves struct dirents, cleanup_dirents(), and __cleanup_dirents to shared/cleanup.h so plugins can use the pattern
netapp: fix resource leak on alloc failure in netapp_ontapdevices() — refactors netapp_ontapdevices() to use __cleanup_dirents
netapp: fix resource leak on alloc failure in netapp_smdevices() — same fix for netapp_smdevices()

igaw:
Thanks!

PR #3703, opened 2026-07-30
scaleflux: clamp code_type before array access in nvme_parse_evtlog()

fhassan-ibm:
code_type is derived from firmware data as (info->code % 0x100), giving a value in the range 0..255. It is then used directly as an index into sfx_evtlog_warning[4] or sfx_evtlog_error[14] with no bounds check, so any code_type value exceeding the array size causes an out-of-bounds read.

Clamp code_type to 0 ("RESERVED") when it would exceed the bounds of the relevant array before either array is indexed.

Fixes: Coverity CID 557357 (Out-of-bounds read, High)


igaw:
Thanks!

PR #3705, opened 2026-07-30
ocp: improve telemetry log reading and fix allocation issues and related bugs

brgoing:
Improves OCP telemetry log reading in the OCP plugin.
- Speeds up telemetry log retrieval by reading in the largest transfer size the drive and host support.
- Switches log-buffer allocations to libnvme_alloc/libnvme_free for aligned allocations.
- Fixes a use-after-free bug plus several memory leaks tied to C9 log-page reads.
- Fixes le64 handling in C9 log-page handling.
- Adds NULL checks after each memory allocation.

### DIFF
--- a/plugins/ocp/ocp-nvme.c
+++ b/plugins/ocp/ocp-nvme.c
@@ -1285,14 +1285,14 @@ exit_status:
 static int get_c9_log_page_data(struct libnvme_transport_handle *hdl,
 		int print_data, int save_bin, const char *output_file,
 		struct telemetry_str_log_format **log_data_out,
-		__u8 **string_buffer_out, __le64 *total_log_page_sz)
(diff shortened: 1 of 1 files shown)


plugins/ocp/ocp-nvme.c
```diff
@@ -588,13 +588,8 @@ static int eol_plp_failure_mode(int argc, char **argv, struct command *acmd,
 ///////////////////////////////////////////////////////////////////////////////
 /// Telemetry Log
 //global buffers
-static __le64 total_log_page_sz;
-static __u8 *header_data;
-static struct telemetry_str_log_format *log_data;
```
igaw, line 593: YES!

igaw:
Nice cleanup. The OCP plugin has a somewhat particular way of handling allocations...

PR #3706, opened 2026-07-30
discoverd: rename discoverd.conf to nvme-discoverd.conf

martin-belanger:
discoverd.conf and the legacy discovery.conf sit in the same directory and are easy to confuse. Rename discoverd's own configuration file to nvme-discoverd.conf.


igaw:
Thanks!

PR #3707, opened 2026-07-30
plugins: invoke the arg parser for the remaining commands

jimmunn:
zns list and the micron/seagate plugin-version commands returned without calling the arg parser, so they silently ignored the global options (--output-format, --verbose, etc.) and offered no --help.

Wire each of them through parse_args() so the global options are accepted and --help works. Output is unchanged.


igaw:
Nice! Thanks!

PR #3708, opened 2026-07-31
plugins/config: add create command

dwsuse:
Add a command which allows to create a config from the command line.

This splits out the changes from #3698


dwsuse:
windows fails because there is no windows version of `test_mkdir_from_fname`

```
==================================== 8/39 ====================================
test:         nvme-cli:shared - fs-util
start time:   10:30:41
duration:     0.02s
result:       exit status 1
command:      PATH=D:/a/nvme-cli/nvme-cli/.build-ci/shared/test;D:\a\_temp\msys64\ucrt64\bin;D:\a\_temp\msys64\usr\local\bin;D:\a\_temp\msys64\usr\bin;D:\a\_temp\msys64\usr\bin;C:\Windows\System32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;D:\a\_temp\msys64\usr\bin\site_perl;D:\a\_temp\msys64\usr\bin\vendor_perl;D:\a\_temp\msys64\usr\bin\core_perl MESON_TEST_ITERATION=1 MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MALLOC_PERTURB_=241 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 D:/a/nvme-cli/nvme-cli/.build-ci/shared/test/test-fs-util.exe
----------------------------------- stdout -----------------------------------
test_basename:
 - regular path [PASS]
 - no slash at all [PASS]
 - trailing slash → empty (unlike POSIX basename()) [PASS]
 - bare "/" → empty [PASS]
 - empty input → empty [PASS]
test_mkdir_p:
 - creates every missing parent [PASS]
 - the deepest directory is real and writable [PASS]
 - re-running on an existing tree is a no-op success [PASS]
 - a trailing slash does not confuse it [PASS]
 - a path longer than PATH_MAX is rejected [PASS]
test_mkdir_from_fname:
 - creates parent directories from filename: got -129, want 0 [FAIL]
 - the deepest parent directory is real and writable [FAIL]
 - re-running on an existing tree is a no-op success: got -129, want 0 [FAIL]
 - bare filename without directories is a no-op success: got -129, want 0 [FAIL]
 - a parent path longer than PATH_MAX is rejected: got -129, want -38 [FAIL]
test_fsync_dir:
 - fsync_dir on a real and a bogus path [PASS]
==============================================================================
```

martin-belanger:
If I'm not mistaken, the one thing missing in Windows is the `dirname()` function. How about implementing `shr_dirname()` as follows (suggested by Claude):

```
char *shr_dirname(char *path)
{
      char *end;
      char *slash;

      if (!path || !*path)
              return ".";

      // strip trailing '/'
      end = path + strlen(path) - 1;
      while (end > path && *end == '/')
              end--;

      if (end == path && *end == '/')
              return "/";                     // was all slashes

      slash = end;
      while (slash > path && *slash != '/')
              slash--;

      if (*slash != '/')
              return ".";

      while (slash > path && *(slash - 1) == '/')
              slash--;

      if (slash == path)
              return "/";

      *slash = '\0';
      return path;
}
```

martin-belanger:
> Passing NULL to dirname() is undefined and will crash on Linux.

I disagree with copilot. I was looking at the source code for glibc's dirname, and it is NULL safe. In fact it returns a pointer to "."

https://github.com/bminor/glibc/blob/master/misc/dirname.c

dwsuse:
There are also other libc implementation but it seems musl does the same thing otherwise the tests would fail on alpine.

PR #3709, opened 2026-07-31
lm-nvme: return errors from lm_migration_recv()

sahmed-ibm:
lm_migration_recv() stores command and write failures in err but returns success unconditionally at the end of the function.

Return err so command execution and output write failures are reported to the caller.


igaw:
Thanks!

PR #3711, opened 2026-07-31
libnvme: Lazy sysfs reads

martin-belanger:
@igaw - That was more work than I thought it would be. Had this be done in C++ or Python, it would have been a lot easier. ;-)

## Why this is large

This PR adds a new code generator, `generate_sysfs_accessors.py`, and uses it to make `struct libnvme_ctrl_sysfs` (the controller's sysfs-derived fields: firmware, model, serial, numa_node, queue_count, cntrltype/cntlid/dctype, phy_slot, dhchap keys, etc.) opaque and lazily-populated instead of eagerly read and embedded directly in `struct libnvme_ctrl`.

The existing `generate_accessors.py` parses accessor definitions out of annotated header fields. That model doesn't work here: nobody outside a small, fixed set of generated `.c` files may ever see this struct's layout, or code could read a field before its first access populates it. So the new generator is driven by a Python dict instead, and reuses `generate_accessors.py`'s `Member` model and emitters rather than duplicating them.

**Scope: controller only.** Subsystems and namespaces still use the old eager, header-annotated, non-lazy accessors. This PR doesn't touch them — the generator and the opacity pattern are proven on one struct first.

## The tradeoff I want you to be aware of

This adds real complexity. Before this PR, everything `libnvme_ctrl` knew about a controller was a plain field you could see by opening `private.h`. After this PR, the sysfs-derived fields are behind an opaque pointer, and the struct itself is defined nowhere but inside a generated `.c` file that nothing else may include. Grepping `private.h` no longer shows you the layout — you have to know to look at `ctrl-sysfs.c`, or at the dict in the generator, to see what's there. This is the necessary cost of the optimization.

## The actual win

Sysfs is now read on first access instead of unconditionally at scan/reconfigure time. Measured against 4 real controllers, `nvme list` drops from 186 to 136 sysfs `openat` calls (-27%) — fields the command never displays (`numa_node`, `queue_count`, `sqsize`) are no longer read at all, and fields it does display are read once per controller that actually needs them instead of every controller unconditionally. `nvme list -v`, which does need those fields for every controller, correctly shows no regression.

## 4 Commits, in review order

1. `libnvme/generator: rename generate-accessors.py for import` — file rename plus four references, no behavior change.
2. `libnvme: extract libnvme_parse_src_addr() for direct testing` — pulls parsing logic out of a getter so it's unit-testable without a live `struct libnvme_ctrl`; also removes the last direct struct access that would've blocked opacity.
3. `libnvme/generator: add dict-driven ctrl-sysfs generator` — the new generator itself. Worth reviewing on its own merits before commit 4, since everything downstream is just this generator's output.
4. `libnvme: make struct libnvme_ctrl_sysfs opaque` — the atomic, all-or-nothing switch: `private.h` forward-declaration, the dict, generated outputs, the four hand-written loader files, meson/`.ld`/SWIG wiring, and every call site the compiler forced to change.

## Test plan

- `-Dwerror=true` (default) and `-Dfabrics=disabled` both build clean.
- `meson test` at baseline on both (100/2/0, 58/0/0 — unchanged from pre-PR).
- `make checkpatch-diff`: 0 errors (a few pre-existing false-positive warning classes only — `__cleanup_*` blank-line, `volatile` diagnostic counters).


libnvme/tools/generator/generate_sysfs_accessors.py
```diff
@@ -0,0 +1,628 @@
...
+touches it. It only needs a new entry when a group's loader needs to vary
+along a build axis that doesn't already exist here. Note its values
+(output/custom filenames) are specific to struct libnvme_ctrl_sysfs, same
+as CTRL_SYSFS -- it stays a separate module-level table rather than a
+CTRL_SYSFS key only because a second dict (namespaces/subsystems) isn't
+scoped yet; see the design note above.
+
+Example usage:
+  ./generate_sysfs_accessors.py --out-dir ../../src/nvme
+"""
```
igaw, line 18: This needs to be shorten, no reasoning here, this is already in the MD file

libnvme/tools/generator/generate_sysfs_accessors.py
```diff
@@ -0,0 +1,628 @@
...
+generate_accessors = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(generate_accessors)
+
+Member = generate_accessors.Member
+
+SPDX_C = '// SPDX-License-Identifier: LGPL-2.1-or-later'
+SPDX_H = '/* SPDX-License-Identifier: LGPL-2.1-or-later */'
+SPDX_LD = '# SPDX-License-Identifier: LGPL-2.1-or-later'
+BANNER = (
+    '/* This file is auto-generated by generate_sysfs_accessors.py. Do\n'
```
igaw:
```
/*
 * This ...
 *
```

libnvme/tools/generator/generate_sysfs_accessors.md
```diff
@@ -0,0 +1,135 @@
+# Generate Sysfs Accessors Tool
```
igaw: can we move this document to the other ones and match the naming scheme. personally I would prefer the lower case version.

libnvme/tools/generator/generate_sysfs_accessors.py
```diff
@@ -0,0 +1,628 @@
...
+)
+LD_BANNER = (
+    '#\n'
+    '# This file is auto-generated by generate_sysfs_accessors.py. Do\n'
+    '# not edit by hand -- edit the CTRL_SYSFS dict instead and\n'
+    '# regenerate.\n'
+    '#'
+)
+
+CTRL_SYSFS = {
```
igaw: This is configuration/input data, right? This really shouldn't be in the generator file. Hmm, it's adding another layer of complexity too. Anyway, I think the input should live at least in `libnvme/src/nvme`.

shared/compiler-attributes.h
```diff
@@ -32,3 +32,14 @@
...
+/**
+ * __shr_likely - Hint that an expression is usually true.
+ * __shr_unlikely - Hint that an expression is usually false.
+ *
+ * Wrap a branch condition to help the compiler order the generated code
+ * for the common case, e.g. a lazy-cache guard that is false on every
+ * call after the first.
+ */
+#define __shr_likely(x)   __builtin_expect(!!(x), 1)
+#define __shr_unlikely(x) __builtin_expect(!!(x), 0)
```
igaw, line 45: separate commit

libnvme/src/nvme/private-tree.h
```diff
@@ -35,6 +35,18 @@ static inline void cleanup_dirents(struct dirents *ents)
 #define NVME_DEFAULT_HOSTID "00000000-0000-0000-0000-000000000000"
 
 char *libnvme_hostid_from_hostnqn(const char *hostnqn);
+/**
```
igaw, line 55: new line missing

libnvme/src/nvme/private-tree.h
```diff
@@ -27,6 +27,24 @@ static inline void cleanup_dirents(struct dirents *ents)
 #define FREE_CTRL_ATTR(a) \
 	do { free(a); (a) = NULL; } while (0)
 
+/* Sentinel marking a lazy sysfs member as "read from sysfs, attribute
```
igaw:
```c
/*
 * Sentinel ...
 *
```

libnvme/src/nvme/tree.c
```diff
@@ -42,6 +42,12 @@ static int libnvme_ctrl_scan_namespace(struct libnvme_global_ctx *ctx,
 static int libnvme_ctrl_scan_path(struct libnvme_global_ctx *ctx,
 		struct libnvme_ctrl *c, char *name);
 
+/* Sentinel address for a lazy sysfs member that has been read and found
+ * absent -- see the NO_SYSFS_ATTR comment in private-tree.h. The content
+ * is never dereferenced; only the address is significant.
+ */
```
igaw: this has already been documented in the header.  Just remove it.

libnvme/src/nvme/private-tree.h
```diff
@@ -27,6 +27,24 @@ static inline void cleanup_dirents(struct dirents *ents)
...
+ * does not exist" -- distinct from both NULL ("not yet read", which is
+ * what a fresh calloc() already gives every pointer field for free) and
+ * a real heap pointer ("read, value cached"). A single extern symbol --
+ * every translation unit that stamps, checks, or frees a lazy member
+ * must compare against the same address, which a `static` per-TU
+ * definition cannot guarantee. Defined once in tree.c.
```
igaw: this documentation is not good.

libnvme/src/nvme/ctrl-sysfs-custom-fabrics.c
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ * Copyright (c) 2026, Dell Technologies Inc. or its subsidiaries.
+ *
+ * Authors: Martin Belanger <Martin.Belanger@dell.com>
+ */
+
+void libnvmf_ctrl_load_fabrics_attrs(struct libnvme_ctrl *c)
```
igaw: I would add return code. This function is complex enough that something might go wrong and we have no means to report it.

libnvme/src/nvme/ctrl-sysfs-custom-win.c
```diff
@@ -0,0 +1,77 @@
...
+	struct nvme_id_ctrl id_ctrl;
+	struct ctrl_map_entry *ctrl_entry;
+	int ret;
+
+	ret = libnvme_ctrl_identify(c, &id_ctrl);
+	if (ret != 0) {
+		libnvme_msg(c->ctx, LIBNVME_LOG_ERR,
+			"Failed to identify ctrl %s, error %d\n",
+			c->name, ret);
+		return;
```
igaw: we really should not eat up this error here, this needs to be propagated.

libnvme/src/nvme/private.h
```diff
@@ -333,36 +339,29 @@ struct libnvme_ctrl {  // !generate-accessors:read=generated,write=none !generat
 	char *name;
 	char *sysfs_dir;
 	char *address;
-	char *firmware;
-	char *model;
 	char *state;			// !access:read=custom
-	char *numa_node;
-	char *queue_count;
-	char *serial;
-	char *sqsize;
```
igaw, line 342: This looks really good. I like the direction

libnvme/src/nvme/private.h
```diff
@@ -333,36 +339,29 @@ struct libnvme_ctrl {  // !generate-accessors:read=generated,write=none !generat
...
-	long reconnect_count;		// !access:read=custom
 	struct libnvme_fabrics_config cfg; // !access:nested:write=none
+
+	/* Opaque: firmware/model/serial/numa_node/queue_count/sqsize/
+	 * cntrltype/cntlid/dctype/phy_slot/dhchap_host_key/
+	 * dhchap_ctrl_key/keyring/command_error_count/reset_count/
+	 * reconnect_count. Lazily populated and freed only through
+	 * libnvme_ctrl_sysfs_alloc/reset/free() and the accessors in
+	 * ctrl-sysfs.h -- see generate_sysfs_accessors.py.
+	 */
```
igaw: unnecessary documentation.

libnvme/tools/generator/generate_sysfs_accessors.py
```diff
@@ -6,19 +6,18 @@
 
 Currently drives only struct libnvme_ctrl_sysfs (see CTRL_SYSFS below). The
 generic name anticipates the same treatment for namespaces and subsystems,
-expected to reuse this script's struct-emission and variant-orchestration
-logic rather than a second, parallel generator. Nothing below is
-generalized ahead of that; VARIANTS and CTRL_SYSFS itself are still
-ctrl-specific.
+expected to reuse this script's struct-emission logic rather than a second,
```
igaw: For the final version, I think this could be merged/squashed to avoid so much code churn. It is adds to the review overhead.

igaw:
Overall I like the direction this is going. I hope we can move more of the sysfs entry under this mechanism eventually. The overhead is real. I ponder if we should think about making `list` a bit more lightweight. It's a kind of dumping ground for anything. 

Two things which need to be addressed:

- Input arguments:
This should be a file in `libnvme/src/nvme`. The format is not that important, whatever works best is okay IMO. We can update anytime if needed. The important part is that we bundle all input at one place.
- Error handling:
Errors need to be propagated to the caller. We can't hide them in the caching layer. This was one of the major problems we had in libnvme 1, that we didn't handle errors everywhere and didn't propagate them all the time.

libnvme/tools/generator/generate_sysfs_accessors.md
martin-belanger: Quick clarification before I act on this: did you mean *upper* case, and moving it to `libnvme/design/`? The file name is already *lower* case, so I'm just wondering...

If you're suggesting moving the doc to `libnvme/design/`, I'd push back gently. The `libnvme/design/` docs (REGISTRY.md, CONFIG.md, INTEGRATION.md, etc.) describe how libnvme itself is designed and how it behaves at runtime — they're meant for anyone consuming or reading about the library. generate_accessors.md and generate_sysfs_accessors.md are a different kind of document: they explain how the generator tooling works, for whoever is maintaining the generators themselves, not for general readers of libnvme. That's why this one lives next to its script in tools/generator/, matching generate_accessors.md's existing precedent, rather than alongside the design docs. Happy to move it if you still want that after this distinction, just didn't want to do it on a misreading.

libnvme/src/nvme/private-tree.h
martin-belanger: This is fixed by the very next commit ("libnvme/generator: add dict-driven ctrl-sysfs generator"), which adds the blank line between libnvme_hostid_from_hostnqn()'s declaration and the libnvme_parse_src_addr() kernel-doc block. So it's only missing in this one commit, not in the series as a whole. I'd rather not amend this commit just for a single blank line that's already corrected one commit later — let me know if you feel strongly about it.

libnvme/tools/generator/generate_sysfs_accessors.md
igaw: casing: the upper case file names always scream `I AM IMPORTANT! LOOK AT ME`. Thus I prefer normal casing. But given those documents are not in the root anymore, I don't care too much. Just saying.

I get your point, but the tooling is kind of intergral piece now. What about moving it to `libnvme/design/tooling` so we have all those documents in one place? They are mostly for the AI anyway.

martin-belanger: With the UPPERCASE, I was only following what every project seems to be doing. For example, systemd has: README.md, NEWS, LICENSE.*, CITATION.cff, AGENTS.md, TODO.md, and pretty much all the files under [systemd/docs/](https://github.com/systemd/systemd/tree/main/docs).

I agree with you that an all-uppercase file screams at you. Not sure why everybody does it.

With regards to the tooling, since those tools are only run manually once in a while to regenerate source files under libnvme/src/nvme/, I thought it'd be best if the documents were right next to the tool and are named the same base name as the python script they describe. 

That being said, I see your point and I can move both `generate-accessors.md` and `generate_sysfs_accessors.md` to libnvme/design/tooling/ as you suggested.

libnvme/tools/generator/generate_sysfs_accessors.py
martin-belanger: Done

libnvme/tools/generator/generate_sysfs_accessors.py
martin-belanger: Done

libnvme/tools/generator/generate_sysfs_accessors.md
martin-belanger: Done

libnvme/tools/generator/generate_sysfs_accessors.py
martin-belanger: Done

shared/compiler-attributes.h
martin-belanger: Done

libnvme/src/nvme/private-tree.h
martin-belanger: Done

libnvme/src/nvme/tree.c
martin-belanger: Done

libnvme/src/nvme/private-tree.h
martin-belanger: Done

libnvme/src/nvme/private.h
martin-belanger: Done

libnvme/tools/generator/generate_sysfs_accessors.py
martin-belanger: Done

libnvme/src/nvme/ctrl-sysfs-custom-fabrics.c
martin-belanger: Done

libnvme/src/nvme/ctrl-sysfs-custom-win.c
martin-belanger: Done

martin-belanger:
@igaw - I believe all review comments have been addressed.

libnvme/src/ctrl-sysfs.ld
```diff
@@ -0,0 +1,45 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+/*
+ * This file is part of libnvme.
+ *
+ * Copyright (c) 2025, Dell Technologies Inc. or its subsidiaries.
```
igaw, line 6: 2026?  Ah it's from the generator :)

The `ctrl-sysfs.ld` script should be in `libnvme/src/` where the rest of the linker scripts are, just so we have everything bundled together.

igaw:
The first three patches look good. The last three patches could be squashed together I suppose.

Overall, it's a good thing that we separate sysfs values from the other properties of a ctrl. I was looking if we have more sysfs values which could be handled this way, e.g. for `subsys`, `ns` or `path`. I think there are a bunch more we could also move under this mechanism. 

```
652:    queue_depth = libnvme_get_path_attr(p, "queue_depth");
664:    ana_state = libnvme_get_path_attr(p, "ana_state");
679:    numa_nodes = libnvme_get_path_attr(p, "numa_nodes");
695:    failover_count = libnvme_get_path_attr(p,
707:    retry_count = libnvme_get_path_attr(p, "diag/command_retry_count");
718:    error_count = libnvme_get_path_attr(p, "diag/command_error_count");
822:    sysfs_stat_path = libnvme_get_path_attr(p, "stat");
1177:   p->ana_state = libnvme_get_path_attr(p, "ana_state");
1181:   p->numa_nodes = libnvme_get_path_attr(p, "numa_nodes");
1185:   grpid = libnvme_get_path_attr(p, "ana_grpid");
1190:   queue_depth = libnvme_get_path_attr(p, "queue_depth");
```

Anyway, I'll take a strace on my test machine and see what kind of overhead could be removed.

libnvme/src/ctrl-sysfs.ld
martin-belanger: Done

martin-belanger:
Moved `ctrl-sysfs.ld` script to `libnvme/src/` and squashed the last 3 commits into one.

I do plan to do other structs (subsystem, namespace, path) once this PR lands.

dwsuse:
FWIW, with the current master for 1028 namespaces and 6 paths:

```
root@dolin:~ # awk -F'"' '/^openat\(/ {n=split($2,a,"/"); print a[n]}' /tmp/list | sort | uniq -c | sort -nr
   6168 queue_depth
   6168 numa_nodes
   6168 ana_state
   6168 ana_grpid
   1028 uuid
   1028 size
   1028 nuse
   1028 nsid
   1028 nguid
   1028 multipath
   1028 metadata_bytes
   1028 logical_block_size
   1028 eui
   1028 csi
     60 address
     12 dhchap_secret
      7 subsysnqn
      7 serial
      7 nvme-subsystem
      7 model
      7 hostnqn
      7 hostid
      7 firmware_rev
      6 transport
      6 tls_mode
      6 tls_key
      6 state
      6 sqsize
      6 slots
      6 queue_count
      6 numa_node
      6 libnvme3.so.1
      6 dhchap_ctrl_secret
      6 dctype
      6 cntrltype
      6 cntlid
      3 nvme6
      3 nvme5
      3 nvme4
      3 nvme3
      3 nvme2
      3 nvme1
      2 LC_CTYPE
      1 subsystype
      1 nvme-subsys1
      1 nvme-cli.conf
      1 nvme
      1 locale.alias
      1 locale-archive
      1 libzstd.so.1
      1 libz.so.1
      1 libpthread.so.0
      1 liblzma.so.5
      1 libkmod.so.2
      1 libkeyutils.so.1
      1 libjson-c.so.5
      1 libjitterentropy.so.3
      1 libcrypto.so.3
      1 libc.so.6
      1 ld.so.cache
      1 iopolicy
      1 gconv-modules.cache
```

igaw:
Thanks!

dwsuse:
and with these changes here:

```
root@dolin:~/nvme-cli # awk -F'"' '/^openat\(/ {n=split($2,a,"/"); print a[n]}' /tmp/list | sort | uniq -c | sort -nr
   6168 queue_depth
   6168 numa_nodes
   6168 ana_state
   6168 ana_grpid
   1028 uuid
   1028 size
   1028 nuse
   1028 nsid
   1028 nguid
   1028 multipath
   1028 metadata_bytes
   1028 logical_block_size
   1028 eui
   1028 csi
      7 subsysnqn
      7 nvme-subsystem
      7 hostnqn
      7 hostid
      6 transport
      6 tls_mode
      6 tls_key
      6 state
      6 libnvme3.so.1
      6 dhchap_secret
      6 address
      3 nvme6
      3 nvme5
      3 nvme4
      3 nvme3
      3 nvme2
      3 nvme1
      2 LC_CTYPE
      1 subsystype
      1 serial
      1 nvme-subsys1
      1 nvme-cli.conf
      1 nvme
      1 model
      1 locale.alias
      1 locale-archive
      1 libzstd.so.1
      1 libz.so.1
      1 libpthread.so.0
      1 liblzma.so.5
      1 libkmod.so.2
      1 libkeyutils.so.1
      1 libjson-c.so.5
      1 libjitterentropy.so.3
      1 libcrypto.so.3
      1 libc.so.6
      1 ld.so.cache
      1 iopolicy
      1 gconv-modules.cache
      1 firmware_rev
```

PR #3712, opened 2026-07-31
ioctl-win: update fw-commit action implementation based on testing

brgoing:
Testing showed that Windows support for commit actions is limited. Update the commit action mappings to Windows firmware activation flags for better behavior.


igaw:
Thanks!

PR #3720, opened 2026-08-03
Refactoring of discover code path

igaw:
The discover code path is unnecessary complex. The logic is spread between different functions, functions are too big, etc.

Thus restructure and refactor discover code path into something more understandable and maintainable.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -2589,6 +2589,51 @@ static void nvme_parse_tls_args(const char *keyring, const char *tls_key,
 	}
 }
 
+static bool dc_should_connect(struct libnvmf_context *fctx,
+		struct nvmf_disc_log_entry *e, bool *pdisconnet)
+{
+	bool disconnect = false;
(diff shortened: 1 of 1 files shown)


libnvme/src/nvme/fabrics.c
```diff
@@ -2613,16 +2653,12 @@ static int _nvmf_discovery(struct libnvme_global_ctx *ctx,
 		fctx->hooks.discovery_log(fctx, connect, log, numrec,
 			fctx->hooks.user_data);
 
-	if (!connect)
-		return 0;
```
Mr-Bossman: why are we removing this?

igaw: From my commit message:

```
Factor out the decision-making logic from _nvmf_discovery. This
logic will later be used to control each Discovery Controller
individually. Thus do not return immediately when !connect.
```

igaw:
I realized that I need to fix the `connect` path because the PDC cleanups have to go on top of it. #3724

PR #3722, opened 2026-08-03
discover API cleanups

dwsuse:
While working on #3720 I went down the `connect` rabbit hole.

WIP


igaw:
I figured out what my problem is with `connect`. It is overloaded, once it is used for the `connect-all` case and once it is needed to do the discovery itself. We can't talk to a DC before connecting. Thus the `connect` from user is a different `connect` than the one used for discovery nodes. This needs to be splitted.

igaw:
Replaced by #3724

PR #3723, opened 2026-08-03
solidigm-telemetry: fix UB sign-extension shift in telemetry_log_get_value

bellis-ibm:
When size_bit equals 64 the expression (0ULL - 1) << size_bit shifts a
64-bit value by its full width, which is undefined behaviour in C.  Guard
the sign-extension with size_bit < 64 and use ~((1ULL << size_bit) - 1)
to build the mask safely.


igaw:
Thanks!

PR #3724, opened 2026-08-04
libnvme/fabrics: handle connect separately from discover

igaw:
connect-all uses the discover APIs to connect to all I/O
controllers. The connect argument of _nvmf_discover is currently
used for both discovery and connection.

_nvmf_discover should always iterate over all discovery entries.
It implicitly assumes that child discovery controllers are connected
first. Therefore, _nvmf_discover does not need a connect
argument.

For the connect-all case, use the user's explicit request to connect
or not connect to all I/O controllers.

Supersede: #3720

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -455,7 +455,9 @@ static void consume_conn(const struct libnvmf_config_conn *conn,
 				MAX_DISC_RETRIES);
 		libnvmf_context_set_default_keep_alive_timeout(fctx,
 				NVMF_DEF_DISC_TMO);
-		err = libnvmf_discover(st->ctx, fctx, st->connect, st->force);
+		libnvmf_context_set_connect(fctx, st->connect);
+		libnvmf_context_set_force(fctx, st->force);
+		err = libnvmf_discover(st->ctx, fctx);
 	} else {
 		err = libnvmf_connect(st->ctx, fctx);
 	}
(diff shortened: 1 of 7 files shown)


fabrics.c
```diff
@@ -899,7 +907,9 @@ int fabrics_discovery(const char *desc, int argc, char **argv, bool connect)
 		if (ret)
 			return 0;
 
-		ret = libnvmf_discovery(ctx, fctx, connect, force);
+		libnvmf_context_set_connect(fctx, connect);
```
martin-belanger: Is this redundant? We already called `libnvmf_context_set_connect(fctx, connect);` at line 872.

fabrics.c
```diff
@@ -235,14 +235,14 @@ static void hook_already_connected(struct libnvmf_context *fctx,
...
 
 static void hook_discovery_log(struct libnvmf_context *fctx,
-		bool connect, struct nvmf_discovery_log *log,
+		struct nvmf_discovery_log *log,
 		uint64_t numrec, void *user_data)
 {
 	struct hook_fabrics_data *hfd = user_data;
 
 	if (hfd->raw)
 		save_discovery_log(hfd->raw, log);
```
martin-belanger, line 266: `save_discovery_log()` opens the target file with `O_TRUNC`, and this hook now fires once per recursion level since `_nvmf_discovery()` no longer early-returns on `connect == false`.

For a multi-level referral chain, `--raw` will end up containing only the deepest discovery controller's log and everything else gets overwritten.

I think that this is an existing bug for `connect-all --raw`, but now it will also get triggered by `discover --raw`.

fabrics.c
igaw: Yes, I've moved the set_connect and set_force up so it is set for all cases. Makes also more sense :)

fabrics.c
igaw: The hook is called before the `connect` test already:

```c
	if (fctx->hooks.discovery_log)
		fctx->hooks.discovery_log(fctx, log, numrec,
			fctx->hooks.user_data);

	if (!connect)
		return 0;
```

What is the best (least user suprise) behavior? Should `save_discovery_log` just append all the logs to one raw file? Or should be create a file per log (and add postfixed in this case, e.g. `$FILENAME-1.bin` `$FILENAME-2.bin`?

I tend to the second solution.

igaw: I went ahead and added the postfix solution.

PR #3726, opened 2026-08-04
ocp: fix resource leak of description in parse_event_fifo

sahmed-ibm:
When parse_ocp_telemetry_string_log() returns an error, the function exits early via return -1 without freeing the heap-allocated description buffer (malloc'd 41 bytes earlier). The existing free(description) under the free_desc label is not reached by this code path.

Set ret = -1 and goto free_desc instead of returning directly, so all cleanup is handled centrally at the existing free_desc label, consistent with other error paths in the same function.

Fixes Coverity CID 557340 (RESOURCE_LEAK).


igaw:
Thanks!

PR #3727, opened 2026-08-04
nvme-print: fix untrusted loop bounds in json_nvme_fdp_configs()

sahmed-ibm:
The json_nvme_fdp_configs() function iterates over FDP configuration descriptors using n as the loop bound, where n is read from the device-supplied log page field log->n. The outer loop used n + 1 as its bound, iterating one past the number of configurations reported by the device.

Iterating beyond the allocated buffer results in out-of-bounds memory access and undefined behavior.

The inner loop used le16_to_cpu(config->nruh) directly as a loop bound, re-evaluating the device-supplied field on each iteration. Reading a device-controlled value as an unsanitized loop bound is a tainted scalar defect.

Fix the outer loop to iterate exactly n times. Extract nruh into a local variable before the inner loop to make it clear the value is read once and used as a sanitized loop bound.

### DIFF
--- a/nvme-print-json.c
+++ b/nvme-print-json.c
@@ -2527,8 +2527,9 @@ static void json_nvme_fdp_configs(struct nvme_fdp_config_log *log, size_t len)
 
 	obj_add_uint(r, "n", n);
 
-	for (int i = 0; i < n + 1; i++) {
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3728, opened 2026-08-04
nvme: fix negative errno passed to libnvme_strerror in get_register_properties

bellis-ibm:
libnvme_exec_admin_passthru() can return a negative errno value.
Passing that negative value directly to libnvme_strerror(), which
expects a non-negative code, produces an incorrect error string.
Negate err before passing it to libnvme_strerror() to produce the
correct human-readable error string.


igaw:
Thanks!

PR #3729, opened 2026-08-04
libnvme: check return value of libnvmf_tid_set_identity in test

bellis-ibm:
The call to libnvmf_tid_set_identity() in
test_tid_setter_invalidates_cache() did not check the return value,
unlike all other call sites in the same file. Add the check so a
failure is reported explicitly rather than silently producing a
misleading cache-invalidation test result.


igaw:
Thanks!

PR #3732, opened 2026-08-04
nvme / libnvme: rename discovery to discover

martin-belanger:
@igaw - Per you request, I reviewed all the instances of "discover" and "discovery".  

Three commits, each independent -- feel free to accept/reject them separately:

1. `fabrics_discovery()`, `libnvmf_discovery()`, `libnvmf_discovery_nbft()`, and the internal `_nvmf_discovery()` are action functions -- to be consistent with sibling action functions like `libnvmf_connect()`/`libnvmf_disconnect_ctrl()`, they need to be `libnvmf_discover()` etc. "discovery" stays everywhere it's a noun (Discovery Controller, Discovery Log, etc).
2. A documentation wording fix in `nvme-discover.txt` found along the way.
3. While auditing the whole codebase for discover/discovery spelling, found an unrelated pre-existing issue: `nvme_host_discover_log` and three sibling structs were missing the "y" against their own kdoc titles and every function that touches them. Confirmed against the actual specs (TP8019 for the AVE Discovery Log, Base Spec 2.3 for the Host Discovery Log) that "Discovery" is correct. Predates this rename entirely.

If you'd rather have these as a single commit, GitHub's "Squash and merge" handles that on merge.

### DIFF
--- a/Documentation/nvme-discover.txt
+++ b/Documentation/nvme-discover.txt
@@ -91,9 +91,9 @@ OPTIONS
 	Suppress already connected errors.
 
 --force::
-	Disable the built-in persistent discover connection rules.
-	Combined with --persistent flag, always create new
-	persistent discovery connection.
+	Disable the built-in persistent discovery controller connection
+	rules. Combined with --persistent flag, always create new
+	persistent discovery controller connection.
 
 --nbft::
 	Only look at NBFT tables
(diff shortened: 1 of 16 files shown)


igaw:
Thank you very much. I am glad you did that. Glad the API is now streamlined.

PR #3733, opened 2026-08-04
solidigm: remove trailing space from clear-pcie-correctable-errors name

jimmunn:
The command name string had a trailing space, so the argv word "clear-pcie-correctable-errors" never matched it and the sub-command could not be invoked. Drop the space so the command is reachable.


igaw:
Thanks!

Issue #3735, opened 2026-08-05
Replace open coded field decoding with libnvme's getters


igaw:
> Ideally, the nvme-print code would be updated with something like
> 
> ```diff
>  static void stdout_id_ctrl_sanicap(__le32 ctrl_sanicap)
>  {
>         __u32 sanicap = le32_to_cpu(ctrl_sanicap);
> -       __u32 rsvd4 = (sanicap & 0x1FFFFFF0) >> 4;
> -       __u32 vers = (sanicap & 0x8) >> 3;
> -       __u32 ows = (sanicap & 0x4) >> 2;
> -       __u32 bes = (sanicap & 0x2) >> 1;
> -       __u32 ces = sanicap & 0x1;
> -       __u32 ndi = (sanicap & 0x20000000) >> 29;
> -       __u32 nodmmas = (sanicap & 0xC0000000) >> 30;
> +       __u32 vers = NVME_CTRL_SANICAP_VERS(sanicap);
> +       __u32 nvers = NVME_CTRL_SANICAP_NVERS(sanicap);
> +       __u32 ows = NVME_CTRL_SANICAP_OWS(sanicap);
> +       __u32 bes = NVME_CTRL_SANICAP_BES(sanicap);
> +       __u32 ces = NVME_CTRL_SANICAP_CES(sanicap);
> +       __u32 ndi = NVME_CTRL_SANICAP_NDI(sanicap);
> +       __u32 nodmmas = NVME_CTRL_SANICAP_NODMMAS(sanicap);
> ```
> 
> So we don't do the decoding open coded anymore.  

 _Originally posted by @igaw in [#3731](https://github.com/linux-nvme/nvme-cli/issues/3731#issuecomment-5188649071)_

PR #3738, opened 2026-08-05
nvme-print-stdout: use libnvme OAES print getters

ikegami-t:
Replace open coded field decoding.


ikegami-t:
A commit for the issue #3735.

igaw:
Excellent job! Thanks a lot!

PR #3739, opened 2026-08-05
solidigm-telemetry: check return value of recursive sldm_telemetry_structure_parse calls

bellis-ibm:
Two recursive call sites in sldm_telemetry_structure_parse ignored
the function's int return value, allowing silent continuation after a
parse failure. Wrap both calls in an if-guard and break out of the
enclosing loop on error.


igaw:
Thanks!

PR #3741, opened 2026-08-05
util: fix resource leak in dump_command_metadata()

sahmed-ibm:
The dump_command_metadata() function calls build_model() which heap-allocates model and its model->plugins sub-array. After json_program() emits the JSON output, both allocations are never freed before the function returns.

This results in model->plugins and model leaking on every successful invocation of the dump-command-metadata subcommand.

Free model->plugins and model after json_program() returns to prevent the resource leak.


igaw:
Thanks!

PR #3742, opened 2026-08-05
libnvme/fabrics: fix negative errno passed to libnvme_strerror in nbft_discovery

bellis-ibm:
nvme_discovery_log() returns a negative errno on failure, but the value
was passed directly to libnvme_strerror() which requires a positive errno.
Negate ret at the call site, consistent with all other libnvme_strerror()
calls in this file.


igaw:
Thanks!

PR #3743, opened 2026-08-05
util/argconfig: fix always-false range check in comma-sep array parser

bellis-ibm:
strtoumax() returns uintmax_t, so comparing the result against ULLONG_MAX
is always false and overflowed values are silently accepted. Add an errno
check after strtoumax() to catch values that exceed UINTMAX_MAX, fixing
the dead range check in the _long instantiation of the macro.


igaw:
Thanks!

PR #3744, opened 2026-08-05
libnvme: convert lazy getters to return error codes

martin-belanger:
A lazy getter could not distinguish an absent attribute from a failed read: a string getter returned NULL for both, and a numeric getter returned 0, indistinguishable from a real zero value. Every lazy getter now returns int (0, -ENOENT, or a negative errno) with the value delivered through an out-param.

Each getter also takes a caller-supplied default value, written to the out-param before any other logic runs. This lets callers use the getter directly in an expression without a separate declaration and initializer.

SWIG keeps the existing Python property syntax; a real failure now raises NvmeError instead of looking like absence.


igaw:
Thanks! I am sure having proper error handling even on such simple interfaces will help in the long run.

Issue #3660, opened 2026-07-28
CI: Add runner for end to end on FreeBSD


igaw:
Start looking into supporting FreeBSD in upstream. First step is to get the build running. For this a CI setup is necessary.

igaw:
The ci-containers project has gained support for creating FreeBSD images: 
https://github.com/linux-nvme/ci-containers/pull/71

Missing tasks is to hook it up into kubevirt-action:
https://github.com/linux-blktests/blktests-ci/issues/46

PR #3745, opened 2026-08-05
Get initial FreeBSD support rolling

dwsuse:
Usually, porting to a new platform, architecture, or compiler uncovers generic problems throughout the codebase. The initial porting attempt to FreeBSD is no exception. We need to fix generic build dependencies and add dummy implementations for FreeBSD. These dummy implementations require further review, as some parts can and should be shared with Linux.

### DIFF
--- a/libnvme/src/nvme/ioctl-freebsd.c
+++ b/libnvme/src/nvme/ioctl-freebsd.c
@@ -1,0 +1,269 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ * Copyright (c) 2026 SUSE Software Solutions
+ *
+ * Authors: Daniel Wagner <dwagner@suse.de>
+ */
+#include <errno.h>
+#include <stddef.h>
+#include <string.h>
+
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+
+#include <shared/compiler-attributes-util.h>
+
+#include <libnvme.h>
+
+#include "private.h"
+#include "loopback.h"
+
+/*
+ * FreeBSD's nvme(4) driver implements the same "Linux compatible" NVMe
+ * ioctls as the Linux kernel (NVME_IOCTL_ID, NVME_IOCTL_ADMIN_CMD,
+ * NVME_IOCTL_IO_CMD, NVME_IOCTL_RESET; see sys/dev/nvme/nvme_linux.h in
+ * the FreeBSD source), so this mirrors ioctl-linux.c almost verbatim.
+ * Two things FreeBSD doesn't have: the 64-bit passthru ioctls (only the
+ * 32-bit struct is supported, which the existing ioctl32/64 probing in
+ * ioctl_admin_passthru()/ioctl_io_passthru() already falls back away
+ * from via -ENOTTY), and Linux's BLKBSZSET/BLKRRPART block-device
+ * ioctls.
+ */
+
+static int nvme_verify_chr(struct libnvme_transport_handle *hdl)
+{
+	static struct stat nvme_stat;
+	int err = fstat(hdl->fd, &nvme_stat);
+
+	if (err < 0)
+		return -errno;
+
+	if (!S_ISCHR(nvme_stat.st_mode))
+		return -EINVAL;
+	return 0;
+}
+
+__shr_public int libnvme_reset_subsystem(
+		struct libnvme_transport_handle *hdl)
+{
+	int ret;
+
+	ret = nvme_verify_chr(hdl);
+	if (ret)
+		return ret;
+
+	ret = ioctl(hdl->fd, LIBNVME_IOCTL_SUBSYS_RESET);
+	if (ret < 0)
+		return -errno;
+	return ret;
+}
+
+__shr_public int libnvme_reset_ctrl(struct libnvme_transport_handle *hdl)
+{
+	int ret;
+
+	ret = nvme_verify_chr(hdl);
+	if (ret)
+		return ret;
+
+	ret = ioctl(hdl->fd, LIBNVME_IOCTL_RESET);
+	if (ret < 0)
+		return -errno;
+	return ret;
+}
+
+__shr_public int libnvme_rescan_ns(struct libnvme_transport_handle *hdl)
+{
+	int ret;
+
+	ret = nvme_verify_chr(hdl);
+	if (ret)
+		return ret;
+
+	ret = ioctl(hdl->fd, LIBNVME_IOCTL_RESCAN);
+	if (ret < 0)
+		return -errno;
+	return ret;
+}
+
+__shr_public int libnvme_get_nsid(
+		struct libnvme_transport_handle *hdl, __u32 *nsid)
+{
+	__u32 tmp;
+
+	errno = 0;
+	tmp = ioctl(hdl->fd, LIBNVME_IOCTL_ID);
+	if (errno)
+		return -errno;
+
+	*nsid = tmp;
+	return 0;
+}
+
+__shr_public int libnvme_update_block_size(
+		__shr_unused struct libnvme_transport_handle *hdl,
+		__shr_unused int block_size)
+{
+	/* FreeBSD has no equivalent of Linux's BLKBSZSET/BLKRRPART. */
+	return -ENOTSUP;
+}
+
+/*
+ * The 64 bit version is the preferred version to use, but for backwards
+ * compatibility keep a 32 version.
+ */
+static int ioctl_passthru32(struct libnvme_transport_handle *hdl,
+		unsigned long ioctl_cmd, struct libnvme_passthru_cmd *cmd)
+{
+	struct linux_passthru_cmd32 cmd32 = {};
+	void *user_data;
+	int err = 0;
+
+	user_data = hdl->submit_entry(hdl, cmd);
+	if (hdl->ctx->dry_run)
+		goto out;
+
+	memcpy(&cmd32, cmd, offsetof(struct linux_passthru_cmd32, result));
+	cmd32.result = 0;
+
+	do {
+		err = ioctl(hdl->fd, ioctl_cmd, &cmd32);
+		if (err >= 0)
+			break;
+		err = -errno;
+	} while (hdl->decide_retry(hdl, cmd, err));
+
+out:
+	cmd->result = cmd32.result;
+	hdl->submit_exit(hdl, cmd, err, user_data);
+	return err;
+}
+
+/*
+ * Not supported by FreeBSD's nvme(4) Linux-compatible ioctls today, but
+ * kept symmetric with ioctl-linux.c: the 64-bit ioctl numbers below are
+ * simply never claimed by the driver, so this returns -ENOTTY and the
+ * generic probing logic in ioctl_admin_passthru()/ioctl_io_passthru()
+ * permanently falls back to the 32-bit path on first use.
+ */
+static int ioctl_passthru64(struct libnvme_transport_handle *hdl,
+		unsigned long ioctl_cmd, struct libnvme_passthru_cmd *cmd)
+{
+	void *user_data;
+	int err = 0;
+
+	user_data = hdl->submit_entry(hdl, cmd);
+	if (hdl->ctx->dry_run)
+		goto out;
+
+	do {
+		err = ioctl(hdl->fd, ioctl_cmd, cmd);
+		if (err >= 0)
+			break;
+		err = -errno;
+	} while (hdl->decide_retry(hdl, cmd, err));
+
+out:
+	hdl->submit_exit(hdl, cmd, err, user_data);
+	return err;
+}
+
+static int ioctl_io_passthru(struct libnvme_transport_handle *hdl,
+		struct libnvme_passthru_cmd *cmd)
+{
+	int err;
+
+	if (hdl->ioctl_io_state == IOCTL_STATE_IOCTL64)
+		return ioctl_passthru64(hdl,
+			LIBNVME_IOCTL_IO64_CMD, cmd);
+
+	if (hdl->ioctl_io_state == IOCTL_STATE_IOCTL32 ||
+			!hdl->ctx->ioctl_probing)
+		goto do_ioctl32;
+
+	err = ioctl_passthru64(hdl, LIBNVME_IOCTL_IO64_CMD, cmd);
+	if (err >= 0 || err != -ENOTTY) {
+		hdl->ioctl_io_state = IOCTL_STATE_IOCTL64;
+		return err;
+	}
+
+	hdl->ioctl_io_state = IOCTL_STATE_IOCTL32;
+
+do_ioctl32:
+	return ioctl_passthru32(hdl, LIBNVME_IOCTL_IO_CMD, cmd);
+}
+
+static int ioctl_admin_passthru(struct libnvme_transport_handle *hdl,
+		struct libnvme_passthru_cmd *cmd)
+{
+	int err;
+
+	if (hdl->ioctl_admin_state == IOCTL_STATE_IOCTL64)
+		return ioctl_passthru64(hdl,
+				LIBNVME_IOCTL_ADMIN64_CMD, cmd);
+
+	if (hdl->ioctl_admin_state == IOCTL_STATE_IOCTL32 ||
+			!hdl->ctx->ioctl_probing)
+		goto do_ioctl32;
+
+	err = ioctl_passthru64(hdl, LIBNVME_IOCTL_ADMIN64_CMD, cmd);
+	if (err >= 0 || err != -ENOTTY) {
+		hdl->ioctl_admin_state = IOCTL_STATE_IOCTL64;
+		return err;
+	}
+
+	hdl->ioctl_admin_state = IOCTL_STATE_IOCTL32;
+
+do_ioctl32:
+	if (cmd->opcode == nvme_admin_fabrics)
+		return -ENOTSUP;
+
+	return ioctl_passthru32(hdl, LIBNVME_IOCTL_ADMIN_CMD, cmd);
+}
+
+__shr_public int libnvme_exec_admin_passthru(
+		struct libnvme_transport_handle *hdl,
+		struct libnvme_passthru_cmd *cmd)
+{
+	if (!hdl)
+		return -ENODEV;
+
+	if (!cmd->timeout_ms && hdl->timeout)
+		cmd->timeout_ms = hdl->timeout;
+
+	switch (hdl->type) {
+	case LIBNVME_TRANSPORT_HANDLE_TYPE_DIRECT:
+		return ioctl_admin_passthru(hdl, cmd);
+	case LIBNVME_TRANSPORT_HANDLE_TYPE_MI:
+		return libnvme_mi_admin_admin_passthru(hdl, cmd);
+	case LIBNVME_TRANSPORT_HANDLE_TYPE_LOOPBACK:
+		return __libnvme_loopback_admin_passthru(hdl, cmd);
+	default:
+		break;
+	}
+
+	return -ENOTSUP;
+}
+
+__shr_public int libnvme_exec_io_passthru(
+		struct libnvme_transport_handle *hdl,
+		struct libnvme_passthru_cmd *cmd)
+{
+	if (!hdl)
+		return -ENODEV;
+
+	if (!cmd->timeout_ms && hdl->timeout)
+		cmd->timeout_ms = hdl->timeout;
+
+	switch (hdl->type) {
+	case LIBNVME_TRANSPORT_HANDLE_TYPE_DIRECT:
+		return ioctl_io_passthru(hdl, cmd);
+	case LIBNVME_TRANSPORT_HANDLE_TYPE_LOOPBACK:
+		return __libnvme_loopback_io_passthru(hdl, cmd);
+	default:
+		break;
+	}
+
+	return -ENOTSUP;
+}
--- a/libnvme/src/nvme/lib-freebsd.c
+++ b/libnvme/src/nvme/lib-freebsd.c
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ * Copyright (c) 2026 SUSE Software Solutions
+ *
+ * Authors: Daniel Wagner <dwagner@suse.de>
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <libnvme.h>
+
+#include <shared/compiler-attributes-util.h>
+#include <shared/fs-util.h>
+
+#include "cleanup.h"
+#include "private.h"
+
+/*
+ * FreeBSD's nvme(4) driver exposes both the controller (/dev/nvmeX) and
+ * its namespaces (/dev/nvmeXnY) as character devices -- there is no
+ * separate block device node the way Linux has /dev/nvmeXnY as a block
+ * device, so only S_ISCHR is checked here.
+ */
+
+static int __libnvme_transport_handle_open_direct(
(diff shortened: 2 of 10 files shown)


dwsuse:
```shell
root@freebsd:~/nvme-cli # meson test -C .build
ninja: Entering directory `/root/nvme-cli/.build'
ninja: no work to do.
 1/41 nvme-cli:shared - array-util                         OK               0.01s
 2/41 nvme-cli:shared - base64                             OK               0.00s
 3/41 nvme-cli:shared - crc32                              OK               0.00s
 4/41 nvme-cli:shared - ini                                OK               0.00s
 5/41 nvme-cli:shared - parse-util                         OK               0.00s
 6/41 nvme-cli:shared - hash-util                          OK               0.00s
 7/41 nvme-cli:shared - string-util                        OK               0.01s
 8/41 nvme-cli:shared - io-util                            OK               0.01s
 9/41 nvme-cli:shared - net-util                           OK               0.00s
10/41 nvme-cli:shared - shr-assert                         OK               0.00s
11/41 nvme-cli:shared - fs-util                            OK               0.01s
12/41 nvme-cli:libnvme - check-public-symbols              OK               0.03s
13/41 nvme-cli:libnvme - cpp-dump                          OK               0.00s
14/41 nvme-cli:libnvme - cpp-misc                          OK               0.00s
15/41 nvme-cli:libnvme - uuid                              OK               0.00s
16/41 nvme-cli:libnvme - tree                              OK               0.00s
17/41 nvme-cli:libnvme - header/endian                     OK               0.00s
18/41 nvme-cli:libnvme - check-public-headers              OK               0.04s
19/41 nvme-cli:libnvme - header/fabrics                    OK               0.01s
20/41 nvme-cli:libnvme - header/ioctl                      OK               0.01s
21/41 nvme-cli:libnvme - header/lib                        OK               0.01s
22/41 nvme-cli:libnvme - header/lib-types                  OK               0.01s
23/41 nvme-cli:libnvme - header/mi                         OK               0.01s
24/41 nvme-cli:libnvme - header/nbft                       OK               0.01s
25/41 nvme-cli:libnvme - header/nvme-cmds                  OK               0.01s
26/41 nvme-cli:libnvme - header/nvme-cmds-base             OK               0.01s
27/41 nvme-cli:libnvme - header/nvme-cmds-fabrics          OK               0.01s
28/41 nvme-cli:libnvme - header/nvme-cmds-mi               OK               0.01s
29/41 nvme-cli:libnvme - header/nvme-cmds-nvm              OK               0.01s
30/41 nvme-cli:libnvme - header/nvme-cmds-zns              OK               0.01s
31/41 nvme-cli:libnvme - header/nvme-types                 OK               0.01s
32/41 nvme-cli:libnvme - header/nvme-types-base            OK               0.01s
33/41 nvme-cli:libnvme - header/nvme-types-fabrics         OK               0.01s
34/41 nvme-cli:libnvme - header/nvme-types-mi              OK               0.01s
35/41 nvme-cli:libnvme - header/nvme-types-nbft            OK               0.01s
36/41 nvme-cli:libnvme - header/nvme-types-nvm             OK               0.01s
37/41 nvme-cli:libnvme - header/nvme-types-zns             OK               0.01s
38/41 nvme-cli:libnvme - header/scan                       OK               0.01s
39/41 nvme-cli:libnvme - header/tree                       OK               0.01s
40/41 nvme-cli:libnvme - header/util                       OK               0.01s
41/41 nvme-cli:nvme-cli - command-metadata-schema          SKIP             0.04s   exit status 77

Ok:                40
Fail:              0
Skipped:           1

Full log written to /root/nvme-cli/.build/meson-logs/testlog.txt
```

dwsuse:
I've merged all cleanup patches. That means the "only" thing for FreeBSD support is to add code to these dummy implementation. I will not do this. Thus I am asking for someone who wants add the missing pieces.

bsdimp:
I've been maintaining the FreeBSD nvme-cli port for years with a small shim library. It compiled most things.  I implemented linux-compatible ioctls in FreeBSD, as well as moving to the linux naming convention for device nodes. The fork is at https://github.com/bsdimp/nmve-cli . The last version I ported was 2.11.

For that port, FreeBSD didn't have if_alg.h, so  I made it a meson detected thing, created stubs for nvme-rpmb, commented out some things that needed bigger changes, Made using LC_MEASUREMENT optional, removed an unneeded endian.h, implemented an open wrapper to catch opening sysfs stuff (minimally), added a needed libgen.h include, removed a few asm/byteorder.h includes, fixed the micron and wdc support to remove linuxisms, fixed the SN340 support (independent of the os), and created small wrappers for converting the linuxisms into freeBSD generically at http://github/bsdimp/freebsd-nvme-shim.

libnvme/src/nvme/mem-freebsd.c
```diff
@@ -0,0 +1,84 @@
...
+#include "mem.h"
+#include "private.h"
+
+/*
+ * FreeBSD: plain page-aligned allocations. Linux's transparent-hugepage
+ * mmap/madvise flags (MAP_HUGETLB, MADV_HUGEPAGE) have no FreeBSD
+ * equivalent, so libnvme_alloc_huge() always takes the small-allocation
+ * posix_memalign() path regardless of size -- correct, just without the
+ * performance benefit of real huge pages.
+ */
```
bsdimp, line 27: FreeBSD manages promotion to superpages automatically, so the performance penalties and benefits of huge pages are automatic. So it's unfair to say 'just without the performance benefits' here.

libnvme/src/nvme/lib-freebsd.c
```diff
@@ -0,0 +1,122 @@
...
+#include <compiler-attributes.h>
+
+#include "private.h"
+
+/*
+ * FreeBSD: opening a device node is plain POSIX and portable, so this
+ * mirrors lib-linux.c's approach without assuming Linux's "nvme%dn%d"
+ * naming scheme. Actual command submission is stubbed out in
+ * ioctl-freebsd.c.
+ */
```
bsdimp, line 29: FreeBSD supports both nvme%dn%d and nvme%dns%d for some time now.

libnvme/src/nvme/lib-freebsd.c
```diff
@@ -0,0 +1,122 @@
...
+		free(hdl);
+		return -ENOMEM;
+	}
+
+	if (!strncmp(name, "NVME_TEST_FD", 12)) {
+		hdl->type = LIBNVME_TRANSPORT_HANDLE_TYPE_DIRECT;
+		hdl->fd = LIBNVME_TEST_FD;
+
+		if (!strcmp(name, "NVME_TEST_FD64"))
+			hdl->ioctl_admin_state = IOCTL_STATE_IOCTL64;
```
bsdimp, line 95: FreeBSD does support the Linux ioctls, but does not support the *64 variants. In the past, these haven't been used by nvme-cli for anything I cared about.

libnvme/src/nvme/scan-freebsd.c
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#include <dirent.h>
+
+#include <compiler-attributes.h>
+
+#include <libnvme.h>
+
+#include "private.h"
+
```
bsdimp, line 20: Is there a spec for all these functions? They look to mirror the old sysfs walking that I too stubbed out.

libnvme/src/nvme/tree-freebsd.c
```diff
@@ -0,0 +1,258 @@
...
+		return -ENODEV;
+
+	return 0;
+}
+
+__shr_public const char *libnvme_ctrl_get_state(libnvme_ctrl_t c)
+{
+	char *state = c->state;
+
+	c->state = strdup("");
```
bsdimp, line 71: what state is supposed to be here?

libnvme/src/nvme/tree-freebsd.c
```diff
@@ -0,0 +1,258 @@
...
+	free(state);
+	return c->state;
+}
+
+__shr_public int libnvme_init_ctrl(__shr_unused libnvme_host_t h,
+		__shr_unused libnvme_ctrl_t c,
+		__shr_unused int instance)
+{
+	return -ENOTSUP;
```
bsdimp, line 80: Same for these.

util/hash-freebsd.c
```diff
@@ -0,0 +1,22 @@
...
+unsigned char *create_hmac_sha256(unsigned char *data, int datalen,
+		unsigned char *key, int keylen)
+{
+	return NULL;
+}
+
+unsigned char *create_md5(unsigned char *data, int datalen)
+{
+	return NULL;
+}
```
bsdimp: Both of these are in FreeBSD libraries in the base.

util/sighdl-freebsd.c
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Authors: Daniel Wagner <dwagner@suse.de>
+ */
+
+#include <signal.h>
+#include <errno.h>
+#include <stddef.h>
+
+#include "sighdl.h"
+
```
bsdimp: Might want to have a comment that FreeBSD is straight up POSIX for signals.

nvme-pci-ids-freebsd.c
```diff
@@ -0,0 +1,27 @@
...
+		const char *ctrl_name, char **sysfs_dir)
+{
+	return -ENOTSUP;
+}
+
+int __nvme_get_pci_ids(const char *sysfs_dir,
+		__u32 *vid, __u32 *did,
+		__u32 *subsys_vid, __u32 *subsys_did,
+		__u32 *class_code)
+{
```
bsdimp: sysfs_dir  is what?

libnvme/src/nvme/ctrl-sysfs-custom-freebsd.c
```diff
@@ -0,0 +1,51 @@
...
+	if (asprintf(&c->sysfs->dctype, "%u", id_ctrl.dctype) < 0)
+		return -ENOMEM;
+
+	return 0;
+}
+
+int libnvme_ctrl_load_phy_slot(__shr_unused struct libnvme_ctrl *c)
+{
+	/* FreeBSD has no PCIe physical slot sysfs equivalent. */
+	return 0;
```
bsdimp: Yes. This is only obtainable if you parse the dmidecode / smbios data that's around, but without a nice wrapper.

libnvme/src/nvme/ioctl-freebsd.c
```diff
@@ -0,0 +1,61 @@
...
+#include <compiler-attributes.h>
+
+#include <libnvme.h>
+
+#include "private.h"
+
+__shr_public int libnvme_reset_subsystem(
+		__shr_unused struct libnvme_transport_handle *hdl)
+{
+	return -ENOTSUP;
```
bsdimp: There's no subsystem level reset, so this is correct.

libnvme/src/nvme/ioctl-freebsd.c
```diff
@@ -0,0 +1,61 @@
...
+}
+
+__shr_public int libnvme_reset_ctrl(
+		__shr_unused struct libnvme_transport_handle *hdl)
+{
+	return -ENOTSUP;
+}
```
bsdimp, line 74: NVME_RESET_CONTROLLER

libnvme/src/nvme/ioctl-freebsd.c
```diff
@@ -0,0 +1,61 @@
...
+
+__shr_public int libnvme_rescan_ns(
+		__shr_unused struct libnvme_transport_handle *hdl)
+{
+	return -ENOTSUP;
+}
```
bsdimp, line 88: There's not one currently, so this is currect.

libnvme/src/nvme/ioctl-freebsd.c
```diff
@@ -0,0 +1,61 @@
...
+
+__shr_public int libnvme_get_nsid(
+		__shr_unused struct libnvme_transport_handle *hdl,
+		__shr_unused __u32 *nsid)
+{
+	return -ENOTSUP;
+}
```
bsdimp, line 102: NVME_GET_NSID

libnvme/src/nvme/ioctl-freebsd.c
```diff
@@ -0,0 +1,61 @@
...
+
+__shr_public int libnvme_update_block_size(
+		__shr_unused struct libnvme_transport_handle *hdl,
+		__shr_unused int block_size)
+{
+	return -ENOTSUP;
```
bsdimp, line 109: As far as I know, the FreeBSD driver handles this correctly by asking at open time.

libnvme/src/nvme/ioctl-freebsd.c
```diff
@@ -0,0 +1,61 @@
...
+}
+
+__shr_public int libnvme_exec_admin_passthru(
+		__shr_unused struct libnvme_transport_handle *hdl,
+		__shr_unused struct libnvme_passthru_cmd *cmd)
+{
```
bsdimp, line 228: NVME_PASSTHROUGH_CMD on the right nvme device node (the non-namespace one) or the NVME_IOCTL_ADMIN_CMD command that has the same semantics as Linux.

libnvme/src/nvme/ioctl-freebsd.c
```diff
@@ -0,0 +1,61 @@
...
+	return -ENOTSUP;
+}
+
+__shr_public int libnvme_exec_io_passthru(
+		__shr_unused struct libnvme_transport_handle *hdl,
+		__shr_unused struct libnvme_passthru_cmd *cmd)
+{
+	return -ENOTSUP;
```
bsdimp, line 268: NVME_PASSTHROUGH_CMD on the right nvme device node (the namespace one) or the NVME_IOCTL_OP_CMD command that has the same semantics as Linux.

bsdimp:
I left several comments, but I won't have a ton of time to write this myself for a while. I do likely have time for questions and review. I tried to find the right folks to chat with when I heard about the Windows port, but didn't fallow up on all the leads because the problems of the day got in the way.

igaw:
> I've been maintaining the FreeBSD nvme-cli port for years with a small shim library. It compiled most things. I implemented linux-compatible ioctls in FreeBSD, as well as moving to the linux naming convention for device nodes. The fork is at https://github.com/bsdimp/nmve-cli . The last version I ported was 2.11.

Thanks for the pointer. As far I can tell the current master is not far of to get native support for FreeBSD. The ioctl interface can just use the Linux implementation then. For nvme-pci devices the 32bit variant is good enough, the 64bit version is only necessary for fabrics. In this case we just disable the 64bit support on FreeBSD, that's easy. The interesting part is mostly, the sysfs replacement IMO. We have now an abstraction in libnvme the attribute layer which is on Linux sysfs and for Windows it uses nvme commands to read the attributes directly. I suppose this will be similar for FreeBSD. 

> For that port, FreeBSD didn't have if_alg.h, so I made it a meson detected thing, created stubs for nvme-rpmb, commented out some things that needed bigger changes, Made using LC_MEASUREMENT optional, removed an unneeded endian.h, implemented an open wrapper to catch opening sysfs stuff (minimally), added a needed libgen.h include, removed a few asm/byteorder.h includes, fixed the micron and wdc support to remove linuxisms, fixed the SN340 support (independent of the os), and created small wrappers for converting the linuxisms into freeBSD generically at http://github/bsdimp/freebsd-nvme-shim.

- 665c13adf558 wdc: On FreeBSD, get the timezone the same as with real glibc
  Is gated behind `NVME_HAVE_TM_GMTOFF` now
- 46a52714d0b3 sn340: Add support
  I suppose this is a backport, nothing to do?
- ffb4c34fdf9f nvme-cli: Make micron and wdc work on FreeBSD
  we touched this part during the Windows port, but likely it needs to addressed differently for FreeBSD, @bbusacker 
- 61720774cc71 meson.build
  the whole build should be fixed in master
- 91c101c0af7c  Hack because LC_MEASUREMENT isn't on FreeBSD
  needs to fixed
- 006779532540 Add libgen.h for basename
  basename  should be replaced shr_basename so it also works on Windows
- 56863b34e583 Don't redfine u8.
  still there, needs to be fixed
- 7be1b7a09566 Don't redefine LITTLE_ENDIAN
  big endian and little endian macros are gated now
- 0e1d96074db4 nvme.h: endian.h isn't needed
  fixed by the Windows port
- 21fe1b76c60e Don't nvme-rpmb.c when no AF_ALG
  fixed by using openSSL instead

So there only a few things left from your v2.11 port which are missing in master. I will look into getting all these changes into master.

And thanks for reviewing the patch, sorry it is ai slop... I asked to implement only empty dummies for an clean build on FreeBSD.

igaw:
> I left several comments, but I won't have a ton of time to write this myself for a while. I do likely have time for questions and review. I tried to find the right folks to chat with when I heard about the Windows port, but didn't fallow up on all the leads because the problems of the day got in the way.

Thanks for trying to get people involved. I'll deal with the list above. Hooking up the ioctl part should also be straight forward if we can just reuse the Linux implementation (with 32bit mode only). Maybe we an even reuse more bits from Linux there. This PR contains the full set of function which are platform depended.

My only real question is what needs to be done for the attributes, is there also a sysfs combat implementation? If not, then we could just add the nvme commands for it (identify ctrl, identify ns, etc).

dwsuse:
I've let claude just rebased and updated it to the latest master. There were a few significant changes in libnvme. And then asked to add the implementation to it and it got it even working:

```shell
root@freebsd:~/nvme-cli # .build/nvme list-subsys
nvme0 - NQN=nqn.2019-08.org.qemu:nvme-virtual-0
        hostnqn=nqn.2014-08.org.nvmexpress:uuid:00000000-0000-0000-0000-000000000000
\
 +- nvme0 pcie 0000:00:02.0
root@freebsd:~/nvme-cli # .build/nvme list-subsys -v
nvme0 - NQN=nqn.2019-08.org.qemu:nvme-virtual-0
        hostnqn=nqn.2014-08.org.nvmexpress:uuid:00000000-0000-0000-0000-000000000000
        iopolicy=
        model=undefined
        serial=
        firmware=
        type=nvm
\
 +- nvme0 pcie 0000:00:02.0
root@freebsd:~/nvme-cli # .build/nvme list
Node                  Generic               SN                   Model                                    Namespace  Usage                      Format           FW Rev
--------------------- --------------------- -------------------- ---------------------------------------- ---------- -------------------------- ---------------- --------
/dev/nvme0n1          nvme0n1               nvme-virtual-0       QEMU NVMe Ctrl                           0x1          1.07  GB /   1.07  GB    512   B +  0 B   11.0.2
root@freebsd:~/nvme-cli # .build/nvme id ctrl /dev/nvme0
NVME Identify Controller:
vid       : 0x1b36
ssvid     : 0x1af4
sn        : nvme-virtual-0
mn        : QEMU NVMe Ctrl
fr        : 11.0.2
...
```

So this is still a bit messy but working starting point.

igaw:
I think as soon we have #3660 figured out, this could go in and then let's improve it in tree. If it is not merged it will bit rot.

libnvme/src/nvme/mem-freebsd.c
dwsuse: Sure, is replaced with something more useful. Sorry bad claude...

libnvme/src/nvme/lib-freebsd.c
dwsuse: comment is updated accordingly.

libnvme/src/nvme/lib-freebsd.c
dwsuse: the documentation is updated. the 64bit version is only useful for fabrics commands. the nvme spec changed the result type from 32bit to 64bit...

libnvme/src/nvme/scan-freebsd.c
dwsuse: these scan function here are for scanning for devices in '/dev/*` and some parts of the sysfs. Iterating over the complete sysfs tree find all attributes. Since for FreeBSD the nvme commands are used to identify controllers and namespaces, the sysfs walking is not needed. Thus it's safe to just return success here.

libnvme/src/nvme/tree-freebsd.c
dwsuse: The controller state, this is mostly interesting for fabrics controllers, e.g. new, connecting, live, ....

It's safe to ignore it for now. Added a comment here.

libnvme/src/nvme/tree-freebsd.c
dwsuse: This function is used to initialize a freshly created libnvme_ctrl object usying the data from sysfs. Because the Identify command is used in this implementation, this not needed. 

This is only used for fabrics controllers.

util/hash-freebsd.c
dwsuse: great, this is gone.

util/sighdl-freebsd.c
dwsuse: this one is also gone :)

nvme-pci-ids-freebsd.c
dwsuse: indeed this is really ugly. I'll update it to 'source' and have documentation explaining what it means, it's platfrom dependend.

libnvme/src/nvme/ctrl-sysfs-custom-freebsd.c
dwsuse: comment updated for the time being.

libnvme/src/nvme/ioctl-freebsd.c
dwsuse: the documentation in the file is updated accordingly

libnvme/src/nvme/ioctl-freebsd.c
dwsuse: implemented

libnvme/src/nvme/ioctl-freebsd.c
dwsuse: okay :)

libnvme/src/nvme/ioctl-freebsd.c
dwsuse: also implemented.

libnvme/src/nvme/ioctl-freebsd.c
dwsuse: okay, this part is used after a namespace has been added/resized for Linux. It's a bit of a platform specific hack. Let's see how this goes.

libnvme/src/nvme/ioctl-freebsd.c
dwsuse: yep, implemented.

libnvme/src/nvme/ioctl-freebsd.c
dwsuse: also added.

igaw:
```shell
root@freebsd:~/nvme-cli # .build/nvme list -v
Subsystem        Subsystem-NQN                                                                                    Controllers
---------------- ------------------------------------------------------------------------------------------------ ----------------
nvme0            nqn.2019-08.org.qemu:nvme-virtual-0                                                              nvme0

Device           Orchestrator Cntlid SN                   MN                                       FR       TxPort Address        Slot   Subsystem    Namespaces
---------------- ------------ ------ -------------------- ---------------------------------------- -------- ------ -------------- ------ ------------ ----------------
nvme0            kernel       0      nvme-virtual-0       QEMU NVMe Ctrl                           11.0.2   pcie   (null)                nvme0        nvme0n1

Device            Generic              NSID       Usage                                             Format           Controllers
----------------- -------------------- ---------- ------------------------------------------------- ---------------- ----------------
/dev/nvme0n1      nvme0n1              0x1          1.07 GB /   1.07 GB (   1.00 GiB /    1.00 GiB) 512   B +  0 B   nvme0
```

bsdimp:
Oh! I'll have to add this to my testing queue.

igaw:
It should work for nvme-pci devices well. There isn't a lot of platform specific stuff left. BTW, feel free to critique the documentation, claude can be very wrong :) 

Anyway, it seems to work on my test VM. I'll hope to get the CI stuff sorted out soon, so this should be in fairly soon.

PR #3746, opened 2026-08-05
ocp: fix untrusted allocation size in parse_event_fifos()

sahmed-ibm:
The parse_event_fifos() function reads event_fifo_start and event_fifo_size directly from the device-supplied telemetry buffer and uses them to compute a FIFO data pointer and size without any bounds validation.

A malicious or corrupt device can supply out-of-range values, causing the computed pfifo_start pointer to exceed the telemetry buffer boundary. Passing an unbounded fifo_size to parse_event_fifo() then drives an inner loop well past the end of the buffer, resulting in an out-of-bounds read and potential heap corruption.

Validate fifo_offset and fifo_size against the known data area size from poffsets before using them to compute the FIFO pointer or calling parse_event_fifo().


igaw:
Thanks!

PR #3747, opened 2026-08-05
nvme: fix use-after-free of args->log in get_log_offset()

fhassan-ibm:
@args->log was assigned from @*log before calling
libnvme_realloc(), which may move the buffer to a new address, leaving @args->log pointing into freed memory.

Move the assignment of @args->log to after the realloc so it always derives from the updated @*log pointer, using the already saved @args->lpo as the offset.

Detected by Coverity: CID 557492 (Use after free, High).

### DIFF
--- a/nvme.c
+++ b/nvme.c
@@ -10730,7 +10730,7 @@ static int get_log_offset(struct libnvme_transport_handle *hdl,
 	if (!*log)
 		return -ENOMEM;
 
-	args->log = *log + args->lpo;
+	args->log = (char *)*log + args->lpo;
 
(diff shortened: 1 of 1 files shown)


igaw:
The commit message doesn't match what the patch does.

fhassan-ibm:
Updated the commit message to accurately reflect the change. The patch only adds a (char *) cast to make the void pointer arithmetic well-defined — the previous message incorrectly described moving the assignment, which was done in an earlier iteration. Sorry for the confusion.

igaw:
Thanks!

Just one thing: please avoid the `@` prefixing for variables. It's very unusual and t doesn't add any value in my opinion.

PR #3748, opened 2026-08-05
ocp: fix out-of-bounds access in json_c7_log()

fhassan-ibm:
res is a char pointer into the 458-byte res_arr buffer. After the first use block (encoding rsvd1[3] into res_arr), res is left pointing 3+ bytes past the start of res_arr.

The second use block then calls memset(res, 0, 458) from that advanced position, writing 3 bytes past the end of res_arr. The subsequent 456-iteration sprintf loop writes from the same advanced position, compressing the usable space.

Reset res = res_arr before the second memset so that both the memset and the following loop operate from the start of the buffer.

Addresses-Coverity: 557324

### DIFF
--- a/plugins/ocp/ocp-print-json.c
+++ b/plugins/ocp/ocp-print-json.c
@@ -1160,6 +1160,7 @@ static void json_c7_log(struct libnvme_transport_handle *hdl, struct tcg_configu
 				  le32_to_cpu(log_data->pro_rlc));
 	json_object_add_value_int(root, "TCG Error Count", le32_to_cpu(log_data->tcg_ec));
 
+	res = res_arr;
 	memset((__u8 *)res, 0, 458);
 	if (log_page_version == 1) {
 		res += sprintf(res, "%d%d", *(__u8 *)&log_data->no_of_ns_prov_locking_obj_ext,


plugins/ocp/ocp-print-json.c
```diff
@@ -1160,6 +1160,7 @@ static void json_c7_log(struct libnvme_transport_handle *hdl, struct tcg_configu
 				  le32_to_cpu(log_data->pro_rlc));
 	json_object_add_value_int(root, "TCG Error Count", le32_to_cpu(log_data->tcg_ec));
 
+	res = res_arr;
```
igaw, line 1163: this will overwrite the first three written bytes from line 1129

igaw (changes requested):

plugins/ocp/ocp-print-json.c
fhassan-ibm: You're right that res = res_arr would overwrite the first three bytes written in the loop at lines 1128–1129. However, those bytes are already captured into the JSON object via json_object_add_value_string(root, "Reserved1", res_arr) on line 1130 before the reset, so the buffer is safe to reuse from the start.

That said, if you have a cleaner approach in mind I'm happy to update the patch — please let me know what you'd prefer.

igaw: Oh, I missed that part.  Sorry about the noise.

Personally, I would introduce helper function which handle the append and not open code this stuff. Those random values all over the place is not a good style, error prone.

But that's not our problem :)

igaw:
Thanks!

PR #3750, opened 2026-08-05
netapp: fix null pointer dereference in netapp_ontapdevices_print_regular()

fhassan-ibm:
If format is neither NNORMAL nor NCOLUMN, formatstr remains NULL and is subsequently passed to printf(), causing a null pointer dereference.

Add an else-return guard so an unexpected format value causes an early return before formatstr is used.

Coverity-Id: 557362


igaw:
Thanks!

PR #3754, opened 2026-08-06
libnvme: Add lazy sysfs read for Paths and Namespaces

martin-belanger:
Follow-on to #3711/#3744: extends the lazy sysfs read pattern from controllers (`libnvme_ctrl`) to paths (`libnvme_path`) and namespaces (`libnvme_ns`).

Two commits, one per struct:

1. `libnvme_path` -- straightforward, same shape as the controller work.
2. `libnvme_ns` -- trickier. Some members differ between Linux and Windows, and on Linux the `lba_count`/`lba_util`/`meta_size` members are read differently depending on whether the kernel's `csi` sysfs attribute exists (added in 6.8), so presence of that one attribute gates how the others are read.

Measured against 4 real controllers (1 PCIe, 3 NVMe-oF/TCP), same method as #3711's controller measurement -- strace'd sysfs `openat` calls, before/after:

| Command | Before | After | Reduction |
|---|---|---|---|
| `nvme list` | 229 | 217 | -12 (-5.2%) |
| `nvme list-subsys` | 227 | 195 | -32 (-14.1%) |


libnvme/src/nvme/path-sysfs.c
```diff
@@ -0,0 +1,66 @@
...
+#include "path-sysfs.h"
+
+struct libnvme_path_sysfs {
+	char *ana_state;
+	char *numa_nodes;
+	int *grpid;
+	volatile int queue_depth;
+	volatile long multipath_failover_count;
+	volatile long command_retry_count;
+	volatile long command_error_count;
```
igaw: these 'volatile' are not necessary

libnvme/src/nvme/sysfs_accessors_specs.py
```diff
@@ -114,7 +124,86 @@
...
+    # struct libnvme_path itself, is a Linux-only concept. Windows still
+    # needs every getter to exist and link (an app must not need
+    # #ifdef _WIN32 to call them), so this is the simplest possible use
+    # of a per-OS override: nothing to resolve, just no source.
+    'members': [
+        {
+            'name': 'ana_state',
+            'attr': 'ana_state',
+            'type': 'char *',
+            'volatile': True,
```
igaw, line 158: volatile is most of the time wrongly used, see kernel documentation on this. if we really need the property of atomic read/write to the memory location we have to look into `READ_ONCE` and `WRITE_ONCE`

libnvme/src/nvme/path-sysfs-linux.c
```diff
@@ -0,0 +1,185 @@
...
+	str = libnvme_get_path_attr(c, "diag/command_error_count");
+	if (!str)
+		return -ENOENT;
+
+	if (sscanf(str, "%ld", &c->sysfs->command_error_count) != 1)
+		return -EINVAL;
+
+	*val = c->sysfs->command_error_count;
+	return 0;
+}
```
igaw, line 184: nice to see this code generated!

libnvme/test/sysfs/data/tree-pcie.out
```diff
@@ -4,7 +4,7 @@ host nqn.2014-08.org.nvmexpress:uuid:ce4fee3e-c02c-11ee-8442-830d068a36c6 hostid
       ctrl nvme0 transport=pcie traddr=0214:90:00.0
   subsystem nvme-subsys1 nqn.1994-11.com.samsung:nvme:PM1735a:2.5-inch:S6RTNE0R900057
     ns 1 nvme1n1
-      path nvme1c1n1 ana=optimized numa=0-1 qdepth=0
+      path nvme1c1n1 ana=- numa=0-1 qdepth=0
         ctrl nvme1 transport=pcie traddr=052e:78:00.0
-      path nvme1c2n1 ana=optimized numa=2-3 qdepth=0
+      path nvme1c2n1 ana=- numa=2-3 qdepth=0
```
igaw, line 9: ah, this is great! it's not lying anymore!

libnvme/src/nvme/ns-sysfs-custom-win.c
```diff
@@ -0,0 +1,211 @@
...
+	if (!id)
+		return -ENOMEM;
+
+	ret = libnvme_ns_identify(n, id);
+	if (ret)
+		return ret;
+
+	nvme_id_ns_flbas_to_lbaf_inuse(id->flbas, &flbas);
+
+	if (!SYSFS_IS_LOADED(n->sysfs->lba_size)) {
```
igaw, line 43: on windows there is not sysfs... that's why the `libnvme_ns_identify` is issued.

igaw: or does `SYSFS_IS_LOADED` just mean if the `lba_size` is valid?

In this case I am slightly confused by the name of `SYSFS_IS_LOADED`.

libnvme/src/nvme/ns-sysfs-custom-win.c
```diff
@@ -0,0 +1,211 @@
...
+ * Authors: Martin Belanger <Martin.Belanger@dell.com>
+ */
+
+#include <ccan/endian/endian.h>
+
+#include "mem.h"
+
+#include "ns-sysfs.c"
+
+#define GETSHIFT(x) (__builtin_ffsll(x) - 1)
```
igaw: maybe something for shared as I think there are a couple more users already in the code base.

igaw:
BTW, what about moving the generated code into a sub directory? We can do this after this series.

libnvme/src/nvme/path-sysfs.c
martin-belanger: You're right that volatile could be removed from the generated struct. Here volatile is used more as a hint that the generated code must always read the sysfs. The value in the sysfs changes over time and therefore cannot be cached into memory forever.

libnvme/src/nvme/sysfs_accessors_specs.py
martin-belanger: Same as previous. This is not the type volatile, but a hint for the generator that sysfs must be read every time the getter is invoked because the sysfs value changes over time.

libnvme/src/nvme/ns-sysfs-custom-win.c
martin-belanger: The lazy sysfs getters need to keep track of 3 possible states:
1. sysfs has never been read and the in-memory cached value in uninitialized
2. we tried to load the attr from sysfs, but the attr does not exist. In this case the in-memory cache is "loaded" with a special singleton pointer: `NO_SYSFS_ATTR`.
3. We read the attr from the sysfs and it is now "loaded" in the in-memory cache. So we can return that in-memory cache instead of reading the sysfs again.

`!SYSFS_IS_LOADED(p)` covers the 1st case
`SYSFS_IS_LOADED(p)` covers the 2nd and 3rd cases
`SYSFS_IS_ABSENT` is used to distinguish between the 2nd and 3rd cases.

libnvme/src/nvme/ns-sysfs-custom-win.c
martin-belanger: I will look into it

martin-belanger:
> BTW, what about moving the generated code into a sub directory? We can do this after this series.

We can certainly do that. Maybe the directory could be called `generated`. And this would give people a hint to refrain from manually editing those files. We can put all the generated code in there, not just the sysfs accessors.

I like the idea. But as you said, after the series land.

libnvme/src/nvme/path-sysfs.c
martin-belanger: addressed

libnvme/src/nvme/sysfs_accessors_specs.py
martin-belanger: addressed

libnvme/src/nvme/ns-sysfs-custom-win.c
martin-belanger: addressed. ccan already provides ilog32 which can be used instead of defining our own GETSHIFT.

martin-belanger:
@igaw - All comment reviews addressed.

martin-belanger:
@igaw - The failed CI workloads seem to be Github server issues.
```
Error: Service Unavailable
Error: Failed to resolve action download info.
```

igaw:
Thanks!

PR #3755, opened 2026-08-06
ocp: fix resource leak of fp in print_ocp_telemetry_normal

LJaved-IBM:
The print_ocp_telemetry_normal() function opens a file @fp for writing the telemetry output. On four error return paths inside the if (fp) block, the file is returned from without being closed: after parse_statistics() failure for DA1, after parse_event_fifos() failure for DA1, and after the same two failures for DA2.

The open file descriptor is leaked each time one of these error paths is taken.

Close @fp on all four error return paths before returning to prevent the resource leak.


igaw:
Thanks!

PR #3756, opened 2026-08-06
move non nvme specific helpers from util to shared

dwsuse:
The first part of the code reorganization is to move the non nvme specific code to shared. While at it also add more tests.

After this move util only contains code which is contains nvme specific parts. This code will be moved together with the all the code in the root dir under src.


igaw:
@bbusacker @shroffni @ikegami-t I've took the liberty to re-license from `GPL 2.0 or later` to `LGPL 2.1 or later`. I hope this is okay. As bonus you get free tests :)

shroffni:
> @bbusacker @shroffni @ikegami-t I've took the liberty to re-license from `GPL 2.0 or later` to `LGPL 2.1 or later`. I hope this is okay. As bonus you get free tests :)

I am fine with this change. Looks good to me.

ikegami-t:
Looks good. Thank you.

PR #3757, opened 2026-08-06
libnvme: Add missing lazy sysfs namespace attributes

martin-belanger:
Follow-on to #3754: converts the last three eager `libnvme_ns` fields -- `eui64`, `nguid`, `uuid` -- to the same lazy sysfs pattern.

Three commits:

1. Generator infra -- generalizes a `char *`-only special case in the sysfs-accessor generator to any pointer-typed member.
2. `eui64`/`nguid`/`uuid` -- needed a bit of custom handling since their sysfs parsing doesn't fit the generator's existing plain-attr/volatile-attr shapes.
3. Doc refresh -- `generate_sysfs_accessors.md`

Measured against 4 real controllers (1 PCIe, 3 NVMe-oF/TCP), same method as before -- strace'd sysfs `openat` calls, this time against a fully-eager baseline (before any of the lazy-sysfs work):

| Command | Before (eager) | After | Reduction |
|---|---|---|---|
| `nvme list` | 295 | 221 | -74 (-25.1%) |
| `nvme list-subsys` | 299 | 199 | -100 (-33.4%) |
| `nvme list -vv` | 299 | 255 | -44 (-14.7%) |


igaw:
Thanks!

PR #3758, opened 2026-08-06
libnvme: fix uninitialized buffer in libnvmf_tid_get_canonical()

LJaved-IBM:
If all fields of the TID are NULL, none of the APPEND() macro invocations write anything to @buf, leaving it uninitialized when passed to strdup(). Initialize @buf to an empty string to ensure strdup() always operates on valid data.


igaw:
Thanks!

PR #3759, opened 2026-08-06
libnvme/test: fix use after free in test_nvmf_sanitize_addrs()

LJaved-IBM:
If libnvme_create_ctrl() fails, @c is NULL. The subsequent calls to nvmf_sanitize_addrs() and libnvme_free_ctrl() are then made with a NULL pointer, resulting in a use after free.

Guard each nvmf_sanitize_addrs() and libnvme_free_ctrl() block with if (c) to skip the sanitize and free steps when controller creation failed.

### DIFF
--- a/libnvme/test/test-fabrics.c
+++ b/libnvme/test/test-fabrics.c
@@ -488,9 +488,11 @@ static bool test_nvmf_sanitize_addrs(struct libnvme_global_ctx *ctx)
 	params.subsysnqn = "nqn.2024-01.com.example:test";
 	params.traddr = "192.168.1.10";
 	params.host_traddr = "storage.example.com";
-	ret = libnvme_create_ctrl(ctx, &params, &c);
-	p = !ret && c && !strcmp(libnvme_ctrl_get_host_traddr(c),
-			"storage.example.com");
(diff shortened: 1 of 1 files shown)


libnvme/test/test-fabrics.c
```diff
@@ -494,12 +494,14 @@ static bool test_nvmf_sanitize_addrs(struct libnvme_global_ctx *ctx)
 	CHECK(p, "hostname host_traddr stored verbatim at creation");
 	pass &= p;
 
-	ret = nvmf_sanitize_addrs(ctx, c);
-	p = ret == -ENVME_CONNECT_TRADDR;
-	CHECK(p, "hostname host_traddr rejected at connect: %d", ret);
-	pass &= p;
+	if (c) {
```
igaw: I would just fail the complete test when `libnvme_create_ctrl` fails. No point to continue after this point, e.g.

```c
libnvme_create_ctrl(ctx, &params, &c);
shr_assert(c);
p = !strcmp(libnvme_ctrl_get_host_traddr(c), "storage.example.com");
```

Tests are allowed to completely fail on the first error.

igaw:
Looks good. Thanks!

PR #3760, opened 2026-08-06
util: fix double close of bin_file in read_binary_file

bellis-ibm:
bin_file was declared with __cleanup_file, which inserts an implicit
fclose() at every scope exit. The additional explicit fclose() call
after fread() caused a double close on every successful code path,
which is undefined behaviour. Remove the redundant explicit call and
let __cleanup_file handle it exclusively.


igaw:
Thanks!

PR #3761, opened 2026-08-06
libnvme/mi: fix overflowed constant in subsystem_health_status_poll

bellis-ibm:
The clear flag is encoded into bit 31 of CDW1 (the CLRS bit per
NVMe-MI spec section 5).  The original expression:
        req_hdr.cdw1 = (clear ? 1 : 0) << 31;
shifts a signed int literal left into its sign bit, which is
undefined behaviour in C.  Additionally, cdw1 is typed __le32
so a raw integer assignment skips the required endian conversion,
producing incorrect wire bytes on big-endian hosts (e.g. IBM POWER).

Fix both issues by using an unsigned literal and wrapping the
result with cpu_to_le32(), consistent with every other cdw
assignment in this file.

Addresses-Coverity-ID: 559299 (Overflowed constant)


igaw:
Thanks!

PR #3762, opened 2026-08-06
micron: clean up and standardize argument parsing and output format handling

brgoing:
* Consolidate on use of micron_parse_options for commands that need the model name.
* Clean up parse_and_open error handling
* Convert to using the global --output-format instead of a local --format argument.
* Default output format to "normal" instead of a mix of "normal" and "json".


igaw:
Thanks!

PR #3763, opened 2026-08-06
plugins/ymtc: fix unterminated string passed to strtol in ReadSysFile

bellis-ibm:
read() was called with sizeof(idstr) which could fill the entire
buffer leaving no room for a null terminator. Pass sizeof(idstr) - 1
instead so the zero-initialised final byte is always preserved as the
terminator that strtol requires.


igaw:
Thanks!

PR #3765, opened 2026-08-07
plugins/lm: fix NULL pointer check for fopen in lm_migration_recv

bellis-ibm:
When opening the output file in lm_migration_recv(), fopen() returns a
FILE pointer on success or NULL on failure. The code incorrectly checked
'if (fd < 0)', which is an integer file descriptor check pattern rather
than a pointer check.

Because pointer addresses are never negative, 'fd < 0' always evaluated
to false, causing fopen() failures to go undetected and subsequently
leading to a NULL pointer dereference in fwrite().

Fix the condition to check 'if (!fd)'.


igaw:
Thanks!

PR #3766, opened 2026-08-07
plugins/solidigm: fix integer overflow in wltracker_show_newer_entries

bellis-ibm:
In wltracker_show_newer_entries() and sldgm_get_workload_tracker(),
log->samplePeriodInMilliseconds and cfg.run_time_s are 32-bit unsigned
integers.

When multiplying log->samplePeriodInMilliseconds by 1000 or WLT2MS (25000)
to compute timestamps, the multiplication is performed using 32-bit 
arithmetic before being added/assigned to 64-bit variables (__u64). For
larger sample periods or runtimes, this results in an integer overflow and
wrap-around prior to promotion.
    
Cast log->samplePeriodInMilliseconds and cfg.run_time_s to __u64 prior to
multiplication to perform 64-bit precision arithmetic.
    

igaw:
Thanks!

PR #3767, opened 2026-08-07
plugins/huawei: fix unsigned return code check in huawei_list

bellis-ibm:
In huawei_list(), ret is declared as unsigned int.

The function calls validate_output_format(), which returns a signed int
(-EINVAL on failure). Because ret is unsigned, storing the negative
return code in ret converts it to a large positive value. As a result,
the check 'ret < 0' always evaluates to false, allowing invalid output
format errors to go undetected.
   
Declare ret as signed int so 'ret < 0' correctly catches negative error
codes.
    

igaw:
Just removed the whitespace changes in the author secction.

Thanks!

PR #3768, opened 2026-08-07
nvme: fix negative error formatting in wait_self_test

bellis-ibm:
In wait_self_test(), when nvme_identify_ctrl() fails, it returns a negative
error code (e.g. -EIO). The function previously called: 
nvme_show_error("identify-ctrl: %s", libnvme_strerror(err));

Passing a negative value to libnvme_strerror() is improper because it
expects a positive status/error code.

Use nvme_show_err(err, "identify-ctrl") instead, which properly formats
negative error values in line with the rest of the codebase.


igaw:
Thanks!

PR #3769, opened 2026-08-07
plugins/ocp: check return value of ocp_get_uuid_index

bellis-ibm:
In get_hwcomp_log_data(), ocp_get_uuid_index() was called without
checking its return value. If ocp_get_uuid_index() fails, uidx
remains uninitialized, causing log page requests to be constructed
with invalid UUID indices.
    
Check ret = ocp_get_uuid_index(hdl, &uidx) and return ret if it
fails, matching the error checking pattern used elsewhere in the OCP
plugin.
    

igaw:
Thanks!

PR #3770, opened 2026-08-07
plugins/seagate: fix dead code caused by checking flags & NORMAL

bellis-ibm:
In plugins/seagate/seagate-nvme.c, NORMAL is defined as 0 in enum
nvme_print_flags.
    
Checking 'if (flags & NORMAL)' performs a bitwise AND with 0, which
always evaluates to false. This caused all normal-mode output statements
throughout the Seagate plugin (including temp_stats, log_pages_supp,
vs_smart_log, vs_pcie_error_log, and stx_vs_fw_activate_history) to be
treated as unreachable dead code and suppressed.
    
Replace 'flags & NORMAL' with '!(flags & JSON)' so normal human-readable
output is correctly displayed when JSON format is not requested.
    

igaw:
Oops, my bad. Thanks for fixing it!

PR #3776, opened 2026-08-07
nvme: redesign --persistent, remove --epcsd/--no-epcsd

martin-belanger:
Redesigns discovery controller persistence around EPCSD, and removes the now-redundant `--epcsd`/`--no-epcsd` flags.

`--persistent[=no|auto|force]` replaces `--persistent`/`--no-persistent` and `--epcsd`/`--no-epcsd`. Bare `--persistent` means `auto`; omitted means `no`. In `auto` mode, persistence is decided per Discovery Log Page entry, driven entirely by that entry's own EPCSD flag -- `no` and `force` are explicit overrides of that default.

EPCSD only affects whether a connection is kept alive afterward, not discovery itself: `discover`/`connect-all` still fetch every Discovery Log Page and follow every referral regardless of EPCSD, exactly as before this change. An EPCSD=0 entry's Discovery Log Page is still fully processed; it's just disconnected once that's done, not skipped.

Four commits, in review order:
- `nvme: add OPT_STRING_OPTIONAL() for optional-argument options` -- generic CLI framework addition, no behavior change on its own.
- `nvme: split discovery.conf lines on whitespace only, not '='` -- fixes a latent tokenizer bug that would otherwise silently break a glued `--opt=value` inside a discovery.conf line.
- `nvme: redesign --persistent, remove --epcsd/--no-epcsd` -- the actual behavior change, in both libnvme and the CLI.
- `doc: document --persistent[=no|auto|force]`

### DIFF
--- a/src/fabrics.c
+++ b/src/fabrics.c
@@ -533,7 +533,13 @@ int nvmf_convert_discovery_line(struct libnvmf_config_emitter *emitter,
 	fa.tos = -1;
 	fa.ctrl_loss_tmo = NVMF_DEF_CTRL_LOSS_TMO;
 
-	while ((ptr = strsep(&p, " =\n")) != NULL && argc < MAX_DISC_ARGS - 1)
+	/*
+	 * Split on whitespace only, not '=' -- a glued "--opt=value" token
+	 * (e.g. "--persistent=force") must reach argconfig_parse() intact.
+	 * getopt_long() parses '=' within a single token natively for
+	 * every option shape, so splitting it out here is never required.
+	 */
+	while ((ptr = strsep(&p, "\n ")) != NULL && argc < MAX_DISC_ARGS - 1)
 		argv[argc++] = ptr;
 	argv[argc] = NULL;
 
--- a/src/fabrics.c
+++ b/src/fabrics.c
@@ -495,6 +495,15 @@ record_err:
 	}
 }
 
+const char *nvmf_resolve_persistent_arg(
+		struct argconfig_commandline_options *opts, const char *arg)
+{
+	if (!argconfig_parse_seen(opts, "persistent"))
+		return NULL;
+
+	return arg ? arg : "auto";
+}
+
 /*
  * Parse one discovery.conf line -- the same argv-style syntax 'nvme
  * discover'/'connect-all' accept, reusing NVMF_ARGS so every short and long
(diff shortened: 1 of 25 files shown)


igaw:
Just from the user point of view, this makes a lot of sense. So I personally would prefer the interface to the user over the for different flags. @Mr-Bossman any objections?

src/fabrics.c
```diff
@@ -551,10 +556,7 @@ int nvmf_convert_discovery_line(struct libnvmf_config_emitter *emitter,
 		return 0;
 
 	return nvme_config_convert_discovery_args(emitter, &fa,
-		persistent ? LIBNVMF_TRISTATE_TRUE :
-			no_persistent ? LIBNVMF_TRISTATE_FALSE : LIBNVMF_TRISTATE_UNSET,
-		epcsd ? LIBNVMF_TRISTATE_TRUE :
-			no_epcsd ? LIBNVMF_TRISTATE_FALSE : LIBNVMF_TRISTATE_UNSET);
+		nvmf_resolve_persistent_arg(persistent_arg));
```
igaw: The arg parser has `argconfig_parse_seen` which I would suggest to use here instead of the NVMF_PERSISTENT_NOT_GIVEN` approach. Then persistent_arg can just be initialized to `NULL`.

Just to keep the consistency.

martin-belanger: Addressed.

Mr-Bossman:
> Just from the user point of view, this makes a lot of sense. So I personally would prefer the interface to the user over the for different flags. @Mr-Bossman any objections?

This looks good. It seems to change the default behavior of `nvme connect` to `--persistent=no` rather than `--persistent=auto` like it was before, this isn't an issue just something to note.

src/fabrics.c
```diff
@@ -512,18 +521,13 @@ int nvmf_convert_discovery_line(struct libnvmf_config_emitter *emitter,
...
-			   "persistent discovery connection"),
-		  OPT_FLAG("no-persistent", 0, &no_persistent,
-			   "explicitly not a persistent discovery connection"),
-		  OPT_FLAG("epcsd",        0, &epcsd,
-			   "Explicit Persistent Connection Support for Discovery"),
-		  OPT_FLAG("no-epcsd",     0, &no_epcsd,
-			   "explicitly does not support Explicit Persistent Connection Support for Discovery"),
+		  OPT_STRING_OPTIONAL("persistent", 'p', "no|auto|force",
+			   &persistent_arg,
+			   "persistent discovery connection mode"),
```
Mr-Bossman: `"persistent discovery connection mode (Default: no)"`

martin-belanger: Just a small clarification. 
Default to "no" when `--persistent` is not provided. If `--persistent` is provided "bare" (ie. w/o any value), then the default is "auto".

I have modified the help text as follows:
```
persistent discovery connection mode (default: no; auto if given bare)
```

igaw:
I hope we got figured this now out :)

Thanks!

PR #3777, opened 2026-08-07
solidigm: fix unchecked return value in solidigm_config_get_nlog_formats()

sahmed-ibm:
The solidigm_config_get_nlog_formats() function looks up the "NLOG_FORMATS" key from the JSON config object using json_object_object_get_ex(). The return value of the call was not checked.

If "NLOG_FORMATS" is absent from the config, the lookup silently fails and nlog_formats remains NULL. While the caller guards against a NULL return, the function itself makes no distinction between a successful lookup that returned NULL and a failed one, violating the consistent error-checking contract of the codebase.

Check the return value and return NULL explicitly when the key is not found to make the error path consistent with all other call sites.


igaw:
Thanks!

PR #3778, opened 2026-08-07
nvme-print: revert outer loop bound change in json_nvme_fdp_configs()

sahmed-ibm:
Per NVMe spec section "5.2.12.1.29 Flexible Data Placement (FDP) Configurations (Log Page Identifier 20h)", the log->n field is a 0-based value. A value of 0 means one configuration, a value of 1 means two configurations, and so on. Therefore the loop must iterate n + 1 times to cover all configurations reported by the device.

The earlier change to use i < n was incorrect and is reverted here. The Coverity TAINTED_SCALAR report on this loop bound is a false positive: the loop bound i < n + 1 is correct per spec.

Fixes: a7bf8a9 ("nvme-print: fix untrusted loop bounds in json_nvme_fdp_configs()")

Partially reverts https://github.com/linux-nvme/nvme-cli/pull/3727

### DIFF
--- a/src/nvme-print-json.c
+++ b/src/nvme-print-json.c
@@ -2527,7 +2527,7 @@ static void json_nvme_fdp_configs(struct nvme_fdp_config_log *log, size_t len)
 
 	obj_add_uint(r, "n", n);
 
-	for (int i = 0; i < n; i++) {
+	for (int i = 0; i < n + 1; i++) {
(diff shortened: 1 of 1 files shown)


igaw:
Och, those 0-based indexes should have been banned in the spec. Thanks for checking after me :)

PR #3779, opened 2026-08-07
tree-fabrics: fix null pointer dereference in ctrl_lookups()

sahmed-ibm:
The ctrl_lookups() function calls libnvme_first_host() to get the first host entry, then immediately passes h to
libnvme_get_subsystem() without checking whether the return value is NULL.

libnvme_first_host() returns NULL when the host list is empty. Dereferencing a NULL h inside libnvme_get_subsystem() causes undefined behavior.

Add shr_assert(h) after the call to libnvme_first_host() to catch a NULL return and abort early, consistent with the existing check used after libnvme_create_host() in the same file.


igaw:
Thanks!

PR #3780, opened 2026-08-07
solidigm-telemetry: fix missing fallthrough annotations in debug info switch

sahmed-ibm:
The sldm_debug_info_parse() function uses an intentional fallthrough chain in its segment-ID switch to set tracker_log_name before reaching the shared tracker parsing body. The DEBUG_INFO_ID_TRACKER_INFO and DEBUG_INFO_ID_TRACKER_BUFFER cases fall through to the next case without a fallthrough annotation.

Unannotated fallthrough triggers -Wimplicit-fallthrough warnings from the compiler and is flagged by static analysis tools as a control flow defect (CID 557320, CID 557398).

Add fallthrough; after each intentional fall-through to match the convention used throughout the rest of the codebase and suppress the warning.


igaw:
Thanks!

PR #3781, opened 2026-08-07
solidigm-telemetry: check return value of sldm_telemetry_structure_parse

sahmed-ibm:
The return value of sldm_telemetry_structure_parse() was not being checked before proceeding to use the parsed_struct object for NLOG and side trace processing. If parsing fails, the incomplete structure data would lead to invalid results in downstream operations.

Check the return value and skip to the next iteration if parsing fails.


igaw:
Thanks!

PR #3782, opened 2026-08-07
sandisk: remove logically dead code in sndk_do_cap_telemetry_log

sahmed-ibm:
The function uses host_gen and ctrl_init flags to select which libnvme telemetry API to call. host_gen is initialized to 1 and only ever set to 0 in the CONTROLLER branch, which also sets ctrl_init to 1.

Because the BOTH and invalid-type branches return early, the only code paths that reach the dispatch block are HOST (ctrl_init=0, host_gen=1) and CONTROLLER (ctrl_init=1). The else branch calling libnvme_get_host_telemetry() is therefore unreachable - no defined telemetry type produces ctrl_init=0 and host_gen=0 simultaneously.

Remove host_gen and the dead libnvme_get_host_telemetry() branch. Replace the else if (host_gen) guard with a plain else, making the two-way dispatch match the two reachable states.

### DIFF
--- a/plugins/sandisk/sandisk-nvme.c
+++ b/plugins/sandisk/sandisk-nvme.c
@@ -42,7 +42,6 @@ static int sndk_do_cap_telemetry_log(struct libnvme_global_ctx *ctx,
 	struct nvme_telemetry_log *log;
 	size_t full_size = 0;
 	int err = 0, output;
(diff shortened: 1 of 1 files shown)


igaw:
Nice catch! Thanks!

PR #3783, opened 2026-08-07
test: fix integer overflow when setting SAVE bit in cdw10/cdw11

bellis-ibm:
Shifting a bool/int left by 31 bits without first promoting to
uint32_t invokes signed integer overflow (undefined behaviour in C). 
Cast the shift operand to uint32_t so the arithmetic happens in the 
unsigned domain, consistent with every other SAVE-bit expression 
in the file.

Fixes Coverity CIDs 557424 and 557388 (INTEGER_OVERFLOW).
Signed-off-by: Brooke Ellis <Brooke.Ellis.ibm.com>


igaw:
Thanks!

PR #3784, opened 2026-08-07
test: fix integer overflow in shift expressions in misc.c

bellis-ibm:
Several __u8, __u16, and bool values are shifted left by 16, 24, or
31 bits without first being promoted to uint32_t. In C, these shifts
happen on signed int, causing signed integer overflow (undefined
behaviour) when the result sets bit 31 or exceeds 16 bits.

Cast each shift operand to uint32_t so the arithmetic stays in the
unsigned domain, matching the pattern used elsewhere in the codebase.

Fixes Coverity CIDs 557419, 557377, 557460, 557498, 557509
(INTEGER_OVERFLOW).


igaw:
Thanks!

PR #3785, opened 2026-08-07
solidigm: remove logically dead code in parse_tracker_chunk_json

bellis-ibm:
entry_info is guaranteed non-NULL at this point because the
!entry_info guard above would have continued to the next loop
iteration if it were NULL. The ternary expression

  entry_info ? get_arg_description(entry_info, i) : NULL

therefore has a dead branch. Replace it with a direct call to
get_arg_description().

Fixes Coverity CID 557386 (DEADCODE).


igaw:
Thanks!

PR #3786, opened 2026-08-07
ocp: sanitize tainted ascii_id_length before memcpy

bellis-ibm:
ascii_id_length is a __u8 field read from device-supplied telemetry
data (pstring_buffer). Passing it directly as the memcpy size without
bounds checking allows a corrupt or malicious value to write beyond
the description buffer.

Replace 'memcpy(desc, src, len + 1)' with an explicit copy of len
bytes followed by a null terminator. Since ascii_id_length is __u8
(max 255) and all callers provide a 256-byte buffer, the copy and
terminator always fit without needing an additional clamp.

The same fix is applied consistently across all three string lookup
functions: get_statistic_id_ascii_string(), get_event_id_ascii_string()
and get_vu_event_id_ascii_string().

Fixes Coverity CIDs 557416, 557331, 557323 (TAINTED_SCALAR).

### DIFF
--- a/plugins/ocp/ocp-telemetry-decode.c
+++ b/plugins/ocp/ocp-telemetry-decode.c
@@ -635,7 +635,8 @@ int get_statistic_id_ascii_string(int identifier, char *description)
 				SIZE_OF_DWORD));
 
 			memcpy(description, pdescription,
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3787, opened 2026-08-07
libnvme: rename sysfs to attrs, move generated files to a subdir

martin-belanger:
Renames the lazy-caching accessor mechanism from "sysfs" to "attrs" terminology, and moves every generator-owned file into a new `src/nvme/generated/` subdirectory.

Most of this PR is purely a mechanical substitution of "sysfs" by "attrs". Not all sysfs were replaced by attrs though. I kept sysfs for things that really reference Linux's sysfs.

Two commits, in review order:
- `libnvme: rename sysfs to attrs and move generated files to a subdir` -- the lazy ctrl/path/ns/subsystem accessor rename and move.
- `libnvme: move accessors.[ch] and accessors-fabrics.[ch] to generated subdir`. Also fixes a real, non-obvious bug this move would otherwise cause: `tid.c`/`tid.h`, `exclusion.c`, and `discoverd/src/tid.h` all reference `accessors-fabrics.h` via the public `nvme/accessors-fabrics.h` path rather than a same-directory include, which was silently falling through to a stale copy in `/usr/include` from a previous install once the real file moved.


igaw:
Great improvement! Thanks a lot!

PR #3788, opened 2026-08-07
Enable nvme-mi-send/recv admin command support unconditionally and add Windows support

brgoing:
Makes the in-band nvme-mi-send / nvme-mi-recv CLI commands available regardless of CONFIG_MI (since they’re NVMe admin passthrough commands, not out-of-band MI transport), adds Windows (WinPE-only) passthrough support for those opcodes, and relocates the MI status string helper into the core libnvme utilities so it’s available without MI transport support.

Changes:
* Remove CONFIG_MI gating for nvme-mi-send / nvme-mi-recv command registration and implementation in the CLI.
* Move libnvme_mi_status_to_string() implementation into libnvme/src/nvme/util.c and export it from the base version script.
* Add WinPE handling for NVMe-MI Send/Recv admin opcodes in the Windows ioctl backend.

### DIFF
--- a/libnvme/src/libnvme-mi.ld
+++ b/libnvme/src/libnvme-mi.ld
@@ -32,7 +32,6 @@ LIBNVME_MI_3 {
 		libnvme_mi_scan_ep;
 		libnvme_mi_scan_mctp;
 		libnvme_mi_set_csi;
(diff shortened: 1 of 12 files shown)


igaw:
Thanks!

PR #3789, opened 2026-08-08
nvme-print-stdout: use libnvme TMPTHH print getter

ikegami-t:
Replace open coded field decoding.


ikegami-t:
Remaining fix for the issue: #3735.

igaw:
Thanks a lot for updating the nvme-cli code to use the library APIs. This is really great stuff!

PR #3791, opened 2026-08-08
nvme: move submit_io to src/io-cmds

ikegami-t:
This is to move other IO commands as same.


igaw:
Good idea! The `nvme.c` files is way to big, so splitting into smaller pieces is a good thing to do.

```shell
[28](https://github.com/linux-nvme/nvme-cli/actions/runs/31263410067/job/93117592364?pr=3791#step:5:229)
../src/io-cmds.c: In function 'submit_io':
../src/io-cmds.c:190:49: error: implicit declaration of function 'basename' [-Wimplicit-function-declaration]
  190 |                                                 basename(argv[optind]));
      |                                                 ^~~~~~~~
```

There is `shr_basename` which should fix this problem.

igaw:
One thing we need to figure out is how to split up the plugin registry. Currently, it needs to be in `nvme.c` which makes it a bit ugly. That's also why `main` is in nvme.c.

ikegami-t:
Just fixed the warning by including `libgen.h` instead of using `shr_basename`. If needed I will do change `basename` to `shr_basenem` by additional patch since currently `basename` used by other implementation also.
About the nvme.c yes I tried to change the static command entry function to global but seems not easy to do. So for now I am thinking to split the IO commands as same fabrics commands as only the static command entry functions are still remained in nvme.c. We may split the IO commands as same with the plugins commands but not sure.
Thank you.

ikegami-t:
Just pushed the commits to fix and update the commit and add a commit to add GL_ENT macro then the IO command functions can be used as a global function.

igaw:
Didn't have time to look into this PR. Next week, sorry about that.

ikegami-t:
No problem. Thank you.

igaw:
Hi, I thought a bit the old plugin infrastructure and decided it's time to get rid of the header include magic for the plugins. The new version is using plain C to register the plugins and follows more what the kernel does here. Anyway, I've merged #3848 and supersede this PR. I hope that's okay for you. 

Anyway, with my PR the nvme.c contents is not distributed into different files. We might have to move a few things around as I didn't spend too much time trying to figure out where it needs to go.

Thanks a lot for your input though!

ikegami-t:
The PR changes merged look nice! Thank you.

Issue #3774, opened 2026-08-07
Update libnvme APIs to spec v2.4


igaw:
The spec v2.4 has been released. Update the types and cmds header files accordingly. 

This is also necessary for #3713 (TP4201 post quantum crypto update)

igaw:
Base spec changes from ECN122, TP4150, ECN122 and TP6039 done.

igaw:
I suppose I need to feed all documents into claude to find the rest of the changes. The above ones seem a bit random although correct. If someone wants to do this, please go ahead.

My workflow is:
- download all specs/tp pdf documents mentioned in the changelog
- install poppler-utils (pdftotext, etc)
- tell claude to read the change log and figure out what is missing. ask for single step changes, e.g. do TP6093.

hreinecke:
I have converted the 2.4 docx version into docbook XML format; I guess you can feed that into claude with no issues. Say the word if you want to try.

igaw:
Don't bother, I've got this all automated reading the officially released PDFs.

PR #3792, opened 2026-08-08
nvme-types-base: add cross-controller reset

ikegami-t:
Update to spec v2.4.


ikegami-t:
Change for the issue: #3774.

igaw:
Thanks!

PR #3795, opened 2026-08-10
nvme: fix to set global variable log_level in logging

ikegami-t:
Previously changed to set only ctx log level so no logging outout.


igaw:
Right, I forgot that we still need to set the global variable. My plan was to get rid of it but we are apparently not there yet. 

Thanks for fixing my mistake!

ikegami-t:
Yes I will do try to use the ctx log level instead for the logging output later. Thank you.

PR #3796, opened 2026-08-10
libnvme: merge attr accessor files into one shared set

martin-belanger:
No functional change. This is the follow-on to #3787 (todo #55, part 2): merges the four per-struct lazy-attribute accessor file sets (`ctrl-attrs.*`, `path-attrs.*`, `ns-attrs.*`, `subsys-attrs.*`) into one shared `attr-accessors.*` set, and consolidates the hand-written OS-specific loader files the same way.

Generated files: `attr-accessors.{c,h,i}` (one merged output per kind instead of four), plus `attr-accessors-{linux,win}.c` for the one spec (`PATH_ATTRS`) with real per-OS getter divergence. `attr-accessors.ld` merges the four old `LIBNVME_{CTRL,PATH,NS,SUBSYS}_ATTRS_3` tags into one `LIBNVME_ATTR_ACCESSORS_3`.

Hand-written files: eight `*-attrs-custom-*.c` loader files collapse to two, `attr-accessors-custom-{linux,win}.c`. Along the way, dropped the separate `*-custom-fabrics.c`/`*-custom-no-fabrics.c` pair entirely -- `CONFIG_FABRICS` is never defined on Windows (`want_fabrics` in the top-level `meson.build` is gated on `host_system == 'linux'`), so `attr-accessors-custom-win.c` just hand-writes the no-op loader body directly with no `#ifdef`, and Linux's real `#ifdef CONFIG_FABRICS`/`#else` branch is inlined in `attr-accessors-custom-linux.c` instead of reaching for a file for one function.

`generate_attr_accessors.py` now merges every `ATTR_SPECS` entry's output into one shared file per kind (one preamble, not four) instead of one file set per spec -- mirrors the "concatenate parts under one shared preamble" pattern `generate_accessors.py` already uses for multiple structs from one annotated header. `attr_accessors_specs.py` spreads a `_SHARED` dict of filenames into every spec so they can't drift apart by typo; `main()` also asserts they agree before generating anything.

Verified the generated output is identical to before the merge (same function/symbol sets, diffed programmatically, not just inspected) before writing anything into the tree. Full build (default, `-Dwerror=true`, `-Dfabrics=disabled`), `meson test`, `-Dcheck-accessors=true`, and live checks against real hardware (`nvme list`, `list-subsys`, Python bindings) all pass.


igaw:
Very nice. This looks really good. Thanks!

PR #3802, opened 2026-08-10
scaleflux: fix divide-by-zero in nvme_parse_evtlog()

sahmed-ibm:
The nvme_parse_evtlog() function uses a progress spinner driven by a modulo expression against log_len divided by 100.

When log_len is less than 100, integer division produces zero, and the subsequent modulo operation is undefined behaviour.

Guard the modulo-based check with a log_len >= 100 condition so the divisor is only evaluated when it is non-zero. The final-byte check remains unconditional, ensuring the spinner fires even for small log buffers.


igaw:
Thanks!

PR #3804, opened 2026-08-10
plugins/ocp: Fixed propagation of nsid when getting error injection.

lgdacunh:
get-error-injection was not respecting --all-ns flag.


igaw:
Thanks!

PR #3806, opened 2026-08-11
build: enable the unit/c test suite on Windows

bbusacker:
The C unit tests under `tests/unit/c` are gated behind
`host_system == 'linux'`, but nothing in that directory is actually
Linux-specific — there are no platform conditionals in it at all, and
both binaries build cleanly with MinGW.

This series fixes the two real portability problems and drops the gate.

### `tests/unit/c: don't silently pass when stderr can't be redirected`

`test-argconfig-parse` redirects stderr to the bit bucket so the expected
parse errors don't clutter the log. Two issues on Windows:

- `"/dev/null"` doesn't exist natively; the spelling is `"NUL"`. Fixed by
  using the existing `DEV_NULL` from `src/common.h`, which already selects
  per-platform, rather than adding another switch.
- When `freopen()` failed the test printed an error but left `test_rc`
  untouched, so it reported success anyway. That's a latent bug on any
  platform, not just Windows.

### `build: enable the unit/c test suite on Windows`

`test-global-config.c` hardcoded `"/tmp"` for its throwaway `.conf` files.
On native Windows `"/tmp"` resolves to the root of the current drive, so it
only exists if somebody happened to create it there; on a drive without it
`mkstemp()` returns -1 and the test aborts. Now honours `TMPDIR`/`TMP`/`TEMP`
first with `"/tmp"` as the fallback, and uses the portable `shr_mkstemp()`
wrapper from `shared/`.

### Testing

Built and run on Windows with MSYS2 UCRT64 / MinGW gcc 16.1.0: full build
clean, `meson test` reports 51 passed / 0 failed / 1 skipped, including both
`argconfig_parse` and `global-config`. Linux behaviour is unchanged — the
only Linux-visible deltas are the `DEV_NULL` swap and the `test_rc` fix.

Each test was also mutation-tested to confirm it isn't inappropriately passing.

`checkpatch` is clean per patch (0 errors; the one warning is its
preference against `Co-authored-by:`, which is already used throughout
this tree).

### DIFF
--- a/tests/unit/c/test-argconfig-parse.c
+++ b/tests/unit/c/test-argconfig-parse.c
@@ -7,11 +7,13 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <ccan/array_size/array_size.h>
+
+#include "fs-util.h"
+
 #include "../src/argconfig.h"
 #include "../src/cleanup.h"
 #include "parse-util.h"
-
-#define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
 
 const char *libnvme_strerror(int errnum);
 
@@ -421,9 +423,11 @@ int main(void)
 
 	test_rc = 0;
 	setlocale(LC_NUMERIC, "C");
-	f = freopen("/dev/null", "w", stderr);
-	if (!f)
+	f = freopen(shr_dev_null(), "w", stderr);
+	if (!f) {
 		printf("ERROR: reopening stderr failed: %s\n", libnvme_strerror(errno));
+		test_rc = 1;
+	}
 
 	for (i = 0; i < ARRAY_SIZE(toval_tests); i++)
 		toval_test(&toval_tests[i]);
--- a/tests/unit/c/test-global-config.c
+++ b/tests/unit/c/test-global-config.c
@@ -20,6 +20,8 @@
 #include <string.h>
 #include <unistd.h>
 
+#include "fs-util.h"
+
 #include "../src/global-config.h"
 #include "../src/args.h"
 
@@ -30,13 +32,19 @@ struct nvme_args nvme_args = {
 	.supported_output_formats = NORMAL,
 };
 
+/*
+ * The throwaway .conf files land in the current directory, which meson
+ * sets to the build directory. A relative template keeps this off "/tmp",
+ * which isn't a usable temp directory on Windows.
+ */
 static char *write_temp(const char *content)
 {
-	char *path = strdup("/tmp/nvme-cli-conf-test-XXXXXX");
+	char *path = strdup("nvme-cli-conf-test-XXXXXX");
 	int fd;
 
 	shr_assert(path);
-	fd = mkstemp(path);
+
+	fd = shr_mkstemp(path);
 	shr_assert(fd >= 0);
 	shr_assert(write(fd, content, strlen(content)) == (ssize_t)strlen(content));
 	close(fd);
--- a/tests/unit/meson.build
+++ b/tests/unit/meson.build
@@ -3,7 +3,5 @@
 # Fast unit tests that need no NVMe hardware and no built nvme binary.
 # CLI-driven tests live in ../cli/; device tests live in ../e2e/.
 
-if host_system == 'linux'
-    subdir('c')
-endif
+subdir('c')
 subdir('py')


tests/unit/c/test-global-config.c
```diff
@@ -30,13 +34,40 @@ struct nvme_args nvme_args = {
 	.supported_output_formats = NORMAL,
 };
 
+/*
+ * Where to put the throwaway .conf files. "/tmp" only exists on Windows
+ * if someone happens to have created it on the current drive, so honour
+ * the environment's temp directory first and keep /tmp as the fallback.
+ */
+static const char *temp_dir(void)
```
igaw: This is something we need in other places as well (the unit test for INI is not running on Windows because of this). Could you add the temp dir handling and also the temp file creation to the `shared` project (including unit tests)? 

Basically get `shr_mkstemp` working out of the box :)

tests/unit/meson.build
```diff
@@ -3,7 +3,7 @@
 # Fast unit tests that need no NVMe hardware and no built nvme binary.
 # CLI-driven tests live in ../cli/; device tests live in ../e2e/.
 
-if host_system == 'linux'
+if host_system == 'linux' or host_system == 'windows'
```
igaw: I would remove the gate completely

igaw:
Glad we get the test coverage increased.

BTW, please drop the Co-Authored tag. Thanks!

tests/unit/c/test-argconfig-parse.c
```diff
@@ -421,9 +420,11 @@ int main(void)
 
 	test_rc = 0;
 	setlocale(LC_NUMERIC, "C");
-	f = freopen("/dev/null", "w", stderr);
-	if (!f)
+	f = freopen(DEV_NULL, "w", stderr);
```
igaw: I think it would make sense also to move the `DEV_NULL` from common.h to 
shared`. In fact I'd like to get rid of `common.h`. Too many places which do almost the same (common.h, nvme.h, utils.h)

dwsuse: I've merged the common.h header into shared. There is no `shr_dev_null()`.

tests/unit/c/test-global-config.c
bbusacker: Dropped the `temp_dir()` helper entirely — with `shr_mkstemp()` now in `shared/fs-util.h` the test doesn't need any temp-directory handling at all. It creates `nvme-cli-conf-test-XXXXXX` relative to the CWD (meson runs tests from the build directory), which is the same pattern `shared/test/test-fs-util.c` and `shared/test/test-table.c` already use. No `/tmp`, no `TMPDIR`/`TMP`/`TEMP` probing, and no `snprintf()` truncation check to get wrong.

On the INI test, two separate things keep it off Windows: `shared/test/meson.build` gates it behind `host_system != 'windows'`, and `test_file()` in `test-ini.c` uses `char path[] = "/tmp/nvme-ini-test-XXXXXX"` with plain `mkstemp()`. I have both fixed on a branch — un-gating `ini`, `crypto-util`, `sig-util` and `shr-assert`, plus the same relative-template `shr_mkstemp()` change in `test-ini.c` — and can send it as a follow-up series once this one lands.

tests/unit/meson.build
bbusacker: Done — the gate is gone entirely, `subdir('c')` is unconditional now.

tests/unit/c/test-argconfig-parse.c
bbusacker: Rebased onto dwsuse's "shared: distribute common.h to common code", so `common.h` is out of the picture here — this is now `freopen(shr_dev_null(), "w", stderr)`.

While I was in there I also dropped this file's hand-rolled `ARRAY_SIZE` in favour of `<ccan/array_size/array_size.h>`, the spelling used elsewhere in the tree. It needs to come from somewhere explicit once the test is built on Windows: the Windows headers provide `ARRAYSIZE`, not `ARRAY_SIZE`.

bbusacker: Thanks — rebased onto that. `shr_dev_null()` is in `shared/fs-util.h`, returning `"NUL"` from `fs-util-win.c` and `"/dev/null"` from `fs-util-linux.c`, so the test just calls it and the `common.h` include is gone.

Re-verified after the rebase: 52 Ok / 0 Fail / 1 skip on Windows (UCRT64/MinGW, gcc 16.1.0) and 119/119 on Linux, with each of the two patches building standalone.

bbusacker:
Force-pushed, rebased onto current master (`a44fffb2c`). Co-Authored-By tag dropped from both commits — thanks for the heads-up.

What changed since the last push:

- **`common.h` is gone from both tests.** Rebasing onto dwsuse's "shared: distribute common.h to common code" made that fall out naturally: `freopen(shr_dev_null(), ...)` in test-argconfig-parse, and `<ccan/array_size/array_size.h>` in place of that file's hand-rolled `ARRAY_SIZE`.
- **The `temp_dir()` helper is deleted.** test-global-config now creates its throwaway `.conf` relative to the CWD via `shr_mkstemp()`, so there's no temp-directory handling left in the test.
- **The meson gate is removed** rather than widened to `linux or windows`.

Net diff is now 3 files, +19/-9 — smaller than the version you reviewed.

Verification: Windows (UCRT64/MinGW, gcc 16.1.0) 52 Ok / 0 Fail / 1 skip, Linux 119/119, and each of the two patches builds standalone.

igaw:
Thanks!

PR #3809, opened 2026-08-11
nvme: tolerate unavailable huge pages during firmware download

### LINKED ISSUE #3793: fw-download fails on 16 KiB-page ARM64 kernel without HugeTLB/THP
### Problem

`nvme fw-download` exits with status 1 when given a firmware image of 512 KiB or larger on my Raspberry Pi 5. This kernel uses 16 KiB base pages and has no HugeTLB or THP support. The failure happens before nvme-cli sends a Firmware Image Download command to the controller.

This differs from #2848 because I cannot reserve huge pages as suggested in #2853. `/sys/kernel/mm/hugepages` does not exist, and `/proc/meminfo` has no HugePages or Hugepagesize entries.

### Environment

```
$ nvme version
nvme version 2.13 (git 2.13)
libnvme version 1.13 (git 1.13)

$ uname -a
Linux homeassistant.lan 6.18.39+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.18.39-1+rpt1 (2026-07-29) aarch64 GNU/Linux

$ getconf PAGE_SIZE
16384

$ ls /sys/kernel/mm/hugepages
ls: cannot access '/sys/kernel/mm/hugepages': No such file or directory
```

The firmware image is 2 MiB. I checked tags v2.14, v2.15, and v2.16. The same allocation code is present in all three. It is also present in current 3.0 development under `libnvme/src/nvme/mem-linux.c`.

### Reproduction

```
# nvme fw-download /dev/nvme0 -f firmware.bin -x 16384
# echo $?
1
```

Version 2.13 prints no error message. `strace` shows the failed allocations:

```
mmap(NULL, 2097152, PROT_READ|PROT_WRITE,
     MAP_PRIVATE|MAP_ANONYMOUS|MAP_HUGETLB, -1, 0) = -1 EINVAL (Invalid argument)
mmap(NULL, 4210688, PROT_READ|PROT_WRITE,
     MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x...
madvise(0x..., 2097152, MADV_HUGEPAGE) = -1 EINVAL (Invalid argument)
+++ exited with 1 +++
```

`nvme_alloc_huge()` treats the failed advisory `madvise()` call as fatal and frees the ordinary allocation. The function is now named `libnvme_alloc_huge()` in the 3.0 branch. The discussion in #2128 says that `MADV_HUGEPAGE` should not fail, but it returns `EINVAL` on this configuration.

### Validated workaround

I split the image into eight 256 KiB files so that each invocation stayed below `HUGE_MIN`. I then downloaded the pieces in order, passing the offset in DWORDs:

```
nvme fw-download /dev/nvme0 -f part-00 -x 16384 -O 0
nvme fw-download /dev/nvme0 -f part-01 -x 16384 -O 65536
nvme fw-download /dev/nvme0 -f part-02 -x 16384 -O 131072
# ... through part-07 at offset 458752
```

All eight commands returned `Firmware download success`. The controller accepted the assembled image with `nvme fw-commit -s 2 -a 0`, and the firmware activated successfully after a cold reset.

For this controller and firmware, nvme-cli did not need to map the whole file or use physically contiguous huge pages. Sending smaller sections with the correct offsets worked.

### Suggested behavior

I see two possible fixes:

1. Keep the aligned ordinary allocation when `madvise(..., MADV_HUGEPAGE)` fails, then let the kernel's DMA mapping determine whether it can use the buffer.
2. Read and send the firmware in transfer-sized sections instead of allocating the whole image at once.

If the allocator cannot be changed, the documentation should mention that some kernels do not expose `nr_hugepages` and show how to perform a segmented download.

Related: #2848, #2850, #2853, #2128.

m4r1k:
## Summary

`libnvme_alloc_huge()` currently discards its aligned fallback buffer when
`MADV_HUGEPAGE` fails. The advice is optional, so keep the allocation and let
the I/O path try to map it.

This also adds an opt-in `fw-download --stream` mode. It reads and submits one
transfer-sized portion at a time, which limits memory use and avoids allocating
a full-image buffer.

## Testing

- `meson compile -C /tmp/nvme-cli-build`
- `meson test -C /tmp/nvme-cli-build --print-errorlogs`: 104 passed, 2 expected failures, 2 skipped
- `checkpatch.pl`: 0 errors, 0 warnings
- Added an allocator test that forces `MAP_HUGETLB` and `MADV_HUGEPAGE` to fail
- Tested both the default and `--stream` paths on the affected 16 KiB-page ARM64 system with a Samsung PM991. Both firmware downloads completed successfully. No firmware commit or activation was performed during this test.

Fixes #3793

### DIFF
--- a/libnvme/tests/mem.c
+++ b/libnvme/tests/mem.c
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * Test the ordinary-memory fallback used when huge pages are unavailable.
+ */
+
+#include <errno.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include <compiler-attributes.h>
+#include <shr-assert.h>
+
+#include <libnvme.h>
+
+static bool madvise_called;
+
+void *mmap(void *addr, size_t length, int prot, int flags, int fd,
+	   off_t offset)
+{
+	if (flags & MAP_HUGETLB) {
+		errno = EINVAL;
+		return MAP_FAILED;
+	}
+
+	return (void *)syscall(SYS_mmap, addr, length, prot, flags, fd, offset);
+}
+
+int madvise(__shr_unused void *addr, __shr_unused size_t length,
+	    __shr_unused int advice)
+{
+	madvise_called = true;
+	errno = EINVAL;
+	return -1;
+}
+
+int main(void)
+{
+	struct libnvme_mem_huge mh = { 0 };
+	unsigned char *buf;
+
+	buf = libnvme_alloc_huge(2 * 1024 * 1024, &mh);
+	shr_assert(madvise_called);
+	shr_assert(buf);
+	shr_assert(mh.p == buf);
+	shr_assert(mh.libnvme_alloc);
+	shr_assert(mh.len == 2 * 1024 * 1024);
+	shr_assert(buf[0] == 0);
+	shr_assert(buf[mh.len - 1] == 0);
+
+	buf[mh.len - 1] = 0xa5;
+	libnvme_free_huge(&mh);
+	shr_assert(!mh.p);
+	shr_assert(!mh.len);
+
+	return 0;
+}
--- a/Documentation/nvme-fw-download.txt
+++ b/Documentation/nvme-fw-download.txt
@@ -12,6 +12,7 @@ SYNOPSIS
 			[--fw=<firmware-file> | -f <firmware-file>]
 			[--xfer=<transfer-size> | -x <transfer-size>]
 			[--offset=<offset> | -O <offset>]
+			[--stream]
 
 DESCRIPTION
 -----------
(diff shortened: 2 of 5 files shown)


libnvme/tests/mem.c
```diff
@@ -0,0 +1,60 @@
...
+{
+	if (flags & MAP_HUGETLB) {
+		errno = EINVAL;
+		return MAP_FAILED;
+	}
+
+	return (void *)syscall(SYS_mmap, addr, length, prot, flags, fd, offset);
+}
+
+int madvise(void *addr, size_t length, int advice)
```
igaw:
```
int madvise(__shr_unused void *addr, __shr_unused ....);
```

instead the void casts.

igaw:
A couple nitpicks:
- Could you wrap your commit message after 72 chars?
- Use `__shr_unused` instead of cast to shutdown the compiler from complaining.

The test needs to be gated so it only runs on Linux and not cross compile (`if host_system == 'linux' and (not meson.is_cross_build()) `)

m4r1k:
Thanks for the fast comment, in theory all done.

igaw:
Thank  your for reporting and fixing it so quickly! Highly appreciated!

PR #3813, opened 2026-08-11
libnvme: group discovery decision state and log it per entry

martin-belanger:
Each Discovery Log Page entry's controller, connect decision, disconnect decision, and already-connected state were separate locals and out-params threaded through dc_should_connect() and nvmf_connect_disc_entry(). Bundle them into struct dc_decision.

Add dc_log_decision(), logged at LIBNVME_LOG_DEBUG (-vv), covering every entry including early skips (duplicate info, transport mismatch, already connected) and the primary's own self-entry decision, so a customer bug report captured with -vv shows what was decided for every entry and why.


igaw:
Thanks!

PR #3815, opened 2026-08-11
libnvme/tests/ioctl: run the mock ioctl tests on Windows

bbusacker:
The `libnvme/tests/ioctl` suite is the only place command-building is checked
without a device, and it has been Linux-only because both halves of the harness
are: the mock is injected with `LD_PRELOAD`, and it intercepts the NVMe passthru
ioctl. Windows has neither.

This series adds a Windows interception layer so the existing `identify`,
`features` and `logs` tests run there against the same expectations, and fixes
the two result-handling bugs the new coverage exposed.

### How it works

libnvme reaches the driver through `DeviceIoControl`. `--wrap` cannot intercept
that -- libnvme is a DLL, so the call is bound through its own import thunk at
load time and never passes through a symbol the test executable links. What does
work is rewriting the thunk: `iat-hook.c` patches libnvme's Import Address Table
slot, which redirects the call without touching any code and leaves other modules
(including libnvme's own device enumeration) alone.

`mock-win.c` then decodes the Windows IOCTL back into the NVMe command it
represents and matches it against the next expected `struct mock_cmd`, exactly as
`mock.c` does. That forward translation in `ioctl-win.c` is lossy, so each decoder
records which fields it actually recovered -- and a bit mask where a field
survives only partly, e.g. Windows' 4-bit `LogSpecificField` against NVMe's 7.
Only recovered fields are compared; a discarded field is skipped rather than
compared against zero, which would turn a real mismatch into a false pass.

Where Windows genuinely cannot behave identically, the test says so instead of
being skipped: `win_err` and `win_no_ioctl` on `struct mock_cmd`, read through
`mock_err()` / `mock_ok()` / `mock_result()`. On Linux every one of those resolves
to the value the test asserted before. The mock rejects an expectation libnvme
could not have produced, so an undeclared divergence fails loudly.

`libnvme/tests/ioctl/README.md` documents all of this, including what a new
decoder has to do.

### Two fixes

- The Windows passthru entry points only assigned `cmd->result` on paths that
  retrieved completion data, leaving stale caller data in the field for any
  command that failed early. Linux always writes it back. Now both do.
- `sndk_do_sn861_drive_resize()` copied `admin_cmd.result` out unconditionally,
  including on error paths, and its caller left the receiving variable
  uninitialised.

### Scope

The other ioctl tests stay off Windows. They exercise families that reach the
driver by other routes -- SCSI CDBs for read, write and flush, dedicated
firmware, format and sanitize IOCTLs -- or that libnvme does not implement on
Windows at all. Each needs its own decoder, so they are left disabled rather than
reporting a pass they didn't earn.

### Verification

- Linux (Ubuntu 24.04, gcc): 119/119 tests pass, and per-test verdicts are
  identical to `upstream/master` -- the only difference in the list is
  registration order, since `subdir('ioctl')` moved ahead of `subdir('sysfs')`.
- Windows (MSYS2 UCRT64, gcc 16.1.0): 55 ok / 0 fail / 1 skipped, 0 compiler
  warnings, with `libnvme - identify`, `libnvme - features` and `libnvme - logs`
  passing.
- Each of the six commits builds standalone on Linux.
- checkpatch: 0 errors. Five of six commits are clean; the nine warnings on the
  last are all in `mock-win.c` and deliberate -- a Windows API function-pointer
  typedef (explained in a comment at the declaration), long `fail()` message
  strings, the `-ENOSYS` -> `ERROR_CALL_NOT_IMPLEMENTED` errno mapping, and a
  "falls through" phrase in prose that isn't a switch fallthrough.
- The IAT approach was verified on both x86_64 and aarch64.


libnvme/tests/ioctl/identify.c
```diff
@@ -621,6 +643,7 @@ static void test_identify_iocs_ns_csi_user_data_format(void)
 		.cdw11 = (TEST_FIDX << 0) | (TEST_CSI << 24),
 		.cdw14 = TEST_UUID,
 		.out_data = &expected_id,
+		WIN_CSI_UNSUPPORTED,
```
igaw, line 646: `.err = WIN_CSI_UNSUPPORTED` ?

libnvme/tests/ioctl/logs.c
```diff
@@ -204,6 +204,7 @@ static void test_get_log_cmd_effects(void)
 			 (((sizeof(expected_log) >> 2) - 1) << 16),
 		.cdw14 = (TEST_CSI << 24),
 		.out_data = &expected_log,
+		WIN_CSI_UNSUPPORTED,
```
igaw, line 207: same here is it `.err` or `.win_err`?

libnvme/design/tests/ioctl-mock.md
```diff
@@ -0,0 +1,83 @@
+# ioctl mock tests
```
igaw, line 1: There is `libnvme/design` folder. I suggest to add a sub folder tests and move  this document there, with a different name obviously.

igaw:
Thank for the effort. IMO the ioctl tests are not really bringing a lot to the table, they are mostly exercising the serializing/deserializing path which is platform neutral. So I am wondering if a Windows port is worth having. Don't get me wrong, I don't mind it as long it doesn't add additional maintenance burden. WDYT?

libnvme/tests/ioctl/identify.c
brgoing: WIN_CSI_UNSUPPORTED is a macro that sets .win_err = -ENOTSUP and .win_no_ioctl = true (defined in mock.h).  This pattern is used for 8 different tests, which is probably why it was implemented as a macro.  Would you prefer the settings to be explicit instead of using the macro?

brgoing:
> Thank for the effort. IMO the ioctl tests are not really bringing a lot to the table, they are mostly exercising the serializing/deserializing path which is platform neutral. So I am wondering if a Windows port is worth having. Don't get me wrong, I don't mind it as long it doesn't add additional maintenance burden. WDYT?

I agree that most of the value from the mock tests is gained by just running them on Linux, since the serializing/deserializing path is shared.  The main benefits received from having separate Windows ioctl tests are code coverage of ioctl-win.c and a check against unexpected changes to the ioctl-win implementation without the need for end-to-end testing.  Writing the tests has also helped highlight some of the differences in behavior vs Linux, and has provided an opportunity to better evaluate those differences and see if different design decisions could be made.

The commands that are currently being mocked and tested on Windows have been well tested on hardware and should be stable, so maintenance shouldn't be an issue from that side.  We _will_ need to make sure that any intended changes to ioctl-win.c that would change the expected behavior include a matching change to the tests.

Do you have any particular concerns regarding the maintenance of these tests?  If you don't think the benefits received merit the complexity of the tests, we can drop this and rely on end-to-end testing.

libnvme/design/tests/ioctl-mock.md
brgoing: Thanks for the feedback.  The README.md file is now design/tests/ioctl-mock.md.

libnvme/tests/ioctl/features.c
```diff
@@ -1472,6 +1473,7 @@ static void test_set_status_code_error(void)
 		.cdw11 = EVENTS,
 		.result = TEST_RESULT,
 		.err = TEST_SC,
+		.win_err = -EIO,    /* Windows returns EIO */
```
igaw, line 1476: One thing which bothers me is the special casing of the error codes. I think the Windows port has to go the extra mile to also mimick the error codes.

For example here, this is not correct. This tests case here wants to test the nvme protocol layer not the transport layer, thus there should only be the status code, not an error code.

libnvme/tests/ioctl/features.c
```diff
@@ -1521,6 +1556,7 @@ static void test_get_status_code_error(void)
 		.cdw10 = TEST_SEL << 8 | NVME_FEAT_FID_KATO,
 		.result = TEST_RESULT,
 		.err = TEST_SC,
+		.win_err = TEST_SC | NVME_SC_DNR,    /* Windows sets DNR */
```
igaw, line 1559: Again, I am bother about this different behavior here. Why is it different? This is a mocking tests, we explicitly want to see the error/status code which is set above. Let me re-read the argument why `.win_err` is necessary.

libnvme/tests/ioctl/identify.c
igaw: Yes, I prefer the explicit case because it avoids an lookup what `WIN_CSI_UNSUPPORTED` is when reading the test case. The tests cases are also open coded for the exact reason.

igaw:
> I agree that most of the value from the mock tests is gained by just running them on Linux, since the serializing/deserializing path is shared. The main benefits received from having separate Windows ioctl tests are code coverage of ioctl-win.c and a check against unexpected changes to the ioctl-win implementation without the need for end-to-end testing. Writing the tests has also helped highlight some of the differences in behavior vs Linux, and has provided an opportunity to better evaluate those differences and see if different design decisions could be made.

I agree having tests to catch platform depended behaviors would be great to have. Though I don't think it's good to mix this into the existing serialization/deserialization for the nvme commands.

> The commands that are currently being mocked and tested on Windows have been well tested on hardware and should be stable, so maintenance shouldn't be an issue from that side. We _will_ need to make sure that any intended changes to ioctl-win.c that would change the expected behavior include a matching change to the tests.

Okay, good to hear.

> Do you have any particular concerns regarding the maintenance of these tests? If you don't think the benefits received merit the complexity of the tests, we can drop this and rely on end-to-end testing.

My main concern was that we would have to deal with platform-specific behavior in the serialization/deserialization tests. I am fine with having dedicated tests for the platform abstraction. This is why I went ahead and already merged #3847. It was not meant to annoy you. I just realized that we no longer need the complex syscall intercept setup for it with the transport abstraction.

brgoing:
> > I agree that most of the value from the mock tests is gained by just running them on Linux, since the serializing/deserializing path is shared. The main benefits received from having separate Windows ioctl tests are code coverage of ioctl-win.c and a check against unexpected changes to the ioctl-win implementation without the need for end-to-end testing. Writing the tests has also helped highlight some of the differences in behavior vs Linux, and has provided an opportunity to better evaluate those differences and see if different design decisions could be made.
> 
> I agree having tests to catch platform depended behaviors would be great to have. Though I don't think it's good to mix this into the existing serialization/deserialization for the nvme commands.
> 
> > The commands that are currently being mocked and tested on Windows have been well tested on hardware and should be stable, so maintenance shouldn't be an issue from that side. We _will_ need to make sure that any intended changes to ioctl-win.c that would change the expected behavior include a matching change to the tests.
> 
> Okay, good to hear.
> 
> > Do you have any particular concerns regarding the maintenance of these tests? If you don't think the benefits received merit the complexity of the tests, we can drop this and rely on end-to-end testing.
> 
> My main concern was that we would have to deal with platform-specific behavior in the serialization/deserialization tests. I am fine with having dedicated tests for the platform abstraction. This is why I went ahead and already merged #3847. It was not meant to annoy you. I just realized that we no longer need the complex syscall intercept setup for it with the transport abstraction.

The new approach looks good.  We'll close this and evaluate how we want to handle these as platform-specific tests separate from the command serialization/deserialization tests.

PR #3816, opened 2026-08-12
netapp: fix heap-buffer-overflow in ONTAP TLV label parsing

itsmeut01:
The TLV parser in `netapp_get_ontap_labels()` trusts the length field from the
C2 vendor log data without bounds checking. A crafted or corrupted log with a
large TLV length (e.g. `0xFFFF`) causes out-of-bounds heap reads when computing
offsets for subsequent TLV entries (`netapp-nvme.c:232`).

Additionally, `ontap_labels_to_str()` passes the unchecked `label_len` directly
as the copy count, which can overflow the fixed-size `ONTAP_LABEL_LEN` (260 byte)
destination buffer.

### Fix

- Add bounds checks after each TLV length computation to ensure offsets stay
  within `ONTAP_C2_LOG_SIZE` (4096)
- Cast TLV length to `size_t` before multiplying by 4 to prevent integer overflow
- Cap copy length in `ontap_labels_to_str()` to `ONTAP_LABEL_LEN - 1`
- Change `label_len`, `i`, `j` from `int` to `size_t` for unsigned arithmetic

### Verification

Tested with a crafted 4096-byte C2 log binary with
`*(uint16_t *)&log_data[34] = 0xFFFF` (TLV length = 0xFFFF * 4 = 262140):

- **Before fix**: 1 invalid read detected by Valgrind at `netapp_get_ontap_labels()` — 210,232 bytes past the allocated block
- **After fix**: 0 errors, clean Valgrind run

### DIFF
--- a/plugins/netapp/netapp-nvme.c
+++ b/plugins/netapp/netapp-nvme.c
@@ -186,9 +186,12 @@ static void ontap_get_subsysname(char *subnqn, char *subsysname,
 		nvme_show_error("Unable to fetch ONTAP subsystem name");
 }
 
-static void ontap_labels_to_str(char *dst, char *src, int count)
+static void ontap_labels_to_str(char *dst, const char *src, size_t count)
 {
-	int i;
+	size_t i, max = ONTAP_LABEL_LEN - 1;
+
+	if (count > max)
+		count = max;
 
 	memset(dst, 0, ONTAP_LABEL_LEN);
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

TBH this parser should be completely rewritten, but then there is much more fun stuff to do :)

itsmeut01:
TS :-)

PR #3821, opened 2026-08-12
micron: fix logically dead code in micron_nand_stats()

sahmed-ibm:
The micron_nand_stats() function conditionally fetches the 0xFB log page only for models where eModel is not M5407 or M5410. The has_fb_log flag is therefore always false for M5410, making the M5410 branch of the spec ternary unreachable.

The dead condition can never be true, so spec is always assigned 1 regardless, which causes no incorrect output but is misleading and flagged as logically dead code by static analysis.

Replace the ternary with a direct assignment of 1 and update the comment to reflect that M5410 is excluded by the guard above.


igaw:
Thanks!

PR #3826, opened 2026-08-12
completions: generate zsh completion from command metadata

jimmunn:
Add zsh support to generate-completions.py, emitting the _nvme function from the same command metadata the bash generator already consumes, so the two shells stay in sync with the compiled binary.

Add test-zsh-completion.sh, which drives the generated _nvme function with the zsh completion builtins stubbed to capture what it offers.

Add TESTING.md, a cross-shell manual-test checklist and per-shell notes, since zsh renders completions through its own framework and needs more live-shell verification than the automated suite can cover.


completions/generate-completions.py
```diff
@@ -453,10 +718,22 @@ def main():
...
     ap.add_argument("input", nargs="?", default="-",
                     help="JSON model file, or '-' for stdin (default)")
-    ap.add_argument("--bash", metavar="FILE", required=True,
+    ap.add_argument("--bash", metavar="FILE",
                     help="write bash completion to FILE ('-' for stdout)")
+    ap.add_argument("--zsh", metavar="FILE",
+                    help="write zsh completion to FILE ('-' for stdout)")
     args = ap.parse_args()
 
+    requested = [(s, getattr(args, s)) for s in GENERATORS if getattr(args, s)]
```
bcapener, line 727:
```suggestion
    requested = [(s, getattr(args, s)) for s in GENERATORS if hasattr(args, s)]
```

bcapener (approved):

completions/generate-completions.py
jimmunn: argparse always creates the attribute defaulting to None, so hasattr(args, s) is unconditionally true — the current getattr(args, s) correctly filters on whether a path was actually supplied. Making this change actually makes some of the tests fail.

igaw:
Thanks!

The test plan will need updates when my https://github.com/linux-nvme/nvme-cli/pull/3824 PR lands.

Issue #3713, opened 2026-08-01
`keys gen-kxchap`: the DHHC-1 payload is the spec's *secret*, not a *key* — and there is no way to emit the representation of a given secret


kariya-mitsuru:
## Environment

- nvme-cli `master` @ `ab7358832710e789bc852df80bd40732a8b205a4` (version `3.0-b.4`)
- Reference: NVM Express Base Specification, Revision 2.3, sections 8.3.5.5.7 and 8.3.5.5.8

The same behaviour is present in 2.x, where the code lives in `gen_dhchap_key()` in `nvme.c` rather than in `gen_kxchap()` / `libnvmf_gen_kxchap_key()`.

## Summary

`DHHC-1:xx:<Base64>:` is defined by the specification as a representation of a **secret**. nvme-cli treats it as a representation of a **key**: it runs the secret-to-key transform at generation time and puts the transformed value in the Base64 payload.

The resulting string is still a well-formed, interoperable DHHC-1 secret representation, so **this is not an interoperability bug** — every consumer, including the Linux kernel, reads the payload as a secret and applies the transform exactly once. But the naming and the resulting behaviour are confusing in a way that has two concrete consequences, described below.

Raising this now because 3.0 is still in beta and the command is being renamed (`gen-dhchap-key` → `keys gen-kxchap`) anyway, so this seems like the natural moment to settle the terminology.

## 1. What the specification says

Section 8.3.5.5.8 is titled **"Secret Representation"**, and defines the format as:

> `DHHC-1:xx:<Base64 encoded string>:`
>
> - `xx` indicates the hash function to be used to transform **the secret** in key (refer to section 8.3.5.5.7) [...] The two ASCII characters "00" indicate no transform (i.e., use the secret as a key); and
> - The Base64 (refer to RFC 4648) string encodes **the secret** (32, 48, or 64 bytes binary) followed by the CRC-32 (refer to RFC 1952) of **the secret** (4 bytes binary).

Section 8.3.5.5.7 defines the transform that the consumer performs:

> transforming the provided secret into a key applying the HMAC function using the hash function **specified in the secret representation** [...] (i.e., `key = HMAC(secret, NQN || "NVMe-over-Fabrics")`)

So the payload is the secret, `xx` is an instruction to the consumer, and the key is derived by the consumer at use time. Nothing in the specification puts a key inside this string.

## 2. nvme-cli is internally inconsistent about this

[...]

PR #3827, opened 2026-08-12
keys: emit the KX-HMAC-CHAP secret representation, and name the commands after it

kariya-mitsuru:
Nine patches against `master` (`4d033951a`). Addresses #3713.

`nvme keys gen-kxchap` transforms the secret into a key before encoding it, so a
secret given with `-s` never reaches the output. This series emits it as it
stands (NVMe Base 2.4, section 8.3.4.5.8), lets the secret rather than the hash
id decide its length, fixes two ways a secret could come out different from the
one supplied, and moves the wording from *key* to *secret*.

More patches than the three sketched in #3713: re-reading my own work turned up
two ways a `-s` came out different from the one given, and `check-dhchap-key`
had the same `-k` trap `gen-dhchap-key` did. Three of the changes reach a
command line written against 2.x and are listed under NEWS.md's
incompatible-changes heading.

## Testing

- each commit builds on its own; `meson test` 109 ok / 2 expected fail / 0 fail
  at each
- `checkpatch.pl --strict`: 0 errors, warnings only on pre-existing over-length
  strings
- `tests/cli/nvme_keys_test.py` grows from 21 to 33 cases, five of them pinning
  the 2.x alias surface; each case pinning new behaviour was confirmed to fail
  on its parent commit
- `--hmac` x `--secret` x `--key-length` swept exhaustively against a 2.16 build:
  apart from the three incompatibilities, every combination either matches 2.16
  or turns one of its errors into a success


igaw:
Thanks a lot. I really appreciated to that this was spitted up into smaller steps. 

Well, that means I need to update blktests once again :)

PR #3828, opened 2026-08-12
libnnvme, nvme-cli: Discover rewrite

martin-belanger:
Rewrites libnvme's Discovery Controller walk (`_nvmf_discover()`, `libnvmf_discover()`, `libnvmf_discover_nbft()`) and its supporting connection-resolution code in `fabrics.c`. This backs `nvme discover`, `nvme connect-all`, and NBFT boot discovery.

Four main things in here:
1. Several real bugs, found and fixed along the way (list below).
2. NBFT discovery now walks through the same code `nvme discover`/`connect-all` use, instead of a second, forked implementation of the same walk.
3. `--force` renamed to `--no-reuse` (kept as a deprecated alias): `--force`'s original job was working around a persistence bug that's since been fixed elsewhere, so what's left is only its narrower meaning, "never reuse an existing connection." This is also the one commit that touches more than one file -- every other commit is mostly just `fabrics.c`.
4. A new TID-keyed visited list, so the Discovery Log Page walk tracks every Discovery Controller it has already visited in this walk -- closes a gap where a referral graph with more than one path to the same DC, or an outright cycle, could be walked more than once, or skipped incorrectly.

**Bugs found and fixed along the way (mostly NBFT related)**, none part of the original plan, all found while working through the surrounding code:
- `_nvmf_discover()` never fired `hooks.connected` for anything it discovered and connected itself, referrals or IOC entries -- `connect-all` has never reported per-device progress for its own discovery-driven connects, even though the hook exists and fires correctly everywhere else.
- Three `fctx->hooks.*` calls (`connected`, `already_connected`, `decide_retry` in `libnvmf_connect()`) had no NULL check, even though the hooks are documented-optional and every other call site in the file guards them. Any external libnvme caller passing NULL for a hook it doesn't need would crash.
- **NBFT**: An NBFT Discovery Descriptor's `host_iface` was freed before its own walk started, then read (use-after-free) by every connect made during that walk.
- **NBFT**: The Discovery Descriptor connection's own DHCP retry cleared the destination address instead of the local one -- a retry that could never succeed.
- **NBFT**: `read_discovery()` (`nbft.c`) treated an absent DC NQN reference as a parse failure, silently dropping the whole Discovery Descriptor. Per the Boot Specification, absent just means "use the well-known NQN" -- the common case, not a malformed table.
- **NBFT**: Once that parsing bug was fixed, `libnvmf_discover_nbft()` still never read the parsed NQN when present, always connecting with the well-known one instead.

**Commits, in order.** Commit 5 is the substantial one -- the actual rewrite; the rest are mostly small and mechanical:

1. `62779e721` -- stop handing per-entry context clones to hooks/recursion (a real unsoundness: the clone shallow-aliases owned pointers a hook could free via a public setter)
2. `0f5e6d8a4` -- unify the four resolve-or-create-a-DC paths into one `dc_open()`
3. `a1150d09d` -- rename `--force` to `--no-reuse` (kept as a deprecated alias); the old name never described what it did
4. `02c75ab01` -- identify a discovery self entry by its spec-defined SUBTYPE instead of a pointer-identity coincidence
5. `32e6c0bae` -- the core rewrite: depth-first walk with a visited-set and depth cap, `dc_decide()` extraction, and its first unit tests
6. `ac3caf4b6` -- fold NBFT's forked walk onto the same machinery (UAF and dead-retry fixes included)
7. `56b9f0221` -- fire `hooks.connected` for discovery-walk connects
8. `663dfcca5` -- NBFT Discovery Descriptor NQN parsing and consumption fixes
9. `9a703c7f5` -- guard the three unchecked hook calls

**Verification**: every commit builds and tests standalone; `-Dwerror=true` and `-Dfabrics=disabled` both clean throughout;  full unit suite green, including the 9 new `dc_decide()` cases. Live-smoke-tested `nvme discover` against a real TCP discovery controller. Not tested against a real NBFT boot environment.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -35,6 +35,7 @@
 #include <ccan/list/list.h>
 #include <ccan/str/str.h>
 
+#include <array-util.h>
 #include <compiler-attributes.h>
 
 #include <libnvme.h>
@@ -2609,8 +2610,8 @@ static int set_discovery_kato(struct libnvmf_context *fctx)
 	int tmo = fctx->ctrl_params.cfg.keep_alive_tmo;
 	/*
 	 * EPCSD isn't known until after the Discovery Log Page comes back, so
-	 * auto mode optimistically requests a KATO here; dc_should_connect()
-	 * disconnects per entry afterward if EPCSD turns out to be unset.
+	 * auto mode optimistically requests a KATO here; dc_decide()
+	 * disconnects afterward if EPCSD turns out to be unset.
 	 */
 	bool wants_kato = fctx->persistent == LIBNVMF_PERSISTENT_AUTO ||
 		fctx->persistent == LIBNVMF_PERSISTENT_FORCE;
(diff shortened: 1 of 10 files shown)


martin-belanger:
Looks like the windows build is timing out `Error: socket hang up`

libnvme/src/nvme/fabrics.c
```diff
@@ -2830,27 +2830,41 @@ static int _nvmf_discover(struct libnvme_global_ctx *ctx,
 		}
 		d.connect = true;
 
+		/*
+		 * Only discovery controllers need their KATO set here.
+		 * NVM subsystems (I/O controllers) get theirs set elsewhere.
+		 */
+		if (e->subtype != NVME_NQN_NVME)
+			set_discovery_kato(&nfctx);
```
igaw, line 2838: this should be a separate commit. nothing to do with the rest of this patch.

libnvme/src/nvme/fabrics.c
```diff
@@ -2830,27 +2830,41 @@ static int _nvmf_discover(struct libnvme_global_ctx *ctx,
...
-				_nvmf_discover(ctx, &nfctx, d.c, false,
-					       false);
-			}
+			/*
+			 * fctx, not nfctx: nfctx is a per-entry clone that
+			 * dies with this iteration. Recursing with it would
+			 * hand every deeper level a stack alias of fctx's
+			 * owned pointers (hostnqn, hostid, tls_key,
+			 * nbft_path) instead of the real, single owner.
+			 */
```
igaw, line 2853: no need to document this here. this is okay for the commit message.

libnvme/src/nvme/fabrics.c
```diff
@@ -3109,8 +3123,8 @@ static bool validate_uri(struct libnvme_global_ctx *ctx,
 }
 
 static int nbft_connect(struct libnvme_global_ctx *ctx,
-		struct libnvmf_context *fctx, struct libnvme_host *h,
-		struct nvmf_disc_log_entry *e,
+		struct libnvmf_context *fctx, struct libnvmf_context *hook_fctx,
```
igaw, line 3126: I really don't like that we have so many contexts now. Ideally there should be one. The `nfctx` is alrready border line, but `hook_fctx` is not a good idea, whatever it solves. 

This strongly indicates, that we need to figure out the role of the context and might have to split it up into different parts (internally)

igaw: I've tried this idea here: #3836

martin-belanger:
This series grew iteratively as I found bugs along the way. It is not as well organized as it could be. Will resubmit a new PR based on https://github.com/linux-nvme/nvme-cli/pull/3836.

PR #3830, opened 2026-08-12
wdc: fix logically dead code in wdc_do_cap_telemetry_log

sahmed-ibm:
The wdc_do_cap_telemetry_log() function uses two flags, host_gen and ctrl_init, to select between three telemetry retrieval paths. host_gen is initialized to 1 and is only ever set to 0 in the CONTROLLER branch, where ctrl_init is also always set to 1.

Because ctrl_init and host_gen can never both be 0, the final else branch calling libnvme_get_host_telemetry() is unreachable. The dead branch is misleading and adds unnecessary complexity.

Remove host_gen entirely and replace the three-way dispatch with a two-way branch on ctrl_init: call libnvme_get_ctrl_telemetry() when set, and libnvme_get_new_host_telemetry() otherwise.


igaw:
Thanks!

PR #3831, opened 2026-08-12
micron: fix vs-internal-log file size mismatches vs log headers and missing telemetry host log

brgoing:
Cleans up a few minor log issues impacting the output of micron vs-internal-log.
* Fixes a possible telemetry log size mismatch when comparing the log header within the saved file to the actual file size.  Since the log header was being read twice - once to determine size and again as part of the entire log pull, the size on the second read could be mismatched to the size actually read if new log entries had been made.  No longer reads the header twice.
* Ensures that the host telemetry log gets created when reading the logs.
* Fixes log size calculation for common logs that was including the log header size twice.

### DIFF
--- a/plugins/micron/micron-nvme.c
+++ b/plugins/micron/micron-nvme.c
@@ -3379,17 +3379,20 @@ static int get_common_log(struct libnvme_transport_handle *hdl, uint8_t id, uint
 		}
 		memcpy(buffer, (uint8_t *)&hdr, sizeof(hdr));
 	} else if (hdr.log_size < hdr.max_size) {
-		buffer = (uint8_t *)libnvme_alloc(sizeof(hdr) + hdr.log_size);
+		/* log_size includes the header, so the payload is the remainder */
+		uint32_t payload = hdr.log_size - sizeof(hdr);
+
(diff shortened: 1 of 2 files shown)


plugins/micron/micron-nvme.c
```diff
@@ -3379,17 +3379,20 @@ static int get_common_log(struct libnvme_transport_handle *hdl, uint8_t id, uint
 		}
 		memcpy(buffer, (uint8_t *)&hdr, sizeof(hdr));
 	} else if (hdr.log_size < hdr.max_size) {
-		buffer = (uint8_t *)libnvme_alloc(sizeof(hdr) + hdr.log_size);
+		/* log_size includes the header, so the payload is the remainder */
+		uint32_t payload = hdr.log_size - sizeof(hdr);
+
+		buffer = (uint8_t *)libnvme_alloc(hdr.log_size);
```
igaw, line 3385: FWIW, libnvme_alloc returns `void *`, thus there is no explicit cast necessary. IMO, it reads simpler without the cast.

igaw:
Thanks!

PR #3832, opened 2026-08-12
plugins/config: add 'nvme config status' command

martin-belanger:
Add a command that reports on the legacy config files. It checks whether config.json/discovery.conf are present and whether they have already been converted to the new INI format (nvme-fabrics.conf), based on whether the rollback symlink is present next to each one. When a file has been converted, it hints that both the file and its symlink are safe to delete once rollback is no longer a concern.

### DIFF
--- a/Documentation/cmds-main.txt
+++ b/Documentation/cmds-main.txt
@@ -181,6 +181,9 @@ linknvme:nvme-config-validate[1]::
 linknvme:nvme-config-show[1]::
 	Show the resolved NVMe-oF INI connection configuration
 
+linknvme:nvme-config-status[1]::
+	Report legacy config.json/discovery.conf conversion status
+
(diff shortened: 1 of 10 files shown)


igaw:
Makes sense. The only  thing I wonder if we should return an error code, so this command could be used in shell scripts without having to parse the output. The problem is no error is `0` and any error code is mapped to `1`, because return codes in POSIX are a very difficult topic.

martin-belanger:
I added a return code as follows:

```
EXIT STATUS
-----------
Returns 0 if neither legacy file has anything to report. Returns a
non-zero status if config.json or discovery.conf, or a '.converted'
marker for either, is present -- regardless of which of the states
listed above applies. Use the printed output, not the exit status, to
tell those states apart.
```

igaw:
Thanks!

PR #3833, opened 2026-08-12
libnvme: normalize a comma-separated FC WWN pair in TID construction

martin-belanger:
fabrics.c already fixes this for discovery-log entries before they're used for candidate matching, but any other caller building a TID directly from FC fields (connect, a config file, ...) got a hard rejection instead. tid_sanitize_addr() already silently canonicalizes IP addresses the same way, so extend the same treatment to FC.


igaw:
Thanks!

PR #3834, opened 2026-08-13
libnvme: fix the pin secret expansion

kariya-mitsuru:
Single patch against `master` (`5fd1ef488`).

`getswordfish()` declares the block counter inside the fill loop, so it is reset before every block and the increment at the end of the body is discarded. Every block comes out as the same `SHA-256(pin || 0)`, and a `pin:` secret longer than one digest just repeats its first 32 bytes rather than advancing the counter per block the way the algorithm it was added with does.

The counter also reached the digest as the raw bytes of an `unsigned int`, which is harmless only while it stays zero — the state this bug happens to keep it in. It is now serialized as four big endian bytes, so the expansion does not depend on the host byte order once the counter advances. A 32 byte secret is a single block and is unchanged; longer ones now differ, but `pin:` is an undocumented testing aid that has never appeared in a stable release.

## Testing

- `meson test`: 111 ok / 2 expected fail / 0 fail
- `checkpatch.pl --strict`: 0 errors, 0 warnings, 0 checks
- `libnvme/tests/raw-secret.c` gains the three valid lengths, checking the expansion against a fixed vector and asserting that a later block does not repeat the first; confirmed to fail on the parent commit
- binaries built before and after compared for `--secret=pin:1234`: the 32 byte secret is byte-identical, the 64 byte one no longer repeats its first block

### DIFF
--- a/libnvme/src/nvme/crypto.c
+++ b/libnvme/src/nvme/crypto.c
@@ -683,6 +683,7 @@ static ssize_t getswordfish(struct libnvme_global_ctx *ctx,
 		const char *seed, void *buf, size_t buflen)
 {
 	unsigned char hash[EVP_MAX_MD_SIZE];
+	unsigned int counter = 0;
 	EVP_MD_CTX *md_ctx;
 	size_t copied = 0;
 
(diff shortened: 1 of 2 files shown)


igaw:
Uff, sorry about that. Thanks for fixing it. Even nice to have a test case. This is is good for those who want to implement their own version of this algorithm.

PR #3835, opened 2026-08-13
ocp: fix null check and unsigned comparison in parse_event_fifo()

sahmed-ibm:
The parse_event_fifo() function casts a byte-offset pointer into a struct nvme_ocp_telemetry_event_descriptor and immediately dereferences it to check the class type, but never guards against a NULL result before that dereference.

Two downstream sub-conditions in the same block attempted to compensate: a redundant NULL check after the pointer was already used, and a greater-than-or-equal-to-zero comparison against event_data_size which is typed as an unsigned byte and can never be negative. Both checks have no effect and were flagged by static analysis.

Add a NULL guard before the first dereference and remove the two now-dead sub-conditions from the if statement to fix both issues.


igaw:
Thanks!

PR #3837, opened 2026-08-13
libnvme: rewrite the discovery walk as a depth-capped, deduped walker

martin-belanger:
Rewrites libnvme's discovery walk (`_nvmf_discover()`) from scratch, replacing the closed #3828, and breaks it into smaller, logical helpers to make the flow easier to follow.

Four main things in here:
1. `--force` renamed to `--no-reuse` (kept as a deprecated alias): `--force` used to double as both "force persistence" and "skip reusing an existing connection"; now that `--persistent=force` exists as its own explicit option, `--force` only means the latter, so its name no longer matches what's left of its job. This is also the widest-reaching commit -- the only one that touches CLI docs, generated accessor bindings, and `src/fabrics.c`, not just `libnvme/src/nvme/fabrics.c`.
2. NBFT discovery now walks through the same code `nvme discover`/`connect-all` use, instead of a second, forked implementation of the same walk.
3. A new TID-keyed visited list, so the Discovery Log Page walk tracks every Discovery Controller it has already visited in this walk -- closes a gap where a referral graph with more than one path to the same DC, or an outright cycle, could be walked more than once.
4. Several real bugs, found and fixed along the way (list below).

## Bugs found and fixed along the way

None of these were part of the original plan; all found while working through the surrounding code.

- Three `fctx->hooks.*` calls (`connected`, `already_connected`, `decide_retry`) had no NULL check, even though every hook is documented-optional and every other call site in the file guards them. An external caller passing NULL for a hook it doesn't need would crash.
- A spec-legal all-zero DCNQNHOR (NVMe Boot Specification rev 1.4: "no unique NQN, use the well-known Discovery NQN") was treated as a hard parse failure, silently dropping the whole Discovery Descriptor -- including the common case, since most descriptors have no unique NQN at all.
- Once that parsing bug was fixed, `libnvmf_discover_nbft()` still never read the parsed NQN when present, always connecting with the well-known one instead.
- When `--device` didn't resolve to a real controller, `fctx->persistent` got silently forced to `NO`, discarding whatever the user actually asked for with `--persistent`.
- An already-connected referral was skipped entirely, without ever walking its own referrals.
- A discovery self entry was identified by pointer identity (`cl == c`) instead of its spec-defined SUBTYPE.
- A referral's KATO was requested after the connect already happened, too late to take effect.
- `hooks.connected` never fired for anything discovered and connected during the walk itself -- `connect-all` never reported per-device progress for its own discovery-driven connects, even though the hook exists and fires correctly everywhere else.
- NBFT: a Discovery Descriptor's `host_iface` was freed before its own walk read it from every connect made during that walk (moot now -- `ctrl_params_dup()`'s `__cleanup_ctrl_params` already prevents this by construction).
- NBFT: the Discovery Descriptor connection's own DHCP retry cleared the destination address (`traddr`) instead of the local one (`host_traddr`) -- a retry that could never succeed.

## Commits

1. **guard three unchecked fabrics hook calls against NULL** -- every hook is optional; three call sites weren't checking.
2. **rename discover/connect-all `--force` to `--no-reuse`** -- `--force` no longer means what its name says now that `--persistent=force` exists separately; `--force` kept as a deprecated alias.
3. **use a Discovery Descriptor's own NQN when specified** -- a spec-legal all-zero DCNQNHOR was mishandled as a hard parse failure, dropping the whole descriptor.
4. **unify discovery ctrl lookup-or-create into `dc_open()`** -- one path for "reuse or create a discovery controller," used by discover, connect-all, and (later) NBFT.
5. **rewrite `_nvmf_discover()` as a depth-capped, deduped walk** -- the core change. Fixes four real bugs: an already-connected referral was skipped without ever walking its own referrals, a self entry was misidentified by pointer identity instead of SUBTYPE, KATO was requested too late to take effect, and `hooks.connected` never fired for controllers discovered and connected during the walk. Adds a referral depth cap and a TID-based visited set for cycle detection. `dc_decide()` is a new pure, unit-tested function (9 cases in `test-fabrics.c`) driving every disconnect decision.
6. **fold NBFT discovery onto the shared walker** -- `nbft_discovery()` was a forked copy of the walk logic, missing the depth cap and visited-set the general path just gained. Replaced with the same `dc_open()`/`dc_walk()` machinery via a new private `connect_leaf` hook, fixing two more pre-existing bugs along the way (see commit message).

## Testing

- `meson compile` clean under default, `-Dwerror=true`, and `-Dfabrics=disabled`.
- `meson test` passes, excluding the pre-existing, unrelated `fabrics-mock-cli` LD_PRELOAD failure that's also present on a clean master.
- New `test_dc_decide()` unit test exercises `dc_decide()`'s disconnect logic directly.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -2536,6 +2536,19 @@ static libnvme_ctrl_t lookup_ctrl(libnvme_host_t h,
 	return NULL;
 }
 
+/*
+ * Same as lookup_ctrl(), but only returns a controller that's actually
+ * connected. lookup_ctrl() can also match a scanned-but-unconnected
+ * draft, which has no kernel-assigned name yet.
+ */
+static libnvme_ctrl_t lookup_live_ctrl(libnvme_host_t h,
+		const struct libnvme_ctrl_params *params)
+{
+	libnvme_ctrl_t c = lookup_ctrl(h, params);
+
+	return (c && libnvme_ctrl_get_name(c)) ? c : NULL;
+}
+
 __shr_public int libnvmf_get_owner_from_tid(struct libnvme_global_ctx *ctx,
 		const struct libnvmf_tid *tid, char **owner)
 {
@@ -2819,6 +2832,32 @@ static bool dc_visited_has(const struct dc_visited *v,
 	return false;
 }
 
+/*
+ * Registers c's own real, connected identity -- not whatever a referring
+ * entry claimed about it -- so a cycle that loops back to c (including
+ * back to the primary) is recognized even if reached via a differently
+ * reported path. Best-effort: a TID construction failure here just means
+ * this one DC isn't deduplicated, not a fatal error for the walk.
+ */
+static void dc_visited_register(struct dc_visited *visited,
+		struct libnvme_ctrl *c, struct libnvmf_context *fctx)
+{
+	struct libnvmf_tid *tid;
+	int err;
+
+	err = libnvmf_tid_from_fields(libnvme_ctrl_get_transport(c),
+			libnvme_ctrl_get_traddr(c), libnvme_ctrl_get_trsvcid(c),
+			libnvme_ctrl_get_subsysnqn(c),
+			libnvme_ctrl_get_host_traddr(c),
+			libnvme_ctrl_get_host_iface(c),
+			fctx->hostnqn, fctx->hostid, &tid);
+	if (err)
+		return;
+
+	if (dc_visited_append(visited, tid))
+		libnvmf_tid_free(tid);
+}
+
 static inline void free_libnvmf_tid(struct libnvmf_tid **tid)
 {
 	libnvmf_tid_free(*tid);
@@ -2843,12 +2882,21 @@ static void dc_connect_leaf_entry(struct libnvme_global_ctx *ctx,
 		return;
 	}
 
-	err = nvmf_connect_disc_entry(h, e, params, &d.c);
+	if (fctx->hooks.connect_leaf) {
+		/*
+		 * NBFT's own leaf-connect quirks (DHCP retry, firing
+		 * hooks.connected) apply instead -- it fires the hook
+		 * itself.
+		 */
+		err = fctx->hooks.connect_leaf(ctx, fctx, h, e, params, &d.c);
+	} else {
+		err = nvmf_connect_disc_entry(h, e, params, &d.c);
+		if (d.c && fctx->hooks.connected)
+			fctx->hooks.connected(fctx, d.c, fctx->hooks.user_data);
+	}
 	if (d.c) {
 		d.connect = true;
 		dc_log_decision(fctx, e, &d, NULL);
-		if (fctx->hooks.connected)
-			fctx->hooks.connected(fctx, d.c, fctx->hooks.user_data);
 	} else if (err == -ENVME_CONNECT_ALREADY) {
 		dc_already_connected(fctx, h, e);
 	} else {
@@ -2860,7 +2908,8 @@ static int _nvmf_discover(struct libnvme_global_ctx *ctx,
 		struct libnvmf_context *fctx,
 		const struct libnvme_ctrl_params *ctrl_params,
 		struct libnvme_ctrl *c, enum dc_ownership own, int depth,
-		struct dc_visited *visited, const __u16 *parent_eflags);
+		struct dc_visited *visited, const __u16 *parent_eflags,
+		bool *disconnected);
 
 /*
  * A referral entry (NVME_NQN_DISC) is a branch point: it opens up another
@@ -2875,6 +2924,7 @@ static void dc_walk_referral(struct libnvme_global_ctx *ctx,
 	__cleanup_libnvmf_tid struct libnvmf_tid *tid = NULL;
 	struct dc_decision d = { 0 };
 	enum dc_ownership child_own;
+	bool child_disconnected = false;
 	libnvme_ctrl_t cl;
 	int err;
 
@@ -2902,8 +2952,8 @@ static void dc_walk_referral(struct libnvme_global_ctx *ctx,
 		return;
 	}
 
-	cl = lookup_ctrl(h, params);
-	if (cl && libnvme_ctrl_get_name(cl)) {
+	cl = lookup_live_ctrl(h, params);
+	if (cl) {
 		d.c = cl;
 		d.already_connected = true;
 		child_own = DC_BORROWED;
@@ -2933,12 +2983,15 @@ static void dc_walk_referral(struct libnvme_global_ctx *ctx,
 	/*
 	 * The child decides its own fate (disconnect or persist)
 	 * internally, the same way this DC just did above for
-	 * itself -- so the free below is unconditional, matching
-	 * how libnvmf_discover() already treats its own primary c.
+	 * itself. Only free it if it actually disconnected --
+	 * a persisted child must stay in the tree, since a sibling
+	 * or later referral elsewhere in this same walk may still
+	 * need to find it via lookup_live_ctrl()/dc_visited_has().
 	 */
 	_nvmf_discover(ctx, fctx, params, d.c, child_own, depth + 1,
-		       visited, &eflags);
-	libnvme_free_ctrl(d.c);
+		       visited, &eflags, &child_disconnected);
+	if (child_disconnected)
+		libnvme_free_ctrl(d.c);
 }
 
 /*
@@ -2975,14 +3028,18 @@ static bool dc_survey_self_entry(struct libnvmf_context *fctx,
 	d.disconnect = dc_decide(fctx, libnvme_ctrl_get_subsysnqn(c), own,
 			self_seen, self_eflags, parent_eflags);
 
-	if (own == DC_BORROWED)
-		dc_log_decision(fctx, self_entry, &d,
-			"pre-existing, not ours to disconnect");
-	else if (self_entry)
+	if (self_entry) {
 		dc_log_decision(fctx, self_entry, &d, NULL);
-	else
+	} else if (own == DC_BORROWED) {
+		dc_log_decision(fctx, NULL, &d,
+			"pre-existing, not ours to disconnect");
+	} else if (parent_eflags) {
+		dc_log_decision(fctx, NULL, &d,
+			"no self entry, using parent's view");
+	} else {
 		dc_log_decision(fctx, NULL, &d,
 			"no self entry, EPCSD assumed 0");
+	}
 
 	return d.disconnect;
 }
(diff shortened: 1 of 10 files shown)


libnvme/src/nvme/fabrics.c
```diff
@@ -3691,41 +3684,58 @@ static struct libnvme_ctrl *discover_lookup_ctrl_by_device(
...
-static int discover_lookup_ctrl(struct libnvme_global_ctx *ctx,
-		struct libnvmf_context *fctx, struct libnvme_host *h,
-		struct libnvme_ctrl **ctrl, bool *already_connected)
+/*
+ * Resolve the primary discovery controller connection: reuse one via
+ * --device if given, reuse one found by matching connection parameters
+ * otherwise, or create a fresh one if --no-reuse was given or neither
+ * lookup found anything. Never touches fctx->persistent; ownership is
+ * reported separately so the caller knows what it may disconnect later.
+ */
```
igaw: This comment is hard to read and it mostly just describes what the code does.

libnvme/src/nvme/fabrics.c
```diff
@@ -3691,41 +3684,58 @@ static struct libnvme_ctrl *discover_lookup_ctrl_by_device(
...
 		/*
-		 * When we found an existing controller it might not have a
-		 * device handle yet
+		 * No existing controller, or --no-reuse was given: create a
+		 * new one.
 		 */
-		err = libnvme_open(ctx, c->name, O_RDONLY, &c->hdl);
+		err = nvmf_create_discovery_ctrl(ctx, fctx, &fctx->ctrl_params,
+				h, &c);
 		if (err) {
```
igaw: Shouldn't there be a :

```c
*own = DC_OWNED;
```

igaw: ah, don't set it at the beginning of the function, output arguments should only be updated when there is no error returned.

libnvme/src/nvme/fabrics.c
```diff
@@ -2709,24 +2708,39 @@ static bool dc_should_disconnect(struct libnvmf_context *fctx,
...
+ * @self_seen: whether this DC's own self entry (SUBTYPE 03h, "current
+ *             discovery subsystem") was found in its own Discovery Log
+ *             Page.
+ * @self_eflags: that self entry's EFLAGS; meaningful only if @self_seen.
+ * @parent_eflags: the EFLAGS this DC's own referral entry carried in its
+ *                 parent's Discovery Log Page, or NULL if this DC has no
+ *                 parent (the primary). Used only as a fallback when
+ *                 @self_seen is false.
+ */
+static bool dc_decide(struct libnvmf_context *fctx, const char *subnqn,
```
igaw, line 2713: According the documentation, why  not `dc_should_disconnect`? Just inline the exisint dc_should_disconnect.

libnvme/src/nvme/fabrics.c
```diff
@@ -2771,16 +2785,221 @@ static void dc_log_decision(struct libnvmf_context *fctx,
...
+ * walk: connect it if --connect was requested, and there is nothing
+ * further to recurse into.
+ */
+static void dc_connect_leaf_entry(struct libnvme_global_ctx *ctx,
+		struct libnvmf_context *fctx, struct libnvme_host *h,
+		struct nvmf_disc_log_entry *e,
+		struct libnvme_ctrl_params *params)
+{
+	struct dc_decision d = { 0 };
+	int err2;
```
igaw: `int err`

libnvme/src/nvme/fabrics.c
```diff
@@ -2771,16 +2785,221 @@ static void dc_log_decision(struct libnvmf_context *fctx,
...
+		struct libnvmf_context *fctx, struct libnvme_host *h,
+		struct nvmf_disc_log_entry *e,
+		struct libnvme_ctrl_params *params, int depth,
+		struct dc_visited *visited, __u16 eflags)
+{
+	__cleanup_libnvmf_tid struct libnvmf_tid *tid = NULL;
+	struct dc_decision d = { 0 };
+	enum dc_ownership child_own;
+	libnvme_ctrl_t cl;
+	int err2;
```
igaw: `int err`

libnvme/src/nvme/fabrics.c
```diff
@@ -2771,16 +2785,221 @@ static void dc_log_decision(struct libnvmf_context *fctx,
...
+		dc_log_decision(fctx, e, &d,
+			"referral depth limit reached, not descending");
+		return;
+	}
+
+	if (!libnvmf_tid_from_fields(params->transport, params->traddr,
+			params->trsvcid, params->subsysnqn,
+			params->host_traddr, params->host_iface,
+			fctx->hostnqn, fctx->hostid, &tid) &&
+	    dc_visited_has(visited, tid)) {
```
igaw: Here I would just use tabs 
```c
	if (!libnvmf_tid_from_fields(params->transport, params->traddr,
				params->trsvcid, params->subsysnqn,
				params->host_traddr, params->host_iface,
				fctx->hostnqn, fctx->hostid, &tid) &&
			dc_visited_has(visited, tid)) {
		dc_log_decision(fctx, e, &d, "already visited this walk");
		return;
	}
```

igaw: btw, when `libnvmf_tid_from_fields` fails, should we just bail out with an error "invalid entry" instead going further?

libnvme/src/nvme/fabrics.c
```diff
@@ -2771,16 +2785,221 @@ static void dc_log_decision(struct libnvmf_context *fctx,
...
+		d.c = cl;
+		child_own = DC_BORROWED;
+	} else {
+		set_discovery_kato(fctx, params);
+		err2 = nvmf_connect_disc_entry(h, e, params, &d.c);
+		if (!d.c) {
+			if (err2 == -ENVME_CONNECT_ALREADY)
+				dc_already_connected(fctx, h, e);
+			else
+				dc_log_decision(fctx, e, &d,
```
igaw, line 2967: I leaning towards also logging 'already connected' here. it would make the log easier to read I think

libnvme/src/nvme/fabrics.c
```diff
@@ -2771,16 +2785,221 @@ static void dc_log_decision(struct libnvmf_context *fctx,
...
+			tid = NULL; /* ownership transferred */
+	}
+
+	/*
+	 * The child decides its own fate (disconnect or persist)
+	 * internally, the same way this DC just did above for
+	 * itself -- so the free below is unconditional, matching
+	 * how libnvmf_discover() already treats its own primary c.
+	 */
+	_nvmf_discover(ctx, fctx, params, d.c, child_own, depth + 1,
```
igaw, line 2991: `5.2.12.3.3 Discovery (Log Page Identifier 70h) base spec v2.3`
Image

I've heard the argument that the total number of hops is eight. This implementation allows eight recursive iterations, as far as I can tell. That would mean 8 × the maximum number of referrals in a log page, right?

libnvme/src/nvme/fabrics.c
```diff
@@ -2771,16 +2785,221 @@ static void dc_log_decision(struct libnvmf_context *fctx,
...
+ */
+static bool dc_survey_self_entry(struct libnvmf_context *fctx,
+		struct libnvme_ctrl *c, enum dc_ownership own,
+		const __u16 *parent_eflags, struct nvmf_discovery_log *log,
+		uint64_t numrec)
+{
+	struct nvmf_disc_log_entry *self_entry = NULL;
+	struct dc_decision d = {
+		.c = c,
+		.primary = !parent_eflags,
```
igaw: wouldn't it better to test explicit the flags which making this dc primary? I'd prefer to have this explicit here.

libnvme/src/nvme/fabrics.c
```diff
@@ -2802,117 +3021,73 @@ static int _nvmf_discover(struct libnvme_global_ctx *ctx,
...
-			dc_log_decision(fctx, e, &d, "already connected");
+		if (e->subtype == NVME_NQN_CURR)
 			continue;
-		}
 
 		/* Skip connect if the transport types don't match */
-		if (strcmp(libnvme_ctrl_get_transport(c),
-			   params.transport)) {
+		transport = libnvmf_trtype_str(e->trtype);
+		if (strcmp(libnvme_ctrl_get_transport(c), transport)) {
```
igaw, line 3122: This is not for now, but this is a policy thing and we might want to allow it and it? I think we should have a fabrics context setter for which transports are acceptable.

libnvme/tests/test-fabrics.c
```diff
@@ -640,6 +640,79 @@ static bool test_unescape_uri(void)
 	return pass;
 }
 
+/* -------------------------------------------------------------------------
+ * dc_decide — pure disconnect-decision logic for one discovery controller
+ * -------------------------------------------------------------------------
+ */
+static bool test_dc_decide(struct libnvme_global_ctx *ctx)
```
igaw, line 647: Very cool that we have this type of tests as well!

igaw:
Generally looks good, just a few nitpicks.

The only big question is on the interpretation what 'referral entries deeper than 8 levels' means.

libnvme/src/nvme/fabrics.c
igaw: After a bit of pondering, my reading is also 8 levels (e.g. recursion levels) and not 8 hob in total. Another reason why this doesn't really matter is that a log page can be huge (roughly 16GB IIRC) thus a single one could already be to much to handle.

martin-belanger: You make an excellent point. It's a bit weird to limit the number of referral levels and not limit the number of entries per log page. I'll run this by Erik to see what he thinks.

libnvme/src/nvme/fabrics.c
martin-belanger: Agreed. This is from the existing legacy code. I kept it, but I'm not sure this is necessary at all.

libnvme/src/nvme/fabrics.c
martin-belanger: Done

libnvme/src/nvme/fabrics.c
martin-belanger: Done

libnvme/src/nvme/fabrics.c
martin-belanger: Done

libnvme/src/nvme/fabrics.c
martin-belanger: done

libnvme/src/nvme/fabrics.c
martin-belanger: Done

libnvme/src/nvme/fabrics.c
martin-belanger: Done. Split into two `if ()` statements

libnvme/src/nvme/fabrics.c
martin-belanger: done

libnvme/src/nvme/fabrics.c
martin-belanger: Done. Will open an issue to potentially stop the loop when we reach a configurable maximum number of DLP entries

libnvme/src/nvme/fabrics.c
martin-belanger: done

igaw:
Thanks!

PR #3838, opened 2026-08-13
plugins/solidigm: Added log names to solidigm log-page-directory.

lgdacunh:
Added log names to list of Solidigm log names.


igaw:
Thanks!

PR #3841, opened 2026-08-14
nvme-print: fix to check init and finish functions

ikegami-t:
Since basically nvme_print is skipped for dry-run. But some print functions not using nvme_print are not skipped. So the json_r object not created then the json print error caused.

### DIFF
--- a/src/nvme-print.c
+++ b/src/nvme-print.c
@@ -26,7 +26,7 @@
 #define nvme_print(name, flags, ...)				\
 	do {							\
 		struct print_ops *ops = nvme_print_ops(flags);	\
-		if (ops && ops->name && !nvme_args.dry_run)	\
+		if (nvme_print_check(ops, ops->name))		\
 			ops->name(__VA_ARGS__);			\
 	} while (false)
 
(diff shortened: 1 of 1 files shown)


src/nvme-print.c
```diff
@@ -49,6 +49,20 @@ static struct print_ops *nvme_print_ops(nvme_print_flags_t flags)
 	return ops;
 }
 
+static bool nvme_print_check(struct print_ops *ops, void *func)
+{
+	if (!ops || !func)
+		return false;
+
+	if (!nvme_args.dry_run)
+		return true;
```
igaw: I would change the order and also the logic slightly:

```c
...
/* 
  * Always run init and the finish hooks to properly initialize the
  * the output plugin.
  */
if (func == ops->show_init || func == ops->show_finish)
		return true;

if (nvme_args.dry_run)
		return false;

return true;
```

This feels a bit more natural do read.

ikegami-t: Right so just fixed as mentioned. Thank you.

igaw:
Thanks!

PR #3842, opened 2026-08-14
libnvme, nvme: guarantee terminated discovery log entry strings

martin-belanger:
This is something I noticed while working on #3837.

Discovery Log Page Entry fields (`trsvcid`, `subnqn`, `traddr`) are fixed-size char arrays copied off the wire with nothing guaranteeing a NUL terminator -- `trsvcid`/`traddr` are normally space-padded per spec, not NUL-padded. Every consumer (`strlen()`, `strdup()`, `printf("%s")`, `json_object_new_string()`) assumes a real C string anyway. Usually harmless (the bytes right after a field are normally zero), but a DC that doesn't zero its reserved bytes can trigger a real out-of-bounds read, most concretely on the last entry's `traddr`. Not new -- traceable back to before `sanitize_discovery_log_entry()` even existed.

1. **`shared: split shr_trim() into shr_rtrim() and shr_ltrim()`** -- needed to sanitize `trsvcid`, `subnqn`, `traddr`.
2. **`libnvme: guarantee terminated discovery log entry strings`** -- force-NUL-terminate + `shr_rtrim()` each field, replacing the old `strchomp()`.
3. **`nvme: harden discovery log print paths against unterminated fields`** -- the `discovery_log` hook has always handed over the raw, unsanitized log page (required for `nvme discover --raw` to stay truly raw), so the stdout/JSON backends were never defended against missing NULL. `log` is now `const` throughout the hook chain to make sure the hook/callback cannot modify the log page.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -77,20 +77,6 @@ static inline void cleanup_ctrl_params(struct libnvme_ctrl_params *params)
 }
 #define __cleanup_ctrl_params __cleanup(cleanup_ctrl_params)
 
-/**
- * strchomp() - Strip trailing spaces
- * @str: String to strip
- * @max: Maximum length of string
- */
-static void strchomp(char *str, int max)
-{
-	int i;
-
-	for (i = max - 1; i >= 0 && str[i] == ' '; i--) {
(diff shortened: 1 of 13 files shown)


igaw:
Thanks!

PR #3843, opened 2026-08-14
ini-util: fix Windows error reporting when opening a directory

brgoing:
Windows reports EACCESS when opening a directory, but tests expect EISDIR.  Reclassifying directory opening error as EISDIR for any platform.


igaw:
Thanks!

PR #3844, opened 2026-08-14
Enable assert-util testing on windows using new shared spawn and wait utilities.

brgoing:
Windows doesn't support fork, so the assert-util tests needed to be redesigned slightly to work across platforms.
* Adds new shr_pipe, shr_spawn, and shr_wait_proc utils with Linux and Windows implementations.
* Redesigns the assert-util tests to use the new shared pipe and spawn utilities.  Since fork is not supported on Windows, the test design now re-spawns the test binary as a child process.
* Enables test-assert-utils on Windows.

### DIFF
--- a/shared/meson.build
+++ b/shared/meson.build
@@ -33,6 +33,7 @@ if host_system == 'windows'
     sources += [
         'fs-util-win.c',
         'crypto-util-win.c',
+        'proc-util-win.c',
         'sig-util-win.c',
     ]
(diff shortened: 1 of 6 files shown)


igaw:
FWIW, when I was extending the test coverage I was not sure if it was worth to adding the process management functions. Well, we might need it somewhere else, so let's add them.

Thanks!

PR #3846, opened 2026-08-16
libnvme: stop reporting every ENOKEY as a missing TLS key

kariya-mitsuru:
Single patch against `master` (`5fd1ef488`).

`ENVME_CONNECT_NOKEY` reports `pre-shared TLS key is missing`, but the kernel also returns `ENOKEY` when the KX-HMAC-CHAP secret is missing, and when a key is present but cannot be looked up. The message now says what the errno means, the way `strerror(ENOKEY)` does; `dmesg` still identifies which key it was.

## Testing

- `meson test`: 111 ok / 2 expected fail / 0 fail
- `checkpatch.pl --no-tree --strict`: 0 errors, 0 warnings, 0 checks
- `ENOKEY` call sites checked against mainline `3d6d817622b0`


igaw:
FWIW, if you think there is need for more error codes to distinguish cases, it would be the right time to add them :)

Thanks anyway!

PR #3851, opened 2026-08-17
nvme-discoverd: gate persistent connections on EPCSD

martin-belanger:
## Summary

nvme-discoverd now honors EPCSD (Explicit Persistent Connection Support
for Discovery) instead of always keeping a persistent connection to every
DC.

- Read the EPCSD bit off a DC's own self entry (SUBTYPE 03h) in its
  Discovery Log Page, previously dropped silently.
- Fall back to the parent's referral-entry view when a self entry is
  absent; default to EPCSD=0 if neither is available.
- EPCSD=0: disconnect the DC and reconnect on a configurable interval
  (`epcsd-poll-interval-minutes`, default 15) to re-check.
- EPCSD=1: unchanged — already persistent, already refreshes via AER.


igaw:
Thanks!

PR #3852, opened 2026-08-17
plugins/exclusion: use shared array-util for entry_collection

martin-belanger:
entry_collection duplicated the doubling-capacity realloc() pattern shared/array-util.h already provides. Switch to SHR_PTRARRAY_DEFINE().


igaw:
Thanks!

Nice that the common library is useful :)

PR #3853, opened 2026-08-17
nvme-discoverd: make the DC give-up horizon configurable

martin-belanger:
DC_GIVEUP_USEC was a hardcoded 72h compile-time constant. Add dc-giveup-hours (default 72, 0 = never give up) to nvme-discoverd.conf instead.


igaw:
Could you rebase it? Thanks!

martin-belanger:
> Could you rebase it? Thanks!

Done

igaw:
Thanks!

PR #3854, opened 2026-08-17
libnvme, tests: cover discovery-walk depth cap, cycles, and self-entry precedence

martin-belanger:
Four cases in the LD_PRELOAD fabrics mock were untested: a referral's self entry overriding its parent's view, an already-connected referral still getting its own log page walked, the referral depth cap, and a referral cycle not being re-walked. Add all four, plus a log-page fetch counter in the mock server to verify the cycle case.


igaw:
Thanks, good extension of the tests. Do we need also double linked circular referrals?  

`A <-> B <->C <-> A`

or do we have them already?

martin-belanger:
No, not in the test suite yet — I'll add a dedicated `A <-> B <->C <-> A` ring test as a follow-up.

PR #3855, opened 2026-08-17
nvme: hyphenate --fast-io-fail-tmo, keep old spelling as alias

martin-belanger:
Every other fabrics option uses hyphens, and the INI config key is already "fast-io-fail-tmo". Add the hyphenated form as the documented option; keep --fast_io_fail_tmo working as a hidden alias, same pattern as --dhchap-secret.


igaw:
Thanks!

PR #3857, opened 2026-08-17
micron: unused value in micron_write_os_config_to_file()

sahmed-ibm:
The micron_write_os_config_to_file() function writes system configuration headers to a log file before spawning diagnostic commands. Inside the loop, fpOSConfig is assigned NULL immediately after calling fclose().

This assignment is never read because fpOSConfig is re-assigned on the next iteration of the loop, resulting in an unused value.

Remove the unused assignment to fpOSConfig to clean up the code.


igaw:
Thanks!

PR #3858, opened 2026-08-17
libnvme, tests: add a fully bidirectional 3-DC referral ring test

martin-belanger:
Only a 2-node back-edge cycle was tested. Add a doubly-linked ring (A<->B<->C<->A, each DC referring to both ring-neighbors), asserting every DC's Discovery Log Page is fetched exactly once.

follow up to: https://github.com/linux-nvme/nvme-cli/pull/3854


igaw:
Thanks!

PR #3859, opened 2026-08-17
scaleflux: fix strcpy into fixed-size buffer in nvme_expand_cap()

sahmed-ibm:
The nvme_expand_cap() function stores the device name in the fixed-size dev_name buffer. When the transport handle is not a controller, the device name is copied into dev_name using strcpy() without checking the length of the source string.

Copying an unbounded string into a fixed-size buffer can overflow the buffer, corrupting adjacent stack memory and causing undefined behavior.

Replace strcpy() with snprintf() bounded by sizeof(dev_name) to prevent the overflow.


igaw:
Thanks!

PR #3860, opened 2026-08-17
libnvme: rename dhchap-secret INI key to kxchap-secret

martin-belanger:
Rename the `dhchap-secret`/`dhchap-ctrl-secret` INI keys to `kxchap-secret`/`kxchap-ctrl-secret`, following TP4201's rename of the DH-HMAC-CHAP protocol to KX-HMAC-CHAP. The secret string format (`DHHC-1:...`) is unchanged.

The INI config format has never had a stable release, so there is no back-compat requirement at the INI level. `nvme config-convert` still reads the legacy `dhchap_key`/`dhchap_ctrl_key` `config.json` fields and writes the new `kxchap-*` INI keys.

Also fixes a stale `--dhchap-secret` reference in the `--concat` error message (the CLI flag itself was already renamed in an earlier PR; `--dhchap-secret` remains as a hidden alias).

### DIFF
--- a/Documentation/nvme-fabrics.conf.txt
+++ b/Documentation/nvme-fabrics.conf.txt
@@ -83,8 +83,8 @@ TUNABLE (type default, '[Host]', endpoint section, or 'controller =' line):
 
 SECURITY ('[Host]' or endpoint section only, never a 'controller =' line):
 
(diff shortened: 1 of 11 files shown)


igaw:
Nice! With this we should have the obvious spec update covered. Thanks!

PR #3862, opened 2026-08-17
Add shr_spawnp and refactor Micron plugin to use shared spawn and wait utilities

brgoing:
The Micron plugin had its own cross-platform spawn helper implementations.  This change adds path searching versions of the shared spawn utilities and uses them in the Micron plugin instead of having separate implementations.  Also adds a shr_read_all method that reads until EOF, which is needed for draining a pipe.
* Adds path-searching shr_spawnp(), based on execvp on Linux and _spawnvp() on Windows.
* Adds shr_read_all to read a until EOL, handling EINTR as needed.
* Refactors Micron's spawn helpers to use the new shared utilities.

### DIFF
--- a/plugins/micron/micron-utils-linux.c
+++ b/plugins/micron/micron-utils-linux.c
@@ -9,21 +9,20 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
-#include <spawn.h>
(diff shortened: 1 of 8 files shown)


igaw:
Nice! Thanks!

PR #3869, opened 2026-08-18
seagate: remove duplicate nvme_set_features_simple call in vs_clr_pci e_correctable_errs

sahmed-ibm:
The vs_clr_pcie_correctable_errs() function dispatches between two different NVMe feature commands based on device model using an if/else block. Each branch correctly assigns err from the appropriate command.

An unconditional call to nvme_set_features_simple() after the if/else block overwrites err before it can be checked. For Jaguar/Panther devices this silently discards the result of nvme_set_features() and re-issues the wrong command; for other devices the command runs twice.

Remove the duplicate call so that err reflects the result of the device-appropriate command selected by the if/else block.


igaw:
Thanks!

PR #3870, opened 2026-08-18
ocp: fix unused value of temp_sz in get_telemetry_dump()

sahmed-ibm:
The get_telemetry_dump() function captures the original unpadded DA1 stats size into temp_sz before the size may be rounded up to a 512-byte boundary for the read. The subsequent print_telemetry_da_stat() call re-derives that same value inline instead of using temp_sz, leaving the stored value unread before it is overwritten by the FIFO loop.

The dead assignment is misleading and inconsistent with the FIFO loop blocks in the same function, which correctly assign temp_sz and pass it to print_telemetry_da_fifo().

Pass temp_sz to print_telemetry_da_stat() so the stored value is used as intended.


igaw:
Thanks!

PR #3871, opened 2026-08-18
solidigm: fix STRING_OVERFLOW in join_options

sahmed-ibm:
The join_options() function builds a pipe-delimited option string by concatenating each entry from a string array into dest using strcat(). No bounds checking is performed against the size of the destination buffer.

Passing a fixed-size stack buffer to strcat() without a length limit allows the written string to silently overflow the buffer if the total length of the joined options exceeds the buffer size, resulting in undefined behavior.

Add a dest_size parameter and replace strcat() with strncat(), limiting each concatenation to the remaining space in the buffer. Update both call sites to pass sizeof() of their buffer.


igaw:
Thanks!

PR #3872, opened 2026-08-18
tests: fall back to legacy commands when needed

schlad:
Grouped subcommands like 'id ctrl'/'id ns'/'id ctrl-list'/'id ns-list'/ 'id nvm-ns'/'log smart' were introduced in the v3.0, so older stable releases  only know the flat hyphenated legacy names, so tests failed before their own body ever ran (e.g. every TestNVMe subclass hits this in the shared setUp(), which calls 'nvme id ctrl').

Add command(), which prefers the current grouped syntax and falls back to the legacy form (id-ctrl, id-ns, list-ctrl, list-ns, nvm-id-ns, smart-log) only when a one-time --help probe shows the installed nvme binary doesn't support it.

### DIFF
--- a/tests/e2e/nvme_copy_test.py
+++ b/tests/e2e/nvme_copy_test.py
@@ -95,7 +95,7 @@ class TestNVMeCopy(TestNVMe):
         array.  Returns 0 if either command fails or the pif field is absent
         (0 = 16-bit guard / no PI, the safe default for format 0/2 copy).
         """
-        id_ns_cmd = f"{self.nvme_bin} id ns {self.ns1} --output-format=json"
+        id_ns_cmd = f"{self.nvme_bin} {self.command('id ns')} {self.ns1} --output-format=json"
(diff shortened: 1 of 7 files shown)


schlad:
This could be useful if e2e tests are run from the latest code but the nvme-cli under tests (for instance the one packaged for SLE), is the older code.

igaw:
yes, makes sense to make the e2e backwards compatible. sorry didn't realize it.

schlad:
> yes, makes sense to make the e2e backwards compatible. sorry didn't realize it.

meanwhile I amended to commits, to sort out the problems pointed by the CI checks

schlad:
I see some checks did not pass, but I assume nothing for me to fix/do, right?

igaw:
The CodeQL builds are failing because the service has been disabled since they want money for it. Was just to lazy to remove it yet. Let me address it now.

PR #3874, opened 2026-08-18
seagate: pass fw activation history log page by pointer

sahmed-ibm:
The print_stx_vs_fw_activate_history() and
json_stx_vs_fw_activate_history() functions both accept fwActivHis of type stx_fw_activ_history_log_page by value. The struct is 4096 bytes due to a large internal padding array, so each call copies the entire struct onto the stack.

Neither function modifies fwActivHis; both only read its fields. The 4096-byte stack copy is unnecessary and exceeds the 512-byte threshold for pass-by-value performance concerns.

Change both functions to accept a pointer-to-const instead and update the two call sites to pass the address of the local fwActivHis variable.


igaw:
Thanks!

PR #3878, opened 2026-08-19
completions: add PowerShell completion generator

jimmunn:
Add PowerShell output to generate-completions.py, the generated completion script, and automated tests.


igaw:
Nice! Thanks!

PR #3881, opened 2026-08-20
nvme-test: fix _probe_command check

brgoing:
--help always exits with a 1, even if the command exists.  Update the _probe_command check to look at the generated output instead to determine sub-command support.


igaw:
Thanks!

PR #3882, opened 2026-08-20
ioctl-win: fix inacurate comment

brgoing:
Update comment to provide reasoning for the PropertyId setting.

Testing showed a different behavior for PropertyId than originally assumed. The comment was out-of-date and no longer matched the implmentation, leading to confusion.


igaw:
Thanks!

PR #3883, opened 2026-08-20
completions: regenerate from a complete Linux build

jimmunn:
Picks up the --fast_io_fail_tmo -> --fast-io-fail-tmo rename and the new supported-pages command. The PowerShell completion was generated on Windows, which omits the Linux-only fabrics and plugin commands; regenerating on Linux restores them. Most of the bash/zsh diff is reordering to match the binary's command registration order.

### DIFF
--- a/completions/_nvme
+++ b/completions/_nvme
@@ -18,6 +18,12 @@ _nvme () {
 	_cmds=(
 		'list:List all NVMe devices and namespaces on machine'
 		'list-subsys:List nvme subsystems'
+		'show-topology:Show the topology'
+		'top:nvme top'
+		'reset:Resets the controller'
+		'subsystem-reset:Resets the subsystem'
+		'ns-rescan:Rescans the NVME namespaces'
+		'virt-mgmt:Manage Flexible Resources between Primary and Secondary Controller'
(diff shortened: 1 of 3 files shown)


jimmunn:
This does raise the question of whether completions for Windows should be generated in Windows or Linux. Currently, the Windows completions will include commands available in Linux but not available on Windows. However, all Windows commands will be included.

igaw:
Can you drop the generated bits? I'll run the generator with the release script. Just add the real change.

Good question. These will be generated on my machine on release. I guess we need two build configuration for generating the completion correctly by extending the release.sh script accordingly

jimmunn:
I am waiting on feedback from a message I posted in Matrix. These files are already old, but I may update the files after we make a decision on how to manage the generated files.

igaw:
As posted on the chat channel, adding a dedicated meson build target to update these files matches the `update-accessors` build target. So this makes sense. 

What I'd like to avoid is a lot of churn in the completion files every time we touch a file.  Currently it looks like it an update shuffles a lot of code around. Could the output made a bit more stable?

jimmunn:
PR 3954 included newer version of these files. Closing this.

PR #3884, opened 2026-08-20
gitignore: ignore __pycache__ anywhere

jimmunn:
Replace the two path-specific __pycache__ entries with a single top-level pattern so Python bytecode caches are ignored wherever they appear.


igaw:
Thanks!

Issue #3885, opened 2026-08-20
OCP telemetry log string parsing off by 1


fyodor-soiguine-sndk:
OCP telemetry sting log STR-ID-3, STR-EV-3 and STR-VUEV-3 state that the length field is 0-based.
Commit [dbb74e8ac4935e98685c8ec3a9d9ae20d0275861](https://github.com/linux-nvme/nvme-cli/commit/dbb74e8ac4935e98685c8ec3a9d9ae20d0275861) took off the +1 adjustment

fyodor-soiguine-sndk:
@bellis-ibm FYI

PR #3886, opened 2026-08-20
ocp: telemetry string parsing off-by-1

fyodor-soiguine-sndk:
Fix off-by-1 string lengths due to 0-based length field(s). #3885


igaw:
Could you add a commit message and your Singed-off-by  tag please?

fyodor-soiguine-sndk:
> Could you add a commit message and your Singed-off-by tag please?

@igaw, done.

igaw:
Thanks!

PR #3887, opened 2026-08-21
keys: name the TLS commands after the PSK, and fix two bugs on the way

kariya-mitsuru:
Fourteen patches against `master` (`0968160a1`). Follows #3827 and finishes what #3713 asked for.

#3827 renamed the KX-HMAC-CHAP half of the `keys` plugin and left the TLS half for a separate series. This is it: `gen-tls`, `check-tls` and `insert-tls` become `gen-tls-psk`, `check-tls-psk` and `insert-tls-psk`, while `import`, `export` and `revoke` keep their names, since they operate on keyring entries rather than on a particular PSK.

Two patches change what the program does, and go first:

- `nvme keys export` and every `--keyfile` write a file `nvme keys import` cannot read back, whenever the identity version is 0 — which is the default.
- `insert-tls-psk` and `gen-tls-psk --insert` report success when the keyring cannot be linked, printing an uninitialized local as the inserted PSK serial and writing it to `--keyfile`.

The rest are documentation and message text: output that now names which of the three PSKs it means, help text and kernel-doc in the specification's vocabulary, a few plain errors gone (a `-h` that never existed, a wrong hash name, a page in no index), and the differences a 2.x command line hits written down — in the alias pages and under NEWS.md's incompatible-changes heading.

**One of them is not about TLS at all.** A cross-reference has to be spelled `linknvme:<page>[<section>]`, and 174 across 55 pages fuse the macro name with the target's leading `nvme`, so they reach the reader verbatim.

## Testing

- each commit builds on its own; `meson test` 105 ok / 2 expected fail / 0 fail at each
- the export bug, against a real keyring (RHEL 9.6, `nvme_keyring` loaded): on `master`, `keys export` and `--keyfile` both write `NVMeTLSkey-0:01:...`, and `keys import` rejects it with `Invalid argument`; with the series both write `NVMeTLSkey-1:01:...` over the same key bytes and import succeeds
- the keyring bug, with `keyctl_link()` denied: `master` prints `Inserted TLS key 00000000`, exits 0 and leaves the keyring empty; the series reports `Permission denied` and exits 1


plugins/keys/keys-plugin.c
```diff
@@ -355,7 +355,8 @@ static int append_keyfile(struct libnvme_global_ctx *ctx, const char *keyring,
 		return err;
 	}
 
-	err = libnvmf_export_tls_key_versioned(ctx, ver, hmac, key_data,
+	/* PSK interchange format version, not the identity version */
+	err = libnvmf_export_tls_key_versioned(ctx, 1, hmac, key_data,
```
igaw, line 358: No, this is assuming that there is only version 1. The code is written to be ready for version 2 etc.

This doesn't fix anything. Don't add policy into the parser.

plugins/keys/keys-plugin.c
```diff
@@ -768,18 +769,19 @@ static void __scan_tls_key(struct libnvme_global_ctx *ctx, long keyring_id,
...
 	if (err)
 		return;
 
-	if (sscanf(desc, "NVMe%01d%c%02d %*s", &ver, &type, &hmac) != 3)
+	if (sscanf(desc, "NVMe%*1d%c%02d %*s", &type, &hmac) != 2)
 		return;
 
-	err = libnvmf_export_tls_key_versioned(ctx, ver, hmac, key_data, key_len,
+	/* PSK interchange format version, not the identity version */
+	err = libnvmf_export_tls_key_versioned(ctx, 1, hmac, key_data, key_len,
```
igaw, line 785: Same here.

completions/_nvme
```diff
@@ -8417,25 +8417,25 @@ _nvme () {
 					               "--output-format-version" "1 2" && return
 					local _keys_gen_tls
 					_keys_gen_tls=(
-						--keyring=':Keyring for the retained key.'
+						--keyring=':Keyring for the TLS PSK.'
```
igaw: the tab completions are auto generated directly from the command args definitions.  No need to update them.

plugins/keys/keys-plugin.c
igaw: and version 0 does exist, although deprecated. 

```
	0 - Obsolete (default); defined in the NVMe TCP Transport
	    Specification 1.0
```

plugins/keys/keys-plugin.c
```diff
@@ -420,18 +420,19 @@ static int do_insert_tls_key(struct libnvme_global_ctx *ctx, const char *keyring
...
+	const char *hmac =
+	    "Hash function for the retained PSK and for the cipher suite the TLS PSK is bound to (1 = SHA-256, 2 = SHA-384).";
+	const char *version = "TLS PSK identity version to use (0 = obsolete; 1 = TLS 1.3 with the PSK digest in the identity)";
+	const char *hostnqn = "Host NQN for the retained PSK and the TLS PSK identity.";
+	const char *subsysnqn = "Subsystem NQN for the TLS PSK identity.";
+	const char *keyring = "Keyring for the TLS PSK.";
+	const char *keytype = "Key type of the TLS PSK.";
+	const char *insert = "Additionally insert the TLS PSK into the keyring.";
+	const char *keyfile = "Append the identity of the TLS PSK, and the TLS PSK itself, to keyfile.";
+	const char *compat = "Use non-RFC 8446 compliant algorithm for deriving the retained PSK and the TLS PSK for older implementations.";
```
igaw, line 434: BTW, these changes here will be picked up the auto tab generator.

libnvme/src/nvme/no-crypto.c
```diff
@@ -15,13 +15,13 @@
...
 				  const unsigned char *key_data,
-				  size_t key_len, char **identity)
+				  size_t key_len, char **encoded_key)
 {
 	return -ENOTSUP;
 }
 
 __shr_public int libnvmf_export_tls_key(struct libnvme_global_ctx *ctx,
-	const unsigned char *key_data, int key_len, char **identity)
+	const unsigned char *key_data, int key_len, char **encoded_key)
```
igaw: the implementation calls the last argument `encoded_keyp`. I think it would make sense to use the same name here

Documentation/nvme-check-tls-key.txt
```diff
@@ -18,15 +18,20 @@ SYNOPSIS
...
+* '--insert'/'-i' and '--keyfile'/'-f' are gone. 2.x used them to derive
+  a TLS PSK and store it in the keyring; the new 'check-tls' never
+  modifies the keyring -- that moved to linknvme-keys-insert-tls[1].
+
+* The identity is no longer the first line of stdout. 2.x printed it
+  alone; the alias prints 'Key is valid (HMAC <id>, length <len>)' first
+  and reports the keyring state after it, so a script taking the
+  identity with 'head -1' reads the wrong line. Searching the keyring is
+  itself new, and a keyring that does not exist is reported as a
+  failure: the command exits 1 where 2.x exited 0.
```
igaw: What about making `check_tls` printing status information in vebose mode? I think it matches also the design we wanted to have for all commands. On success no output, return code 0.

igaw:
Overall great cleanup. 

The nitpicking on the key version is that the library or the nvme-cli doesn't need to care about the version at at. If it is the wrong version the kernel should reject it what it does. Sure we can add a warning or so but I really would like to avoid anything hard coded unless it is necessary.

About the additional output of the key check command: the idea here is to follow the POSIX style of commans. On success only return 0 (unless --verbose is given and the status has to go to stderr, not stdout). On error, write status to stderr and return -1.

plugins/keys/keys-plugin.c
kariya-mitsuru: The PSK interchange format version and the TLS PSK identity version are unrelated fields.

In NVMe TCP 1.0 the interchange format version was already 1, while the identity version was 0. Version 0 of the interchange format has never existed.

As the commit message says, the current code puts the identity version into the interchange format version field, which is simply wrong: a key with the default identity version 0 is written out as `NVMeTLSkey-0:...`, and `nvme keys import` then rejects it.

This is not a parser. It is the version written into the keyfile: the value never reaches the kernel, and the kernel does not accept the interchange format at all. And `libnvmf_import_tls_key_versioned()`, which is what reads this format back, rejects anything other than version 1.

So I believe this fix is correct.

plugins/keys/keys-plugin.c
kariya-mitsuru: Same as on the `append_keyfile()` thread, except that this one writes to the output of `nvme keys export` rather than to a keyfile.

completions/_nvme
kariya-mitsuru: Sorry, I edited them without noticing the "Do not edit." in the header. Dropped both files -- `_nvme` and `bash-nvme-completion.sh` -- from this series and force-pushed; the branch no longer touches `completions/` at all.

libnvme/src/nvme/no-crypto.c
kariya-mitsuru: Done -- and not only in `no-crypto.c`: `crypto.c` and `crypto.h`, including the kernel-doc, now use `encoded_keyp` for this out parameter as well, matching the name `libnvmf_export_tls_key_versioned()` already had.

Documentation/nvme-check-tls-key.txt
igaw: IIRC, I had some discussion with @martin-belanger about this command. I forgot the outcome of the discussion. Let me check on Monday what was the thinking here.

kariya-mitsuru: Thanks -- I will wait for your answer on whether this should change at all, and if so how.

One thing worth knowing before you decide: this only covers what nvme-cli itself prints. libnvme's own messages go to stdout, because global-ctx.c calls libnvme_set_logging_file(ctx, stdout) for every command, and -v/-vv raise its level to INFO/DEBUG.

For check-tls-key at least: without --verbose, an error still puts libnvme's message on stdout -- a bad CRC prints "CRC mismatch (key ..., crc ...)" there, while nvme-cli's own "Key decoding failed" goes to stderr. With -vv, libnvme prints on stdout on a successful run too. I have not checked the other commands.

plugins/keys/keys-plugin.c
igaw: Ah, got it. Well, the commit message is hard to read and it really lacks the important message: that these are two different version fields.  

And adding these comments into the code should be avoided. This should be commit message, which is exactly the  thing I would have needed in the first place.

```c
 /* PSK interchange format version, not the identity version */
```

igaw: FWIW, I'll update the commit message if the rest is okay.

completions/_nvme
igaw: Indeed, the generated tab completion should have an info which screams 'generated'.

libnvme/src/nvme/no-crypto.c
igaw: Thanks for synching these up.

Documentation/nvme-check-tls-key.txt
igaw: Faulty memory, the discussion on the output behavior was on the  'config validate, convert' commands. 

Since `nvme keys` is new and I think we should get these commands correct and accept breaking with 2.x here. 

```shell
$ .build/nvme keys gen-kxchap-secret | .build/nvme keys check-kxchap-secret | echo $?
0
$ .build/nvme keys gen-kxchap-secret | .build/nvme keys check-kxchap-secret
Secret is valid (HMAC 0, length 32, CRC 01b11c41)
```

If we follow here the POSIX command style, the status should only be printed with `-v`. 

Let's create a new issue on this. As this PR is for the renaming etc.

stdout vs stderr:  libnvme should also log to stderr. Let's fix on top of these changes.

igaw:
Thanks a lot!

PR #3889, opened 2026-08-21
rpmb: move into a plugin and split into subcommands

dwsuse:
RPMB is not part of the NVMe base specification -- it reuses Security Send/Receive as transport but the protocol itself is a Micron-originated vendor extension. Move it out of the always-built src/ core into an optional nvme-cli plugin (plugins/rpmb/), alongside security and sed.

While moving it, split the single nvme rpmb --cmd=<action> command into one subcommand per action (info, program-key, read-counter, read-data, write-data, read-config, write-config), each with only the options it needs, instead of one flag-driven mega-command. There is no deprecated alias for the old --cmd= form since there's no 1:1 mapping to preserve.

Also fixes a few latent bugs found while touching this code: a malloc()/libnvme_free() allocator mismatch in read_rpmb_key(), a missing NULL check after calloc() in auth_data_write_chunk(), rpmb --cmd=info always returning a failure exit code, and the target option being parsed with OPT_UINT into a char field.


dwsuse:
> RPMB is not part of the NVMe base specification

This statement is not correct. it is part of the spec. Though it still makes sense to move into a plugin.

PR #3890, opened 2026-08-21
tests: fix resource leak in run_handle_plugin()

sahmed-ibm:
run_handle_plugin() duplicates stdout and stderr with dup() before redirecting them for output capture. When the first dup() call succeeds but the second fails, the condition triggers an error jump to done: without closing the already-opened duplicate fd.

This leaks a file descriptor each time the error path is taken.

Close saved_stdout and saved_stderr individually on the error path, guarded by a >= 0 check, so only successfully opened handles are closed regardless of which dup() call failed.

Fixes: CID 562695, CID 562696


igaw:
Thanks!

PR #3891, opened 2026-08-21
libnvme: fix resource leak in read_discovery()

sahmed-ibm:
The read_discovery() function allocates discovery early in the parse path. Two error branches — HFI not found and security descriptor not found — jump to the shared cleanup label without setting r to a non-zero value first.

When get_heap_obj() for the NQN field succeeds and returns 0, r holds 0 at the point of the jump. The cleanup guard "if (r) free(discovery)" then evaluates false and the allocation is not freed, causing a memory leak.

Set r to -EINVAL before each of the two affected goto statements so the cleanup guard always frees discovery on every error path.


igaw:
Nice catch! Thanks!

PR #3892, opened 2026-08-21
shared: fix resource leak in test_add_columns_invalid_width

sahmed-ibm:
The test_add_columns_invalid_width() function calls shr_table_init_with_columns() and checks that it returns NULL for an invalid column width. If the allocation unexpectedly succeeds and returns a non-NULL t, the pointer goes out of scope at return with no corresponding free call.

This results in a resource leak each time the allocation succeeds on this path.

Free t after the check if it is non-NULL to prevent the leak.


igaw:
Thanks!

PR #3893, opened 2026-08-21
shared: fix resource leaks in test_read_file

sahmed-ibm:
The test_read_file() function calls shr_read_file() twice expecting NULL: once for a missing file and once for an empty file. The results are recorded with check_bool() but execution always continues. If shr_read_file() unexpectedly returns non-NULL on either path, buf is overwritten or goes out of scope without being freed first.

This results in a memory leak each time the unexpected path is taken.

Add free(buf) after each expected-NULL check to release any allocation before buf is reused or the enclosing block exits, matching the pattern used by all other shr_read_file() calls in the function.


igaw:
Thanks!

PR #3894, opened 2026-08-21
shared: check return value of remove() in dir_is_writable()

sahmed-ibm:
The dir_is_writable() helper writes a temporary marker file to confirm that a directory is accessible, then calls remove() to clean it up. The return value of remove() was not checked.

If remove() fails the function still returns true, leaving a stale marker file behind. Subsequent test runs may encounter the leftover file and produce incorrect results.

Check the return value of remove() and return false if cleanup fails, so the test framework surfaces any filesystem error.


igaw:
Thanks!

PR #3896, opened 2026-08-23
shared: add shr_parse_csv_u128 parser

ikegami-t:
Since needed to parse __u128 comma-separated list.

### DIFF
--- a/shared/parse-util.c
+++ b/shared/parse-util.c
@@ -85,3 +85,62 @@ DEFINE_SHR_PARSE_CSV_FUNC(ushort, unsigned short, UINT16_MAX)
 DEFINE_SHR_PARSE_CSV_FUNC(uint, unsigned int, UINT32_MAX)
 DEFINE_SHR_PARSE_CSV_FUNC(ulonglong, unsigned long long, ULLONG_MAX)
 DEFINE_SHR_PARSE_CSV_FUNC(uchar, unsigned char, UINT8_MAX)
+
+#ifdef NVME_HAVE_INT128
+int shr_parse_csv_uint128(char *string, uint128_t *val, unsigned int max_length)
+{
+	unsigned int digit;
+	int ret = 0;
+	uint128_t v;
+	char *temp;
+	char *src;
+	int base;
+
+	if (!string || !*string)
+		return 0;
+
+	temp = strtok(string, ",");
+
+	while (temp) {
+		if (ret >= max_length)
+			return -EINVAL;
+
+		errno = 0;
+		v = 0;
+		base = 10;
+		src = temp;
+
+		if (src[0] == '0' && (src[1] == 'x' || src[1] == 'X')) {
+			base = 16;
+			src += 2;
+		}
+
+		if (*src == '\0')
+			return -EINVAL;
+
+		while (*src) {
+			if (*src >= '0' && *src <= '9')
+				digit = *src - '0';
+			else if (base == 16 && *src >= 'a' && *src <= 'f')
+				digit = *src - 'a' + 10;
+			else if (base == 16 && *src >= 'A' && *src <= 'F')
+				digit = *src - 'A' + 10;
+			else
+				return -1;
+
+			if (v > (UINT128_MAX - digit) / base)
+				return -ERANGE;
+
+			v = v * base + digit;
+			src++;
+		}
+
+		val[ret] = v;
+		ret++;
+
+		temp = strtok(NULL, ",");
+	}
+
+	return ret;
+}
+#endif /* NVME_HAVE_INT128 */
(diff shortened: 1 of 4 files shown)


ikegami-t:
FYI: The function basically suggested by the google AI mode.

ikegami-t:
The armhf cross build failure will be fixed later.

shared/parse-util.c
```diff
@@ -85,3 +85,64 @@ DEFINE_SHR_PARSE_CSV_FUNC(ushort, unsigned short, UINT16_MAX)
...
+				digit = *src - '0';
+			else if (base == 16 && *src >= 'a' && *src <= 'f')
+				digit = *src - 'a' + 10;
+			else if (base == 16 && *src >= 'A' && *src <= 'F')
+				digit = *src - 'A' + 10;
+			else
+				return -1;
+
+			next_v = v * base + digit;
+			if (next_v < v) {
```
igaw: usually it's better to check if the operation will overflow before doing it.

Maybe something like this below (though I am not sure if this is correct)?
```c
	if (v > (UINT128_MAX - digit) / base)
		return -ERANGE;

	v = v * base + digit;
```

shared/parse-util.c
```diff
@@ -85,3 +85,64 @@ DEFINE_SHR_PARSE_CSV_FUNC(ushort, unsigned short, UINT16_MAX)
...
+	int base;
+
+	if (!string || !*string)
+		return 0;
+
+	temp = strtok(string, ",");
+
+	while (temp) {
+		if (ret >= max_length)
+			return -1;
```
igaw:
```c
return -EINVAL;
```

shared/parse-util.c
```diff
@@ -85,3 +85,64 @@ DEFINE_SHR_PARSE_CSV_FUNC(ushort, unsigned short, UINT16_MAX)
...
+		base = 10;
+		src = temp;
+
+		if (src[0] == '0' && (src[1] == 'x' || src[1] == 'X')) {
+			base = 16;
+			src += 2;
+		}
+
+		if (*src == '\0')
+			return -1;
```
igaw:
```c
return -EINVAL;
```

shared/parse-util.c
```diff
@@ -85,3 +85,64 @@ DEFINE_SHR_PARSE_CSV_FUNC(ushort, unsigned short, UINT16_MAX)
...
+			else if (base == 16 && *src >= 'a' && *src <= 'f')
+				digit = *src - 'a' + 10;
+			else if (base == 16 && *src >= 'A' && *src <= 'F')
+				digit = *src - 'A' + 10;
+			else
+				return -1;
+
+			next_v = v * base + digit;
+			if (next_v < v) {
+				fprintf(stderr, "%s out of 128-bit range\n",
```
igaw: Please remove the `fprintf` in the helper function. This is the callers job, thus it's important to return proper return error codes

shared/parse-util.h
```diff
@@ -67,3 +73,9 @@ int shr_parse_csv_uchar(char *string, unsigned char *val,
 		(unsigned char *)(val) + BUILD_ASSERT_OR_ZERO(		\
 			sizeof(*(val)) == sizeof(unsigned char)),	\
 		max_length)
+
+#define shr_parse_csv_u128(string, val, max_length)			\
```
igaw, line 82: while at it, could you move `shr_parse_csv_u8` up, so the definition are sorted their size? Thanks!

shared/tests/test-parse-util.c
```diff
@@ -303,6 +361,7 @@ int main(void)
 	pass &= test_csv_u32();
 	pass &= test_csv_u64();
 	pass &= test_csv_u8();
+	pass &= test_csv_u128();
```
igaw, line 368: Could you reorder the test so the are in order of the their size, e.g.

```c
test_csv_u8
test_csv_u16
test_csv_u32
test_csv_u64
test_csv_u128
```

Thanks

ikegami-t:
Just fixed for the review comments. Thank you.

ikegami-t:
For the armhf cross build just disalbed the function for now.

shared/parse-util.h
```diff
@@ -11,6 +11,13 @@
 
 #include <ccan/build_assert/build_assert.h>
 
+#ifdef NVME_HAVE_INT128
+typedef unsigned __int128 uint128_t;
+
+#define UINT128_CONST(high, low) (((uint128_t)(high) << 64) | (low))
+#define UINT128_MAX UINT128_CONST(UINT64_MAX, UINT64_MAX)
+#endif /* NVME_HAVE_INT128 */
```
igaw, line 19: Seems a reasonable approach. Glad you caught this!

igaw:
Thanks!

PR #3897, opened 2026-08-23
nvme: continue without PI when get_pi_info() fails

prabhakarpujeri:
In `submit_io()` (shared by `nvme read`, `nvme write` and `nvme compare`), a failure of `get_pi_info()` (e.g. the Identify Namespace command for the NVM command set is not supported by the drive) only set `pi_available` to false while `logical_block_size` stayed 0. Execution then continued with `logical_block_size == 0` and hit a divide-by-zero (SIGFPE) when computing the block count, unless `--block-count` was given.

Per review: `get_pi_info()` is allowed to fail, in which case PI is simply not available and the I/O should still be submitted. So the command is only aborted when the logical block size cannot be determined at all:

- `get_pi_info()` now returns the logical block size and metadata size as soon as the base Identify Namespace succeeds, so the block size is known even when PI information is not; when the requested PI tags are invalid it clears them again so the command still fails with `-EINVAL`.
- `submit_io()` no longer aborts on a `get_pi_info()` failure when the block size is known; PI is marked unavailable and the I/O is submitted without PI tags.
- As in `write_zeroes()` and `verify()`, an `NVME_SC_INVALID_FIELD` result from `init_pi_tags()` is tolerated instead of failing the command.

### DIFF
--- a/src/nvme-cmds-io.c
+++ b/src/nvme-cmds-io.c
@@ -317,6 +317,8 @@ static int get_pi_info(struct libnvme_transport_handle *hdl,
 	nvme_id_ns_flbas_to_lbaf_inuse(ns->flbas, &lba_index);
 	lbs = 1 << ns->lbaf[lba_index].ds;
 	ms = le16_to_cpu(ns->lbaf[lba_index].ms);
+	*logical_block_size = lbs;
+	*metadata_size = ms;
 
 	nvm_ns = libnvme_alloc(sizeof(*nvm_ns));
 	if (!nvm_ns)
@@ -349,8 +351,15 @@ static int get_pi_info(struct libnvme_transport_handle *hdl,
 			lbs += ms;
 	}
 
-	if (invalid_tags(lbst, ilbrt, sts, pif))
+	if (invalid_tags(lbst, ilbrt, sts, pif)) {
+		/*
+		 * The requested PI is invalid; clear the block size so
+		 * that the caller fails the command.
+		 */
+		*logical_block_size = 0;
+		*metadata_size = 0;
 		return -EINVAL;
+	}
 
 	*logical_block_size = lbs;
 	*metadata_size = ms;
@@ -1117,6 +1126,9 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	} else {
 		err = get_pi_info(hdl, cfg.nsid, cfg.prinfo,
 			cfg.ilbrt, cfg.lbst, &logical_block_size, &ms);
+		if (err && !logical_block_size)
+			return err;
+		/* PI is not available if get_pi_info() failed */
 		pi_available = err == 0;
 	}
 
@@ -1213,7 +1225,7 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	if (pi_available) {
 		err = init_pi_tags(hdl, &cmd, cfg.nsid, cfg.ilbrt, cfg.lbst,
 			cfg.lbat, cfg.lbatm);
-		if (err)
+		if (err && err != NVME_SC_INVALID_FIELD)
 			return err;
 	}
 	gettimeofday(&start_time, NULL);


src/nvme-cmds-io.c
```diff
@@ -1117,7 +1117,9 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	} else {
 		err = get_pi_info(hdl, cfg.nsid, cfg.prinfo,
 			cfg.ilbrt, cfg.lbst, &logical_block_size, &ms);
-		pi_available = err == 0;
+		if (err)
+			return err;
```
igaw, line 1130: It's not that simple, `get_pi_info` is allowed to fail, in this case PI is not available.

prabhakarpujeri: Understood, thanks for pointing that out. Reworked: a `get_pi_info()` failure now just means PI is unavailable, so `submit_io()` continues without PI tags instead of aborting. To keep the block count calculation safe, `get_pi_info()` sets the logical block size / metadata size as soon as the base Identify Namespace succeeds (clearing them only when the requested PI tags are invalid, which still fails the command with `-EINVAL`). The command now only aborts when the block size cannot be determined at all. Also aligned with `write_zeroes()`/`verify()`: an `NVME_SC_INVALID_FIELD` from `init_pi_tags()` is tolerated.

src/nvme-cmds-io.c
```diff
@@ -1213,7 +1225,7 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	if (pi_available) {
 		err = init_pi_tags(hdl, &cmd, cfg.nsid, cfg.ilbrt, cfg.lbst,
 			cfg.lbat, cfg.lbatm);
-		if (err)
+		if (err && err != NVME_SC_INVALID_FIELD)
```
igaw, line 1228: this is not needed as it is guarded by `pi_available`

src/nvme-cmds-io.c
```diff
@@ -1117,6 +1126,9 @@ static int submit_io(int opcode, char *command, const char *desc, int argc, char
 	} else {
 		err = get_pi_info(hdl, cfg.nsid, cfg.prinfo,
 			cfg.ilbrt, cfg.lbst, &logical_block_size, &ms);
+		if (err && !logical_block_size)
+			return err;
+		/* PI is not available if get_pi_info() failed */
```
igaw, line 1131: this comment is exactly what is written in the next line. don't over document.

src/nvme-cmds-io.c
```diff
@@ -349,8 +351,15 @@ static int get_pi_info(struct libnvme_transport_handle *hdl,
...
 	}
 
-	if (invalid_tags(lbst, ilbrt, sts, pif))
+	if (invalid_tags(lbst, ilbrt, sts, pif)) {
+		/*
+		 * The requested PI is invalid; clear the block size so
+		 * that the caller fails the command.
+		 */
+		*logical_block_size = 0;
+		*metadata_size = 0;
```
igaw, line 360: This is not a good API, it overloads the meaning of the output values.

src/nvme-cmds-io.c
```diff
@@ -349,8 +351,15 @@ static int get_pi_info(struct libnvme_transport_handle *hdl,
...
 		return -EINVAL;
```
igaw, line 361: The problem here is that we need to return somehow 'invalid tags' here and handle let the callsite to handle it. `-EINVAL` is not a clear error code. Thus I suggest to define an error code for nvme-cli as we have in libnvme

```c
enum libnvme_connect_err {
	ENVME_CONNECT_RESOLVE	= 1000,
	ENVME_CONNECT_ADDRFAM,
```

So for this case

```c
enum cli_error_codes {
	ECLI_INVALID_TAGS	= 2000,

```

which makes it simple on the callsite to filter out.

igaw: FWIW, there is a second call site of `invalid_tags` which should also handled correctly. 

I am working on extending our test framework so we can add tests cases for this which will also fix this part. So no need to update this patch.

PR #3898, opened 2026-08-23
nvme: return -EINVAL for invalid prinfo in submit_io

prabhakarpujeri:
The prinfo range check in `submit_io()` returns `err`, which is always zero at this point (only successful `parse_and_open()` / get-namespace-id results reach this line). Passing `--prinfo=0x10` or higher to `nvme read`/`write`/`compare` therefore exits with status 0 without submitting any I/O, silently faking success.

Return `-EINVAL`, matching the identical check in `write_zeroes()` and `verify_cmd()`.


igaw:
Thanks!

Issue #3900, opened 2026-08-24
completions: add 'generated' notice 


igaw:
Indeed, the generated tab completion should have an info which screams 'generated'.

_Originally posted by @igaw in https://github.com/linux-nvme/nvme-cli/pull/3887#discussion_r3842696340_

igaw:
@jimmunn could you update the tab generator to add some info at the beginning of the files? Thanks!

igaw:
#3904 has been merged

PR #3904, opened 2026-08-24
completions: make the generated-file notice scream

jimmunn:
The one-line "generated ... do not edit" comment was easy to miss. Replace it with a boxed all-caps banner in the bash, zsh, and PowerShell headers so it is obvious the files are generated and how to regenerate them.

Link: https://github.com/linux-nvme/nvme-cli/issues/3900


igaw:
Thanks a lot!

PR #3905, opened 2026-08-24
shared: fix string not null terminated in shr_read_file()

sahmed-ibm:
shr_read_file() allocates a buffer buf of exactly size bytes and fills it with fread(). The buffer is never null terminated before being returned to the caller.

Callers that pass buf to string functions such as strstr() trigger undefined behavior because those functions require a null-terminated string and will read past the end of the allocation.

Store the raw ftell() result in file_size, set size to file_size + 1 so the caller knows the true allocation size, allocate size bytes, fread file_size bytes, and null terminate at buf[file_size] to ensure the returned buffer is always null terminated.

### DIFF
--- a/shared/fs-util.c
+++ b/shared/fs-util.c
@@ -87,6 +87,7 @@ unsigned char *shr_read_file(const char *dir, const char *path, long *size, int 
 {
 	__cleanup_free char *file_path = NULL;
 	unsigned char *buf = NULL;
+	long file_size;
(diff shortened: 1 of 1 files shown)


igaw:
The data which is read from the file is generally binary, thus it would be wrong to assume it's a string, and that's also why the test fail.

sahmed-ibm:
Oh I didn't realize, that makes more sense. Thank you for the clarification!

PR #3907, opened 2026-08-25
shared: check the new row's val pointer in shr_table_get_row_id()

prabhakarpujeri:
The failure test allocates `t->rows[row].val` but checks `t->rows->val`, i.e. row 0. For every row after the first, row 0's `val` is already allocated and non-NULL, so a failed calloc for a later row goes undetected: the row id is returned as valid and `shr_table_set_value_*()` then writes through NULL.

Test against the row that was just allocated.


igaw:
Thanks!

PR #3908, opened 2026-08-25
amzn: clamp detail-I/O histogram counts to the compiled-in array sizes

prabhakarpujeri:
`num_of_hists` and `num_bins` come from the device log page and are used to index the fixed-size arrays `hist_io_sizes[8]`, `io_hist_array[8]`, `bins[64]` and `counts[32]`. A device reporting larger values walks past the end of these arrays in `amzn_print_detail_io()` and `amzn_json_create_detail_io()`.

Clamp both to the compiled-in limits, the same way `amzn_print_latency_histogram()` already bounds its own loop with `b < 64`.

### DIFF
--- a/plugins/amzn/amzn-nvme.c
+++ b/plugins/amzn/amzn-nvme.c
@@ -468,6 +468,20 @@ static void amzn_print_stats(struct amzn_latency_log_page *log,
 		amzn_print_normal_stats(log, detail);
 }
 
+/* the log page counts come from the device; cap to the buffer sizes */
+static void amzn_sanitize_log_page(struct amzn_latency_log_page *log)
+{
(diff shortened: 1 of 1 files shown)


igaw:
It make sense to validate input data but I think we should do this when we read/parse it. From a quick look, a log page is fetched in `get_stats` and then parsed in `amzn_compute_stats_diff`. I would expect checks there.

prabhakarpujeri:
Moved the checks into `get_stats` where the log page is fetched — both for the initial read and each poll iteration before `amzn_compute_stats_diff` parses it. Consumers now see capped counts, no per-consumer clamping.

igaw:
Thanks!

PR #3910, opened 2026-08-25
intel: check intel_vu_log allocation in get_internal_log()

prabhakarpujeri:
`get_internal_log()` mallocs an `intel_vu_log` and is later memcpy()ing the response header into it without ever checking the allocation. On OOM this is an immediate NULL deref in `memcpy(intel, buf, ...)`. Bail out with -ENOMEM before use.

### DIFF
--- a/plugins/intel/intel-nvme.c
+++ b/plugins/intel/intel-nvme.c
@@ -1352,10 +1352,10 @@ static int get_internal_log(int argc, char **argv, struct command *acmd,
 	int err, output, i, j, count = 0, core_num = 1;
 	struct libnvme_passthru_cmd cmd;
 	struct intel_cd_log cdlog;
-	struct intel_vu_log *intel = malloc(sizeof(struct intel_vu_log));
 	struct intel_vu_nlog *intel_nlog = (struct intel_vu_nlog *)buf;
-	struct intel_assert_dump *ad = (struct intel_assert_dump *) intel->reserved;
-	struct intel_event_header *ehdr = (struct intel_event_header *)intel->reserved;
+	struct intel_assert_dump *ad;
+	struct intel_event_header *ehdr;
+	__cleanup_free struct intel_vu_log *intel = NULL;
 	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
 	__cleanup_nvme_transport_handle struct libnvme_transport_handle *hdl = NULL;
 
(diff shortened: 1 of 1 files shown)


plugins/intel/intel-nvme.c
```diff
@@ -1359,6 +1359,9 @@ static int get_internal_log(int argc, char **argv, struct command *acmd,
 	__cleanup_nvme_global_ctx struct libnvme_global_ctx *ctx = NULL;
 	__cleanup_nvme_transport_handle struct libnvme_transport_handle *hdl = NULL;
 
+	if (!intel)
+		return -ENOMEM;
+
```
igaw: can you move this down after deceleration section. I would also move the malloc down, this is also not really nice. and while at it, just introduce the `__cleanup_free` stuff, e.g.


```c
__cleanup_free struct intel_vu_log *intel = NULL;

[...]

intel = malloc(sizeof(*intel));
if (!intel)
   return -ENOMEM;
```

Thanks!

prabhakarpujeri: Reworked as suggested: `__cleanup_free struct intel_vu_log *intel = NULL;`, malloc moved past `parse_and_open`, derived pointers set right after the NULL check, and the manual free() calls dropped.

igaw:
Looks way better. Thanks!

HaroPanosyan:
Just to note, around 5-6 years ago Intel sold its SSD business to SK-Hynix which started a new company called Solidigm. If you need to use intel plugin then all good, but "solidigm" plugin is where we continue to make updates and fixes.

igaw:
Unfortunately, it's a common pattern that companies abandon projects/products and the maintainers of open source projects are left to deal with the fallout. In this case, we don't know can we just drop the intel plugin, can it be replaced by the solidigm one? Who is taking up the job to maintain it?  

I am not sure how to deal with this, but I think I'll have some discussion during ALPSS with the rest of the community and see what's the outcome of it.

Yeah, I know that I am barking at the wrong tree here, it's just really bad for everyone.

fyodor-soiguine-sndk:
Don't forget WDC no longer makes SSDs either.

igaw:
Yes, that's true but Sandisk still maintains the wdc plugin. @jeff-lien-sndk thank  you!

jeff-lien-sndk:
@igaw @fyodor-soiguine-sndk We will make any critical bug fixes in the WDC plugin if a customer pushes for it,  but do not plan to add any new features to the WDC plugin.  We are recommending that our users move to the Sandisk plugin.

PR #3913, opened 2026-08-25
libnvme: nbft: reject heap objects too small for the structures they decode

prabhakarpujeri:
## Problem

Port of linux-nvme/libnvme#1125 to libnvme3.

The SSNS extended-info descriptor now validates its disclosed length (good), but three heap-object reads can still overrun:

1. `read_hfi_info_tcp()` dereferences the trinfo heap object as a full `struct nbft_hfi_info_tcp` (flags, mac, addresses, …) without verifying the object's declared length covers even the fields read unconditionally.
2. the SSNS `subsys_traddr_obj` is handed to `format_ip_addr()`, which always `memcpy()`s a 16-byte `struct in6_addr` regardless of the declared object length.
3. the newer HFI `trinfo_version >= 2` ext-info object is dereferenced/read via `read_hfi_info_dhcp()` with the same under-declared trap.

`get_heap_obj()` validates only heap containment (`in_heap`) plus string NUL-termination; a file can point a struct-typed heap object at `heap_end - 2` and later reads walk past the heap.

## Fix

- `__get_heap_obj()` gains a `min_len` parameter rejecting non-string heap objects shorter than the structure the caller dereferences.
- `get_heap_obj()` derives the minimum from the output pointer type via `_Generic` (strings, byte arrays, and the SSNS ext-info reader keep min 0 — the latter does its own spec-length check); call sites lose their `(char **)` casts.
- Explicit `>= sizeof(struct in6_addr)` check for the SSNS transport address before `format_ip_addr()`.

## Testing

- Full meson suite green (99 pass; crypto-util skip is env-related: no OpenSSL on this host, also on pristine master).
- Checked the NBFT fixture dumps are unchanged by the patch semantics (valid tables remain accepted; only under-sized heap objects — never correctly decodable — are rejected).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -177,6 +177,11 @@
   instead of alongside current commands, and running one prints a
   warning that it will be removed in the next major version.
 
+* The project now builds with `-std=gnu11` instead of `-std=gnu99`,
+  raising the minimum required compiler support from C99 to C11. A
+  toolchain new enough for C11 is now required to build nvme-cli and
+  libnvme.
+
 * `nvme disconnect-all` with no options no longer disconnects every
   fabric controller. It now only disconnects controllers with no
   recorded owner in the new ownership registry. A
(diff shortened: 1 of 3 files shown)


igaw:
droped the reference to libnvme 1.

Didn't know about `_Generic`, nice C11 feature BUT we only require gnu99 as min version.

igaw:
I am taking care of the patch, please don't update it.

I don't have a problem to update the min requirement for the compiler. As far I can tell all major distro ship compilers which are able to deal with C11. And if not, it's about time to update.

igaw:
Thanks!

PR #3914, opened 2026-08-25
libnvme: examples: fix select() usage in telemetry-listen

prabhakarpujeri:
## Problem

Port of linux-nvme/libnvme#1126 to libnvme3 (the example has the same misuse vector).

`wait_events()` passes the *count* of uevent fds to `select()` as `nfds`; the argument is max-fd+1, and with stdio occupying 0-2 the uevent fds always start at 3, so for the common 1-3 controller case the watcher never wakes on any telemetry AEN. Additionally, `select()` clears inactive fds in the passed set, and the same set is reused — after the first event the loop only monitors fds that happened to be active. With no controller it spins at 100% CPU, and `calloc()` isn't checked.

## Fix

Track maxfd while `FD_SET`ing; pass `maxfd + 1` to `select()` on a re-armed local copy of the set; skip the wait loop when no controller was found; check `calloc()`.

## Testing

- Full meson suite green. The select-nfds semantics were verified with a standalone snippet in the 1.x PR (same mechanism applies verbatim).

### DIFF
--- a/libnvme/examples/telemetry-listen.c
+++ b/libnvme/examples/telemetry-listen.c
@@ -110,7 +110,7 @@ static void check_telemetry(libnvme_ctrl_t c, int ufd)
 	}
 }
 
-static void wait_events(fd_set *fds, struct events *e, int nr)
+static void wait_events(fd_set *fds, struct events *e, int nr, int maxfd)
 {
 	int ret, i;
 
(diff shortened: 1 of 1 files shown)


igaw:

igaw:
Thanks!

PR #3918, opened 2026-08-25
libnvme: account for the appended spaces in libnvmf_get_entity_version()

prabhakarpujeri:
## Problem

Port of linux-nvme/libnvme#1136. `libnvmf_get_entity_version()` rests on the invariant `num_bytes + bufsz == capacity` maintained by `read_file()`, but each of the three `' '` separators is appended to `buffer` without checking or decrementing `bufsz`. After such an append the invariant is off by one, so with proc content long enough to fill the buffer the next separator lands one byte past the allocation, and the following `min(len, bufsz)` + `memcpy()` still trusts the inflated `bufsz` (one more byte of overrun then).

## Fix

Guard the separator appends with `if (bufsz)` and decrement `bufsz` in step so the invariant holds, keeping the trailing `memset(&buffer[num_bytes], 0, bufsz)` in bounds.

## Testing

- Full meson suite green (99 pass; crypto-util skips here for missing OpenSSL, also on pristine master).
- Arithmetic/invariant verified by review against the 1.x PR; the overrun requires unusually long `/proc/sys/kernel/osrelease` content so it is rare in practice, but the function's contract becomes provable.

### DIFF
--- a/libnvme/src/nvme/util-fabrics.c
+++ b/libnvme/src/nvme/util-fabrics.c
@@ -115,7 +115,10 @@ size_t libnvmf_get_entity_version(char *buffer, size_t bufsz)
 	/* /proc/sys/kernel/osrelease contains the Linux
 	 * version (e.g. 5.8.0-63-generic)
 	 */
-	buffer[num_bytes++] = ' '; /* Append a space */
+	if (bufsz) {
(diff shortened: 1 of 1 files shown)


igaw:

Thanks!

PR #3922, opened 2026-08-25
build: fail the release if completion generation fails

jimmunn:
The generate-completions.py call in release.sh was unchecked; guard it like the surrounding build steps so a failure aborts the release.


igaw:
Thanks!

PR #3924, opened 2026-08-26
nvme-print: dump command datum always

ikegami-t:
Since host to controller data transfer needed to dump for the dry-run.


ikegami-t:
Note: The dump data changed to not output for the dry-run by the commit: 2b2cbf7.

igaw:
Thanks!

PR #3926, opened 2026-08-26
libnvme: fix crash on a malformed exclusion entry

martin-belanger:
Any exclusion entry with an unknown key, a token without `=`, or an empty value crashes every caller of `libnvmf_exclusion_match()`: `nvme connect-all`, `nvme-discoverd`, and anything else that consults the list.

```
printf '[exclusions]\nexclusion = bogus=x\n' > /etc/nvme/exclusions.conf
nvme connect-all ...     # SIGSEGV
```

`entry_matches()` parsed each entry with a NULL context to keep the match path quiet, but `tid_parse()` logs those three cases and `libnvme_msg()` dereferences the context.

The fix threads the caller's real context down instead, and reports a bad entry at DEBUG. Matching runs on every connect attempt, so one static typo must not produce a log line per connect. `libnvmf_tid_parse()` and `libnvmf_tid_parse_strict()` now return `-EINVAL` for a NULL context rather than documenting it as "parse silently", which was never true.

The second commit makes the typo visible where an admin would look for it. `nvme exclusion list` showed only drop-in names, so the entries of the main hand-edited `/etc/nvme/exclusions.conf` could not be listed at all. It now prints every list with its entries and marks the ones that can never match:

```
$ nvme exclusion list
(default)
  nqn=nqn.2014-08.org.nvmexpress:uuid:abc
  bogus=x  # invalid: never matches

user
  transport=tcp;traddr=10.0.0.1
```

`--verbose` logs the reason for each one.

Regression tests cover the four malformed forms plus a hostname in `traddr`, which was silently rejected with no diagnostic before.

### DIFF
--- a/Documentation/nvme-exclusion-list.txt
+++ b/Documentation/nvme-exclusion-list.txt
@@ -12,9 +12,15 @@ SYNOPSIS
 
 DESCRIPTION
 -----------
-Without --name, print the name of every named exclusion list under
-/etc/nvme/exclusions.conf.d/.  With --name, print the entries contained
-in that one list.
+Without --name, print every exclusion list and its entries: the main
+hand-edited /etc/nvme/exclusions.conf first, shown as '(default)', then
+each named list under /etc/nvme/exclusions.conf.d/.  With --name, print
+the entries of that one list.
+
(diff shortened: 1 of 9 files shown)


igaw:
Thanks!

PR #3927, opened 2026-08-26
libnvme: add the global context to the host ID generators

martin-belanger:
Three public APIs took no global context:

```c
char *libnvmf_generate_hostid(void);
char *libnvmf_generate_hostnqn_from_hostid(char *hostid);
char *libnvmf_generate_hostnqn(void);
```

Every public API takes a context, whether or not it uses one. These were the only exported functions left without one — I checked the whole library, not just this call path.

Threading the context down has a concrete benefit beyond the rule. The host identifier is read from sysfs, and the chain below these three had no context to pass to `make_sysfs_dir()`, so it called `make_sysfs_dir(NULL, ...)` and got the production path back. `libnvme_set_test_sysfs_dir()` did not apply to `/proc/device-tree/ibm,partition-uuid` or `/sys/firmware/dmi/entries`. The `product_uuid` read did not go through `make_sysfs_dir()` at all — it opened a literal in `fabrics.c`. Since `product_uuid` is the first source tried, that one alone was enough to defeat the sandbox on a typical x86 machine.

With the context in place, a unit test can point host-ID generation at a fixture instead of the machine it runs on. `test-fabrics` now does exactly that: it writes a known UUID into a temporary `test-sysfs-dir` and asserts that `libnvmf_generate_hostid()` returns it, and that `libnvmf_generate_hostnqn()` builds the matching NQN. That test was not expressible before.

The three symbols change signature, so this is an ABI break. Their names are unchanged, so the version scripts need no edit.

Callers updated: `libnvmf_host_get_ids()`, `nvme gen-hostnqn`, `nvme show-hostnqn`, and the `no-fabrics.c` stubs. Verified with the default build, `-Dfabrics=disabled`, the full test suite, valgrind and checkpatch.

Two related items left out of this PR, both smaller and neither an ABI change:

- `libnvmf_read_hostnqn()` and `libnvmf_read_hostid()` take a context but still read the compile-time `SYSCONFDIR "/nvme/hostnqn"` and `.../hostid` literals, so a sandbox does not reach them either. Same call path as this fix.
- `libnvmf_config_load*()` falls back to `SYSCONFDIR "/nvme/nvme-fabrics.conf"` the same way.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -90,12 +90,13 @@ const char *arg_str(const char * const *strings,
 #define NVMF_HOSTNQN_FILE	SYSCONFDIR "/nvme/hostnqn"
 #define NVMF_HOSTID_FILE	SYSCONFDIR "/nvme/hostid"
 
-static int uuid_from_device_tree(char *system_uuid)
+static int uuid_from_device_tree(struct libnvme_global_ctx *ctx,
+				 char *system_uuid)
 {
 	__cleanup_fd int f = -1;
 	ssize_t len;
 
-	f = open(libnvme_uuid_ibm_filename(), O_RDONLY);
+	f = open(libnvme_uuid_ibm_filename(ctx), O_RDONLY);
 	if (f < 0)
 		return -ENXIO;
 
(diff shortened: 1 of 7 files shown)


igaw:
Nice! I forgot about those. Thanks a lot!

PR #3928, opened 2026-08-26
ioctl-win: retry skipped log pages with generic passthrough

brgoing:
It was discovered that IOCTL_STORAGE_QUERY_PROPERTY returns ERROR_INVALID_FUNCTION for logs not listed in the supported logs list without ever submitting the command. It was also discovered that IOCTL_STORAGE_PROTOCOL_COMMAND works for getting log pages without this early filtering behavior, although this is not documented.

Microsoft documents IOCTL_STORAGE_QUERY_PROPERTY as the preferred method for getting logs, and it handles large log requests more cleanly. Keep it as the default, but try IOCTL_STORAGE_PROTOCOL_COMMAND if ERROR_INVALID_FUNCTION is returned, allowing the controller to respond. This behavior better matches Linux's passthrough behavior.

* Retries get log page commands that fail with ERROR_INVALID_FUNCTION using generic IOCTL_STORAGE_QUERY_PROPERTY.
* Uses IOCTL_STORAGE_QUERY_PROPERTY for CSI values other than NVM.
* Fixes retry check in libnvme_get_log_dynamic_chunk to better handle NVME statuses.  Error found during testing.

### DIFF
--- a/libnvme/src/nvme/ioctl-win.c
+++ b/libnvme/src/nvme/ioctl-win.c
@@ -254,25 +254,39 @@ __shr_public int libnvme_update_block_size(struct libnvme_transport_handle *hdl,
 }
 
 /*
+ * IOCTL_STORAGE_PROTOCOL_COMMAND supports a maximum transfer size of 512 pages.
+ * However, the driver does not correctly enforce this limit and larger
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #3930, opened 2026-08-27
libnvme: python: stop freeing tree-owned objects in destructors

martin-belanger:
The Host, Subsystem and Ns constructors are lookup-or-create. They can return an object the tree already owns, so a Python handle going out of scope frees an object the tree still references. A later tree walk or a second handle to the same object then hits freed memory, which crashes the interpreter (SIGSEGV).

Keep the destructor declarations with an empty body. This may occasionally leave a few orphaned objects that will live until the tree is freed, but it's better than getting a SIGSEGV on a double free.


igaw:
Thanks!

PR #3932, opened 2026-08-27
shared/tests: fix improper use of negative row id in test_basic_table

sahmed-ibm:
test_basic_table() calls shr_table_get_row_id() and records whether the returned row is non-negative, but does not guard the code that follows. shr_table_get_row_id() can return -ENOMEM when either of its internal allocations fails.

A negative row is then passed as an array index to shr_table_set_value_str(), shr_table_set_value_int(), and shr_table_add_row(), all of which use it to dereference the rows array. This results in undefined behaviour.

Return early after the failed check, freeing the table to avoid a resource leak, so that the negative row is never used as an array index.


igaw:
Thanks!

PR #3933, opened 2026-08-27
shared: fix missing negative-return check in test_invalid_format_type

sahmed-ibm:
The test_invalid_format_type() function calls shr_table_get_row_id() to allocate a new table row and stores the result in row. The return value was never checked before row was used as an array index and passed to shr_table_set_value_int().

If shr_table_get_row_id() fails and returns -ENOMEM, the subsequent use of row as a negative array index results in an out-of-bounds memory access and undefined behavior.

Guard the result with a check_bool() assertion and return early if row is negative, matching the pattern already used in every other test function in the file.


igaw:
Thanks!

PR #3935, opened 2026-08-27
shared: fix improper use of negative row id in test_multi_type_and_centered

sahmed-ibm:
shr_table_get_row_id() can return a negative error code on allocation failure. The test records the check result but does not bail out, so ra and rb are passed directly to shr_table_set_value_str() and the other set_value helpers.

Those helpers guard with a signed comparison against num_rows, which a negative ra or rb passes. The negative value is then used as an array index into t->rows[], resulting in an out-of-bounds access and undefined behavior.

Return early when ra or rb is negative so the set_value calls are only reached with a valid row id.


igaw:
Thanks!

PR #3938, opened 2026-08-27
doc: nvme-fabrics.conf is shared by the nvme-cli tools, not by everyone

martin-belanger:
nvme-fabrics.conf(5) and design/CONFIG.md both said nvme-stas reads this file, alongside "nvme connect-all" / "nvme config" and nvme-discoverd. That is wrong, and it has already been read as an invitation to make nvme-stas take parameters out of it.

The nvme-cli tools and nvme-discoverd work together from this one shared file. nvme-stas is a separate orchestrator: same format, same parser, its own file, so a host can run both without either acting on the other's configuration. Say format rather than file in the three places that got this wrong, and stop calling a connection configuration host-wide: only the registry and the exclusion list are one per machine.

### DIFF
--- a/Documentation/nvme-fabrics.conf.txt
+++ b/Documentation/nvme-fabrics.conf.txt
@@ -14,9 +14,14 @@ SYNOPSIS
 DESCRIPTION
 -----------
 An INI-format file describing which NVMe over Fabrics (NVMe-oF)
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #3939, opened 2026-08-27
libnvme: carry the fabrics context credentials into the controller

sithglan:
## What

`libnvmf_create_ctrl()` forwards only `fctx->ctrl_params` to
`libnvme_create_ctrl()`. The authentication and transport encryption settings
live on `struct libnvmf_context` *next to* `ctrl_params`, not inside it:

```c
	/* NVMe controller parameters */
	struct libnvme_ctrl_params ctrl_params;
	...
	/* authentication and transport encryption configuration */
	const char *hostkey;
	const char *ctrlkey;
	const char *keyring;
	char *tls_key;
	const char *tls_key_identity;
```

so a controller created from a context that carries any of them comes out with
none of them. `build_options()` then emits an argstr ending in `,tls` with no
key at all and the connect fails with `ENOKEY`:

```
connect ctrl, 'nqn=…,ctrl_loss_tmo=600,tls'
Failed to write to /dev/nvme-fabrics: Required key not available
```

`libnvmf_connect()` never hit this: it called `libnvme_create_ctrl()` itself
and applied the credentials in its own body afterwards. Everything that goes
through the public helper did — including the libnvme3 Python bindings, whose
`nvme.Ctrl()` constructor accepts `hostkey`, `ctrlkey`, `keyring`, `tls_key`
and `tls_key_identity`, stores them on the fabrics context via
`libnvmf_context_set_crypto()`, and then loses them again.

Reported by @martin-belanger while reviewing
[nvme-stas#500](https://github.com/linux-nvme/nvme-stas/pull/500#issuecomment-5443118566),
where nvme-stas had to work around it by setting the keys on the `nvme.Ctrl`
object after construction.

## The change

The credential handling moves into `libnvmf_create_ctrl()`, so the function
returns a controller that reflects the whole context rather than half of it,
and `libnvmf_connect()` calls it instead of repeating the sequence.
`nvme_parse_tls_args()` moves up unchanged to be visible from there. No
behaviour change for `libnvmf_connect()`; `examples/discover-loop.c` and the
Python bindings get the credentials they were already asking for.

## Tests

Two regression tests, both of which fail without the fix:

- `libnvme/tests/test-fabrics.c` — `test_create_ctrl_credentials()`: named keys
  land on the controller as strings, numeric ones land in `cfg.keyring_id` /
  `cfg.tls_key_id`, and the fabrics config still comes across.
- `libnvme/libnvme3/tests/test-ctrl-crypto.py` — the reported symptom at the
  binding level: keys handed to `nvme.Ctrl()` must be readable back off the
  controller, and a controller built without keys must not grow any.

`meson test`: 115 ok, 0 fail (python bindings enabled).

## End-to-end

Patched and unpatched builds side by side against a NetApp ONTAP 9.19.1
subsystem over NVMe/TCP with TLS 1.3 (kernel 7.1.8), connecting the way
nvme-stas does — credentials in the `nvme.Ctrl()` dict — with the `.nvme`
keyring emptied first so the dict is the only source of the PSK:

| | argstr | result |
|---|---|---|
| before | `…,ctrl_loss_tmo=600,tls` | `Required key not available` |
| after | `…,keyring=0x1c398c1a,tls_key=0x0fadb176,tls` | controller `live` |

A 512 KiB write/read round-trip over the TLS path compares identical. Passing
`tls_key_identity` as well also works and reuses the retained key rather than
re-importing it (key serial unchanged).

Note that two things mask the bug on a normal host, in case this looks
untriggerable: `libnvmf_add_ctrl()` copies the keys over from a sibling
controller on the same subsystem, and the kernel resolves a pre-provisioned
PSK by identity on its own. It shows up on the first connection to a subsystem
whose PSK has not been provisioned into the keyring by other means.

### DIFF
--- a/libnvme/libnvme3/meson.build
+++ b/libnvme/libnvme3/meson.build
@@ -99,6 +99,7 @@ if want_python
         [ 'registry', [ files('tests/test-registry.py'), ] ],
         [ 'config', [ files('tests/test-config.py'), ] ],
         [ 'setattr-guards', [ files('tests/test-setattr.py'), ] ],
+        [ 'ctrl-crypto', [ files('tests/test-ctrl-crypto.py'), ] ],
     ]
     foreach test: py_tests
         description = test[0]
--- a/libnvme/libnvme3/tests/test-ctrl-crypto.py
+++ b/libnvme/libnvme3/tests/test-ctrl-crypto.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# This file is part of libnvme.
+"""Tests that the crypto keys given to Ctrl() land on the controller.
+
+Regression: libnvmf_create_ctrl() forwarded only ctrl_params, so 'keyring',
+'tls_key', 'tls_key_identity', 'hostkey' and 'ctrlkey' were accepted by the
+constructor, stored on the fabrics context, and then silently dropped -- the
+connect went out without a key and failed with ENOKEY.
+"""
+
+import unittest
+from libnvme3 import nvme
+
+
+class TestCtrlCrypto(unittest.TestCase):
+
+    def setUp(self):
+        self.ctx = nvme.GlobalCtx()
+        (hostnqn, hostid) = nvme.host_get_ids(self.ctx)
+        self.ctx.hostnqn = hostnqn
+        self.ctx.hostid = hostid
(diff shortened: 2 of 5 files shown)


igaw:
Good catch! It’s very cool that you’re doing end-to-end tests with storage servers. Thanks a lot!

PR #3940, opened 2026-08-27
shared: fix tautological range check in DEFINE_SHR_PARSE_CSV_FUNC

sahmed-ibm:
The CSV parser macro uses strtoumax() to parse each token into a uintmax_t v, then checks overflow with errno == ERANGE or v > ret_max before narrowing the value into the destination type.

For the ulonglong instantiation, ret_max is ULLONG_MAX and v is uintmax_t, which are the same underlying type with the same maximum value. The comparison v > ULLONG_MAX is always false regardless of v, making it dead code that provides no protection.

Replace the v > ret_max comparison with (ret_t)(v) != v, which detects truncation on narrowing assignment for all destination types and avoids the tautological comparison for ulonglong. With ret_max no longer used in the macro body, remove it from the macro parameter list and all call sites, and drop the now-unused <limits.h> include.

### DIFF
--- a/shared/parse-util.c
+++ b/shared/parse-util.c
@@ -8,7 +8,6 @@
 
 #include <errno.h>
 #include <inttypes.h>
-#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3941, opened 2026-08-28
nvme: make --quiet suppress informational output

martin-belanger:
--quiet had no effect. map_log_level() returned LIBNVME_LOG_ERR for it, the same level verbose == 0 already returns, and show_message() never consulted nvme_args.quiet.

nvmf-connect@.service passes --quiet on every udev-triggered connect-all. However, connect-all --quiet would log a "skip connect" message for every Discovery Log Page change AEN. That skip is the designed outcome, not an error, so it now uses nvme_show_message().

Errors still print under --quiet.


igaw:
Thanks!

PR #3942, opened 2026-08-28
libnvme: leave registry ownership alone on a reused connection

martin-belanger:
nvme connect wrote the ownership registry even when it did not create the controller. A bare connect against a controller nvme-stas or nvme-discoverd already holds deleted that controller's ownership entry, with no flag, no prompt and no output. The controller keeps running, but it now reads as unowned and loses the protection the registry gives it.

Passing --owner=NAME to claim a controller is a stated intent and keeps working. Without --owner no intent is stated, so an existing entry is left alone.

A controller the kernel just created still gets its entry cleared, since a recycled instance number can leave a stale one behind. An empty --owner= now disowns instead of recording an empty owner.


igaw:
Thanks!

PR #3943, opened 2026-08-28
libnvme, python: expose a configuration's top-level defaults and host identity

martin-belanger:
@igaw: My nvme-stas work to incorporate all the goodies of libnvme made me realize that something was missing from the public API (sorry :man_facepalming:). 

Two read-side additions to the Python bindings. 

### `config_defaults()` — the defaults for a discovered controller

Returns `{'dc': {...}, 'ioc': {...}}`, kept apart because the caller picks by the class of controller it is about to connect. No new C API.

### `config_host()` — the top-level `[Host]`

Returns the default persona's `hostnqn`, `hostid` and `hostsymname`.

**This one adds public C API**, which I would rather flag than slip past an RC. `libnvmf_config_get_hostnqn()`, `libnvmf_config_get_hostid()` and `libnvmf_config_get_hostsymname()`. Three getters, three exported symbols, no existing signature or behaviour changed.

### DIFF
--- a/libnvme/libnvme3/nvme.i
+++ b/libnvme/libnvme3/nvme.i
@@ -777,6 +777,47 @@ PyObject *config_defaults(struct libnvme_global_ctx *ctx, const char *file)
 	return dict;
 }
 
+PyObject *config_host(struct libnvme_global_ctx *ctx, const char *file)
+{
+	__cleanup_config struct libnvmf_config *config = NULL;
+	PyObject *dict;
+	const char *val;
+	int ret;
+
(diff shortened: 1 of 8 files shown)


igaw:
> @igaw: My nvme-stas work to incorporate all the goodies of libnvme made me realize that something was missing from the public API (sorry 🤦‍♂️).

No worries. One of the reason for delaying the release was to give you time to finish the port of nvme-stas and figure out the problems. It seems to pay off :)

Cheers!

PR #3946, opened 2026-08-29
feat: add lba-range-type command

ikegami-t:
This is for LBA Range Type feature to get and set.


ikegami-t:
FYI: The host-behavior-support documentation provided by the google AI mode basically.

igaw:
Thanks!

PR #3947, opened 2026-08-30
doc: fix to add copy command namespace-id option description

ikegami-t:
The option missed to describe and describe as a mandatory option.

### DIFF
--- a/Documentation/nvme-copy.txt
+++ b/Documentation/nvme-copy.txt
@@ -9,6 +9,7 @@ SYNOPSIS
 --------
 [verse]
 'nvme' [<global-options>] 'copy' <device>
+			[--namespace-id=<nsid> | -n <nsid>]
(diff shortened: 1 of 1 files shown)


Documentation/nvme-copy.txt
```diff
@@ -8,7 +8,7 @@ nvme-copy - Send an NVMe Simple Copy command, provide results
 SYNOPSIS
 --------
 [verse]
-'nvme' [<global-options>] 'copy' <device>
+'nvme' [<global-options>] 'copy' <device> [--namespace-id=<nsid> | -n <nsid>]
```
igaw: Could you move the [] to a new line so it looks more consistent?

ikegami-t: Right then just fixed as so. Thank you.

igaw:
Thanks!

PR #3948, opened 2026-08-31
libnvme: python: decode discovery log page fields the way C does

martin-belanger:
The Python typemap hand-rolled its own strings for six enum fields that libnvme already has decoders for, and they had drifted: "fc" against "fibre-channel", "tls1.3" against "tls13", and so on.

treq was not merely different but wrong: NVMF_TREQ_DISABLE_SQFLOW is a flag that ORs with the base value, and the switch treated it as a value of its own, so a target reporting "required, sq flow control disable supported" matched nothing.


igaw:
Thanks! (no, I am not here :))

PR #3949, opened 2026-08-31
ocp: print physical media units written/read as 128-bit decimal value

brgoing:
Physical media units written/read are 128-bit values.  They were being converted and printed as two separate 64-bit values.  Update to use uint128 utility methods for endian conversion and printing.


igaw:
Thanks!

PR #3950, opened 2026-08-31
utils: fix resource leak in copy_opt_val on strdup failure

sahmed-ibm:
The copy_opt_val() function deep-copies an opt_val table by iterating over each entry and duplicating its str field with strdup(). The dst array is allocated with calloc() before the loop begins.

When strdup() fails for any entry, the function returned -ENOMEM immediately without freeing the already-allocated dst array or the str copies made in previous iterations, leaking all of that memory.

Free all previously duplicated str entries and dst itself before returning on the OOM path to prevent the resource leak.


igaw:
Thanks!

PR #3951, opened 2026-09-01
registry: fix unchecked return value from libnvmf_registry_attr_for_each

sahmed-ibm:
The print_device() callback calls libnvmf_registry_attr_for_each() to iterate over per-controller registry attributes, but discards the return value. The function returns a negative errno if the device directory cannot be opened at the time of the call.

Silently ignoring the error means registry_list() always returns 0 even when attribute enumeration fails for one or more devices, hiding real I/O errors from the caller.

Change the callback signature of libnvmf_registry_device_for_each() from void to int so that each callback can propagate errors directly. print_device() now returns the result of libnvmf_registry_attr_for_each() and the loop in libnvmf_registry_device_for_each() breaks on the first non-zero return, propagating it to the caller. Update the test callback collect_device() to match the new int-returning signature.

### DIFF
--- a/libnvme/libnvme3/nvme.i
+++ b/libnvme/libnvme3/nvme.i
@@ -794,7 +794,7 @@ PyObject *registry_retrieve(struct libnvme_global_ctx *ctx,
 	return str;
 }
 
-static void _registry_collect_device(const char *device, void *user_data)
+static int _registry_collect_device(const char *device, void *user_data)
 {
 	PyObject *str = PyUnicode_FromString(device);
 
@@ -802,6 +802,7 @@ static void _registry_collect_device(const char *device, void *user_data)
 		PyList_Append((PyObject *)user_data, str);
 		Py_DECREF(str);
 	}
+	return 0;
 }
 
 PyObject *registry_devices(struct libnvme_global_ctx *ctx)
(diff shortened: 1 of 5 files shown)


igaw:
```
 libnvme/libnvme3/nvme_wrap.c:4182:53: error: passing argument 2 of ‘libnvmf_registry_device_for_each’ from incompatible pointer type [-Wincompatible-pointer-types]
 4182 |         ret = libnvmf_registry_device_for_each(ctx, _registry_collect_device, list);
      |                                                     ^~~~~~~~~~~~~~~~~~~~~~~~
      |                                                     |
      |                                                     void (*)(const char *, void *)
In file included from libnvme/src/libnvme.h:33,
                 from libnvme/libnvme3/nvme_wrap.c:3369:
../libnvme/src/nvme/registry.h:108:23: note: expected ‘int (*)(const char *, void *)’ but argument is of type ‘void (*)(const char *, void *)’
  108 |                 int (*callback)(const char *device, void *user_data),
      |                 ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

igaw:
Thanks!

PR #3953, opened 2026-09-01
nvme-discoverd: take a systemd time span for the give-up horizon

martin-belanger:
`nvme-discoverd`'s give-up horizon was `dc-giveup-hours`, a whole number of hours. That could not express "never" or anything shorter than an hour, and its `0` meant "never give up" — the opposite of what a timeout of zero normally means. nvme-stas configures the same horizon as `dc-giveup-timeout`, taking a systemd time span. This adopts that name and that encoding.

```
dc-giveup-timeout = 72hours        # the new default
dc-giveup-timeout = 3 days 5 hours
dc-giveup-timeout = 90             # unit-less is seconds
dc-giveup-timeout = infinity       # never give up
dc-giveup-timeout = 0              # give up on the first failure
```

`shr_parse_time()` is copied from systemd's `parse_time()` (`src/basic/time-util.c` at commit `21de611f695b`) rather than reimplemented, so a span means here exactly what it means in a unit file. Both files are LGPL-2.1-or-later; systemd's carries no copyright notice, only the SPDX identifier. Adaptations and their date are recorded above the copied code.

### Notes for review

- `now + dc_giveup_timeout_usec` wraps for a huge finite span. Therefore, a span that would wrap is treated as forever.
- `nvme-discoverd` is `disabled` by default, so `scripts/build.sh` does not compile it. Verified with `-Dnvme-discoverd=enabled -Dwerror=true`.


igaw:
The musl build fails

martin-belanger:
> The musl build fails

Thanks for notifying me. I completely missed that the tests were failing under musl.

igaw:
Thanks!

PR #3954, opened 2026-09-01
Alphabetize completions

jimmunn:
Alphabetize the generated bash and ksh completions files so that file changes are minimized when the C code changes. The PowerShell completions file is already sorted.

Once this is merged, I can create a PR for the integration changes that includes the meson targets and Github workflow to monitor for drift between the generated files and the committed files, similar to how accessors are managed.


igaw:
Thanks!

PR #3955, opened 2026-09-01
doc: add feat commands remainted

ikegami-t:
All generated by google AI mode.


igaw:
These need also be added to the meson.build file.

ikegami-t:
Just fixed to add as mentioned. Thank you.

igaw:
Thanks!

PR #3956, opened 2026-09-02
plugins/solidigm: Improve Telemetry debug info parsing

lgdacunh:
- Add debug info structures.
- Fix optional segment fields.
- Fix debug info segment tag parsing.
- Allow empty MemberList in telemetry configuration files.


igaw:
Thanks!

PR #3957, opened 2026-09-02
fabrics: propagate connect-all/discover errors to the exit status

### LINKED ISSUE #2800: nvme connect-all exits with status 0 despite failed discovery or connection attempts
Image

nvme -version
nvme version 2.3 (git 2.3)
libnvme version 1.3 (git 1.3)

This was not the case for older versions. Because of this nvmf-autoconnect.service is not working as expected.

Please let me know if there is a workaround. I'm using Proxmox, and in my use case the nvme target comes up after the initiator boots.

prabhakarpujeri:
Fixes #2800 on the 2.x branch.

`connect-all` on 2.x drops every error from `__discover()` and friends:

- `discover_from_conf_file()` ignores the `__discover()` return, overwrites `ret` with the disconnect result, and silently skips creation failures of the discovery controller (`goto next`)
- `__discover()` returns 0 even when per-entry connects fail (only the "already connected" case was handled)
- `_discover_from_json_config_file()` drops errors the same way

So with several targets in `discovery.conf` and one down, exit status is 0 and `nvmf-autoconnect.service` (Restart=on-failure) never retries — exactly the regression igaw analyzed on the issue (1.x accumulated the error per discovery.conf entry; 2.x ignored `__discover`'s return).

Note: 3.x already has correct aggregation through the fabrics rework (`eedb1c15d`); this is the targeted 2.x fix, keeping last-error-wins semantics matching 1.x.

### Test (local, 2.x branch)

discovery.conf: `-t tcp -a 127.0.0.1 -s 18080` (refused instantly)

```
# pristine:  nvme connect-all --no-nbft ; echo $?   -> 0
# fixed:     nvme connect-all --no-nbft ; echo $?   -> 1
```

`meson test` suite: 39 pass, 0 fail (2 pre-existing expected failures). The NBFT path was exercised only with --no-nbft in this test loop; NBFT on the 2.x tip has a separate crash, see the other PR.

### DIFF
--- a/fabrics.c
+++ b/fabrics.c
@@ -295,6 +295,7 @@ static int __discover(nvme_ctrl_t c, struct nvme_fabrics_config *defcfg,
 	nvme_subsystem_t s = nvme_ctrl_get_subsystem(c);
 	nvme_host_t h = nvme_subsystem_get_host(s);
 	uint64_t numrec;
+	int err, ret = 0;
 
 	struct nvme_get_discovery_args args = {
 		.c = c,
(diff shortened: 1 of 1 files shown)


igaw:
I've deleted my previous comment because I thought it was for master, this PR is for 2.x

igaw:
Thanks a lot!

PR #3959, opened 2026-09-02
keys: print check command results only with --verbose

### LINKED ISSUE #3902: `nvme keys check*` only status with verbose flag?
Faulty memory, the discussion on the output behavior was on the  'config validate, convert' commands. 

Since `nvme keys` is new and I think we should get these commands correct and accept breaking with 2.x here. 

```shell
$ .build/nvme keys gen-kxchap-secret | .build/nvme keys check-kxchap-secret | echo $?
0
$ .build/nvme keys gen-kxchap-secret | .build/nvme keys check-kxchap-secret
Secret is valid (HMAC 0, length 32, CRC 01b11c41)
```

If we follow here the POSIX command style, the status should only be printed with `-v`. 

Let's create a new issue on this. As this PR is for the renaming etc.

stdout vs stderr:  libnvme should also log to stderr. Let's fix on top of these changes.

_Originally posted by @igaw in https://github.com/linux-nvme/nvme-cli/pull/3887#discussion_r3842864616_

prabhakarpujeri:
Fixes #3902, per igaw's POSIX-style direction on the `keys check-*` commands (status line only with `-v`; exit status carries the verdict).

Both check commands (`check-kxchap-secret`, `check-tls-psk`) move their verdict prints from `nvme_show_result` to `nvme_show_verbose_result`. Errors still go to stderr unconditionally via `nvme_show_error`, and exit codes are unchanged (including the existing "not loaded for identity" case — exit semantics are intentionally untouched here).

Acceptance, verified locally against the build:

```
$ nvme keys check-kxchap-secret --keydata=<valid>
(exit 0, no output)
$ nvme keys check-kxchap-secret -v --keydata=<valid>
Secret is valid (HMAC 0, length 32, CRC ...)
(exit 0)
$ nvme keys check-kxchap-secret --keydata=garbage
Invalid secret header 'garbage'
(exit 1)
```

`nvme_keys_test.py` updated: runs asserting the verdict text pass `-v`, and two new cases assert stdout stays empty / exit stays 0 without `-v`. `meson test`: 106 pass, 0 fail.

### DIFF
--- a/plugins/keys/keys-plugin.c
+++ b/plugins/keys/keys-plugin.c
@@ -271,7 +271,7 @@ static int check_kxchap(int argc, char **argv, struct command *acmd, struct plug
 	if (err)
 		return err;
 
-	nvme_show_result("Secret is valid (HMAC %d, length %d, CRC %08x)", hmac, decoded_len, crc);
+	nvme_show_verbose_result("Secret is valid (HMAC %d, length %d, CRC %08x)", hmac, decoded_len, crc);
 
 	if (!cfg.identity)
 		return 0;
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #3961, opened 2026-09-02
shared: fix resource leaks on early return in table tests

sahmed-ibm:
test_multi_type_and_centered(), test_invalid_format_type(), and test_shr_table_print() all allocate a table via
shr_table_init_with_columns(). On the early-return paths taken when shr_table_get_row_id() returns a negative value, t is never freed before returning.

This results in a resource leak each time the error path is taken.

Free t on all early-return paths to prevent the leak.


igaw:
Thanks!

PR #3962, opened 2026-09-02
libnvme: check return value of sscanf in libnvme_ns_generic_to_blkdev()

sahmed-ibm:
libnvme_ns_generic_to_blkdev() calls libnvme_ns_is_generic() to validate the format of generic, then calls sscanf() again to parse instance and head_instance without checking the return value.

If sscanf() fails to match both fields, instance and head_instance are left uninitialized when passed to sprintf(), resulting in undefined behavior.

Check the return value of sscanf() and return NULL if it does not match the expected two fields.


igaw:
Thanks!

PR #3964, opened 2026-09-02
libnvme: simplify file existence check in test_creates_missing_directory()

sahmed-ibm:
The test called access() on nested_main to verify the file existed, then immediately called unlink() to remove it.

The access() call is redundant: unlink() already returns 0 on success and -1 (ENOENT) on failure, directly indicating whether the file was present. Using the return value of unlink() as the pass/fail indicator removes an unnecessary syscall and makes the intent clearer.

This also resolves a Coverity TOCTOU diagnostic (CID 562382) which flags the access()-then-unlink() pattern as a security best practices violation regardless of context.


igaw:
Thanks!

PR #3965, opened 2026-09-02
exclusion: fix insecure temporary file creation in excl_edit()

sahmed-ibm:
The excl_edit() function creates a scratch copy of the exclusion list under $TMPDIR using mkstemp().  The call is made without first setting a restrictive umask, so the file permissions are determined by the process umask at the time of the call.

If the inherited umask is permissive (e.g. 0022 or 0000), mkstemp() can create the file with group- or world-readable bits set, exposing the contents of the exclusion list to other users on the system before the editor session completes.

Set the umask to 0177 immediately before calling mkstemp() and restore the original umask immediately after, ensuring the scratch file is always created with mode 0600 regardless of the inherited umask.


igaw:
Thanks!

PR #3966, opened 2026-09-02
ocp: remove logically dead code in parse_event_fifo()

sahmed-ibm:
The parse_event_fifo() function assigns pevent_descriptor by casting the sum of pfifo_start and offset_to_move. Since pfifo_start is validated non-NULL at function entry and offset_to_move is always non-negative, this pointer arithmetic can never produce NULL. The null check immediately following the assignment is therefore always false and the break is unreachable.

The if-branch handles all events where the class type is not STATISTIC_SNAPSHOT_CLASS_TYPE, so the following branch implicitly handles the statistic snapshot case. The explicit class type check in the else-if condition is therefore redundant.

Remove the dead null check and break, simplify the else-if to a plain else, and remove the now-unreachable else block.


igaw:
Thanks!

PR #3967, opened 2026-09-02
tests: use pathlib in the e2e test suite

### LINKED ISSUE #3876: e2e tests cases should use Python's pathlib
background: https://lwn.net/Articles/1088781/

https://docs.python.org/3/library/pathlib.html

Pujara006:
Closes #3876.

Replaces `os.path` and the `os.*` filesystem calls in the e2e test suite with
`pathlib` equivalents.

### The sentinel

Worth flagging as the non-mechanical part of this. `test_log_dir` used the
string sentinel `"XXX"` to mean "not set up yet", checked in two places:

```python
if self.test_log_dir == "XXX" or not self.device_data:
log_root = self.test_log_dir if self.test_log_dir != "XXX" else self.log_dir
```

A `Path` never compares equal to a `str`, so both of these would have silently
inverted once `test_log_dir` became a `Path` — no exception, just the wrong
branch. The sentinel and its comparisons are replaced with `None` / `is None`.

### Idioms used

| before | after |
|---|---|
| `open(f, 'a').close()` | `f.touch()` |
| `os.path.exists()` + `os.remove()` | `Path.unlink(missing_ok=True)` |
| `os.path.exists()` + `os.makedirs()` | `Path.mkdir(parents=True, exist_ok=True)` |
| `os.path.exists()` + `stat.S_ISBLK(os.stat(...).st_mode)` | `Path.is_block_device()` |

The last one removes the only use of the `stat` module, so that import is
dropped. Collapsing the `exists()`/`makedirs()` pairs also closes the TOCTOU
window between the two calls.

### On scope

The issue says "e2e test cases", but most of the `os.path` usage lives in the
shared `tests/nvme_test.py` base class, which `tests/cli/` also imports. I've
included it, because the two can't actually be separated: the e2e test cases
build their data-file paths by concatenating `test_log_dir` with a basename,
which raises `TypeError` the moment it becomes a `Path`. Splitting them would
produce a commit that doesn't run.

Happy to rework this if you'd rather keep the base class untouched — though I
think that would mean leaving the e2e cases on `os.path` too.

### Verification

Done without a device, so this is all host-side:

- Every module under `tests/e2e/` imports cleanly, at each commit in the series
  (checked in throwaway worktrees, so the requirement holds per-commit and not
  just at the tip).
- Unit-tested `setup_log_dir()` and `discover_plugin_names()` directly, bypassing
  `setUp()` (which needs a device), including the sentinel fallback behaviour.
- **Paths interpolate correctly into the nvme command strings.**
  `_build_nvme_rw_cmd()` builds `--data="{data_file}"` via f-string; confirmed a
  `Path` renders as the plain path and not as `PosixPath('...')`.
- **`Path.is_block_device()` matches the old `os.path.exists()` +
  `stat.S_ISBLK(os.stat(...).st_mode)`** across a real block device, a character
  device, a regular file, a directory, and a missing path.
- `filecmp.cmp()` accepts `Path` arguments; `touch()` leaves an existing file's
  contents intact, matching `open(f, 'a').close()`; `create_data_file()` works
  with a `Path` argument.
- `flake8 tests/` goes from 117 to 109 findings on the touched files — 8 fewer,
  none new. The `/` operator produces shorter lines than `+ "/" +`.

One behaviour difference worth naming: `Path.is_block_device()` returns `False`
on `OSError` (e.g. `EACCES`) where `os.stat()` would raise. In that
device-polling loop the new behaviour seems preferable — it keeps polling rather
than propagating — but it is a change, and I did not test the permission-denied
case.

Two things I could not verify, and one I'd like your view on:

- **I don't have NVMe hardware**, so the e2e suite itself has not been run, and
  the micron plugin paths (`is_file()`, `stat().st_size`,
  `unlink(missing_ok=True)`) were never executed. That needs CI or a reviewer
  with a device.
- **`mypy --strict` goes from 1510 to 1521.** All eleven are the same shape:
  subclasses call `setup_log_dir()` (which assigns `test_log_dir`) and then use
  it, and mypy can't see through the setter, so `Optional[Path]` stays
  possibly-`None` at the use sites. These aren't new bugs — mypy is surfacing a
  precondition that was invisible while the sentinel was an untyped string. I
  left them rather than scatter `assert`s through the test cases, but happy to
  take a different approach.
- `tests/pyproject.toml` declares `requires-python = ">=3.10"` while
  `run_py_linters.py` runs mypy with `--python-version 3.8`. Everything here is
  3.8-safe either way (`unlink(missing_ok=True)` is exactly 3.8), but you may
  want to reconcile those separately.

### DIFF
--- a/tests/e2e/nvme_compare_test.py
+++ b/tests/e2e/nvme_compare_test.py
@@ -64,15 +64,15 @@ class TestNVMeCompareCmd(TestNVMeIO):
             self.skipTest("because: Optional NVM Command 'Compare' (NVMCMPS) not supported")
         self.start_block = 1023
         self.setup_log_dir(self.__class__.__name__)
-        self.compare_file = self.test_log_dir + "/" + "compare_file.txt"
-        self.write_file = self.test_log_dir + "/" + self.write_file
+        self.compare_file = self.test_log_dir / "compare_file.txt"
+        self.write_file = self.test_log_dir / self.write_file
         self.create_data_file(self.write_file, self.data_size, "15")
         self.create_data_file(self.compare_file, self.data_size, "25")
         self.compare_meta_file = None
         if self.ms > 0 and not self.ns_meta_ext:
-            self.write_meta_file = self.test_log_dir + "/" + self.write_meta_file
+            self.write_meta_file = self.test_log_dir / self.write_meta_file
             self.compare_meta_file = \
-                self.test_log_dir + "/" + "compare_meta_file.bin"
+                self.test_log_dir / "compare_meta_file.bin"
             self.create_meta_file(self.write_meta_file, self.ms)
             self.create_meta_file(self.compare_meta_file, self.ms)
 
--- a/tests/e2e/nvme_format_test.py
+++ b/tests/e2e/nvme_format_test.py
@@ -77,7 +77,7 @@ class TestNVMeFormatCmd(TestNVMe):
         self.nsze = ncap
         self.ctrl_id = self.get_ctrl_id()
         self.lba_format_list = []
-        self.test_log_dir = self.log_dir + "/" + self.__class__.__name__
+        self.test_log_dir = self.log_dir / self.__class__.__name__
         self.setup_log_dir(self.__class__.__name__)
         self.delete_all_ns()
 
--- a/tests/e2e/nvme_qpif_test.py
+++ b/tests/e2e/nvme_qpif_test.py
@@ -58,15 +58,15 @@ class TestNVMeQPIFTest(TestNVMeIO):
 
         self.start_block = 1023
         self.storage_tag = 0x1
-        self.test_log_dir = self.log_dir + "/" + self.__class__.__name__
+        self.test_log_dir = self.log_dir / self.__class__.__name__
         self.setup_log_dir(self.__class__.__name__)
-        self.write_file = self.test_log_dir + "/" + self.write_file
-        self.read_file = self.test_log_dir + "/" + self.read_file
+        self.write_file = self.test_log_dir / self.write_file
+        self.read_file = self.test_log_dir / self.read_file
         self.create_data_file(self.write_file, self.data_size, "15")
-        open(self.read_file, 'a').close()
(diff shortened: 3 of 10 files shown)


igaw:
Thanks a lot going working on this. Looks good to me even as a single commit. Though I am not a real Python hacker :)

I'll merge it and we should see any fallouts in the nightly CI runs. 

I suppose we need to do a flake8/mypy cleanup session with help from our friend claude.

Pujara006:
Thanks for merging.
Happy to take the flake8/mypy cleanup. A few numbers from while I
was in there, in case they help scope it:
- flake8 on tests/ sits at ~109 findings, almost all E501. That's
  mechanical, but it'd be a big reformatting diff for not much
  benefit — is that something you actually want, or would you
  rather leave line lengths alone?
- mypy --strict reports ~1510 errors across tests/, since the suite
  is essentially untyped. Getting to zero is a large job. The
  alternative is to relax strictness for tests/ and tighten it
  file-by-file as things get annotated. I'm happy either way —
  which would you prefer?
- Worth reconciling as part of this: tests/pyproject.toml declares
  requires-python = ">=3.10", while run_py_linters.py runs mypy with
  --python-version 3.8.

igaw:
looks good. the e2e run through without an issue

   https://github.com/linux-nvme/nvme-cli/actions/runs/33841023107/job/100923168877

there is a problem with the writezero test case but that one was already there before.

- flake8: It's a bit of code churn, but I think it the long run it's better to do it. If we disable it, all new contributions tend to ignore it. I know  some of the lines will look funky after the change, so exceptions are allowed IMO but we should have the default be enabled.
- mypy: I've done this for nvmetcli a few weeks ago and it actually found a couple of problems. So getting this done for the code base would also be highly appreciated. claude was really good doing this job, it took only a couple of hours to get done. It was  one big commit which I think is acceptable.  Whatever you prefer. I am happy that someone takes a bit of care of this code base!
- IIRC we should ask for Python 3.10 as this is the current older actively maintained Python version.

PR #3970, opened 2026-09-02
completions: test the generator against a synthetic fixture

jimmunn:
Drive the bash, zsh, and PowerShell completion test suites from a hand-written synthetic metadata fixture instead of a curated subset of real nvme output, and add Python unit tests asserting the generator's output is schema-valid, order-invariant, and alphabetically sorted.


jimmunn:
Created in wrong repo

PR #3971, opened 2026-09-03
libnvme: host identity docs and Python binding cleanupHost ids doc fixes

martin-belanger:
Three small changes around the host identity code, found while re-reading an old review of #3596. No behaviour change.

**Document the ctx default in the host identity docs.** The kdocs for `libnvmf_read_hostnqn()` and `libnvmf_read_hostid()` named `$SYSCONFDIR/nvme` as the only source. They now say that a context default takes priority over the file, that an empty default suppresses the file lookup, and that neither function returns the identity a connect uses. The Python module docstring gains a `Functions` section covering `read_hostnqn()`, `read_hostid()` and `host_get_ids()`, which it did not mention at all.

**Build the host identity Python objects with `Py_BuildValue()`.** `host_get_ids()` assembled its return tuple by hand. `Py_BuildValue()` converts a NULL `char *` to `None` and sets the exception itself on failure, so the reference bookkeeping is not needed. `read_hostnqn()` and `read_hostid()` use it too. Verified identical for NULL, non-NULL and both-NULL inputs, with no change to `None`'s refcount over 10000 calls.

**Flag `read=none` members in the generated SWIG interface.** SWIG has no write-only counterpart to `%immutable`, so a member annotated `read=none` still gets a Python getter while the C side has none. `ctx.hostnqn` and `ctx.hostid` are the only two today. The generator now marks the mismatch in the generated file instead of leaving it silent.

Built with `-Dpython=enabled`, `meson test` clean, checkpatch clean per commit.

### DIFF
--- a/libnvme/libnvme3/accessors.i
+++ b/libnvme/libnvme3/accessors.i
@@ -139,8 +139,8 @@ struct libnvme_global_ctx {
 	bool force_4k;
 	bool mi_probe_enabled;
 	bool ioctl_probing;
-	const char * hostnqn;
-	const char * hostid;
+	const char * hostnqn;	// no C getter; SWIG emits one anyway
+	const char * hostid;	// no C getter; SWIG emits one anyway
 };
 
 %pythoncode %{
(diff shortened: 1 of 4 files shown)


igaw:
Thanks!

PR #3973, opened 2026-09-03
nvme-print: fix null pointer dereference in strset_iterate_sorted_

sahmed-ibm:
The strset_iterate_sorted_() function collects set members into a dynamically grown array via name_collect(). When the array needs to grow, realloc() is called and its return value is stored directly back into nc->names without checking for NULL.

On allocation failure nc->names becomes NULL and is immediately dereferenced on the next line, causing a crash.

Store the realloc() result in a temporary pointer and return false on failure so iteration aborts cleanly.


igaw:
Thanks!

PR #3974, opened 2026-09-03
shared: add shr_xstrcmp function

ikegami-t:
Allocation-checking strcmp() that returns -1 or 1 for a NULL input.

### DIFF
--- a/libnvme/src/nvme/fabrics.c
+++ b/libnvme/src/nvme/fabrics.c
@@ -37,6 +37,7 @@
 
 #include <shared/array-util.h>
 #include <shared/compiler-attributes-util.h>
+#include <shared/string-util.h>
 
 #include <libnvme.h>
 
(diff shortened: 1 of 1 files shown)


libnvme/src/nvme/fabrics.c
```diff
@@ -2475,7 +2476,7 @@ __shr_public bool libnvmf_is_registration_supported(struct libnvme_ctrl *c)
 			return false;
 
 	libnvme_ctrl_get_dctype(c, &dctype, NULL);
-	return !strcmp(dctype, "ddc") || !strcmp(dctype, "cdc");
+	return !shr_xstrcmp(dctype, "ddc") || !shr_xstrcmp(dctype, "cdc");
```
martin-belanger: It looks like what you are looking for is `shr_streq0()` (or even `streq()` from ccan)

ikegami-t: Right then just fixed to use `shr_streq0()` instead. By the way I am not sure if it is really correct behavior or not about the function return true when both NULLs are set as strings. Thank you.

igaw:
Thanks!

PR #3975, opened 2026-09-03
libnvme: validate the host NQN and host identifier

martin-belanger:
`libnvmf_read_hostnqn()` and `libnvmf_read_hostid()` accepted any non-empty string, and `config-ini.c` carried its own private copies of an NQN check and a host identifier check. This adds one definition of each in `shared/` and points everything at it.

**Two validators in `shared/`.** `shr_nqn_valid()` applies the rules from the NVMe Base Specification, section 4.7: at most 223 bytes, an `nqn.` prefix, a `yyyy-mm` date code whose month is 01 to 12, a non-empty remainder, and a canonical UUID when the `nqn.2014-08.org.nvmexpress:uuid:` form is used. `shr_hostid_valid()` requires a canonical UUID that is not all zeros, a value the specification gives no meaning as a host identifier. 

Two deliberate departures from section 4.7, both noted in the header. The prohibition on `org.nvmexpress` as a format 1 reverse domain is not enforced: names of that shape are widely deployed as host NQNs, and 18 distinct ones appear in this tree. And where the specification permits any UTF-8 string, the check accepts printable ASCII only, excluding the space. That is a judgement call rather than a spec rule: a reverse domain name is ASCII, since an internationalized domain reaches the wire as punycode, and a UUID is hex, so no NQN anyone issues is affected. What it buys is catching a truncated read of `/etc/nvme/hostnqn` or a trailing newline, which are the realistic ways a bad value gets there.

**Both read functions now validate.** An invalid value held by the context, including the empty string, falls through to `$SYSCONFDIR/nvme` instead of being used. A file that exists but holds a malformed value is reported at `LIBNVME_LOG_ERR` and treated as absent. A missing file stays silent, as before. This changes what an empty context value means: it used to suppress the file lookup, and is now simply an invalid value like any other.

**The config parser uses the same code.** Its two static checks are gone, and no configuration that parses today stops parsing. One behaviour change follows: a `uuid:`-form NQN whose UUID is malformed is now rejected, where it was previously accepted. Several test fixtures relied on that, spelling a readable label as `nqn.2014-08.org.nvmexpress:uuid:dc-A`. Those labels were never UUIDs, so the fixtures drop the `uuid:` component and stay domain-format names.

Not covered here: `--hostnqn` and `--hostid` reach `libnvmf_host_get_ids()` as arguments ranked above both of these sources, and TIDs accept a `hostnqn=`/`hostid=` pair unchecked. Both are worth doing separately. 

Built with `-Dpython=enabled`. Every commit builds and passes the full suite on its own; checkpatch is clean of errors per commit.

### DIFF
--- a/libnvme/libnvme3/tests/test-config.py
+++ b/libnvme/libnvme3/tests/test-config.py
@@ -97,7 +97,7 @@ controller = transport=tcp;traddr=192.168.1.20;trsvcid=4420
 """)
         _write(os.path.join(self.conf + '.d', 'prod.conf'), """
 [Host]
-hostnqn     = nqn.2014-08.org.nvmexpress:uuid:1111
+hostnqn     = nqn.2014-08.org.nvmexpress:1111
 hostid      = 46ba5037-7ce5-41fa-9452-48477bf00080
 hostsymname = prod-persona
 
@@ -110,7 +110,7 @@ controller = transport=tcp;traddr=10.0.0.9;trsvcid=4420
         self.assertEqual(conns[0]['hostsymname'], 'default-persona')
         self.assertEqual(conns[1]['hostsymname'], 'prod-persona')
         self.assertEqual(conns[1]['hostnqn'],
-                         'nqn.2014-08.org.nvmexpress:uuid:1111')
+                         'nqn.2014-08.org.nvmexpress:1111')
         self.assertTrue(conns[1]['source'].endswith('prod.conf'))
 
     def test_multipath_yields_one_connection_per_controller_line(self):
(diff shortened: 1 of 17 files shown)


igaw:
there is conflict, could you rebase and while at it, it would be possible to use the newly introduced version of these to tests also in config.c, right?

martin-belanger:
> there is conflict, could you rebase and while at it, it would be possible to use the newly introduced version of these to tests also in config.c, right?

Good idea. I will use the new shared APIs to test the config.

martin-belanger:
@igaw - Rebased, and the config parser now uses the shared validators — `config-ini.c`'s own `nqn_valid()` and `hostid_valid()` are gone. `shr_nqn_valid()` picked up the two rules it had that mine did not, the month range and the printable-ASCII check. Added `shr_hostid_valid()` for the not-all-zeros rule.

One behaviour change worth flagging: a `uuid:`-form NQN with a malformed UUID is now rejected by the config parser. Some fixtures used that shape for readable labels like `:uuid:dc-A`, so those drop the `uuid:` component. Description updated.

igaw:
Thanks a lot! Having tests for those helpers is really good!

PR #3976, opened 2026-09-03
python-libnvme: stop scanning the topology when a GlobalCtx is created

martin-belanger:
The Python `GlobalCtx()` constructor scanned the NVMe topology unconditionally. The C API does not: `libnvme_create_global_ctx()` only allocates the context, and every C caller scans when it needs the tree.

That deviation cost more than a wasted sysfs walk. Every public API takes a context, including many that have nothing to do with the device tree, so a context created for configuration, registry or identity work paid for a tree it never used. The scan also created a host the caller never asked for. In nvme-stas, `GlobalCtx(owner=...)` followed by an explicit `Host(...)` left **two** hosts in the tree: the automatically-invented one and the real one the user is asking for. In `discover-loop.py`, the explicit `Host(ctx)` returned the object the scan had already made, so it created nothing at all.

**`GlobalCtx()` no longer scans.** Callers that want the tree call `scan_topology()`. nvme-stas needs no change — it never walks the tree.

**`scan_topology()` is now exposed.** The bindings offered only `refresh_topology()`, which frees every host, subsystem, controller and namespace before scanning, so any object the caller still holds becomes invalid. `scan_topology()` only adds: a device already in the tree is updated in place, nothing is removed. Both docstrings now say which is which. `refresh_topology()` stays, since dropping devices that have gone away may be needed for certain applications.

**`GlobalCtx()` accepts `hostnqn` and `hostid`**, matching the existing `owner` keyword, so a caller can build a fully configured context in one call.

The example keeps its `scan_topology()` call, with a comment explaining that it is optional and what it buys: a controller created afterwards inherits configuration from a peer controller on the same subsystem.

This is an intentional behaviour change to a public interface, taken before 3.0. Built with `-Dpython=enabled`, `meson test` clean, checkpatch clean per commit.

### DIFF
--- a/libnvme/examples/discover-loop.py
+++ b/libnvme/examples/discover-loop.py
@@ -55,6 +55,13 @@ def discover(ctx, host, ctrl, iteration):
             discover(ctx, host, new_ctrl, iteration + 1)
 
 ctx = nvme.GlobalCtx()
+
+# scan_topology() is optional. It parses sysfs and builds a tree of the
+# Host, Subsystem and Ctrl objects that already exist. With that tree in
+# place, the controller created below inherits configuration from a peer
+# controller on the same subsystem.
+ctx.scan_topology()
+
 host = nvme.Host(ctx)
 subsysnqn = nvme.NVME_DISC_SUBSYS_NAME
 transport = 'tcp'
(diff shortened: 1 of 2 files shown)


igaw:
Thanks!

PR #3978, opened 2026-09-03
Completions integration

jimmunn:
Wires the completion generator into meson and adds a CI gate so the committed bash/zsh/PowerShell scripts can't silently drift from the CLI.

- `meson compile -C <dir> update-completions` regenerates the committed `bash-nvme-completion.sh`, `_nvme`, and `nvme-completion.ps1` from the built `nvme` binary; `check-completions` runs the same generation read-only.
- New "Check completions" CI workflow runs `check-completions` and fails when the committed files are out of sync.
- Regeneration documented in `CONTRIBUTING.md` and `completions/README` (Linux/all-plugins build required — reduced `-Dplugins=` builds omit commands).
- Includes a regen so the new gate passes green on this branch.

**Design note — why two targets instead of the accessor pattern:** I used two direct targets, `update-completions` and `check-completions`, so neither needs a build option or reconfigure. The accessor generator needs its configure-time boolean (`-Dcheck-accessors=true`) because it flips *several* targets at once; completions generate in one step, so a dedicated check target is simpler.


igaw:
Looks good.

Could also move these two instruction from 
`~/meson.build` to `~/completions/meson.build`?

```meson
    install_data(
        'completions/bash-nvme-completion.sh',
        rename: 'nvme',
        install_dir: datadir / 'bash-completion/completions',
    )
    install_data(
        'completions/_nvme',
        install_dir: datadir / 'zsh/site-functions',
    )
```

Then completion related stuff is in one meson.build file.

jimmunn:
@igaw , good idea on moving the meson instructions. I've done that and committed. By the way, I will be out next week. If there are any other minor changes you want and you don't want to wait, you may want to make them yourself. I'll be back on Monday, September 14th.

igaw:
Thanks! Enjoy!

PR #3979, opened 2026-09-03
Fix issues found during WinPE testing

brgoing:
Fixes issues found during WinPE testing:
- Fixes device enumeration by using low-level, numeric bus and address values instead of parsing from a string property that can be composed differently based on platform and locale
- Fixes storage protocol commands with opcodes indicating a data transfer to the device but having zero data length by passing a minimal buffer to the driver.  The driver requires a data buffer configuration even if no data transfer is required by the command options.
- Fixes IO storage protocol commands, which were being sent to the driver as admin commands.

All WinPE-specific command handlers are now tested and working:
- Namespace Management
- NVMe-MI Send/Receive
- Format
- Compare


igaw:
Thanks!

PR #3980, opened 2026-09-03
libnvme: return an empty string for missing namespace attributes

mittal-ishaan:
libnvme_ns_get_firmware(), libnvme_ns_get_serial() and libnvme_ns_get_model() returned NULL when a device does not report the attribute, and "nvme list" died in strdup().
 ```
~# nvme list
Segmentation fault         (core dumped) nvme list

 ~# nvme list -o json | jq '.Devices[] | select(.Firmware == null) | .DevicePath'
 "/dev/nvme32n1"

 ```
 
libnvme drops blank attributes, so a controller reporting an all-spaces firmware revision (GCP persistent disks, model "nvme_card-pd") has no firmware string at all.

Return an empty string instead. "nvme list -o json" now prints "" rather than null for a missing attribute.

### DIFF
--- a/libnvme/src/nvme/tree.c
+++ b/libnvme/src/nvme/tree.c
@@ -1563,9 +1563,9 @@ __shr_public const char *libnvme_ns_get_model(struct libnvme_ns *n)
 	const char *val;
 
 	if (!n->c)
-		libnvme_subsystem_get_model(n->s, &val, NULL);
+		libnvme_subsystem_get_model(n->s, &val, "");
 	else
-		libnvme_ctrl_get_model(n->c, &val, NULL);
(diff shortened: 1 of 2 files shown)


igaw:
I am not against making the table code more robust, that's fine. 

Though we could just do the right thing with the defaults, that means instead of

```c
libnvme_ctrl_get_firmware(n->c, &val, NULL);
```

do a 
```c
libnvme_ctrl_get_firmware(n->c, &val, "");
```

mittal-ishaan:
@igaw I updated the defaults to now have an empty string for all three serial, model and firmware.

igaw:
Thanks!

Issue #3901, opened 2026-08-24
OCP C0 SMART log: JSON v2 output is missing form_factor and die_in_use_bad_nand_block fields (no case 6 block)


krishna3554:
## Summary

The OCP SMART / Health Information extended log (C0) JSON printer `json_smart_extended_log_v2()` is missing the `case 6:` field block that both the plain-text printer and the legacy JSON v1 printer emit. As a result, for drives reporting a C0 `log_page_version >= 6`, the fields **form factor**, **die in use bad NAND block (raw)** and **die in use bad NAND block (normalized)** are silently absent from the default `-o json` output while they are present in the default text output.

Static-analysis finding against current `master`; not executed here.

## Location

- File: `plugins/ocp/ocp-print-json.c`
  - Function: `json_smart_extended_log_v2()` - the version-gating `switch (smart_log_ver)` around lines 471-475 goes straight from `default:` into `case 5:` with no `case 6:` section.
- Reference implementations that do print the block:
  - `plugins/ocp/ocp-print-stdout.c`, `stdout_smart_extended_log()`, `case 6:` at ~lines 181-187 (`Form factor`, `Die in use badnandblock-Raw`, `Die in use badnandblock-Normal`)
  - `plugins/ocp/ocp-print-json.c`, `json_smart_extended_log_v1()`, which contains the equivalent `default: case 6:` block (~lines 300-311)
- Field definitions: `plugins/ocp/ocp-smart-extended-log.h` (`form_factor` [223], `die_in_use_bad_nand_block_raw[6]` [363:358], `die_in_use_bad_nand_block_normalized` [365:364])

## Problem

`ocp_smart_add_log()` passes `nvme_args.output_format_ver` (default = 2, see `src/args.c`) to `ocp_smart_extended_log()`, so JSON output normally goes through `json_smart_extended_log_v2()`. Inside that function the per-field gating switch is:

```c
switch (smart_log_ver) {
case 0 ... 1:
    break;
default:
case 5:
    /* pcie errata, media dies, power ... */
    fallthrough;
...
```

There is no `case 6:` arm and no code anywhere else in the function that emits `form_factor` or either `die_in_use_bad_nand_block_*` value. The other two printers gate exactly these three fields behind `case 6:` (i.e. printed when the drive reports log page version >= 6), so v2 JSON diverges from them.

## Trigger / Reproduction

Based on static analysis:

```sh
[...]

PR #3981, opened 2026-09-04
ocp: add missing case 6 fields to ocp smart log v2 json output

brgoing:
Adds missing case 6 field block to OCP SMART extended log v2 JSON printer.  Adds the missing fields form factor, die in use bad nand block raw, and die in use bad nand block normalized.  These fields were already present in the text output.

Adds ocp plugin tests to verify the fix and prevent similar regressions.

Reported in issue #3901.


igaw:
Thanks!

PR #3982, opened 2026-09-04
ioctl-win: add support for dsm command with deallocate attribute

brgoing:
Windows supports the deallocate dataset management command using SCSI UNMAP.  Add a dataset management handler that maps the supported functionality to SCSI_UNMAP, and update dsm test for Windows.


igaw:
Thanks!

PR #3983, opened 2026-09-06
nvme-types-base: fix kernel-doc-check warnings

ikegami-t:
Since not described NVME_SMART_CW_IPS_SHIFT/MASK values.


igaw:
Thanks!

PR #3984, opened 2026-09-06
feat: add remaining feature commands

ikegami-t:
Almost implementation suggested by Copilot.


ikegami-t:
https://github.com/copilot/share/083a0300-01e4-8c07-8150-1a42a4e56123

ikegami-t:
Still considering to add async-event-fonf command fields options instead of the `result` later.

libnvme/src/nvme/nvme-types-base.h
```diff
@@ -10226,6 +10226,34 @@ enum nvme_feat {
 	NVME_FEAT_RATE_LIMITING_TGT_MASK	= 0xff,
 };
 
+// IRQ Coalescing Feature - Shift and Mask definitions
```
igaw, line 10229: Can you add kdoc documentation here?

libnvme/src/nvme/nvme-types-base.h
```diff
@@ -10226,6 +10226,34 @@ enum nvme_feat {
...
+enum {
+	NVME_FEAT_IRQ_COAL_THR_SHIFT		= 0,
+	NVME_FEAT_IRQ_COAL_THR_MASK		= 0xff,
+	NVME_FEAT_IRQ_COAL_TIME_SHIFT		= 8,
+	NVME_FEAT_IRQ_COAL_TIME_MASK		= 0xff,
+};
```
igaw, line 10235: And the GET macros are missing for these.

igaw: BTW, the nvme-cli-ai project can help here. See the nvme-spec skill for this.

ikegami-t: Sorry already existed NVME_FEAT_IRQC_***_SHIFT/MACRO definitions, etc. then just deleted the definitions as duplicated. Thanks for the information I will do use it in future.

libnvme/src/nvme/nvme-types-base.h
ikegami-t: As commented below the definitions are not necessary then deleted. Thank you.

igaw:
Thanks!

PR #3985, opened 2026-09-07
shared: add crash handler that prints a backtrace

### LINKED ISSUE #3716: print stack trace when crashing
In case the library or nvme-cl crashes it would really help to get a stack trace.

Here two simple examples. 
```c
#define _GNU_SOURCE
#include <execinfo.h>
#include <stdio.h>
#include <stdlib.h>

void print_stacktrace(void)
{
    void *buffer[64];
    int nptrs = backtrace(buffer, 64);

    printf("Stack trace (%d frames):\n", nptrs);

    char **symbols = backtrace_symbols(buffer, nptrs);
    if (!symbols) {
        perror("backtrace_symbols");
        exit(EXIT_FAILURE);
    }

    for (int i = 0; i < nptrs; i++)
        printf("  %s\n", symbols[i]);

    free(symbols);
}

void foo(void) { print_stacktrace(); }
void bar(void) { foo(); }

int main(void)
{
    bar();
    return 0;
}
```

```shell
$ gcc -g -rdynamic stacktrace.c -o stacktrace
$ ./stacktrace
Stack trace (7 frames):
  ./stacktrace(print_stacktrace+0x1f) [0x4004d5]
  ./stacktrace(foo+0x9) [0x400575]
  ./stacktrace(bar+0x9) [0x400581]
  ./stacktrace(main+0x9) [0x40058d]
  /lib64/libc.so.6(+0x35b5) [0x7fa9106355b5]
  /lib64/libc.so.6(__libc_start_main+0x88) [0x7fa910635668]
  ./stacktrace(_start+0x25) [0x4003f5]

```

```c
#include <libunwind.h>
#include <stdio.h>

void print_stacktrace(void)
{
    unw_cursor_t cursor;
    unw_context_t context;

    unw_getcontext(&context);
    unw_init_local(&cursor, &context);

    printf("Stack trace:\n");
    while (unw_step(&cursor) > 0) {
        unw_word_t offset, pc;
        char func[256];

        unw_get_reg(&cursor, UNW_REG_IP, &pc);
        printf("  0x%lx: ", (long) pc);

        if (!unw_get_proc_name(&cursor, func, sizeof(func), &offset))
            printf("%s + 0x%lx\n", func, (long) offset);
        else
            printf("?\n");
    }
}

void foo(void) { print_stacktrace(); }
void bar(void) { foo(); }

int main(void)
{
    bar();
    return 0;
}
```

```shell
$ gcc -g stacktrace.c -lunwind -lunwind-x86_64 -o stacktrace
$ ./stacktrace
Stack trace:
  0x4005b8: foo + 0x9
  0x4005c4: bar + 0x9
  0x4005d0: main + 0x9
  0x7fee978de5b5: __libc_start_call_main + 0x75
  0x7fee978de668: __libc_start_main + 0x88
  0x400405: _start + 0x25

```

Hasshey:
Print the call stack via backtrace_symbols_fd() on SIGSEGV/SIGABRT/SIGFPE/ SIGILL/SIGBUS, then re-raise with default disposition so core dumps and kill status are preserved.

Lives in shared/ so the CLI, libnvme, discoverd and nvmf-autoconnect can all use it. libnvme installs it automatically via a constructor (libnvme_init_crash_handler) so every consumer of libnvme gets crash backtraces without any code changes. The nvme CLI also calls nvme_install_crash_handler() at startup as a belt-and-suspenders measure for builds that don't link libnvme.

Closes #3716

### DIFF
--- a/libnvme/src/meson.build
+++ b/libnvme/src/meson.build
@@ -17,6 +17,7 @@ sources = [
 if host_system == 'linux'
     sources += [
         'nvme/attr-accessors-custom-linux.c',
+        'nvme/crash-handler.c',
         'nvme/ioctl-linux.c',
         'nvme/lib-linux.c',
         'nvme/mem-linux.c',
--- a/libnvme/src/nvme/crash-handler.c
+++ b/libnvme/src/nvme/crash-handler.c
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of libnvme.
+ *
+ * Initialize the crash handler when libnvme is loaded.
+ */
+
+#include <shared/crash-util.h>
+
+static __attribute__((constructor)) void libnvme_init_crash_handler(void)
+{
+	shr_install_crash_handler();
+}
--- a/meson.build
+++ b/meson.build
@@ -650,6 +650,9 @@ if want_nvme
         endif
     else
         link_args_list = ['-ldl']
+        if cc.has_link_argument('-rdynamic')
+            link_args_list += '-rdynamic'
+        endif
     endif
 
     nvme_exe = executable(
--- a/shared/crash-util-linux.c
+++ b/shared/crash-util-linux.c
@@ -1,0 +1,188 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * This file is part of nvme-cli.
+ *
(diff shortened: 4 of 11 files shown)


igaw:
A bunch of the build targets do not build.

igaw:
Are you going to work on this anymore?

PR #3987, opened 2026-09-07
build: show all leak kinds in the valgrind test setup

martin-gpy:
By default valgrind only prints "definite" and "possible" leaks. A block still reachable from a global at exit is therefore invisible in the test log, with no stack to work from. Pass --show-leak-kinds=all so every leak kind is listed with a full stack. What fails the run remains at valgrind's default itself: reachable blocks are mostly third-party state we do not control, so they are reported for inspection rather than gating CI.


igaw:
Okay, let's see how this goes. Thanks!

PR #3990, opened 2026-09-07
Fix the Windows build outside Latin-1 locales

h1219-kim:
Building v3.0 on Windows 11 (MSYS2 UCRT64, ANSI codepage 949) hits two problems.

`scripts/win-ucrt64-setup.sh` does not install git, but `scripts/build.sh` calls
`git rev-parse` on its first line. The CI workflow installs git in its own
pacman list, so only people following `Documentation/BUILDING.md` reach this.

The rest is the Python tooling reading files and subprocess output without an
explicit encoding, so Python falls back to the ANSI codepage.

This is not only a CJK problem: the codepage in effect silently changes a
committed artifact. `completions/_nvme` carries a U+2019 (from the Virtium
help text) and `generate-completions.py` writes it through a locale-encoded
stream, so the same generator and the same input give a different file
depending on who ran it:

```
utf-8    b'\xe2\x80\x99'   3 bytes
cp949    b'\xa1\xaf'       2 bytes
cp1252   b'\x92'           1 byte
```

Where the codepage is a multi-byte one, decoding raises instead of garbling
and two tests abort. Clean release build, `PYTHONUTF8` unset, before:

```
30/69 nvme-cli:libnvme - check-public-headers       FAIL   exit status 1
67/69 nvme-cli:nvme-cli - command-metadata-schema   FAIL   exit status 1
Ok: 67   Fail: 2
```

and after:

```
Ok: 69   Fail: 0
```

`libnvme/tools/check-public-symbols.py` already passes `encoding='utf-8'`;
this does the same in the rest of the tooling that runs on Windows.

### DIFF
--- a/completions/generate-completions.py
+++ b/completions/generate-completions.py
@@ -981,7 +981,7 @@ GENERATORS = {"bash": generate_bash, "zsh": generate_zsh,
 
 
 def open_out(path):
-    return sys.stdout if path == "-" else open(path, "w")
+    return sys.stdout if path == "-" else open(path, "w", encoding="utf-8")
 
 
 def main():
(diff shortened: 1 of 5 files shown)


igaw:
Thanks!

PR #3993, opened 2026-09-08
seagate: clamp fw-activate-history entry count, bound JSON string copies

prabhakarpujeri:
## Problem

The Seagate 0xC2 vendor log page is device-supplied, and the
`fw-activate-history` display paths trusted two fields in it:

1. `numValidFwActHisEnt` (a device u32) was used directly as the loop
   bound against the compiled-in `fwActHisEnt[20]` table — a count >
   20 walks past the table in both the plain and JSON views.
2. The JSON view copied `previousFW` / `newFW` with
   `sprintf(char[8], "%s", …)` from fixed-width `u8[8]` fields that carry
   no NUL guarantee — a malformed page gives a device-controlled stack
   overflow (the plain view copies by size into 9-byte buffers and is
   safe).

## Fix

- Clamp `numValidFwActHisEnt` to the table size immediately after the
  single log read, so both display paths can only walk real entries.
- Make the JSON view copy the fixed-width strings by size into 9-byte
  zeroed buffers, exactly mirroring the plain-print path construction.

The clamp mutates only the local stack struct and only on the success
path; nothing prints the raw count afterwards, so the display simply
tops out at the table size.

## Validation

- Full meson build clean (nvme.link, all plugins).
- ASan/UBSan harness replicating the exact old and new expressions with
  adversarial fixtures (`numValidFwActHisEnt = 0xffffffff`, 8-byte FW
  fields without NULs): pre-fix path reports `stack-buffer-overflow` in
  `old_json_path`; fixed path passes clamp-max-20, pass-through on
  small counts, and 8-byte NUL-termination checks (4/4).

### DIFF
--- a/plugins/seagate/seagate-nvme.c
+++ b/plugins/seagate/seagate-nvme.c
@@ -1337,8 +1337,8 @@ static void json_stx_vs_fw_activate_history(const stx_fw_activ_history_log_page 
 	if (fwActivHis->numValidFwActHisEnt > 0) {
 		for (i = 0; i < fwActivHis->numValidFwActHisEnt; i++) {
 			struct json_object *lbaf = json_create_object();
-			char prev_fw[8] = { 0 };
-			char new_fw[8] = { 0 };
+			char prev_fw[9] = { 0 };
+			char new_fw[9] = { 0 };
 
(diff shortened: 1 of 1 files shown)


igaw:
Thanks!

PR #3995, opened 2026-09-08
 gen-hostnqn: derive a stable host identifier without DMI

### LINKED ISSUE #3977: libnvme: no deterministic host identifier without root, or without DMI
`libnvmf_generate_hostid()` tries DMI, then device-tree, then a random UUID. DMI is the only deterministic source, and it needs root: `/sys/class/dmi/id/product_uuid` and `/sys/firmware/dmi/entries/*/raw` are mode `0400`. An unprivileged caller gets `EACCES` and the code silently falls through to `libnvme_random_uuid()`. So the same machine reports a stable identifier as root and a different one on every other call as a regular user.

Why that matters needs no elaboration here: targets gate access on the host NQN, and credentials are bound to it.

The scope is narrow, since anything that opens a controller needs root anyway. It is not empty: `nvme gen-hostnqn` touches no device, runs fine unprivileged, and answers differently every time with nothing saying why. Library consumers see the same.

## Proposal

Add `/etc/machine-id` as the first source:

```
/etc/machine-id (derived) -> DMI -> device-tree -> random
```

`/etc/machine-id` is systemd's local machine identifier: 32 lowercase hex characters, 128 bits, written at installation or first boot and constant afterwards. It has been in systemd since v20 (commit `d7ccca2e`, March 2011) and every major distribution ships it. Man page: https://man7.org/linux/man-pages/man5/machine-id.5.html

It is mode `0444`, so it is the only deterministic source an unprivileged caller can read. It is also the only one that works everywhere. DMI comes from PC firmware, so it exists on x86 and on ARM64 servers following SystemReady, but not on POWER, s390x, or embedded ARM and RISC-V boards. The device-tree fallback does not fill that gap: `PATH_UUID_IBM` is `/proc/device-tree/ibm,partition-uuid` (`tree-linux.c:36`), which covers IBM POWER LPARs only. Every other machine without DMI goes straight to a random UUID, whatever its privileges.

**The raw value must not be used.** systemd's documentation says the machine ID "must not be exposed in untrusted environments, in particular on the network", and that a machine-tied identifier must instead be derived by hashing it under a fixed, application-specific key. A host NQN is sent to the target and is visible across the fabric. The derivation is `sd_id128_get_machine_app_specific()`: HMAC-SHA256 keyed by a public application id, truncated to 128 bits, with the UUID variant and version bits set.

## A second problem in the same function

`uuid_from_product_uuid()` validates the string length only (`fabrics.c:236`). A machine reporting an all-zero or all-FF System UUID, which is common on virtual machines and whitebox boards, gets that as its identifier, shared with every identical machine. Two hosts, one NQN, which is worse than a random value.

This validation used to exist. PR #654 (2020), which is where the DMI-first ordering came from, checked the UUID format and screened for fake vendor UUIDs by rejecting a value where one character dominates, which catches all-zeros and all-FFs. Neither check came along when the logic moved into C. Worth restoring, along with a log line naming the source that won.

## Implementation

`sd_id128_get_machine_app_specific()` is a public libsystemd API, but I would not link it here. libnvme builds for Windows and for systems without systemd, and nvme-cli links libsystemd only for nvme-discoverd. Lifting the implementation avoids that, with precedent: `shr_parse_time()` in `shared/time-util.c` came from systemd, and `shared/` carries the same LGPL-2.1-or-later license. systemd uses its own SHA-256 and HMAC rather than OpenSSL, so lifting also avoids gating this on `CONFIG_OPENSSL`, which `shr_hmac_sha256()` sits behind today.

**Open question:** should machine-id outrank DMI? machine-id identifies an installation, so it survives package upgrades and distribution upgrades but changes if the machine is reimaged; DMI identifies hardware and survives both. I lean toward machine-id first anyway, because with DMI first the same command returns different identifiers as root and as a normal user, and that split is silent.

martin-belanger:
`nvme gen-hostnqn` returns a random UUID whenever DMI is unreadable, and a different one on every call. That happens without root, and on machines with no DMI at all, such as containers and some virtual machines. Nothing says it happened.

Fixes #3977.

- Add `/etc/machine-id` to the chain, between the firmware sources and the random fallback. The firmware sources describe the hardware and survive a reinstall; the machine ID describes the installation. The chain is `DMI -> device tree -> machine ID -> random`.
- Skip a system UUID of all zeros or all ones. Some firmware and some virtual machines report one, and every machine reporting it would share a host identifier.
- Report the source that won, and warn when falling through to random. Normal operation is unchanged: `nvme gen-hostnqn` prints the hostnqn on stdout and nothing else. The source line is `LIBNVME_LOG_DEBUG`, below the default level; the fallback warning goes to stderr.
- Require root for `nvme gen-hostnqn`. The firmware files are readable by root only, and the output belongs in a root-owned file, so an unprivileged run cannot produce or install a usable value. **This is a behavior change**: the command used to return a random UUID instead.

The machine ID must not be published, so it is never used directly. The UUID is `HMAC-SHA256` over a fixed application ID, keyed by the machine ID, shaped into a version 4 UUID. This is the derivation `sd_id128_get_machine_app_specific(3)` performs, and it produces the same value, verified against `systemd-id128`.

The HMAC comes from a copy of systemd's own SHA-256 and HMAC code rather than from OpenSSL. `gen-hostnqn` runs from distribution post-install scripts and must also work in a build configured with `-Dopenssl=disabled`, where `shr_hmac_sha256()` returns NULL. systemd carries this code for the same reason, since libsystemd cannot link OpenSSL either. Both files are LGPL-2.1-or-later. No libsystemd dependency is added.

Tested: each commit builds with `-Dwerror=true` and passes the test suite on its own. New unit tests cover the FIPS 180-2 and RFC 4231 vectors, the machine ID parsing including the empty and `uninitialized` states, and the placeholder being skipped in favor of the machine ID.

A natural follow-up, not included: `shr_hmac_sha256()` and `shr_md5()` could move onto this implementation, which would drop OpenSSL from `shared/` entirely.

### DIFF
--- a/Documentation/nvme-gen-hostnqn.txt
+++ b/Documentation/nvme-gen-hostnqn.txt
@@ -12,9 +12,31 @@ SYNOPSIS
 
 DESCRIPTION
 -----------
-Generate a random host NQN in the form:
+Generate a host NQN in the form:
 nqn.2014-08.org.nvmexpress:uuid:1b4e28ba-2fa1-11d2-883f-0016d3cca427
-and prints it to stdout.
+and print it to stdout.
+
+The UUID is taken from the first of these sources that provides a usable
+value:
+
+1. The system UUID reported by DMI, from /sys/class/dmi/id/product_uuid or
+   /sys/firmware/dmi/entries.
+2. The firmware device tree, from /proc/device-tree/ibm,partition-uuid.
+3. The local machine ID, /etc/machine-id. The machine ID itself is never
+   published: the UUID is derived from it with HMAC-SHA256 and a fixed
+   application identifier, as systemd's
+   sd_id128_get_machine_app_specific(3) does.
+4. A random UUID, if none of the above is available. This value is not
+   stable across invocations, and a warning is printed.
+
+The first two sources describe the hardware and survive a reinstall. The
+machine ID describes the installation and does not. A UUID that is all zeros
+or all ones is a placeholder reported by some virtual machines and firmware,
+and is skipped.
+
+This command must be run as root. The DMI and device tree files are readable
+by root only, so an unprivileged run cannot reach the stable sources, and its
+output cannot be written to /etc/nvme/hostnqn either.
 
 OPTIONS
 -------
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,23 @@
 <!-- SPDX-License-Identifier: GPL-2.0-only -->
 # NEWS
 
+## Changes in 3.1 (unreleased)
+
+### Feature removals and incompatible changes
+
+* `nvme gen-hostnqn` must now be run as root. The DMI and device tree
+  files it reads are readable by root only. An unprivileged run
+  returned a random identifier instead. See `nvme-gen-hostnqn(1)`.
+
+### libnvme
+
+* `libnvmf_generate_hostid()` now falls back to the local machine ID,
+  after DMI and the device tree and before a random UUID. The machine
+  ID itself is never published. The UUID is derived from it with
+  HMAC-SHA256 and a fixed application identifier.
+
+* A system UUID of all zeros or all ones is now discarded as invalid.
+
 ## Changes in 3.0 (2026-09-07)
 
 ### Feature removals and incompatible changes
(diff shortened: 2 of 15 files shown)


igaw:
Can you also add an entry to NEWS.md about this change? We should document the behavior change there.

martin-belanger:
Added under `Changes in 3.1 (unreleased)`

igaw:
Thanks a lot!

PR #3996, opened 2026-09-08
nvme-print: trim padding in the stdout discovery log output

martin-gpy:
traddr in a discovery log entry is a 256-byte fixed-width field, and per the base spec ASCII strings are padded with spaces rather than NUL-terminated. "%-.*s" bounds the read to the field size but does not trim, so all 256 bytes are printed: the address followed by ~245 spaces. The result is a 265-column line that the terminal wraps, which looks like blank lines between traddr and eflags.

json_discovery_log() already handles this with buf2str(), which copies the field out bounded by its own size and right-trims it. Use the same helper for the stdout renderer so both are in sync.

buf2str() was static in nvme-print-json.c, which is only compiled when json-c is found, so move it to nvme-print.c and declare it in nvme-print.h. No functional change to the JSON output.

### DIFF
--- a/shared/string-util.h
+++ b/shared/string-util.h
@@ -9,6 +9,7 @@
 
 #include <ctype.h>
 #include <stdbool.h>
+#include <stdio.h>
 #include <string.h>
 #include <strings.h>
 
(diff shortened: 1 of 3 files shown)


igaw:
I'd say this helper could even go into shared so we can reuse for nvme-discoverd.

martin-gpy:
> I'd say this helper could even go into shared so we can reuse for nvme-discoverd.

Ok, replaced the previous commit with a new one now. Please take a look.

igaw:
Thanks a lot!

PR #3999, opened 2026-09-09
shared: add shr_nqn_equal and use it for identity matching

igaw:
The NVMe Base Specification (4.7/4.8) requires NQN comparison free of
locale-specific text processing, and Boot Specification 1.4 section
3.1 already requires an SMBIOS-derived UUID-format NQN's UUID to be
lower case.

Firmware that gets this wrong and hands libnvme an upper-case UUID in a
host NQN via NBFT is normalized.


martin-belanger:
Did some digging with Claude's help.

# PR #3999 review — spec compliance

`shared: add shr_nqn_equal and use it for identity matching` (igaw, 3 commits).

Specs read: Base 2.4, Boot 1.4, NVMe/TCP 1.3, RFC 9562.

## Verdict

Commit 2 (URI scheme/protocol lowercase) is required by the spec. Keep it.

Commit 1 (case-insensitive NQN comparison) contradicts the spec at three levels, and the case it is meant to tolerate is a firmware bug the Boot Specification already forbids.

## 1. The RFC citation is the wrong one

The commit message and the `shr_nqn_equal()` comment cite RFC 4122. Base 2.4 §4.7 defines the UUID-format NQN against **RFC 9562**, and Boot 1.4 §1.5.9.3 says "The UUID textual representation shall follow the textual format defined in RFC 9562." RFC 9562 obsoleted RFC 4122 in May 2024; the Base 2.1 change log lists the reference swap explicitly.

The substance survives. RFC 9562 §4: "the alphabetic characters may be all uppercase, all lowercase, or mixed case". RFC 9562 is in fact more permissive than RFC 4122, which also required lowercase on output.

## 2. The spec says to compare NQNs as binary strings

Base 2.4 §4.7, last paragraph:

> NVMe hosts, controllers and NVM subsystems process (e.g., compare for equality) UTF-8 NVMe Qualified Names without any text processing or text comparison logic that is specific to the Unicode character set or locale (refer to section 4.8).

§4.8 splits this by direction. Text processing:

> a) may occur as part of entry of UTF-8 strings into hosts and NVM subsystems as shown at points 1 and 3 in Figure 142; and
> b) should not occur as part of receiving UTF-8 strings via NVMe, as shown at points 2 and 4 in Figure 142.

So normalization at an entry point is allowed. Case folding at comparison time, on a value received from a controller, is not.

By that split, two hunks in this PR are fine and the rest are not:

- Fine: `src/fabrics.c`, `nvme disconnect --nqn=` and `nvme dim --nqn=`. That is user input at a configuration interface, Figure 142 point 1.
- Not fine: `tree.c`. `libnvme_lookup_subsystem()`, `libnvme_scan_subsystem()` and `libnvme_lookup_host()` compare values read back from sysfs, which is the controller's own NQN. That is point 2.

## 3. The motivating case is a firmware bug

Boot 1.4 §3.1, "Creating HostNQN and HostID Using SMBIOS System UUID":

> String encoded values of SMBIOS UUIDs **shall** be normalized to lowercase for maximal compatibility and ease of comparison between values.

The Boot 1.4 change log records this as ECN123, marked mandatory: "This change makes it clear in the Boot Specification that those strings need to be lowercase in format."

Firmware that puts an upper-case UUID in a host NQN is not exercising a latitude the spec grants. It is violating a `shall`. Tolerating it in libnvme hides a conformance failure that the firmware vendor can fix.

## 4. Security: the NQN is a cryptographic input

This is the part that case-insensitive matching cannot repair.

**TLS PSK identity** (NVMe/TCP 1.3 §3.6.1.1). The identity string contains NQNh and NQNc verbatim, elements 6 and 8 of the concatenation. The PSK digest in element 10 is

```
Base64(HMAC(PSK, NQNh || " " || NQNc || " " || "NVMe-over-Fabrics"))
```

and that digest is then the Context input to `HKDF-Expand-Label(PRK, "nvme-tls-psk", Context, L)`. One character of case difference changes the identity string, the digest, and the derived TLS PSK.

**DH-HMAC-CHAP** (Base 2.4 §8.3.4):

```
R1 = HMAC(Kh, Ca1 || S1 || T_ID || SC_C || "HostHost" || NQNh || 00h || NQNc)
R2 = HMAC(Kc, Ca2 || S2 || T_ID || SC_C || "Controller" || NQNc || 00h || NQNh)
```

Same conclusion. The response depends on the exact bytes.

The practical effect of this PR is worse than the problem it addresses. libnvme decides two spellings are one identity, then hands the crypto path whichever spelling it happened to store. A visible NQN mismatch becomes an `unknown_psk_identity` alert (TCP 1.3 §3.6.1.4) or a CHAP authentication failure, which is much harder to diagnose.

## 5. The target does not do the same comparison

Verified in the kernel tree (read, not tested), `drivers/nvme`:

| Site | Comparison |
|---|---|
| `target/core.c:1548` `nvmet_host_allowed()` | `strcmp` |
| `target/core.c:1490` hostnqn on controller reuse | `strncmp` |
| `host/core.c:3244` subsystem NQN | `strcmp` |

The Linux target ACL is case-sensitive. A host NQN differing only in case is denied whatever libnvme believes. So this change does not make a connection work; it makes libnvme's topology view disagree with the kernel's. Two subsystems that differ only in case are distinct to the kernel and merged by `libnvme_lookup_subsystem()`.

## 6. `strcasecmp()` is locale-dependent

`shr_nqn_equal()` calls `shr_streqcase0()`, which calls `strcasecmp()`. glibc's `strcasecmp()` folds through the current `LC_CTYPE`. Measured here:

| Locale | `strcasecmp("I", "i")` |
|---|---|
| `C` | 0 |
| `tr_TR.UTF-8` | -32 |

nvme-cli itself is safe by accident: it never calls `setlocale(LC_ALL, "")`, so it stays in the C locale. Library consumers are not. Measured: `python3` sets `LC_CTYPE` from the environment at startup, so under nvme-stas the comparison inside libnvme becomes locale-dependent.

Base 2.4 §4.8 names locale explicitly. If any form of this change survives, it must use ASCII-only folding, not `strcasecmp()`. The same applies to the two new `shr_streqcase0()` calls on `hostid`.

## 7. Folding the whole NQN is wider than the stated problem

Base 2.4 §4.7, first NQN format, ends with:

> A colon (:) prefixed string that the owner of the domain name assigns that does not exceed the maximum length. The naming authority is responsible to ensure that the NQN is worldwide unique.

The suffix is assigned by the domain owner. `nqn.2014-08.com.example:Disk1` and `nqn.2014-08.com.example:disk1` are two legal, distinct NQNs. This patch merges them. The commit message only cites the UUID hex digits, so the code is broader than its own rationale.

## 8. The stored spelling silently wins

`libnvme_lookup_host()` returns the existing object, so `libnvme_get_host(ctx, "NQN.2014-08...UPPER", ...)` now returns the host holding the lower-case spelling and the caller's argument is discarded. Same shape in `libnvme_subsystem_create_ctrl()`. Which spelling reaches the Connect command and the key derivation is no longer the one the caller asked for, and nothing says so.

## Suggested alternative

Normalize on input, compare exactly. That is what §4.8 describes and it fixes the reported case without any of the above.

Entry points, in order of usefulness:

1. `libnvme/src/nvme/nbft.c:868`, where the firmware-supplied host NQN enters libnvme. This is the reported bug, and Boot 1.4 §3.1 already says the value should have been lowercase.
2. DHCP root-path parsing, already covered by commit 2 for the scheme.
3. INI configuration and command-line NQN arguments.

Lowercase only the UUID hex digits of the `nqn.2014-08.org.nvmexpress:uuid:` form, using an ASCII table. Leave the domain-name form alone. Then every comparison stays `strcmp`, matches the kernel, and the crypto paths keep the bytes the peer expects.

## Commit 2 is correct as written

Boot 1.4 Figure 20, `<protocol>` token:

> As specified in RFC 3986, the schema is case insensitive, and this value shall be normalized to lowercase equivalent letters.

A `shall`, and it matches RFC 3986 §3.1 for the scheme itself. The lowercasing is in-place on a value the parser owns, and it does not touch the NQNs that appear elsewhere in the same URI. The comment in the code says so.

libnvme/src/nvme/fabrics.c
```diff
@@ -2599,6 +2600,13 @@ __shr_public int libnvmf_register_ctrl(
 	return nvmf_dim(c, tas, NVMF_TRTYPE_TCP, nvme_get_adrfam(c), "", NULL, result);
 }
 
+/* Lower-cases str in place. */
```
martin-belanger: This could be useful in the shared library

martin-belanger:
Reviewed the latest pushed commits. LGTM.

igaw:
Thanks for the review!

PR #4000, opened 2026-09-09
libnvme, nvme-discoverd: Fix DUPRETINFO handling

martin-belanger:
DUPRETINFO is acted on in two places where it changes behavior it should not.

The bit is defined for Current Discovery Subsystem entries only. Base Specification 2.4 Figure 320: for entries with a SUBTYPE other than 3h it shall be cleared to 0.

- `dc_walk_referral()` tested it on a subtype 1h entry (referral). 
- nvme-discoverd dropped every entry carrying the bit before looking at its subtype. A Discovery controller advertising EPCSD and DUPRETINFO together on its self entry never reached `self_callback`, so `dc_effective_epcsd()` returned false and the controller was parked and polled every `epcsd-poll-interval-minutes` despite supporting persistent connections. This one is user-visible.

Both checks are removed rather than moved to subtype 3h. Nothing connects to a self entry, so there is no duplicate connection for the bit to suppress.

That stays true deliberately, and the reason is now a comment in `_nvmf_discover()`. Everything the spec says about acting on a self entry is permissive: a host *need not* read the log page from every port, *may* read several for redundancy, *may* enable the Discovery Log Page Change AEN on one of them. Against real Discovery controllers the self entry describes the port already connected, so acting on it is a re-connect to the endpoint already in hand. A multi-port controller may also list its other ports, but a log page entry is destination-only and carries no source address or interface, so the only local path available is the one already in use. Choosing another would mean a routing lookup, and where interfaces are not separated by subnet that resolves to the default route.

Tested: each commit builds and passes the test suite on its own, with `-Dnvme-discoverd=enabled -Dwerror=true`.

### DIFF
--- a/discoverd/meson.build
+++ b/discoverd/meson.build
@@ -41,6 +41,10 @@ executable(
     install_dir: sbindir,
 )
 
+if get_option('tests')
+    subdir('tests')
+endif
+
 configure_file(
     input: 'systemd/nvme-discoverd.service.in',
     output: 'nvme-discoverd.service',
--- a/discoverd/src/dlp.c
+++ b/discoverd/src/dlp.c
@@ -43,36 +43,17 @@ static struct libnvmf_tid *tid_from_dlpe(const struct nvmf_disc_log_entry *e,
 		       e->subtype == NVME_NQN_DISC);
 }
 
-int dlp_fetch(struct discoverd_ctx *ctx, const char *devname,
-	      const struct libnvmf_tid *dc_tid,
-	      void (*ioc_callback)(const struct libnvmf_tid *t,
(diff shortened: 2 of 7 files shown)


dwsuse:
would it make sense to add a test case for this?

martin-belanger:
Test cases added.

I did not cover the referral walk itself — it connects, and there is no mock for that today.

igaw:
> Test cases added.

Thanks! 

> I did not cover the referral walk itself — it connects, and there is no mock for that today.

Maybe something should look into. My plan is to have all this covered directly in our test framework so we don't have to rely on manual tests with external hardware.

PR #4002, opened 2026-09-09
log: fix overflow for very large telemetry logs

kpmckay:
The data_written and data_remaining variables in get_telemetry_log() are defined as ints. The data_remaining variable is assigned from a size_t representing the total log size. If this is greater than 2 GiB, the log cannot be extracted as data_remaining goes negative. The data_written variable takes an ssize_t from write().

Change data_remaining to size_t to prevent the overflow, and data_written to ssize_t to match write(). Modify the format specifiers in nvme_show_error() accordingly.

### DIFF
--- a/plugins/log/log-plugin.c
+++ b/plugins/log/log-plugin.c
@@ -38,6 +38,7 @@
 #include <ccan/minmax/minmax.h>
 #include <shared/compiler-attributes-util.h>
 #include <shared/fs-util.h>
+#include <shared/io-util.h>
 #include <shared/mmio-util.h>
 #include <shared/parse-util.h>
 #include <shared/sig-util.h>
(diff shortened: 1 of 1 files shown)


plugins/log/log-plugin.c
```diff
@@ -594,13 +595,13 @@ static int get_telemetry_log(int argc, char **argv, struct command *acmd,
 			nvme_show_error("ERROR: %s: : write failed with error : %s",
 					__func__, libnvme_strerror(errno));
 			break;
-		} else if (data_written <= data_remaining) {
+		} else if ((size_t)data_written <= data_remaining) {
```
igaw: casting types is a code smell, is there no other way to get this sorted out properly?

igaw: shr_write_all

kpmckay: Ah ok, replaced the write loop with shr_write_alI(). No casts and removes duplicated functionality.

igaw:
Thanks!

PR #4003, opened 2026-09-09
NEWS: drop the --epcsd entries from the 3.0 release notes

martin-belanger:
`745ada20f` removed `--epcsd`/`--no-epcsd` a month before the release, but the shipped 3.0 NEWS.md still announce them, and also announce an `epcsd` setting for the INI config. Neither the options nor the INI key exist in the tree, so a reader following the NEWS gets an unknown-option error.

Two changes: the `--epcsd` bullet is removed, and the `nvme-fabrics.conf` bullet now mentions only `persistent`, which is real.

The 3.0 section of NEWS.md is already tagged (so maybe it's too late for the change).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -278,10 +278,6 @@
   writes a matching entry to the exclusion list before disconnecting.
   See `nvme-disconnect(1)`.
 
-* `nvme discover` and `nvme config-create` gained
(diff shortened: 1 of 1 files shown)


igaw:
I think it is better in the long run to have a correct document. I don't see it as fixed document, we have version control for this :)

PR #4004, opened 2026-09-09
libnvme: correct the key-class comment above keys[]

martin-belanger:
The comment above `keys[]` says `LIBNVMF_KEY_TUNABLE` is the only class a `controller =` line accepts. That stopped being true when `LIBNVMF_KEY_DC_TUNABLE` was added.

`config-ini.h` has always described the class correctly, so the two files contradicted each other.

It has already misled Claude who took the comment at face value instead of digging deeper in the code.

Comment only, no behavior change.


igaw:
Thanks!

PR #4011, opened 2026-09-10
build: regenerate and verify every man page section

### LINKED ISSUE #4010: Build error v3.0: "nvme-cli.conf.5 does not exist."
When trying to build v3.0, using the build template of the distro I use, I get the following error message:

`../Documentation/meson.build:516:26: ERROR: File nvme-cli.conf.5 does not exist.`

If I comment-out  the following lines (391-394) in /Documentation/meson.build, like so:

```
adoc_sources_man5 = [
#    'nvme-cli.conf',
#    'nvme-fabrics.conf',
]
```

the build runs trough fine, but that obviously isn't as it's supposed to be done.

Apparently, it's the ".5" that's the issue, because if I add ".txt" to make the lines read "*.conf.txt" - because the source contain those files with "conf.txt" - the error message still comes up, just with the file suffix now being ".conf.txt.5".
I'm not  a dev / programmer, so I haven't yet found out which part of the code makes the build process think that there should be be ".conf.5" file instead of the stated ".conf" file.

martin-belanger:
update-docs.sh copied only *.1 and *.html out of the build directory, and did not enable nvme-discoverd, so the section 5 and 8 pages were never written to Documentation/. Nothing then checked that the regenerated documentation covers every page meson installs, and release.sh ignored update-docs.sh's exit status.

Reported against v3.0.

Fixes: #4010

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,14 @@
   Specification's requirement for that token. DHCP root-path and NBFT
   data are not consistent about its case.
 
+### Build
+
+* `-Ddocs=man` and `-Ddocs=all` install the man pages shipped in
+  `Documentation/` unless `-Ddocs-build=true` is also given. The
+  section 5 and 8 pages were never written there, so those builds
+  failed to configure. They are regenerated from now on, and a
+  release aborts if any page is missing.
+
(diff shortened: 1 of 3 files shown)


igaw:
Thanks!

PR #4013, opened 2026-09-10
doc: require xmlto when building man pages

martin-belanger:
`xmlto` was looked up with `required: false` and the whole man page build sat behind if `xmlto.found()`, so `-Ddocs=man -Ddocs-build=true` without `xmlto` installed configured, built and installed successfully while producing no man pages.


igaw:
Thanks!

PR #4015, opened 2026-09-10
mi-types: add nvme_mi_mi_opcode definitions missing

ikegami-t:
Just added NVMe MI spec rev 2.2 opcodes.


igaw:
Thanks! I haven't checked yet if anything else is missing in the MI headers. There might be some more missing stuff.

PR #4016, opened 2026-09-10
samsung: Fix security vulnerabilities in vs-internal-log

h1219-kim:
`shared: add shr_sanitize_name()` comes first: `shr_valid_name()` only validates, and a caller holding a field the device supplies needs to make a usable name rather than turn the device away. It goes next to the validator instead of becoming a third copy of that character set — the micron plugin already carries a local `sanitize_serial()` that can move over separately.

Then three fixes for `nvme samsung vs-internal-log`, one issue per commit:

- The drive-supplied serial number goes into the dump file names unvalidated, so one holding `../` writes the dumps outside the directory `-O` asked for. The old check refused a `'`, ran only from `compress_dump_files()`, and so never ran without `-z`.
- `tar` and `rm` ran through `system()`. The single-quoting covers `/bin/sh` but not the `cmd.exe` that `system()` runs on Windows, where `'` is not a quoting character and `&`, `|`, `>` and `^` still apply. They now spawn through `shr_spawnp()` with a fixed argv, as the micron plugin does.
- The staging directory was created with `shr_mkdir_p()` and mode 0777, so an existing symlink or somebody else's directory at that predictable name would collect the dumps. It is created with `shr_mkdir()` and mode 0700 instead, and a path that is already taken fails.

`tests/cli/nvme_samsung_test.py` grows from 40 to 51 cases, and `shared/tests/test-string-util.c` covers the new helper.

### DIFF
--- a/Documentation/nvme-samsung-vs-internal-log.txt
+++ b/Documentation/nvme-samsung-vs-internal-log.txt
@@ -54,8 +54,12 @@ OPTIONS
 
 -z::
 --compress::
-	Collect the dumps into a temporary directory and save them as a
-	single compressed .tar.gz archive.
+	Collect the dumps into a 'temp_samsung_dumps' directory next to the
+	output path and save them as a single compressed .tar.gz archive,
+	removing that directory afterwards. The directory is created for
+	this run alone, so the command fails if anything already sits at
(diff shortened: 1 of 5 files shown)


igaw:
Thanks!

PR #4017, opened 2026-09-11
micron: add metadata files to vs-internal-log package

brgoing:
* Generates metadata files containing information regarding the device and the captured logs and includes them in the vs-internal-log package. 
* Consolidates log file names into one central location and removes duplication of string constants.
* Fixes error reporting for bad log package paths.
* Adds testing for new functionality and improves vs-internal-log testing overall.


igaw:
Thanks!

PR #4018, opened 2026-09-12
nvme-types-base: change file header description NVMe revision to 2.4

ikegami-t:
This was suggested by ChatGPT.

### DIFF
--- a/libnvme/src/nvme/nvme-types-base.h
+++ b/libnvme/src/nvme/nvme-types-base.h
@@ -23,7 +23,7 @@
  * NVMe Base Specification type definitions
  *
  * Based on NVM Express Base Specification,
- * Revision 2.3, August 1, 2025 (Ratified)
+ * Revision 2.4, July 31, 2026 (Ratified)
  *
  * This file contains core NVMe types organized by functional area:
  * - Helper Macros: NVME_GET, NVME_SET, NVME_CHECK, NVME_VAL
--- a/libnvme/src/nvme/nvme-types-base.h
+++ b/libnvme/src/nvme/nvme-types-base.h
@@ -321,6 +321,7 @@ static inline bool nvme_is_64bit_reg(__u32 offset)
  * @NVME_CAP_CMBS_SHIFT:	Shift amount to get the controller memory buffer supported
  * @NVME_CAP_NSSS_SHIFT:	Shift amount to get the NVM subsystem shutdown supported
  * @NVME_CAP_CRMS_SHIFT:	Shift amount to get the controller ready modes supported
+ * @NVME_CAP_NSSES_SHIFT:	Shift amount to get the NVM subsystem shutdown enhancements supported
  * @NVME_CAP_MQES_MASK:		Mask to get the maximum queue entries supported
  * @NVME_CAP_CQR_MASK:		Mask to get the contiguous queues required
  * @NVME_CAP_AMS_MASK:		Mask to get the arbitration mechanism supported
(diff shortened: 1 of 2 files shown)


libnvme/src/nvme/nvme-types-base.h
```diff
@@ -311,7 +311,7 @@ static inline bool nvme_is_64bit_reg(__u32 offset)
  * @NVME_CAP_AMS_SHIFT:		Shift amount to get the arbitration mechanism supported
  * @NVME_CAP_TO_SHIFT:		Shift amount to get the timeout
  * @NVME_CAP_DSTRD_SHIFT:	Shift amount to get the doorbell stride
- * @NVME_CAP_NSSRC_SHIFT:	Shift amount to get the NVM subsystem reset supported
+ * @NVME_CAP_NSSRS_SHIFT:	Shift amount to get the NVM subsystem reset supported
```
igaw, line 314: This is technically breaking the API, but I think this is okay do so, as it's the spec not a function etc.  

We should at least add at NEWS.md entry for this though.

src/nvme-print-stdout.c
```diff
@@ -1384,38 +1384,55 @@ static void stdout_subsystem_list(struct libnvme_global_ctx *ctx, bool show_ana)
 	stdout_subsystem(ctx, show_ana);
 }
 
-static void stdout_registers_cap(struct nvme_bar_cap *cap)
+static void stdout_registers_cap(uint64_t cap)
```
igaw, line 1387: Good improvement. Though we should also remove `struct nvme_bar_cap` then

src/nvme-print-stdout.c
```diff
@@ -1384,55 +1388,71 @@ static void stdout_subsystem_list(struct libnvme_global_ctx *ctx, bool show_ana)
...
+	stdout_prop_cap(PROP_CAP_MPSMAX, "%u bytes",
+			1 << (12 + NVME_CAP_MPSMAX(cap)));
+	stdout_prop_cap(PROP_CAP_MPSMIN, "%u bytes",
+			1 << (12 + NVME_CAP_MPSMIN(cap)));
+	stdout_prop_cap(PROP_CAP_CPS, NVME_CAP_CPS(cap) == NVME_CAP_CPS_NONE ?
+			"Not Reported" :
+			NVME_CAP_CPS(cap) == NVME_CAP_CPS_CTRL ?
+			"Controller scope" :
+			NVME_CAP_CPS(cap) == NVME_CAP_CPS_DOMAIN ?
+			"Domain scope" : "NVM subsystem scope");
```
igaw, line 1436: I suggest to follow the old formatting of this snippet, this is really hard to read

igaw:
```c
	stdout_prop_cap(PROP_CAP_CPS,
	       NVME_CAP_CPS(cap) == NVME_CAP_CPS_NONE ? "Not Reported" :
	       NVME_CAP_CPS(cap) == NVME_CAP_CPS_CTRL ? "Controller scope" :
	       NVME_CAP_CPS(cap) == NVME_CAP_CPS_DOMAIN ? "Domain scope" :
	       "NVM subsystem scope");
```

src/nvme-print-stdout.c
```diff
@@ -1428,12 +1428,7 @@ static void stdout_registers_cap(uint64_t cap)
 			1 << (12 + NVME_CAP_MPSMAX(cap)));
 	stdout_prop_cap(PROP_CAP_MPSMIN, "%u bytes",
 			1 << (12 + NVME_CAP_MPSMIN(cap)));
-	stdout_prop_cap(PROP_CAP_CPS, NVME_CAP_CPS(cap) == NVME_CAP_CPS_NONE ?
-			"Not Reported" :
-			NVME_CAP_CPS(cap) == NVME_CAP_CPS_CTRL ?
-			"Controller scope" :
-			NVME_CAP_CPS(cap) == NVME_CAP_CPS_DOMAIN ?
-			"Domain scope" : "NVM subsystem scope");
```
igaw: I see here it is removed again, in this case you don't need to fix the previous patch.

src/nvme-print.c
```diff
@@ -2123,7 +2123,7 @@ const char *nvme_support_str(bool support)
 	if (support)
 		return "Supported";
 
-	return "Not Supported";
+	return "Not supported";
```
igaw: I would leave this one. Not really hurting and below a `Not Reported` is used. So  this would make it inconsistent.

ikegami-t: Okay but for the json output it is used as `Not supported` but not `Not Supported` so it is not consistent between the stdout and json outputs. (The json output will be changed to `Not Supported` from `Not supported` by using this function to get the string. Also for the json outputs the string `Not enabled`, `Not enforced` and `Not ready`, etc. are used as same.)

ikegami-t:
Fixed the commits for the review comments. Thank you.

igaw:
Thanks!

PR #4021, opened 2026-09-15
nvme-print-json: print unique capability LBA formats in id-ns

mh2022-meta:
json_nvme_id_ns() walks the LBA format list up to NLBAF, while stdout_id_ns() walks it up to NLBAF + NULBAF. On a namespace that reports unique capability LBA formats (NULBAF != 0) the trailing formats are therefore missing from "nvme id-ns -o json".

With a drive reporting NLBAF=1 and NULBAF=1 the normal output lists three formats:

  lbaf  0 : ms:0   lbads:12 rp:0 (in use)
  lbaf  1 : ms:0   lbads:9  rp:0x2
  lbaf  2 : ms:64  lbads:12 rp:0x1

while the JSON "lbafs" array only contains lbaf 0 and lbaf 1. The same object is also self-inconsistent, as json_nvme_nvm_id_ns() already uses NLBAF + NULBAF when emitting the "elbafs" array.

Use the same bound as the stdout printer.


igaw:
Thanks!

PR #4023, opened 2026-09-15
plugins/seagate: clamp supported-log-pages count and keep JSON output clean

prabhakarpujeri:
## Summary

Two small fixes to `plugins/seagate/seagate-nvme.c`:

1. **Clamp device-reported `NumLogPages` in `seagate_do_controller_health()` (0xC5 vendor log page).** The value comes straight from the device (u32) and was used as the loop bound against the fixed 4096-byte `log_page_map`. A device that reports a count larger than `MAX_SUPPORTED_LOG_PAGE_ENTRIES` (= 341) would cause an out-of-bounds walk of the map in both the plain and the JSON renderers. The count is now clamped right after the read:

```c
if (le32_to_cpu(…NumLogPages…) > MAX_SUPPORTED_LOG_PAGE_ENTRIES)
    NumLogPages = cpu_to_le32(MAX_SUPPORTED_LOG_PAGE_ENTRIES);
```

2. **Keep the JSON view of `stx-vs-fw-activate-history` parseable.** `json_stx_vs_fw_activate_history()` contained two printf()s that fired inside the JSON envelope — a per-entry "Timestamp: %lu" dump inside the walk loop, and a plain-text "No Firmware activate history available" fallback before the enclosing array was closed. Both polluted stdout under `--output-format=json`. The plain-text renderer (`print_stx_vs_fw_activate_history`) already prints its own copies of both lines, so dropping them here is a no-op for the plain view.

## Testing

- `meson setup build-main && ninja -C build-main` — clean.
- ASan+UBSan mock harness (`tests/mock/seagate_log_pages_harness.c`, not part of this PR) drives the 0xC5 path with a hostile `NumLogPages` (= 0xFFFFFFFF, then = 342) and with a malformed history entry: previously crashed/printed past the log-page bounds; now bounded at 341 entries and the JSON stays parseable.

No device-observable behaviour change for correctly-behaved drives in the plain-text view; JSON view goes from "broken JSON" to valid JSON for those drives.


igaw:
Thanks!

PR #4025, opened 2026-09-15
plugins/micron: clamp num_entries in vs-fw-activate-history to the table size

prabhakarpujeri:
## Summary

`num_entries` in the micron fw activation history log page is device-supplied.
Both printers used it as the loop bound (without even `le32_to_cpu()`) against
the fixed `entries[20]` table in `struct micron_fw_activation_history_table`,
which lives inside the 4096-byte log page:

```c
for (count = 0; count < table->num_entries; count++) {
    if (!display_fw_activate_entry(count, &table->entries[count],
            formatted_output, sizeof(formatted_output), NULL))
        printf("%s
", formatted_output);
}
```

A device reporting a count above 20 runs both loops past the array, emitting
bytes past the log buffer into stdout or into the JSON envelope.

This is the exact same pattern as #3993 (seagate fw-activate-history entry-count
clamp) that igaw merged on 2026-09-08.

## Fix

Clamp `table->num_entries` to `ARRAY_SIZE(table->entries)` right after the
existing version/log-page validation, so both views see the bounded value.  The
clamp also applies `le32_to_cpu`/`cpu_to_le32` at that point.

## Testing

- `meson setup build-main && ninja -C build-main`: clean.
- Hostile-input harness (ASan+UBSan) with `num_entries = 0xFFFFFFFF`:
  - **pre-fix**: `heap-buffer-overflow` — read of size 1 at 52 bytes past a
    4116-byte allocation;
  - **post-fix**: loop runs exactly 20 iterations, exit 0.


igaw:
Thanks!

PR #4027, opened 2026-09-15
completions: document no-trailing-space insertion checks in TESTING.md

jimmunn:
Add two manual checklist items verifying that option completion leaves no space after '=' and inserts a completed value with no leading space.


igaw:
Thanks!

PR #4028, opened 2026-09-15
discoverd: validate DLPE target before host-side inheritance

martin-belanger:
A Discovery Log Page Entry carries no host-side addressing, so host_traddr and host_iface are always inherited from the DC. That inheritance is only valid if the entry's own target is actually reachable through the DC's connection.

For an I/O subsystem or referral entry, a transport mismatch alone invalidates it: host_iface is TCP-only and host_traddr's format is transport-specific, so there is no way to derive parameters for a transport the DC was not discovered on.

For a self entry, a DC with multiple ports may report one per interface (Base spec 2.4, Figure 320, subtype 03). Only the entry matching transport, traddr and trsvcid describes the connection actually in use, so it is the only one kept - to read its EFLAGS, never to connect.


igaw:
Don't we need the same fix in libnvmf version of this?

martin-belanger:
libnvmf already discards entries that do not match the transport type of the DC. This is just to make nvme-discoverd behave the same as libnvmf. And I plan to do the same for nvme-stas as well.

igaw:
Thanks!

PR #4032, opened 2026-09-15
discoverd: honor persistent=force against EPCSD=0

martin-belanger:
persistent=force means keep a Discovery Controller connected regardless of what its own EPCSD bit reports, for a target whose EPCSD self-report cannot be trusted. discoverd's EPCSD-based park decision ignored this entirely: any DC reporting EPCSD=0 got parked unconditionally, with no way to override it.

The resolved persistent value is now threaded through to the controller's tracked state, both on a fresh connect and on warm-restart adoption, and checked before parking. The override notice logs once per DC, not on every Discovery Log Page fetch -- repeated DLP-changed AENs would otherwise repeat it indefinitely.


discoverd/src/main.c
```diff
@@ -71,6 +71,11 @@ struct active_ctrl {
 	bool parent_epcsd_known;
 	bool parent_epcsd; // meaningful only if parent_epcsd_known
 
+	// This DC's resolved "persistent" setting is "force": keep it
+	// connected regardless of what its own EPCSD bit reports.
+	bool force_persistent;
+	bool force_persistent_logged; // the override notice fired once already
```
igaw: I don't like this. It introduces a new way to controlling logging and from previous experience it ends up in a big mess eventually.

discoverd/src/main.c
```diff
@@ -376,8 +384,17 @@ static void fetch_and_process_dlp(const char *devname,
...
 		 epcsd);
 
-	if (e && e->is_dc && !epcsd)
-		epcsd_park(e);
+	if (e && e->is_dc && !epcsd) {
+		if (e->force_persistent) {
+			if (!e->force_persistent_logged) {
+				disc_info("%s - EPCSD=0, but persistent=force: staying connected",
+					  libnvmf_tid_str(dc_tid));
+				e->force_persistent_logged = true;
```
igaw: You just want this to log once, this was not clear from the comment above. Though could this not made a bit more generic, `disc_info_once`?

martin-belanger:
Quick follow up. Two options:

1. Drop the log entirely. No once-logic needed at all.
2. Add a generic `disc_info_once()`. It would still need `bool force_persistent_logged` on the DC object (`struct active_ctrl`). The "already logged" state is per DC.

I think the log matters: without it, nothing in the log explains why a DC stays connected when `EPCSD=0`. Which do you prefer?

igaw:
The first option is also okay for me. I don't even mind that it would log more than once. My 'no' was based on my understanding this is a user visible knob.

I see. the log once is per DC object, in this case we would need a variable in the struct.

martin-belanger:
Kept the log once per DC, but introduced a `log_msg_once()` to make it cleaner

igaw:
Thanks!

PR #4033, opened 2026-09-15
discoverd: use __cleanup_tid in two loops

martin-belanger:
find_devname_for_tid() and the startup_audit() adoption loop each build a fresh TID per iteration with more than one exit path. __cleanup_tid frees it automatically on every one, dropping the repeated tid_free() calls and closing off a future leak if a new exit path is added without one.


igaw:
Thanks!

PR #4034, opened 2026-09-15
libnvme: pick the right self entry on a multi-homed DC

martin-belanger:
dc_survey_self_entry() kept whichever Current Discovery Subsystem entry (subtype 03h) it saw last in the log page. A DC with more than one port can report one self entry per port, so this could read EPCSD from the wrong port. dc_entry_is_self() now also matches transport, traddr and trsvcid against the connection in use.


igaw:
Thanks!

Issue #3663, opened 2026-07-28
`sndk_get_pci_ids` should be replaced with `nvme_get_pci_ids`.


igaw:
> FWIW, `sndk_get_pci_ids` should be replaced with `nvme_get_pci_ids`. 

 _Originally posted by @igaw in [#2994](https://github.com/linux-nvme/nvme-cli/issues/2994#issuecomment-5103522676)_

PR #4036, opened 2026-09-16
sandisk: use common PCI ID lookup helper

allen4791:
Link: https://github.com/linux-nvme/nvme-cli/issues/3663
This PR replaces the sandisk plugin's own PCI ID lookup function with the shared nvme_get_pci_ids() helper that other plugins (e.g. wdc) already use, so sandisk doesn't need to keep a separate copy of the same logic.
Validation
Compared the old and new PCI ID lookup implementations to confirm they resolve to the same sysfs paths and parse values the same way.
I built and ran the full existing test suite on Ubuntu 22.04 to confirm nothing else broke, though none of those tests actually
cover the sandisk plugin.

This is my first PR to nvme-cli, so please let me know if there's anything I'm missing.


igaw:
Looks good. The only thing I would like to change is to drop the 'Link: #3663' reference in the commit message. The  commit message should be self contained and only point to external reference on exceptions. This is because we never know when github decides to go evil or so. :)

allen4791:
I've dropped the link line from the commit message. And @brandon-paupore-sndk mentioned that #4039 covers the same change. I will close this PR if it turns out to be redundant. Thanks again for the advice!

igaw:
I've merged the version from @brandon-paupore-sndk as he is the maintainer of the plugin. 

Nevertheless thanks a lot for your contributions!

PR #4039, opened 2026-09-16
sandisk: port wdc vs-smart-add-log and PCI ID retrieval updates

brandon-paupore-sndk:
This enables the per device command support checking to work properly for devices which are only supported in the newer plugin, and also updates the old device+vendor ID checking to use the new mainline function across the plugin.

I did see #4036 come up today while I had these ready to go, but this should cover those same changes.


brandon-paupore-sndk:
I wrapped lines mostly to 100 columns for the ported command details to roughly match the rest of the plugin, but most of it is large formatted strings that might lose some clarity from going even further. Otherwise there were a few checkpatch findings I'm not sure are real issues:
- WARNING: Missing a blank line after declarations
  - There are blank lines, not sure if `__cleanup_nvme_global_ctx struct` might be causing something there?
- ERROR: need consistent spacing around '*' (ctx:WxV)
  - The spacing seems to be consistent, maybe the detection is tripping on the `__packed` attribute?
- WARNING: Duplicate signature
  - The commit signatures seem fine
  - From some brief experimenting this seems to hit on every commit, even within the main/master branch

plugins/sandisk/sandisk-utils.h
```diff
@@ -259,6 +263,164 @@
...
+	__u64 ext_smart_svn; __u8 ext_smart_pfbs; __u8 ext_smart_rsvd3[3];
+	__u8 ext_smart_dcc[16]; __u64 ext_smart_tnu; __u16 ext_smart_fcc;
+	__u8 ext_smart_bbpg; __u8 ext_smart_rsvd4[3]; __u64 ext_smart_seec;
+	__u64 ext_smart_rfsc; __u8 ext_smart_bsnbc[8]; __u8 ext_smart_eest[16];
+	__u16 ext_smart_ttc; __u64 ext_smart_uio; __u8 ext_smart_pmur[16];
+	__u32 ext_smart_rtoc; __u32 ext_smart_wtoc; __u32 ext_smart_ttoc;
+	__u8 ext_smart_rsvd5[4]; __u64 ext_smart_plrc; __u64 ext_smart_pscc;
+	__u16 ext_smart_maj; __u16 ext_smart_min; __u16 ext_smart_pt; __u16 ext_smart_err;
+	__u32 ext_smart_ftlus; __u32 ext_smart_tcgos; __u8 ext_smart_rsvd6[178];
+	__u16 ext_smart_lpv; __u8 ext_smart_lpg[16];
```
igaw: `sndk_nvme_c0_eol_log_page` and `sndk_nvme_ext_smart_log` should follow the common code style, every field is on a new line.

plugins/sandisk/sandisk-nvme.c
```diff
@@ -587,7 +574,126 @@ static int sndk_vs_smart_add_log(int argc, char **argv,
...
+		nvme_show_error("ERROR: SNDK: log page list is malformed");
+		ret = -1;
+		goto out;
+	}
+
+	if (!num) {
+		page_mask |= SNDK_ALL_PAGE_MASK;
+	} else {
+		for (i = 0; i < num; i++) {
+			if (log_page_list[i] == 0xc0)
```
martin-belanger: This should be a `switch()` statement, or at least: 

```
if() {
} else if() {
} else if() {
}
```

brandon-paupore-sndk:
Agreed with both comments, and updated as suggested.

igaw:
Thanks!

PR #4041, opened 2026-09-18
nvme-print: add identify controller data structure CIU and CIRN

ikegami-t:
Since the fields added into struct nvme_id_ctrl but not printed.


src/nvme-print-stdout.c
```diff
@@ -2651,6 +2654,26 @@ static void stdout_id_ctrl_tmpthha(__u8 tmpthha)
...
+		value = NULL;
+
+	va_end(ap);
+
+	printf("%-*s: %s\n", 10, name, value ? value : alloc_error);
+}
+
+static void stdout_id_ctrl_mupa(__u8 mupa)
+{
+	stdout_id_ctrl_field("%#x", mupa);
```
igaw: shouldn't be the first argument the field name, e.g. `mupa`?

BTW, I do not see why this needs to be so complex anyway, why is `mupa` special and not just a simple

```
printf("mupa       : %#x\n", le16_to_cpu(ctrl->mupa));
```

I am really not fan of these  micro abstraction of everything. It makes it really hard to read and the benefit is so small.

ikegami-t: The reason is just to make easy to add fields for future as I thought. And we can change the existing outputs also to use the macro and functions to unify if needed.

igaw: I see your point, though if we start cleaning up the stdout, we should do it with a plan. I also think it would make sense to introduce generic helpers for the output so that everything looks nice (currently it's not really nicely aligned etc. This could be something similar we do with the table code, where the output is first collected and rendered so everything aligns correctly.

This would be major refactoring of the this code and I would prefer doing it in a relative short time period to avoid even more chaos.

ikegami-t: Thank you so much for suggestions. So is it okay to change other id-ctrl fields outputs to use the `stdout_id_ctrl_field` as same for now? I am thinking to change as below. But still I am thinking also to fix or improve `stdout_id_ctrl_field` for your comments. Also I think the human readable outputs also can be changed to use a common print function.
```diff
+static void stdout_id_ctrl_vid(__u16 vid)
+{
+	stdout_id_ctrl_field("%#x", vid);
+}
 ...
+static void stdout_id_ctrl_cmic(__u8 cmic, bool human)
+{
+	stdout_id_ctrl_field("%#x", cmic);
+
+	if (human) {
+		printf(...);
+		...
+	}
+}
 ...
 static void stdout_id_ctrl(struct nvme_id_ctrl *ctrl, const char *product_name,
 			   void (*vendor_show)(__u8 *vs, struct json_object *root))
 {
 	bool human = stdout_print_ops.flags & VERBOSE, vs = stdout_print_ops.flags & VS;
 
 	if (human && product_name)
 		printf("%s\n\n", product_name);
 
 	printf("NVME Identify Controller:\n");
-	printf("vid       : %#x\n", le16_to_cpu(ctrl->vid));
-	printf("ssvid     : %#x\n", le16_to_cpu(ctrl->ssvid));
-	printf("sn        : %-.*s\n", (int)sizeof(ctrl->sn), ctrl->sn);
-	printf("mn        : %-.*s\n", (int)sizeof(ctrl->mn), ctrl->mn);
 ...
-	printf("cmic      : %#x\n", ctrl->cmic);
-	if (human)
-		stdout_id_ctrl_cmic(ctrl->cmic);
-	printf("aaa...);
-	printf("bbb...);
-	if (human)
-		stdout_id_ctrl_bbb(...);
 ...
+ 	stdout_id_ctrl_vid(le16_to_cpu(ctrl->vid));
+ 	stdout_id_ctrl_ssvid(le16_to_cpu(ctrl->ssvid));
+ 	stdout_id_ctrl_sn(ctrl->sn);
+ 	stdout_id_ctrl_mn(lctrl->mn);
 ...
+ 	stdout_id_ctrl_cmic(ctrl->cmic, human);
+ 	stdout_id_ctrl_aaa(...);
+ 	stdout_id_ctrl_bbb(..., human);
 ...
```

ikegami-t: Just changed `id-ctrl` to use `stdout_id_ctrl_field`. About the generic helpers mentioned if needed we can rechange `stdout_id_ctrl_field` or `stdout_id_ctrl_print` for them I think.

igaw: Let's take a step back here. First, can you just send a PR for adding the new fields to the libnvme API and maybe just the missing printfs without adding new helpers. Just following the existing pattern.

The refactoring/cleaning/re-writing of the nvme-print-stdout.c should be a complete new series. And also here we should take a step back. I've been pondering on this for awhile and as I suggested above, I think we  should base this code on the existing shr_table API. I've done some prototyping and here is a draft: #4049

ikegami-t: Okay I will do that. Yes the table format output seems good. Thank you.

ikegami-t: Just done without adding new helpers. Thank you.

PR #4046, opened 2026-09-18
tests: add sub-test support to runner and add runner unit tests

brgoing:
Adds failure reporting support for unittest sub tests. Sub test failures previously produced no TAP line output. Sub test failures are now reported on the TAP line for the parent test, with only one TAP line generated per test to preserve the expected test count. Also adds unit tests for validating runner behavior.


igaw:
Thanks!

PR #4047, opened 2026-09-18
global-ctx: add nvme_sysfs_ctrl_path() for mockable sysfs paths

brgoing:
nvme-cli code that reads controller sysfs attributes directly (rather than through libnvme) hardcoded "/sys/class/nvme/<ctrl>", so it could not be redirected under --set-options test-sysfs-dir= the way libnvme itself can. Add nvme_sysfs_ctrl_path() to mirror libnvme's test sysfs root and use it in the Micron PCIe BDF lookup and the PCI-IDs sysfs lookup.


src/global-ctx.c
```diff
@@ -121,6 +138,9 @@ static int nvme_apply_option(struct libnvme_global_ctx *ctx, const char *kv)
 		ret = libnvme_set_test_base_dir(ctx, val);
 	} else if (!strcmp(key, "test-sysfs-dir")) {
 		ret = libnvme_set_test_sysfs_dir(ctx, val);
+		if (!ret)
+			snprintf(test_sysfs_dir, sizeof(test_sysfs_dir), "%s",
```
igaw: The static code analyzers will be almost certaintly be unhappy about how snprintf is used without checking for overflow. I would just use `asprintf` instead and also make `test_sysfs_dir` just a pointer type instead an pre allocated array. See how we have done in libnvme.

brgoing: Thanks for the feedback.  I have updated the implementation.

igaw: Nice!

igaw:
Thanks!

PR #4048, opened 2026-09-21
nvme-print: fix out-of-bounds reads in ANA log parsers

itsmeut01:
- Fix out-of-bounds reads in both `stdout_ana_log()` and `json_ana_log()` where device-provided `ngrps` and `nnsids` values are trusted without validating them against the returned ANA log length
- Add bounds checks before each descriptor access and before the NSID iteration loop to prevent reads past the end of the buffer
- Switch `offset` from `int` to `size_t` to avoid signed arithmetic on buffer positions

Test:

- Verified with standalone ASan harness: crafted 64-byte ANA log with `ngrps=5`, `nnsids=100` — unfixed parser triggers ASan `heap-buffer-overflow READ of size 4`, fixed parser returns safely
- Boundary confirmed: malformed descriptors exceeding buffer are rejected, valid descriptors within bounds are still parsed and printed correctly
- Both stdout (`-o normal`) and JSON (`-o json`) paths covered


igaw:
Thanks!

